PR #2591, opened 2012-07-02
Fix breaking error that occurs when a command is called programatically with both an argument array and a callback.

ghost:
Previously, calling a command programatically with both an array of arguments and a callback, such as:

```
npm.uninstall([ 'step', 'npm' ], function (err, data) {  });
```

Would result in something like:

```
npm WARN uninstall invalid argument: ["step","npm"]
```

While the same operation, with the callback omitted, such as:

```
npm.uninstall([ 'step', 'npm' ], function (err, data) {  });
```

Would succeed without error.

For command calls, npm checks to see if there is only one argument present, and if so, if said argument is an array. If this check passes, npm considers this array to be the list of arguments for the command. Specifying a callback as well a command array, however, breaks this logic, because there are multiple arguments specified in the function call.

Following this check is another check to see if a callback is specified. If this check passes, the callback is popped off the `arguments` array, which, in the case described above, would leave it with one element only: the array of arguments to pass to the command. Simply swapping the order of these checks corrects the issue here.

~~I have, however, **not** been able to get npm's tests to run properly, even on an unmodified clone from `isaacs/npm`. Someone else, then, would have to verify that this change doesn't break anything else.~~ UPDATE: See below.

Bug appears to have been introduced around d2e23deb.

### DIFF
--- a/lib/npm.js
+++ b/lib/npm.js
@@ -464,12 +464,12 @@ Object.getOwnPropertyNames(npm.commands).forEach(function (n) {
       var args = Array.prototype.slice.call(arguments, 0)
         , cb = defaultCb
 
-      if (args.length === 1 && Array.isArray(args[0])) {
-        args = args[0]
-      }
-
       if (typeof args[args.length - 1] === "function") {
         cb = args.pop()
+      }
+
+      if (args.length === 1 && Array.isArray(args[0])) {
+        args = args[0]
       }
 
       npm.commands[n](args, cb)


othiym23:
I agree with @timoxley, with the further note that the API is probably going to be seeing some fairly dramatic changes in the not-too-distant future, and making assumptions about the future shape of the API is really not a safe idea.

Thanks for the time you spent putting this together, @mintplant!

PR #2612, opened 2012-07-06
Added file/url completion for "install" command.

helgeholm:
Tried to implement according to the function comment block.


othiym23:
At long last, this has been rebased, squashed, tweaked, and landed as d7b785393dffce93bb70317fbc039a6428ca37c5. Thank you very much for putting this together back when we were all younger, more innocent, and hopeful for the future.

deestan:
Oh wow, I had completely forgotten about this.  Much embarrassed.  Excellent work on the rebase! :)

PR #2716, opened 2012-08-14
Support Graceful Restarts

scien:
isaacs/npm#1999

npm restart currently runs prestop, stop, poststop, restart, prestart, start, poststart.  According to the docs, and the original intention of the feature, a restart script should be able to be provided so the application can choose to restart their service gracefully (see isaacs/npm#184) instead of issueing a full stop/start.

```
Note: npm restart will run the stop and start scripts if no restart script is provided.
```

A full stop/start should only happen if a restart script is not provided.

### DIFF
--- a/lib/run-script.js
+++ b/lib/run-script.js
@@ -80,9 +80,8 @@ function runScript (args, cb) {
 function run (pkg, wd, cmd, cb) {
   var cmds = []
   if (!pkg.scripts) pkg.scripts = {}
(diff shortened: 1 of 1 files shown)


othiym23:
This has been waiting for a test without getting one for over a year. Closing as abandoned. Thanks for taking the time to put it together, @scien, and sorry we let it languish for so long without providing feedback. If you want to take another stab at this, a new version, with tests, would be welcome.

PR #2960, opened 2012-11-25
fix: npm run-script fails silently. closes #2959

andrewrk:


othiym23:
We've gone back and forth on this one a number of times, and at this point this PR is hopelessly stale. Sorry we let it get so long in the tooth, and if the current behavior is still not to your liking, open a new PR and we can try this again.

andrewrk:
can you just fix it? as I've demonstrated it's a simple fix.

PR #2999, opened 2012-12-14
Write installed modules into package.json

lloydwatkin:
I've hit the issue a couple of times where I develop code, and then forget to `npm i <package> --save` as I go along. So what I've done here is write a small feature which reads all the package.json files in node_modules (first level only), extracts the version number and writes the results to package.json.

If you provide 'insert' after the command then existing dependencies list is not clobbered but any packages already in package.json have their version number overwritten.

I did intend to extend by allowing a user to, if desired, to choose whether the dependency was dev or standard but I'm tight for time at present. 

If this is something that gets accepted then I'll possibly come back and add later.

I added docs in doc/cli/backfill.md but if more are required please let me know where.


Jxck:
+1

watilde:
Good idea:)

dennismi:
+1

sholladay:
+1

I wish for this constantly. `npm` generally makes experimenting very easy, but sometimes I run into bugs when moving beyond that stage, because this process is currently very manual.

othiym23:
I'm genuinely really sorry, @lloydwatkin, but this patch has now drifted so far from the current code base that I don't think it's possible to clean it up and merge it as-is. I also think that the requirements for the feature have become a little clearer with some of the changes to `npm@3`, especially given the way that `npm@3` handles shrinkwrap files more transparently / in a more integrated fashion. Therefore, I'm closing this PR in favor of #9230 (which is also part of the [official npm road map](https://github.com/npm/npm/wiki/Roadmap)). If you wanted to take another stab at this, in light of what's in #9230, I bet you could probably team up with @iarna to make it a pretty painless effort to split up the work she's talked about doing to support `npm init`. Thanks for your patience and contribution!

lloydwatkin:
@othiym23 hey no problem, these things happen. It looks like something similar may be coming along very shortly which will be great. I'm not sure I've got time at present but if some becomes available I'll certianly get in touch with @iarna. Thanks for all your efforts on `npm` its a great tool!

PR #3062, opened 2013-01-16
Configurable number of spaces for indent

scttnlsn:
I'm a big fan of `npm install --save` but tend to use 4 spaces for indenting (whereas npm uses 2).  I added a config option (`indent`) which allows this to be changed.  If this is something you're interested in merging into npm I'll submit another pull request to npmconf setting the default `indent` value to 2 (and eliminate all the `npm.config.get('indent') || 2` here).

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -314,7 +314,7 @@ function save (where, installed, tree, pretty, cb) {
       }
     })
 
-    data = JSON.stringify(data, null, 2) + "\n"
+    data = JSON.stringify(data, null, npm.config.get('indent') || 2) + "\n"
     fs.writeFile(saveTarget, data, function (er) {
       cb(er, installed, tree, pretty)
     })
@@ -350,7 +350,7 @@ function prettify (tree, installed) {
        return o
     })
 
-    return JSON.stringify(tree, null, 2)
+    return JSON.stringify(tree, null, npm.config.get('indent') || 2)
   }
   if (npm.config.get("parseable")) return parseable(installed)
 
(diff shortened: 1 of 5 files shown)


syrnick:
+1. please reconsider. Many packages propose npm install --save(-dev) as a way to install them.

blittle:
Reason to reconsider, if you manually ever modify package.json in your editor which goes by different indentation scheme, then when you rerun npm on your package.json and it converts it back, it messes with the git history on the lines when only spacing is getting changed.

ljharb:
(also https://github.com/npm/npm/pull/3299 would do it automatically without an option)

sindresorhus:
:+1: #3299 is such a small change for such a large gain. Can you please reconsider?

othiym23:
The underlying reasoning that @isaacs outlined, however bluntly, when he closed #3299 is as relevant now as it was then, so no, we're not going to reconsider. `package.json`, `.npmrc`, and `npm-shrinkwrap.json` are configuration files first and everything else a distant second, and I'm desperately searching for ways to _reduce_ complexity within the CLI, not increase it.

sylvainpolletvillard:
The only complexity I see here is trying to explain to dozens of frustrated users why https://github.com/npm/npm/pull/3299 adds complexity. As @sindresorhus said, such a small change for such a large gain.

nikolas:
Yeah, this inflexibility is a real annoying part about npm! Please re-open this @isaacs

nsisodiya:
I want tabs in package.json, how i can do it ?

othiym23:
> I want tabs in package.json, how i can do it ?

Simply put, you can't, at least if you want to use `npm version`, `npm install --save`, or any of the other commands where npm will rewrite `package.json`.

danielbayley:
I think that ideally `npm` should be un-opinionated and not force any particular formatting on users. But, since the core team refuse to budge on this type of issue it seems to me that the simplest solution would be to just have a wrapper function in your shell that overrides `npm install --save` to alter the formatting after `npm` has done it's part.

I personally take this a step further and just use [CSON](https://github.com/bevry/cson) and write to `package.son` (optionally added to `.gitignore`) which I setup in [Atom](https://atom.io) to automatically do `json2cson package.cson > package.json` on save. This way I can also use comments.

dominikwilkowski:
Adding a config for the indentation makes sense and isn't much effort. Please reconsider!

PR #3299, opened 2013-03-30
Less package json change

ljharb:
`npm version` changes my entire package.json file by stripping trailing newlines, as well as reindenting it with 2 spaces. This change preserves the trailing newline as well as the original "tabs or N spaces" indentation.


glen-84:
FWIW, JSPM and Typings both support this (using the existing indentation format and EOL sequence).

matejkramny:
why is this not a community decision.? The community clearly wants to have this in, yet its being rejected for irrational reasons

PR #3853, opened 2013-09-03
Rewrite of login/adduser to allow private registries and protected user dbs

scherler:
fixes https://github.com/isaacs/npmjs.org/issues/117
- Enabling private repos by passing the couch object to addUser to not do the work twice
- using couch login instead of a put
- enabling to create new user if you have been admin before
- adding note if a 404 occurs - an indicator for a private rep
- enabling defaultRoles to be configured instead on an empty array
  - this allows to create new user with default roles defined 
    - in .npmrc `defaultRoles = devs,clients`
    - in cli `npm adduser -defaultRoles=devs,clients`
  - if not defined in the config we use the current value
- pointing out that you as well can login as an admin user

### DIFF
--- a/lib/adduser.js
+++ b/lib/adduser.js
@@ -123,7 +123,7 @@ function save (c, u, cb) {
   }
 
   // save existing configs, but yank off for this PUT
-  registry.adduser(u.u, u.p, u.e, function (er) {
+  registry.adduser(u.u, u.p, u.e, registry.couchLogin, function (er) {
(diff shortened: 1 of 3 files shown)


othiym23:
Much has changed around auth and the way that npm interacts with the registry client, so this PR is not something we can merge anymore. Sorry for not getting to this sooner, and sorry for not giving you better feedback on it at the time. Thanks for the time you put into this, though!

PR #3953, opened 2013-09-30
Exclude peerDependencies of devDependencies from shrinkwrap

runk:
Ref https://github.com/isaacs/npm/pull/3730

### DIFF
--- a/lib/shrinkwrap.js
+++ b/lib/shrinkwrap.js
@@ -8,6 +8,7 @@ var npm = require("./npm.js")
   , fs = require("fs")
   , path = require("path")
   , readJson = require("read-package-json")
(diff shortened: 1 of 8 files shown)


othiym23:
`peerDependencies` are no longer being installed implicitly in `npm@3` (which will be in beta within the next month), and shrinkwrap handling is changing pretty epically, so I think this patch is effectively now dead. If it turns out there are inconsistencies in the new implementation, it would be much easier to create a new patch against `npm@3` than to try to rebase this patch against the new installer code, so you should do that. Thanks so much for taking the time and effort to put this together (with a test!), and I'm sorry we didn't provide you with better feedback sooner.

PR #3994, opened 2013-10-15
support --save-peer

tkellen:
I think I covered everything here.  This adds support for installing things directly to `peerDependencies`.

### DIFF
--- a/doc/cli/npm-install.md
+++ b/doc/cli/npm-install.md
@@ -7,7 +7,7 @@ npm-install(1) -- Install a package
     npm install <tarball file>
     npm install <tarball url>
     npm install <folder>
-    npm install <name> [--save|--save-dev|--save-optional]
+    npm install <name> [--save|--save-dev|--save-peer|--save-optional]
     npm install <name>@<tag>
     npm install <name>@<version>
     npm install <name>@<version range>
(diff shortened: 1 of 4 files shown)


jcreamer898:
Since NPM 3 has a much greater need for peer dependencies now, seems like this would be a great feature... Thoughts @iarna?

xpepermint:
+1

QuantumInformation:
+1

dvonlehman:
+1

darkxeno:
+1

gajus:
Yes, @iarna can you please take a look into merging this?

capaj:
I too would appreciate if this was merged.

shellscape:
bummed at the lack of traction on this one. sad panda.

ffxsam:
Let's get this ball rolling again, shall we?

jvivs:
PR #3995 has relevant discussion, but was closed and redirected to the [roadmap](https://github.com/npm/npm/wiki/Roadmap). Alas, whatever was there was removed and is being rewritten as of Jan 27, 2017.

Given the frequency of interacting w/ `peerDependencies` when working with `react`, it'd be great to know where the cli team stands and what the plans are to improve cli interactions w/ peerDeps (if any).

👍

fearphage:
I don't see why you wouldn't want to install it.

> if you do npm install x --save-peer in /path/to/foo, where should x go to be a peer of foo? foo isn't even located in node_modules, so there's no reasonable place to put x.

It should go into `node_modules` like everything else. Peer dependencies mean that my script is going to call `require(x)` so make sure `x` is accessible. Installing in `node_modules` (like everything else) satisfies the role of a peer dependency as far as I can tell.

PR #4010, opened 2013-10-18
force reinstall flag for npm

eleith:
@KevinSheedy made an [excellent suggestion](https://github.com/isaacs/npm/pull/2388#issuecomment-21664472) on improving upon @rvagg's <code>--skip-installed</code> flag for npm.

this is my attempt at making that happen.

the goal of this patch is to make npm be smart about making network requests upon <code>npm install XYZ</code> commands. currently, it fetches <code>XYZ</code> everytime, even when the library already exists.

for continuous integration and provisioned development environments, this feature will come in handy. particularly the many many times i'm on a bus and i need to re-run some setup scripts...

this PR is meant to override https://github.com/isaacs/npm/pull/2388 which is a bit stale by now and proposes an odd syntax that @isaacs [pointed out](https://github.com/isaacs/npm/pull/2388#issuecomment-21657832).

### DIFF
--- a/doc/cli/npm-install.md
+++ b/doc/cli/npm-install.md
@@ -7,7 +7,7 @@ npm-install(1) -- Install a package
     npm install <tarball file>
     npm install <tarball url>
     npm install <folder>
-    npm install <name> [--save|--save-dev|--save-optional] [--force-install]
+    npm install <name> [--save|--save-dev|--save-optional] [--force-reinstall]
     npm install <name>@<tag>
     npm install <name>@<version>
     npm install <name>@<version range>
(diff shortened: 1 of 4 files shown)


othiym23:
Unfortunately, the npm installer has changed since this patch was first made in such a way that the patch no longer applies. In addition, there are some major changes to the npm installer landing shortly that also conflict with the code (if not the spirit) of this patch. Fortunately, the new installer in the `multi-stage` branch (which is what will become `npm@3` very soon) will include something very like `force-reinstall` as its default behavior, so if you're interested in seeing how that works in practice, check out `multi-stage`.

I'm very sorry that it took us so long to get to get to this patch that it was rendered superfluous, and thanks for taking the time and care to put this patch together.

PR #4016, opened 2013-10-20
Package verification using gpg

rlidwka:
It's a step towards package signing that was in roadmap for a quite long time. Since nobody seems to be doing that, I started to implement it myself, so...

For all current packages behaviour of npm is not changed.

However, if there's a `dist.signature` field near a `dist.tarball` field we're downloading (i.e. with `npm install`), npm will assume that `dist.signature` is a url pointing to a detached pgp signature, will download it and verify a tarball with it using external `gpg` program.

Web of trust is not used, all keys in the default public keyring are treated equally.

I also uploaded a [test package](http://registry.npmjs.org/test-package-signing/valid) you can use to check if npm behaves correctly in different cases. That's how you can test it:

``` perl
# Add a public key to your keyring, it can be a bad idea to add pgp
# keys this way, but it looks good enough for testing purposes
gpg --keyserver pgp.mit.edu --recv-key 0x7FC5D0F9888C20BC

# This is a only one correctly signed package, should be installed without issues
npm install test-package-signing@valid

# This package has an incorrect signature belonging to another file
npm install test-package-signing@badsig

# This is a package signed with a different key you don't have
npm install test-package-signing@diffkey

# This package has an garbage instead of signature
npm install test-package-signing@invalid

# This package has a wrong signature url (no sig uploaded)
npm install test-package-signing@nosig
```

It's a verification only. npm is still unable to sign anything, and it would require another few weeks to implement properly and a couple of pull requests to different npm dependencies. So I'd rather see your feedback to this one before continuing.

### DIFF
--- a/lib/cache.js
+++ b/lib/cache.js
@@ -83,6 +83,7 @@ var mkdir = require("mkdirp")
   , chmodr = require("chmodr")
   , which = require("which")
   , isGitUrl = require("./utils/is-git-url.js")
+  , gpg = require("./utils/gpg.js")
 
 cache.usage = "npm cache add <tarball file>"
             + "\nnpm cache add <folder>"
@@ -261,7 +262,7 @@ function add (args, cb) {
   switch (p.protocol) {
     case "http:":
     case "https:":
-      return addRemoteTarball(spec, null, name, cb)
+      return addRemoteTarball(spec, null, null, name, cb)
 
     default:
       if (isGitUrl(p))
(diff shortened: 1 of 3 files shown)


othiym23:
I really appreciate the thorough work you've done putting this patch together, Alex, as well as the thoughtful work Eric did reviewing it. However, we're not going to take this patch. Package integrity is an important part of a robust ecosystem of shared code, which is why we have package shasums. Also, being able to establish a chain of custody for a package, and being able to make attestations about who did what to the package, are also very useful parts of building a trustworthy packaging system.

However, this is just a piece of an overall solution, and it brings with it a lot of the baggage that comes along whenever GnuPG or PGP get involved. Without a web of trust (sigh), a PKI (ugh), or some other mechanism to tie identities to trust metrics, this is essentially a complicated, very expensive, and fragile version of the shasum check npm already has.

I know that @isaacs has some ideas on how we might take advantage of existing npm infrastructure to improve package verification, and I imagine @evilpacket and the rest of the crew at ^lift have some ideas as well. It would be great to get a discussion going (perhaps on node-forward/discussions#29) about the right way to do this, but I do believe pretty strongly that we need a strategy going into this kind of work, because implemented piecemeal, it's potentially more dangerous than doing nothing at all.

I do really appreciate you taking the time to put this together, and I apologize that it took the npm team over a year to get around to giving you feedback. Thank you for your time and effort.

tobytteh:
🤢

omeid:
@othiym23 So the argument is that code signing is too hard, so let's stick with checksum which obviously doesn't address the issue of authenticity and doesn't help with problems like #16839 (which is not the first, and I won't be the last either)?

judge2020:
Gee sure would have helped, https://github.com/eslint/eslint-scope/issues/39.

Solution:

Allow maintainers to enable an option that would require each published package to be signed with a trusted, per-repo or per-account GPG key.

PR #4058, opened 2013-10-30
Npm exec implementation for #3313

can3p:
Hi

I've extracted the exec code from the lifecycle.js and added an npm exec command to run scripts in the package context. In part it fixes #3313. In reality I couldn't eliminate lifecycle at all, because it contains some specific logic like hooks.

I know that @grncdr works on the same task, but it happened that I've made this work too and I'll be happy if it will be useful.

I'm ready to fix things if needed.

### DIFF
--- a/lib/exec.js
+++ b/lib/exec.js
@@ -1,6 +1,8 @@
 module.exports = exec
 
 var path = require("path")
+  , pkgexec = require('./utils/pkgexec')
+  , log = require("npmlog")
   , readJson = require("read-package-json")
 
 exec.usage = "npm exec <command> [<options>]"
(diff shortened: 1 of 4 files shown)


othiym23:
This is another patch that got overtaken by events elsewhere in the npm project. I'm sorry we let it sit for so long. A few things have happened:

* `npm@2` added the ability to pass arguments to commands defined in `"scripts"`
* a few of us agreed that it would be handy for npm to include a separate command, `npm-exec`, that allows you to run commands in the environment provided to child processes by `run-script`, but without all of the command-line argument parsing that regular `npm` does. This is in part because decoupling that argument parsing from npm for a single command was deemed sufficieintly complicated that it made more sense to put in a separate command.
* many tweaks have been made to `npm run-script` since the release of `npm@2`, so this has drifted further from where the code is now.

There is clearly still demand for an `npm-exec`, so if you wanted to take this code and use it as the basis for that and open it as a new PR, that would be awesome. :fireworks: This patch, however, is dead – `npm exec` isn't something that makes sense right now, and the code here and the code in master have gone their separate ways. I apologize for letting this patch sit so long without feedback from the core team, and I thank you for taking the time to put together such a complete patch.

PR #4091, opened 2013-11-05
smaller/equals version than current will throw an error

seriousManual:
referencing #4087 if the version is smaller or equals
an error is thrown, can be overwritten be --force


othiym23:
I agree with @luk- – this doesn't make sense as an error or a warning. There are many good uses for versions that are lower than preceding, e.g. releasing a maintenance version of an older release series from the same repo. Thanks for putting the time into this and the discussion on #4087, though!

PR #4320, opened 2013-12-15
Patch for a hole in loglevel='silent' when using npm programmatically

mikermcneil:
#### Issue

When using `npm.commands.install` programmatically, a "pretty" `console.log` describing the installed modules is run, even when `config.loglevel=silent`. 

Here's a simple test script which demonstrates the issue (modified from the example [here](https://github.com/isaacs/npm#using-npm-programmatically))

``` javascript
var npm = require('npm');
npm.load({
    registry: 'https://registry.npmjs.org',
    loglevel: 'silent'
}, function(err) {
    if (err) throw new Error(err);

    npm.commands.install(['lodash'], function(err, data) {
        // ...
    });
});
```
#### Proposed Solution

Checks if `npm.config.loglevel === 'silent'`, and skips writing a log if it does.

@isaacs If there's another way you'd rather go about approaching this, I'm happy to-- just figured this was the intended behavior ( it's what I was expecting when I configured `silent`, at least )  Thanks!

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -95,7 +95,9 @@ function install (args, cb_) {
       var tree = treeify(installed || [])
         , pretty = prettify(tree, installed).trim()
 
-      if (pretty) console.log(pretty)
+      if ( npm.config.get('loglevel') !== 'silent' && pretty) {
+        console.log(pretty);
+      }
       save(where, installed, tree, pretty, hasArguments, cb_)
     })
   }


mikermcneil:
> broken out into the various smaller modules

:+1: 

In the mean time, I'm working on http://node-machine.org/machinepack-npm

trusktr:
@mikermcneil 

> It only works if npm is already installed on the system (which it almost certainly is-- otherwise how did you get enpeem?).

People who install Meteor don't need to have npm installed globally. Node and NPM are bundled with Meteor.

@thejohnfreeman That workaround

kevin-smets:
For me, the workaround of overriding console.log did not work... Would've been nice though..

I had to resort to more overriding in order to have npm be silent:

```
const LOG_PREFIX = "%log-prefix%";

function log(message) {
    console.log(LOG_PREFIX() + message)
}

// Overwrite stdout so npm will shut up
process.stdout.write = (function (write) {
    return function (string, encoding, fd) {
        if (string.startsWith("dr: ")) {
            arguments[0] = string.replace(LOG_PREFIX, "");
            write.apply(process.stdout, arguments);
        }
    }
})(process.stdout.write);

log("blerb");
```

That way, anything not going through `log()` will not reach stdout at all. You can easily store the `process.stdout.write` to a variable or as a return function to restore it.

mikermcneil:
@kevin-smets what worked for us was [spinning up the `npm install` as a child process](https://github.com/mikermcneil/machinepack-npm/blob/master/machines/install-dependencies.js#L72) then gobbling up the output from that:

```javascript
// Run `npm install`
require('machinepack-process').spawn({
  command: 'npm install',
  dir: '/code/sandbox/some-project/'
}).exec(function (err, bufferedOutput) {
  // ...
});
```

>(see also http://node-machine.org/machinepack-process/spawn)

kevin-smets:
@mikermcneil that's indeed a possibility, but I need to use it programmatically, because I depend on the json output. So running it as a cmd is not really an option for my use case sadly. But thanks for the suggestion.

othiym23:
@kevin-smets is this a case when the `--json` option doesn't produce the output you want? The CLI team strongly discourages programmatic use of npm's internal APIs, as they're poorly designed for third-party use, and are a poison forest of jagged edge cases if you don't understand everything about npm's internal architecture.

kevin-smets:
Ugh, can't believe I missed that flag. I will run it as a child process >_<

Thanks for the help!

PR #4353, opened 2013-12-20
add duplicate-sub-deps option flag to npm install + test

mzedayda:
```
npm install . --duplicate-sub-deps
```

will install all dependencies even if they are already met on higher levels

Can be useful when folder structure counts. 
also makes install process less unpredictable, see #4037


othiym23:
This is going pretty much completely the opposite direction from where `npm@3` is headed. `npm@3` tries to install a maximally flat tree (in effect, a pre-deduplicated tree) from the get-go. As such, this doesn't really fit in with how the installer should work, and so we're not going to take this patch. Thanks for the time you put into this, and if there's a specific use case you were trying to satisfy with this patch that the new `multi-stage` work doesn't handle, please let us know.

PR #4357, opened 2013-12-20
Add flag to force `update` to install latest version

refack:
But only in "our" package's modules (not recursively, cause that's just crazy ;-p )

### DIFF
--- a/lib/update.js
+++ b/lib/update.js
@@ -32,9 +32,13 @@ function update (args, cb) {
         , dep = ww[1]
         , want = ww[3]
         , what = dep + "@" + want
+        , latest = ww[4]
         , req = ww[5]
         , url = require('url')
 
+      // sometimes we want to ignore the version in package.json and update to latest
+      if (npm.config.get("force-latest") && !~where.indexOf('node_modules'))
+        what = dep + "@" + latest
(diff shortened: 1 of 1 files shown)


smikes:
How about this:

```
npm --depth 0 outdated
npm --depth 0 update --save
```

The first command will list what's about to happen; the second command will try to update all the direct dependencies to `@latest`, honoring semver.

othiym23:
Why is `--depth 0` necessary in either case? Is this presuming a pre-`npm@2.6.1` world?

smikes:
Yes, at present I am assuming that at best people are running npm@latest, not npm@next

Also even with npm@2.6.1 once when I did update --dev --save-dev the behavior looked suspiciously like an infinite loop.

othiym23:
A lot has changed around `npm outdated` and `npm update` recently, and I don't think this patch makes sense. @smikes, if you (or somebody else!) want to put together a patch to make

```
npm outdated
npm update --save
npm update --save-dev
```

work the way you describe above, that would have my support / get merged quickly.

@refack, I apologize for the lengthy delay between you submitting this patch and us reviewing it, and I also apologize that all that time went by only for me to decide not to merge it. Thanks for taking the time and energy to put this together!

Issue #3869, opened 2013-09-06
have npm report the error with a package.json


Pomax:
since npm relies on a well formatted package.json file, it would be nice if it could report the actual error it sees when trying to parse a bad package.json rather than merely reporting that there is some error without saying which line/column or the type of error encountered. If there is a JSON parse error perhaps the package.json file could be thrown at jshint to get the precise error?

ilanbiala:
Any progress on this? Seems like something so simple but so important when all that is usually missing is a comma or something visually insignificant.

othiym23:
Yes! There has been! @smikes has done is all a huge favor!

Closed by abdd040da90932535472f593d5433a67ee074801. This is included in `npm@2.7.1`, which is the version currently being tested (`npm@next`).

PR #4373, opened 2013-12-24
make JSON.parse return error positions

rlidwka:
See issue #3869. Since it is marked patch-welcome, here is a patch. :)
### Setup:

```
$ echo '{ "name": "test", some garbage' > package.json
```
### Was:

```
$ npm install
npm ERR! install Couldn't read dependencies
npm ERR! Failed to parse json
npm ERR! Unexpected token s
npm ERR! File: /tmp/package.json
npm ERR! Failed to parse package.json data.
npm ERR! package.json must be actual JSON, not just JavaScript.
```
### Will be:

```
$ ~/npm/cli.js install
npm ERR! install Couldn't read dependencies
npm ERR! Failed to parse json
npm ERR! Unexpected token 's' at 1:19
npm ERR! { "name": "test", some garbage
npm ERR!                   ^
npm ERR! File: /tmp/package.json
npm ERR! Failed to parse package.json data.
npm ERR! package.json must be actual JSON, not just JavaScript.
```
### Implementation:

It was suggested earlier to simply wrap JSON.parse in try..catch whereever it's used and use another library to figure out an exact error.

Sadly, it needs to be done for every JSON.parse call there is. Including those parsing npm-shrinkwrap.json and God knows what else. Ok, we can wrap it into it's own function.

Sadly, npm is scattered across multiple packages, so for example read-package-json will need to be updated as well. When I got to this point, I said "screw it, making minor pull requests to multiple packages is just stupid" and wrote a nice patch.
### Monkeypatch warning:

Yes, I perfectly understand the implications of it. Every single library including core will have this fallback. But since it's only a fallback, [what could possibly go wrong](https://en.wikipedia.org/wiki/Murphy%27s_law)?
### Disadvantages:
1. If the error is deep inside json, parser will bloat up your stack trace with internal calls. Probably fixable with another try..catch.
2. If somebody uses reviver (?) with side effects (??), it may be called x2 times. Maybe check for error type or something.
### Tests:

None at the moment, but I can add some upon request.

### DIFF
--- a/bin/npm-cli.js
+++ b/bin/npm-cli.js
@@ -70,6 +70,36 @@ if (reqVer && !semver.satisfies(nodeVer, reqVer)) {
 
 process.on("uncaughtException", errorHandler)
 
+;(function monkeyPatch() {
+  // This module parses json using jju as a fallback.
+  //
+  // The reason to do that is because jju provide an error message
+  // with an exact position of an occured error.
+
+  try {
+    var jju = require('jju')
+
+    var _parse = JSON.parse
+    JSON.parse = function(text, rev) {
+      try {
+        return _parse(text, rev)
+      } catch(err) {
+        // this call is expected to throw always
+        jju.parse(text, {
+          mode: 'json',
+          reviver: rev,
+        })
+
+        // if an alternate parser didn't throw, re-throw original error
+        // this shouldn't normally happen
+        throw err
+      }
+    }
+  } catch(err) {
+    // module is not installed :(
+  }
+})()
+
 if (conf.usage && npm.command !== "help") {
   npm.argv.unshift(npm.command)
   npm.command = "help"
--- a/package.json
+++ b/package.json
@@ -131,6 +131,9 @@
     "ansistyles",
     "path-is-inside"
   ],
+  "optionalDependencies": {
+    "jju": "~0.4.1"
+  },
   "devDependencies": {
     "ronn": "~0.3.6",
     "tap": "~0.4.0",


mzgol:
I don't like the fact that it changes a global. This always leads to unpleasant surprises.

If `JSON.parse` is too little informative, it'd be better to switch to a library for parsing all over the code base. I know it's tedious but at least safe.

Another option would be to submit a PR to v8 to provide a more informative message. Firefox, for example prints:
```
SyntaxError: JSON.parse: expected double-quoted property name at line 1 column 19 of the JSON data
```
Still worse than what `jju` provides but who said that couldn't be tweaked further as well? And that's way better than what's currently in v8 anyway.

smikes:
Since this has languished for over a year, I'm guessing it's a "no".  

I have written a module that provides a drop-in replacement for `JSON.parse()` that calls `jju.parse()` as a fallback, and does no monkeypatching.  PR for `read-package-json` is here: https://github.com/npm/read-package-json/pull/42

nathan7:
Maybe just export `stringify` too and make it a drop in for the JSON object? It means we can add it all over the place easily with just `var JSON = require(…)`

othiym23:
I'm not comfortable with this kind of global monkeypatch either. This means that the npm tests have to try *everything*, because the many subsidiary modules don't have their tests written with this change in mind. npm's tests don't test everything (far from it), and as a result we're going to need a more targeted solution. Thanks for putting this together and sparking the discussion, though.

PR #4631, opened 2014-02-09
fix for npm not cleaning up lockfiles on ctrl+c

rlidwka:
lockfile is capable of cleaning up after itself

However, npm error handler is calling process.exit() inside process.on('exit') which screws things up.

node < 0.10 doesn't have that feature, so npm on it will still be buggy... well it's a good reason for update, is it?

### DIFF
--- a/lib/utils/error-handler.js
+++ b/lib/utils/error-handler.js
@@ -9,7 +9,7 @@ var cbCalled = false
   , path = require("path")
   , wroteLogFile = false
   , exitCode = 0
-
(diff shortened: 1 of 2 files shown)


othiym23:
We made a bunch of changes to the error-handler as part of dealing with race conditions, and we pretty comprehensively overhauled how npm uses locks as part of `npm@2`. This patch no longer really makes sense in the context of the current code base, so I'm going to close it. Sorry it took so long for us to get back to you on this one, and thanks for taking the time to put it together.

PR #4667, opened 2014-02-13
Don't *always* fetch remote repositories of git-dependencies

guybrush:
this PR trys to fix #4191

currently when you do `npm i` and you have a git-dependency 
it will create a local cache of the repo if not already present.
then it will do `git archive` on that local cached repo to create 
a tarball for the git-ref you defined in the package.json
(which is `master` by default). now when you run `npm i` again it 
will fetch the remote repo to update the cache, no matter what!
even if you did not change the git-ref.

so can we just skip the fetching-part once we have a local cached
repo and we didnt even change the git-ref? the problem is:
how do we know if the git-ref you put in your dependency-url is
a commit-hash or branch/tag-name?

consider following package.json:

```
{ "name" : "foo"
, "version" : "0.0.0"
, "dependencies" : { "bla" : "ghUser/ghRepo#blub" }
}
```

how can we know if `blub` is a commit-hash, a branch or a tag? we cant.

an idea would be to introduce another seperator for the git-url:
`ghUser/ghRepo@branchOrTag` which indicates that we mean a branch and
not a commit-hash?

this PR currently implements: just look in the locally cached repo for a commit
with the hash `blub` and if we find any we just assume that it must
be a commit-hash. then we just skip fetching the remote repo and
just pull out a tarball of the cached repo. 

also i am currently only skipping the fetch if the specified git-ref is longer
than 4 chars (because what if someone defines `ghUser/ghRepo#a`).

further improvement would be to even skip creating the tarball if
the hash in the current installed package's package.json points already
to `blub`. but i did not look into the relevant codepath yet, so no idea.

### DIFF
--- a/lib/cache.js
+++ b/lib/cache.js
@@ -550,24 +550,30 @@ function cloneGitRemote (p, u, co, origUrl, silent, cb) {
 
 function archiveGitRemote (p, u, co, origUrl, cb) {
   var git = npm.config.get("git")
-  var archive = [ "fetch", "-a", "origin" ]
-  var resolve = [ "rev-list", "-n1", co ]
+  var argsFetch = [ "fetch", "-a", "origin" ]
+  var argsResolve = [ "rev-list", "-n1", co ]
+  var argsArchive = [ "archive", co, "--format=tar", "--prefix=package/" ]
   var env = gitEnv()
 
   var errState = null
   var n = 0
   var resolved = null
-  var tmp
-
-  exec(git, archive, {cwd: p, env: env}, function (er, stdout, stderr) {
-    stdout = (stdout + "\n" + stderr).trim()
-    if (er) {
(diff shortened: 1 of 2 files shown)


othiym23:
Oof. The first big cache refactor last year totally wiped out this change, and further tweaks to how git remotes are handled have made it pretty much un-transferrable to the new code base. I'm sorry. We should have merged this sooner. I'm going to close this out (with my apologies for letting it rot like this), but I will say that the entire cache portion of the code base will probably get rewritten at some point this year (it's on [the roadmap](https://github.com/npm/npm/wiki/Roadmap)!), at which point we can revisit how to reduce how much work git does to stay current. 

Again, my apologies for letting this get this far from master.

guybrush:
@othiym23 no worries! you are doing a superb job, thanks for that!

PR #4791, opened 2014-02-27
Added support for install --development

davglass:
This basically installs the packages that `npm prune --production` would remove.

Use case:

In a CI environment, I need to:
- install production dependencies
- analyze them and their size
- display any warnings or helpful information
- install the devDependencies
- run their tests
- prune --production
- package for deployment

Now, I could do a `--production` install, then install without but I actually want to "switch" registries in between these. I only want "production" dependencies installed from a private local registry, but devDependencies can come from anywhere as they will be deleted with a prune. If a production dependency isn't in my local registry, it should 404 but doing a two pass install without scoping it would let it install anyway and then `prune` wouldn't work as I would expect.

I couldn't get all the tests to run on any of my dev machines from master or this branch, but I did add 7 passing tests for this functionality:

```
 ./node_modules/.bin/tap ./test/tap/install-development.js ./test/tap/install-production.js 
ok ./test/tap/install-development.js .................... 4/4
ok ./test/tap/install-production.js ..................... 3/3
total ................................................... 7/7
ok
```

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -249,6 +249,12 @@ function readDependencies (context, where, opts, cb) {
     if (er && er.code === "ENOENT") er.code = "ENOPACKAGEJSON"
     if (er)  return cb(er)
 
+    if (npm.config.get("development")) {
+      data.dependencies = {}
+      //Reset after the first pass so devDependencies have dependencies
+      npm.config.set("development", false)
+    }
+
     if (opts && opts.dev) {
       if (!data.dependencies) data.dependencies = {}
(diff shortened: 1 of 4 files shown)


bengl:
How about `--only [dependencies|devDependencies|...]`?

(Yeah it duplicates `--production`, but that could be deprecated in the future.)

othiym23:
I'm going to go ahead and close this in favor of #9024, which @bengl has helpfully put together for us. I still think getting some version of this landed is a good idea!

PR #4887, opened 2014-03-16
Enable overriding node-gyp binary

ackalker:
Enable overriding the node-gyp binary used by npm by specifying a config
option, similar to specifying the python interpreter.

Example usage:

```
npm install --node-gyp=<path-to-custom-node-gyp>
```


othiym23:
Rebased and landed as 145af6587f45de135cc876be2027ed818ed4ca6a. I created #7479 to capture the need to document this new option, which will probably involve some monkeying with the config tests because the value isn't used within npm itself. This is a topical change, given that it allows people to use unpatched npm with new versions of io.js by pointing at a patched `node-gyp`. Thanks for putting this together, and thanks for your patience!

ackalker:
You're welcome :-)

fengmk2:
Great job! We can use pangyp now!

PR #4901, opened 2014-03-18
Added client certificate parameters to tar download during an npm install

gziskind:
Added parameters to pass client certificate information down to the https request when a npm attachment is fetched. 

Without this change, the HTTP get for the npm module attachment fails when the NPM repository mirror you are pulling from requires client certificates that are passed via npm config settings.


othiym23:
This has been rendered redundant by changes to `npm-registry-client` since you put this patch together (it now includes certificate auth, if available, for all SSL requests), but thank you for taking the time to put this together! Sorry for not reviewing this sooner!

PR #5005, opened 2014-04-04
Add message about cleaning cache upon error

danielkcz:
Full credits goes to http://blogs.msdn.com/b/matt-harrington/archive/2012/02/23/how-to-fix-node-js-npm-permission-problems.aspx

There is plenty issues floating around in repos of other packages. Unfortunately it seems that not many people are reading [Troubleshooting](https://github.com/npm/npm/wiki/Troubleshooting) which mentions it. I wasn't looking there too honestly.


othiym23:
Closing as resolved, for the reasons stated above. Newer versions of io.js, at least, deal with this problem by creating the cache directory properly. Thanks for sending this patch npm's way, though!

PR #5347, opened 2014-05-26
Use sudo for installing in Unix.

g8d3:


othiym23:
I see what you're getting at here, and maybe there's a case to be made for adding a brief discussion of when you might or might not need to use sudo to install / use npm to the README, but just flipping it from not including `sudo` to including it doesn't actually reduce overall confusion, because of how many npm users are on OS X and using Homebrew, where using sudo will just lead to more confusion down the road. Thanks for submitting the patch, though!

PR #5354, opened 2014-05-27
shrinkwrap should not fail if errors are only in dev-dependencies

refack:


othiym23:
In the absence of a test, and given the changes coming to `npm shrinkwrap` in the release of `npm@3` (which is coming soon), it doesn't make sense to merge this patch. Thanks for putting it together, and sorry it took us so long to get back to review this!

PR #5379, opened 2014-05-30
Expect sanity from Linux in  `lifecycle-signal-test`

mmalecki:
Do not expect Linux to return 1 as error code and no signal indication
when child is killed by SIGSEGV.

I'm sure there was a reason behind this, so just throwing it out here for a review.


othiym23:
LGTM. Thanks for passing this along, Maciej! Landed as 37038d7db47a986001f77ac17b3e164000fc8ff3.

PR #5518, opened 2014-06-20
Passing args into run-scripts

bcoe:
this is the start of an approach for passing arguments into run-scripts that @isaacs and I have been talking about. It allows you to run:

`npm run-script start -- --foo=3`

The problem is that it breaks the ability to run a script in the context of a package, e.g., `npm start <package> <package> <package>`, `npm run-script start <package>`.

We will still have the ability to run commands in the context of another package, with with the syntax:

`npm explore <pkg> -- run test`, which seems like a reasonable compromise

thoughts

See #3494

### DIFF
--- a/doc/cli/npm-restart.md
+++ b/doc/cli/npm-restart.md
@@ -3,7 +3,7 @@ npm-restart(1) -- Start a package
 
 ## SYNOPSIS
 
-    npm restart <name>
+    npm restart [-- <args>]
 
 ## DESCRIPTION
 
--- a/doc/cli/npm-run-script.md
+++ b/doc/cli/npm-run-script.md
@@ -3,8 +3,8 @@ npm-run-script(1) -- Run arbitrary package scripts
 
 ## SYNOPSIS
 
-    npm run-script [<pkg>] [command]
-    npm run [<pkg>] [command]
+    npm run-script [command] [-- <args>]
+    npm run [command] [-- <args>]
 
 ## DESCRIPTION
 
--- a/doc/cli/npm-start.md
+++ b/doc/cli/npm-start.md
@@ -3,7 +3,7 @@ npm-start(1) -- Start a package
 
 ## SYNOPSIS
 
-    npm start <name>
+    npm start [-- <args>]
 
 ## DESCRIPTION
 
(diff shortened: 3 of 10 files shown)


machineghost:
I just wanted to +1 (:+1: )  the evils of the extra `--`.  I lost about half an hour trying the natural/logical syntax (ie. `npm test --foo`), before finally giving up and Googling my way to this thread.

In general well-designed software strives to make its users refer to the manual as infrequently as possible. But with the current extra `--` syntax a user has to either:

* stumble on to this thread, or
* `npm help package.json`, then
  * `npm help scripts`, then
  * `npm help run-scripts`
  * (wasting the time to read through the non-helpful parts of those first two before you can even get to the third)

Removing the `--` requirement will make the tool much more friendly to non-Node-experts, by making `npm test --foo` "just work."  Kudos to anyone who can find the time to implement it.

ashtonwar:
For anyone who doesn't like the extra `--` I wrote a [small utility](https://github.com/ashtonwar/npm-run-improved).

```sh
$ npmr command -vf # npm run command -- -vf
```

`npm install -g npm-run-improved`

stefanmaric:
Maybe I'm a bit late.

> We are not going to support passing args into the middle of the script, sorry.

I was looking to create parameterized npm scripts that interpolate args in arbitrary places, not only at the end. I came to a quick&dirty solution that I haven't seen anywhere else so I'm posting it here for future readers.

You just need to wrap your script in an immediately invoked function:

```json
{
  "print": "f(){ echo Hello $1! && echo Bye $2! ;};f",
  "test": "npm run print -- world stranger"
}
```

You get:

```shell
$ npm test -s # -s is to suppress npm messages for this example
Hello world!
Bye stranger!
```

You can also use `$*` which refers to all arguments.

This of course won't work on Windows unless you're using the Windows Subsystem for Linux (WSL, aka Bash for Windows), but since it is portable `sh` it should work everywhere else.

I would use this for dead-simple scripts only. Otherwise, you're better set with a separated `node` or `sh` script file as @isaacs said.

kolya-ay:
@stefanmaric, great hack! Exactly what I looked for, thank you

PR #5610, opened 2014-07-02
Failing test for multiple registry cache bug

Raynos:
There is a bug with how the cache in npm works.

If you use multiple registries it will sometimes install
a module from the cache but that module would be wrong
because it installed a module from the cache with a different
registry.

What we expect is that when we do

`npm install --registry=foo module@version`

That the package.json in `node_modules/module/package.json`
would contain the package.json from the foo registry.

i.e. we expect `_resolved` to contain the `foo` registry.

If it does not then we get a cascading bug where if you
run `npm shrinkwrap` it will put the wrong `"resolved"`
field in the shrinkwrap, i.e. a `"resolved"` field pointing
to a different registry then what you expected

cc @othiym23

### DIFF
--- a/test/common-tap.js
+++ b/test/common-tap.js
@@ -7,6 +7,10 @@ process.env.npm_config_loglevel = "error"
 var bin = exports.bin = require.resolve("../bin/npm-cli.js")
 var once = require("once")
 exports.npm = function (cmd, opts, cb) {
+  if (typeof cb !== 'function') {
+    throw new Error('must call common.npm(arr, opts, func)')
+  }
+
   cb = once(cb)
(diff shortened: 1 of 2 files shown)


othiym23:
This has been moved to the [npm roadmap](https://github.com/npm/npm/wiki/Roadmap), which we're using instead of the confusing `next-*` labels now. Also, this PR is pretty old and stale now (and only includes a test), so I'm going to close it out. The good news is that getting this right is an important part of the cache rewrite, which is a project I'd like to see us tackling before the end of the year. Thanks for putting this together, Jake! We'll probably cherry-pick out the test when we get around to doing this work!

PR #5621, opened 2014-07-03
Check git config to see if https:// should be used instead of git://

### LINKED ISSUE #5257: Can't force git to use https:// instead of git://
Git has a simple way to do this in the config file.
```
[url "https://"]
	insteadOf = git://
```

However npm seems to throw out the environment in lib/cache.js
```
var gitEnv_
function gitEnv () {
  // git responds to env vars in some weird ways in post-receive hooks
  // so don't carry those along.
  if (gitEnv_) return gitEnv_
  gitEnv_ = {}
  for (var k in process.env) {
    if (!~['GIT_PROXY_COMMAND','GIT_SSH','GIT_SSL_NO_VERIFY'].indexOf(k) && k.match(/^GIT/)) continue
    gitEnv_[k] = process.env[k]
  }
  return gitEnv_
}
```
So I can't inject a GIT_CONFIG or anything.

I've hand-hacked it into the `addRemoteGit` function in `lib/cache.js` with
```
  u = u.replace(/^git:/, "https:")
```
but this variable should be pulled from a config somewhere

ceeeKay:
Should fix #5257 and related issues for users behind firewalls blocking git://
as they should have the url.https://.ineadof option configured in their git config already

I've tested this change as part of building atom.io and confirmed it works for me.

### DIFF
--- a/lib/cache/add-remote-git.js
+++ b/lib/cache/add-remote-git.js
@@ -58,6 +58,19 @@ module.exports = function addRemoteGit (u, silent, cb_) {
     u = u.replace(/^ssh:\/\//, "")
   }
 
+  // See if git is configured to use https:// instead of git://
+  var git = npm.config.get("git")
+  var args = [ "config", "--get", "url.https://.insteadof" ]
+  var env = gitEnv()
+  exec(git, args, {env: env}, function (er, stdout, stderr) {
+    stdoutTrimmed = (stdout + "\n" + stderr).trim()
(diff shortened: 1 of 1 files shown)


othiym23:
After reviewing the patch and thinking about it a little, I've decided this is an unnecessarily complicated solution to the problem. See https://github.com/npm/npm/issues/5257#issuecomment-76371401 on #5257 for further detail. Thanks for bringing this issue to our attention, thanks for putting together this patch, and I'm sorry it took us so long to get around to providing this feedback!

PR #5629, opened 2014-07-04
local path support in package.json

dylang:
Adds support for `package.json` to have local paths, such as relative paths to packages which is helpful for testing. Also add support for `--save`, `--save-dev` to save local paths to `package.json`.

This replaces #5570.

As [recommended](https://github.com/npm/npm/pull/5570#issuecomment-47494553):
- The prepublish validation has been removed.
- Tests added. All tests should pass when couchdb is installed.
- Short circuit checks file path before assuming github shortcut for both command line and package.json.

fixes #2442
fixes #4018

and possibly others.


clayallsopp:
hey folks, not super familiar with npm's implementation, but is there a strong reason not to use symlinks for local paths? (or at least have an option for it, such as `symfile://` etc)

our use case is we're usually iterating on modules simultaneously, and it's a bummer to `npm install` every time we want to get the cross-module code changes to appear

(there are external tools that manage this, such as one linked earlier, but might be nice to have built-in support)

othiym23:
@clayallsopp You may want to take a look at [`npm link`](https://docs.npmjs.com/cli/link).

clayallsopp:
@othiym23 yup am aware of that, but I think it's suboptimal for our kind of workflow.

I think the ideal API is a declarative way of doing this in package.json:

```
{
  "dependencies": {
    "lib-core" : "symlink://../lib-core",
    "lib-api"   : "symlink://../lib-api"
  }
}
```

this seams reasonably close to the `file://`API, except that it creates symlinks instead of copies

using `npm link` is currently done inside a `preinstall` script, but it's a bit cludgy (i.e. iterate through all the folders, recursively figure out where the links need to be made, etc)

(not sure if this is a use case you'd want to address and totally understand if it's not)

tnrich:
@clayallsopp , I think what you're suggesting is a great idea. I'd like to develop modules locally, but dislike having to constantly re-install them. Using npm link via a pre-install is indeed very clunky.

tnrich:
I just found the npm package linklocal which auto-symlinks locally installed npm modules and does just what I'm looking for. Thanks @timoxley for that!

PR #5886, opened 2014-08-06
Add --no-verify

Raynos:
`npm version` should not fail if someone has a "funny" pre commit hook.

### DIFF
--- a/lib/version.js
+++ b/lib/version.js
@@ -97,7 +97,7 @@ function checkGit (data, cb) {
         , flag = sign ? "-sm" : "-am"
       chain
         ( [ git.chainableExec([ "add", "package.json" ], {env: process.env})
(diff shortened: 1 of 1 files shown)


othiym23:
I'm in the process of cleaning up the backlog of open patches against npm, and this test has been sitting open, waiting for a test, since last August. I'm going to close it out for now, and if you feel like reopening it with a test later, I'd be happy to merge it. Thanks for putting it together in the first place, Jake!

fyockm:
I'm running into the same problem. I've been using https://github.com/nlf/precommit-hook to run tests prior to allowing commit, however I don't want this to happen for `npm version` because: time.

@Raynos did you find a workaround for this problem?

@othiym23 i'd be happy to write a test, but not sure what kind of test you want? one that fails if there's a bad pre-commit hook? one that will pass with @Raynos change?

/cc @nlf

PR #5962, opened 2014-08-15
Fix `npm dedupe` for private repositories

jandre:
Fixes an issue where `npm dedupe` does not de-duplicate any private git
repositories.

Consider a tree like this:

```
- proj/
    package.json
    node_modules/
      proj2/
        node_modules/
          private_dep/
      private_dep/
```

Where:
- 'proj/package.json' has a dependency to 'private_dep' that is installed
  with path "git+ssh://git@github.com:acct/private_dep.git#0.1.0", e.g.:
  
  "dependencies": {
   "private_dep": "git+ssh://git@github.com:acct/private_dep.git#0.1.0"
  }
- proj2, which is a dependency of proj, also has a dependency to
  private_dep at the same version and path (with the exact same config
  in its package.json)

Technically 'private_dep' is the same version, but npm dedupe uses the
version in private_dep/package.json (e.g. 0.1.0) to do the resolution
rather than the path.

This code will use the git paths in node_modules/ package.json and
the calculated '_from' field to do the comparisons:

  "_from":  "git+ssh://git@github.com:acct/private_dep.git#0.1.0"

*\* PLEASE VERIFY I AM DOING THIS IN A LOGICAL WAY, NOT SURE IF I HAVE UNTANGLED THE RUBE GOLDBERG MACHINE CORRECTLY **

### DIFF
--- a/lib/dedupe.js
+++ b/lib/dedupe.js
@@ -86,7 +86,7 @@ function dedupe_ (dir, filter, unavoidable, dryrun, silent, cb) {
         return d && d.parent && !d.parent.duplicate && !d.unavoidable
       }).map(function M (d) {
         return [d.path, d.version, d.dependents.map(function (k) {
-          return [k.path, k.version, k.dependencies[d.name] || ""]
+          return [k.path, k.version, k.dependencies[d.name] || "", d._from, d.private]
         })]
       })]
     }).map(function (item) {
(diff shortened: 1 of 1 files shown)


othiym23:
I'm clearing out the open patches against npm, and as there's still no test on this patch, and `npm dedupe`'s behavior is changing pretty dramatically in the upcoming major release of npm (currently living on the `multi-stage` branch), we can't merge this change. Again, though, thanks for taking the time and care to put it together in the first place.

PR #5970, opened 2014-08-17
Undo commits when git tag fails on npm version.

sorribas:
Sometimes I do `npm version something` and right away I realize I forgot something. So I run `git reset --hard HEAD^`, I do whatever I wanted to do, commit, and then run `npm version something` again. But I sometimes forget to remove the tag that `npm version` created, so npm fails because the tag already exists. 

That's perfectly fine, but the thing is that even though npm failed to create the tag, it still creates the commit and changed the package.json file. I think it should rollback the entire operation if it fails to create the tag and that's what this PR does.

### DIFF
--- a/lib/version.js
+++ b/lib/version.js
@@ -105,7 +105,14 @@ function checkGit (data, cb) {
 
           , git.chainableExec([ "tag", "v" + data.version, flag, message ]
             , {env: process.env}) ]
-        , cb )
+        , checkTag )
+      function checkTag(err) {
+        var re = /tag \'.*\' already/g
+        var args = arguments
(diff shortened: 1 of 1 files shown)


othiym23:
@sorribas, are you still interested in updating this patch to work the way that @iarna describes? We're trying to get all of the open pull requests on npm merged or closed, and so we'll close this issue in a week or so if you aren't interested in fixing it up.

othiym23:
I'm going to close this for now. If you want to give it another shot later, file a new PR and we can pick up the thread again. Thanks for bringing things as far as you did!

PR #6208, opened 2014-09-17
- Improved "outdated" command to update remote tarballs when the remote content changes

lmartorella:
Hi,
We are working with some remote tarballs, referred using its direct http/https URL in the dependency/devDependencies list.
We found that npm caches the content of that tarball using the URL as unique key.
This is absolutely fine for "tagged" release that never changes. However if we want to support a snapshot-like based deployment (in our continous-integration environment), we would like to release "-SNAPSHOT" remote tarballs, where a CI build machine pushes new content in that tarballs at each build.

Simply using "npm cache clean " doesn't work, since the command cleans the local cache, but not the "node_modules" local folder.

So I've tried to enhance npm with the minimum set of changes, in order to make the "outdated" command to actively check the remote repos.
Changes:
- the local 'package.json' is enriched with a new property (_remote) containing some useful information about the GET response headers (currently etag and modification date), with the original URL as well.
- added new file 'check-remote-tarballs' with the logic to issue a HEAD request
- added new file 'remote-options' with the logic to issue the HEAD request with the right headers, in order to stimulate the HTTP 304 (Not Modified) response.

As a side note, I've found that the "add-remote-tarball" actually uses the npm registry API to issue the GET command for a remote tarball. Is that intended? The npm registry module adds logic to support registry authentication, ca, session, etc... that are actually useless if the remote tarball is on a non-registry machine.
In addition, we needed to patch our IIS server to server .tgz files with the non-standard MIME type "application/x-tar" (explicitly required by npm registry module).
The added HEAD support cannot use the npm registry module (unless we add the HEAD support there), so I've directly used the 'request' dependency.

Thank you for this great tool!
Luciano


lib/cache/check-remote-tarball.js
```diff
@@ -0,0 +1,83 @@
...
+        log.info("OPTIONS fetch failed", url)
+        return cb(er, false)
+      }
+      // Check against etag
+      cb(null, sc == 200) // Otherwise a 304 (Not Modified) will be returned
+    })
+  })
+}
+
+function makeOptionsCall(remote, opts, cb) {
```
othiym23: Docs say `OPTIONS`, code says `HEAD`.

othiym23:
I've spent the last half hour or so looking at this PR and thinking, and I'm still not comfortable with this change. Let me see if I can restate in my own words the problem you're trying to solve: 

> We have a CI system that produces snapshot builds of a dependency on a regular basis. We provide these snapshots as bare package tarballs, and we need npm to notice when they've changed, and to pull down the changed tarballs.

So the goal is for the caching code to notice when a tarball dependency has changed and pull it down. There are two things going on here that seem weird to me:

1. We have since landed npm/npm-registry-client@07bc33502b93554cd7539bfcce37d6e2d5404cd0, which adds `if-modified-since` header checking to `npm-registry-client`, which does most of what you're asking for, including saving that information to the cache for later use. I completely forgot about this PR when I landed that one, but I think this addresses your problem, at least as I've stated it above.
2. This code shouldn't be in the `outdated` / `update` code path at all. This is something that should be handled exclusively by the cache and the installer (in fact, in `npm@3`, which is getting perilously close to beta, `outdated` and `update` use much more or the installer code). I think when I was first reviewing this patch, I was confused about what it does because of where the changes had been made. `npm outdated` is mostly meant to do checks on the versions (the symbolic versions, i.e. the semver, not the timestamps) of packages on the registry. It's not a general-purpose set of routines for checking the liveness of dependencies.

Either way, this patch has gone stale, which is my fault and for which I apologize. I'm going to close it, and if you think I've inaccurately understood the problem you were trying to solve, or believe that important aspects of it remain unsolved, please do file a new pull request or issue to that effect. Thanks for putting the time into this, and for your patience! I'm sorry I didn't get back to you sooner.

PR #6239, opened 2014-09-19
Improving error output

NickHeiner:
It would be helpful for me at least if the peer dependency error message contained a little more information.


iarna:
LGTM, we'll merge this into `2.x`. `3.x` already has something equivalent.

iarna:
Landed in [`2.x`](https://github.com/npm/npm/tree/2.x) as https://github.com/npm/npm/commit/7827249b0999f560ce41d9d6b0f50b113c0bb9f7

PR #6411, opened 2014-10-06
Simple fix to address bug #6325.  Added a limit to number of concurrent git commands

rvermillion:
This is a simple patch that limits the number of git commands that can run at once. Default is 10 or the value of the "git-max-concurrent" config property.  This will resolve the bug where "npm install" will exceed the maximum connections of the SSH server and give the "ssh_exchange_identification: read: Connection reset by peer" error.

### DIFF
--- a/lib/utils/git.js
+++ b/lib/utils/git.js
@@ -17,9 +17,38 @@ function prefixGitArgs() {
   return process.platform === "win32" ? ["-c", "core.longpaths=true"] : []
 }
 
+var execCount = 0,
+    maxConcurrent = +(npm.config.get("git-max-concurrent") || 10),
+    execQueue = []
+
 function execGit(args, options, cb) {
   log.info("git", args)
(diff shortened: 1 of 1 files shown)


othiym23:
Between this issue and #4553, there are now two proposals for how to serialize various bits of npm's I/O. I feel like these are kind of just rearranging the deck chairs on a boat coming up hard on an iceberg. If we're going to try and keep npm from starving its host system or its network connection for resources, I think we probably need to do it more holistically than just constraining one piece of the caching or installation logic. npm has very simpleminded control flow internally, and this is part of what makes it so fast. If given a choice between making it more sophisticated and making it slower, I'm going to go with making it more sophisticated. I appreciate the work you put into putting this together, but I think we probably need something more than the approach in this PR. Thanks for your time, and I'm definitely open to discussing how we might make npm better at using resources efficiently without being greedy.

PR #6423, opened 2014-10-07
added er.parent logging to ETARGET error

davglass:
Just adding in `er.parent` similar to the `E404` error so that you can track down version issues when installing modules.

I didn't see any tests dealing with the logging so I didn't add one. All tests pass locally tho :smile:


othiym23:
Landed in 9748d5cd195d0269b32caf45129a93d29359a796. Thanks for your patience, Dav! And thanks for putting this together in the first place!

davglass:
:metal:

PR #6505, opened 2014-10-17
install peerDependencies from top

othiym23:
The simplest possible change to ensure that peerDependencies get installed properly – be sure to install them at the top level if they're included at the top level. Fixes #6323.


balupton:
Seems this broke:

```
v1.4.11 (2014-05-22):
Allow peer deps to be satisfied by grandparent
```

PR #6534, opened 2014-10-21
remove the user-scaring anti-pattern warning

zeke:
Not all usage of npm scripts is anti-patterny, but the [existing documentation](https://www.npmjs.org/doc/misc/npm-scripts.html#note-install-scripts-are-an-antipattern) makes it seem that way. Everybody uses npm scripts, and as such we should not make them sounds like a bad idea. Yes, there is something to be said for discouraging people from misusing these hooks, but it's not worth scaring people away and making them feel like they shouldn't be using scripts at all.

This change removes that ANTIPATTERN warning and the ramblings that follow it about the good old days of node-waf and other things that only a tiny sliver of npm users care about. There's now a "Common Uses" heading now which continues to advocate the use of `prepublish` as before.

cc @seldo

### DIFF
--- a/doc/misc/npm-scripts.md
+++ b/doc/misc/npm-scripts.md
@@ -38,46 +38,24 @@ run-script <pkg> <stage>`. *Pre* and *post* commands with matching
 names will be run for those as well (e.g. `premyscript`, `myscript`,
 `postmyscript`).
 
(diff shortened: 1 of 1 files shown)


othiym23:
Sure, why not. Landed as 8ac6f2123a6af13dc9447fad96ec9cb583c45a71. Thanks, Zeke.

zeke:
:tada:

Issue #6248, opened 2014-09-20
`npm link` and `npm unlink` should be mirror operations, and `npm unlink` != `npm uninstall`


othiym23:
See #4005 for how the confusion this causes has caused a little forest of forlorn 👍s to sprout up.

othiym23:
Whee.... finally landed this in 7bf85cc372ab5698593b01e139c383fa62c92516.

adamski:
Wish I'd known about this before... just lost a few hours work after realising I was working on my node_modules folder rather than the linked module, did npm link, copied the work over only to realise they are then identical, did npm unlink to remove the link, which then removed the module from my node_modules folder along with all my work... boo hoo :(

**EDIT** saved by AppCode's local history feature, phew!

wearhere:
Follow-up issue: https://github.com/npm/npm/issues/20995 —`npm link` and `npm unlink` are not yet total mirror operations.

PR #6535, opened 2014-10-22
npm rm <noargs> should globally remove the local pkg

### LINKED ISSUE #4005: npm unlink does not unlink from global path
Steps to reproduce:

1. checkout a node project
1. cd <said project>
1. npm link
1. npm unlink

The expected result is to have the project no longer linked globally but alas, unlink says that the project is not installed under its own node_modules folder (and yes, indeed it is not) and doesn't unlink it globally.

othiym23:
(Creating this as a PR on isaac's behalf because there's no real reason not to land it. Needs a test and review but is otherwise good to go.)

Fix #4005 #6248

What's interesting is that the comment in the code seems to indicate
that this was _always_ the intent.  But somewhere along the line, that
seems to have broken.

This makes 'npm link' be un-done by running 'npm unlink'.

### DIFF
--- a/lib/uninstall.js
+++ b/lib/uninstall.js
@@ -28,12 +28,11 @@ function uninstall (args, cb) {
   if (args.length) return uninstall_(args, nm, cb)
 
   // remove this package from the global space, if it's installed there
-  if (npm.config.get("global")) return cb(uninstall.usage)
-  readJson(path.resolve(npm.prefix, "package.json"), function (er, pkg) {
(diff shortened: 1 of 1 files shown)


othiym23:
Landed with minor tweaks in 7bf85cc372ab5698593b01e139c383fa62c92516, finally. Thanks for fixing this way back when, @isaacs. Thanks, me, for finally getting around to merging it!

Issue #6582, opened 2014-10-28
Please give an error when files don't exist


petkaantonov:
I have defined a `files` field in my `package.json` and I don't think npm should publish a broken version when any of those files are missing.

othiym23:
This has been moved to the [npm roadmap](https://github.com/npm/npm/wiki/Roadmap), which we're using instead of the confusing `next-*` labels now.

zkat:
Hey @petkaantonov!

I had originally poked at a version of this change with #12191 (which was connected to #8791) during a Files and Ignores push we had earlier this year, but it got waylaid because it required some bigger changes to `fstream` with the way I was doing it, and other things ended up taking priority.

That said! The team talked it over today, and we agreed that this might be a good thing to have (done a bit differently than the original one -- most notably making it an *error*, as you requested here). This would make it a breaking change, but we don’t have a schedule for when we’re going to implement it right now.

Most likely, what we’ll end up doing is add something at the end of `pack` that checks the final tarball’s file list and errors if any single entry/glob in `files` fails to match at least once. I assume that would fulfill your request here :)

So yeah! We’re gonna leave this open, since we do intend to get to it ourselves ~~some time~~, but I can’t tell you exactly when it’ll land. Thanks for your time!

PR #6618, opened 2014-11-03
add test fixture for missing file

smikes:
Clone of test/packages/npm-test-files , with an entrry in `files` that does not exist.

Repro steps to verify #6582 

```
cd test/packages/npm-test-missing-files
node ../../.. pack .
```

I expect an error message and npm should not produce an output .tgz file.

Instead npm produces a tgz file which does not contain all the files specified in files[]

```
$ tar tzf npm-test-missing-files-1.2.5.tgz
package/package.json
package/.npmignore
package/include4
package/sub/include
package/sub/include2
package/sub/include4
package/test.sh
```


imshivs:
Can anyone point me to the new PR? I'm digging through the discussion that began in #6582.

smikes:
@imshivs I think I never created the new PR, actually.

PR #6659, opened 2014-11-06
Remove unused shell variable from ZSH completion

allait:
Since `$IFS` is not actually changed in the compadd version of ZSH completion, there's no need to save and restore it.

Right now calling `_npm_completion` leaves the global `$si` variable around.

### DIFF
--- a/lib/utils/completion.sh
+++ b/lib/utils/completion.sh
@@ -24,13 +24,11 @@ if type complete &>/dev/null; then
   complete -F _npm_completion npm
 elif type compdef &>/dev/null; then
   _npm_completion() {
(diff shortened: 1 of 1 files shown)


othiym23:
Because I wanted to preserve symmetry with the other place `si` is used, and because I don't know what side effects there might be on `$IFS` inside compctl, I fixed this problem the simple way, by declaring `si` as a local, which I landed as 3f6061d75650441ee690472d1fa9c8dd7a7b1b28. Thanks for raising the issue, though, and thanks for your time in putting together this patch! I'm sorry it took so long for us to get back to you.

PR #5252, opened 2014-05-11
Make npm-debug.log path configurable.

timoxley:
If this is acceptable, I'll send the appropriate PRs for the `npm-log` config in npmconf and add docs.

Allows for location of `npm-debug.log` to be specified:

```
> npm install asdasdas --log-file=/var/log/npm/myapp.log
npm WARN package.json npm-debug-log-test@1.2.3 No description
npm WARN package.json npm-debug-log-test@1.2.3 No repository field.
npm WARN package.json npm-debug-log-test@1.2.3 No README data
npm ERR! 404 404 Not Found: asdasdas
npm ERR! 404
npm ERR! 404 'asdasdas' is not in the npm registry.
npm ERR! 404 You should bug the author to publish it
npm ERR! 404 It was specified as a dependency of 'npm-debug-log-test'
npm ERR! 404
npm ERR! 404 Note that you can also install from a
npm ERR! 404 tarball, folder, or http url, or git url.

npm ERR! System Darwin 13.2.0
npm ERR! command "node" "/Users/timoxley/Projects/libs/npm/bin/npm-cli.js" "install" "asdasdas" "--log-file=/var/log/npm/myapp.log"
npm ERR! cwd /Users/timoxley/Projects/libs/npm/test/tap/npm-debug-log
npm ERR! node -v v0.10.24
npm ERR! npm -v 1.4.10
npm ERR! code E404
npm ERR!
npm ERR! Additional logging details can be found in:
npm ERR!     /var/log/npm/myapp.log
npm ERR! not ok code 0
> ls /var/log/npm
myapp.log
```
## Possible Improvements
- a mkdirp on the log directory
- if specify an existing directory, default to writing `npm-debug.log` in that directory.
- alias/rename to `logfile` vs `log-file`
- fallback if it can't write to log just write to default `process.cwd()/npm-debug.log`

Related https://github.com/npm/npm/issues/1548

Issue #1548, opened 2011-10-13
npm-debug.log could be written to /tmp or similar


kapouer:
Hi,
npm-debug.log sometimes get written in strange places (i even found one in /usr/bin).

It could be very nice to write it to a temp directory, like /tmp on linux, or env[TEMP] on windows,
and of course tell to the user where it is.

A second idea is to timestamp the file name, so that crashes are logged separately.

joshuabc:
Big +1 on adding a configuration option to disable this.

kloplop321:
I keep finding this in all the weirdest places too, and no one on my team remembers how the heck it is getting into places like /etc/php5/fpm/pool.d and such.

DavidSouther:
@kloplop321 that's a really weird place! Presumably, it'll only show up in places that `npm` has executed and encountered an issue. Perhaps there's an automated tool that sometimes works in that part of  your directory structure?

kloplop321:
Well, no one remembers executing it there, and we don't have much automated besides crash detection. Same for our app directories, which are not node based. Though we do have a node-based bot that handles callback URLs for git pushes and such.

jaythomas:
@LeviSchuck, npm creates the debug log in your current working directory, not where you executed it. The problem is your current working directory is not specific to the shell instance you ran npm in. So if you are using, say, a terminal multiplexer like gnu-screen/tmux... or even running a terminal emulator with multiple tabs of bash/zsh open, npm will be open in one screen but writing logs to your current working directory in another screen (ie you could be in "/etc/php5/fpm/pool.d" in your other screen.)

faiq:
npm-debug.log should also never be made when something is wrong with the package, and it clearly isn't npm's fault

monolithed:
`npm-debug.log` is a log file like `httpd.log`, `nginx.log` etc, and the location should be configurable (`npm_modules` as well)!

DavidSouther:
@monolithed I hesitantly disagree - `httpd.log` and `nginx.log` are for long-running daemon processes, which I use to either gather metrics or triage an error. `npm-debug.log` is only generated when the most recent NPM command has encountered a failure, and is highly contextually dependent on where it happened.

Perhaps a hybrid solution would be to delete npm-debug.log on the next successful NPM command?

[...]

PR #6744, opened 2014-11-19
Put npm-debug.log file in the cache folder, not cwd

othiym23:
New version of #6181: This should prevent mishaps like #1548, while avoiding overly configurable options that make it hard to predict where the file will show up.

Fix #1584
Close #5252

### DIFF
--- a/lib/utils/error-handler.js
+++ b/lib/utils/error-handler.js
@@ -14,6 +14,10 @@ var cbCalled = false
   , writeStream = require("fs-write-stream-atomic")
 
 
+function getLogFile () {
+  return path.resolve(npm.config.get("cache"), "npm-debug.log")
+}
+
 process.on("exit", function (code) {
   // console.error("exit", code)
   if (!npm.config || !npm.config.loaded) return
@@ -30,7 +34,7 @@ process.on("exit", function (code) {
 
       log.error("",
                 ["Please include the following file with any support request:"
-                ,"    " + path.resolve("npm-debug.log")
+                ,"    " + getLogFile()
                 ].join("\n"))
       wroteLogFile = false
     }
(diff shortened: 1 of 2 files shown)


watilde:
I got it about what will happen for `npm-debug.log`. It's sensitive about disk size, so I think it's good if set a limit of a size saved log in npm.config. (also other properties too..?

Apart from that, maybe one change has needed at [here](https://github.com/npm/npm/blob/debug-log-in-cache/lib/utils/error-handler.js#L84):
```diff
-  else rm("npm-debug.log", reallyExit)
+  else rm(getLogFile(), reallyExit)
```

othiym23:
This has been moved to the [npm roadmap](https://github.com/npm/npm/wiki/Roadmap), which we're using instead of the confusing `next-*` labels now. Also, @isaacs apparently decided it needed to be closed out when he had a little boo boo with his git branches a few weeks ago, so we'd need to redo it anyway if we wanted to land it. Either way, this is a supportability feature that should be tackled as such.

computersarecool:
So wait @othiym23 I am not sure what the outcome is here.

 The `npm-debug` file now gets put deep in this `../_logs` file correct? The location is not configurable - it has just been moved. Is this correct?

kenany:
@computersarecool Right, the path is currently not configurable.

blachawk:
Is this still the case?  any plans in setting up a custom path location for logs?  Sucks for me as this is buried in a hidden folder in windows called AppData.  Yeah I can get to it but I don't want to take a spaceship to the moon just get a better view of the sun either.

danielo515:
So log output is not configurable ? What a shame.
This is incredibly inconvenient for CI environments, where several CI engines are able to report you error and success logs located on certain directories.

Issue #6758, opened 2014-11-21
Option to always create .cmd shims for Windows, even when installing on Unix


plestik:
Hi,

I have a WIndows host - Unix guest setup. `npm install` is run on the Unix guest, which means no .cmd shims are created. But, I prefer to run `gulp` on my Windows host, which isn't possible without .cmd wrappers. Is it possible to force the creation of the cmd shims? Alternatively, is there a module for this purpose?

Perhaps expanding `bin-links` with `'both'` as an option (in addition to `true` and `false`?

othiym23:
Another way to get the same effect when moving between platforms is to run `npm rebuild` before running gulp. On Windows, it will create the `.cmd` shims, and on all other platforms, it will take care of (re)generating the necessary symbolic links to get the paths set up correctly to use binaries from package scripts. In fact, that’s close enough of a best practice that I’m going to close this feature request. If that’s not sufficient for your use case, open an issue or let the CLI team know what other requirements you have so we can talk it through. Thanks for your time!

buffcode:
`npm rebuild` will also try to rebuild dependencies (in our case a C# library/adapter), which won't be possible on production machines.
In a perfect world we can create a dist-package on our linux CI enviroment (including the shim files) and just copy it to a Windows production-environment for use.

PR #6777, opened 2014-11-24
New option: bin link type

dgieselaar:
This is an implementation of https://github.com/npm/npm/issues/6758. It adds a `bin-link-type` option:
### bin-link-type
- Default: `auto`
- Type: String

Tells npm what kind of bin links should be created. It defaults to `auto`, which means it is based on the environment. If set to `win`, `.cmd` shims will be created. If it's `unix`, only symlinks will be used. If it's `both`, well, both will be created. This is useful when you have a VM setup with a Windows host and a Unix guest, and want to run `npm install` on your guest and still have the `.cmd` shims available on your Windows host.


othiym23:
Sorry for letting this PR go stale, but for the reasons I stated above, I don't think this change makes sense to land in npm proper.

> writing a module which uses the cmd-shim library and provides a cli. The user can then execute the wrapper with a postinstall script.

This sounds like a good place to start – it would look at the desired state, and then either do nothing (if things are already correct) or replace the links with shims (or vice versa). It totally makes sense as a postinstall script. Good luck!

plestik:
Yeah, no problem. I've stumbled on other issues with this approach (there are more platform-specific things than just the bin links), so I changed directions. @smikes thanks for your help, learned a lot! You as well @othiym23.

PR #6793, opened 2014-11-25
add caching based on last-modified/if-modified-since headers

### LINKED ISSUE #6791: Add 'last-modified'/'if-modified-since' date-based caching.
The current ETag caching works well for the public registry; however, some nginx and other load-balancer and proxy implementations might have different behavior, and only respond with a 304 if an if-modified-since header is sent with the request.

The header should be created the same way the if-none-match header is currently created, but instead of the cached etag header, it should send the value of the last-modified header.

I believe most browsers currently operate in this fashion as well.

What do you think?

Also see https://github.com/npm/npm-registry-client/issues/86

lxe:
Fixes #6791

NOTE: the client tests are in https://github.com/npm/npm-registry-client/pull/87

### DIFF
--- a/lib/cache/caching-client.js
+++ b/lib/cache/caching-client.js
@@ -37,9 +37,9 @@ CachingRegistryClient.prototype._invalidatingRequest = function (uri, params, cb
       var invalidated = client._mapToCache(uri)
       // invalidate cache
       //
-      // This is irrelevant for commands that do etag caching, but ls and
-      // view also have a timed cache, so this keeps the user from thinking
-      // that it didn't work when it did.
+      // This is irrelevant for commands that do etag/lastModified caching,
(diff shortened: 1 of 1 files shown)


lxe:
@othiym23 This merely mimics the behavior that browsers do. Although `npmjs.org` uses etag to do proxying, many things like proxies, and private registry mirrors might use if-modified-since. This change should allow the npm actually send these headers (similarly to/alongside with etag), and save on a lot of unnecessary 200 responses.

I'm gonna try to get an integration test going.

lxe:
Travis is throwing unrelated errors

othiym23:
Landed (finally) as 98e1e1080df1f2cab16ed68035603950ea3d2d48. Thanks for persevering!

PR #6807, opened 2014-11-28
Rebased 2716 graceful restart

smikes:
Supersedes #2716 which became nonmergeable due to age
Includes @scien's change (unsquashed because I don't know how to squash and retain credit)
Formatting fix
Unit test

### DIFF
--- a/lib/run-script.js
+++ b/lib/run-script.js
@@ -116,7 +116,7 @@ function run (pkg, wd, cmd, args, cb) {
   if (!pkg.scripts) pkg.scripts = {}
 
   var cmds
-  if (cmd === "restart") {
+  if (!pkg.scripts.restart && cmd === "restart") {
     cmds = [
       "prestop", "stop", "poststop",
       "restart",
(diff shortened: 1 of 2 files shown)


smikes:
Rebased against current master and pushed.  This would still require a major version bump and documentation changes, so it is not a candidate for 2.2.1

othiym23:
I changed my mind. :grin: Because the existing behavior is out of sync with the docs, I decided that it was within scope of a patch-level change, and pulled it in now. Landed as d3df7c7847ac51e8b547e4568bdab5aea84fafc4.

I also merged @watilde's #7471 first (2f6a1df3e1e3e0a3bc4abb69e40f59a64204e7aa), so I mostly pulled in this for the test code. As a result, @scien, congratulations, you are now credited as the author of a test file that @smikes wrote and I brought up to snuff with @feross's `standard` module. What a world!

Thanks to everyone for their contributions to this effort! This is a weird little wart of npm's behavior that is now very-well tested and doing what it claimed to be doing all along.

PR #6849, opened 2014-12-04
doc/tag: candidate explanation of tag workflow

smikes:
Some suggestions re: tag workflow.  Open to major revision/bikeshedding.

There is a distinction between 'tag' (eg a git tag) which identifies a specific release (and is very closely related to 'version') and 'dist-tag' which identifies both a stream of releases and the most current version in that release, but I am having a hard time articulating it.

Is the name "tag" or "dist-tag" still open to bikeshedding or is it basically set in stone?

### DIFF
--- a/doc/api/npm-tag.md
+++ b/doc/api/npm-tag.md
@@ -18,6 +18,6 @@ is the package name and version is the version number (much like installing a
 specific version).
 
 The second element is the name of the tag to tag this version with. If this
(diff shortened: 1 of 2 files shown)


smikes:
Rebased and squashed.

othiym23:
Rebased (again) but otherwise landed as-is as 17f66ceb1bd421084e4ae82a6b66634a6e272929. Thanks! Don't freeze! :snowflake:

Issue #2556, opened 2012-06-22
npm install --link does not work with git:// dependency URLs


per-gron:
## Steps to reproduce
- Create a new directory and add this `package.json` file to it:
  
  `{
    "name": "npm-test2",
    "version": "1.0.0",
    "dependencies": {
      "npm-test": "git://github.com/pereckerdal/npm-test.git"
    }
  }`
- In that directory, invoke `npm install --link`
## Expected result

The `npm-test` package should be installed.
## Actual result

npm spits out an error `404 'npm-test' is not in the npm registry.`
## Additional notes

Invoking `npm install`, without the `--link` flag, works as expected.

The same problem exists for git+ssh:// URLs

Here's a gist with npm-debug.log and npm output: https://gist.github.com/2972216

DonRichards:
I also have the same issue. Trying to import a private gist file npm install git+ssh://gist@github.com:#######.git

PR #6978, opened 2014-12-20
candidate fix for #2556 `install --link` with git dependency goes to registry

smikes:
This is an attempt at a failing test for #2556 .  I say "attempt" because although I can get a failure on the command line with this setup, I can't make it fail programmatically.

When run on the command-line, `installOne` is going into the `localLink` branch, which causes the failure.

However in the attached unit test, `installOne` just succeeds without going into the link branch.

Output of a successful test run and a failed command-line run in this gist: https://gist.github.com/smikes/daf021d1eb588536bad5

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -854,8 +854,10 @@ function targetResolver (where, context, deps) {
 function installOne (target, where, context, cb) {
   // the --link flag makes this a "link" command if it's at the
   // the top level.
+  var isGit = npa(target._resolved).type === "git"
+
   if (where === npm.prefix && npm.config.get("link")
-      && !npm.config.get("global")) {
+      && !npm.config.get("global") && !isGit) {
     return localLink(target, where, context, cb)
   }
   installOne_(target, where, context, function (er, installedWhat) {
--- a/test/tap/git-dependency-install-link.js
+++ b/test/tap/git-dependency-install-link.js
@@ -1,0 +1,184 @@
+var fs = require('fs')
+var resolve = require('path').resolve
+
+var chain = require('slide').chain
+var osenv = require('osenv')
+var mkdirp = require('mkdirp')
+var rimraf = require('rimraf')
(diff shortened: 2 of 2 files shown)


smikes:
Blocking: Sam needs to revisit this test, possibly add `prefix`, see if it would work to use same-process `npm.commands.foo` test instead of `common.npm([...])` subprocess test

Blocking: fix for underlying bug

smikes:
Unblocked and ready for review.

othiym23:
@iarna, I'd like your eyes on this one. The fix feels a little like a proximate fix rather something that addresses the underlying issue, but it works, according to this excellent test that @smikes wrote. This also looks like cheese that will need to be moved soon.

iarna:
Yeah, I'll see how this interacts with multi-stage, since ALL THE CHEESE related to this moved.

iarna:
A HAND MOVING ALL YOUR CHEESE – FOREVER

smikes:
A shining red dot that you can never catch, forever

smikes:
Even if the fix is junk, the test should be salvageable

othiym23:
Well, if it's a proximate fix, I can still land it in `npm@2`, and @iarna can figure out how to land it (or not) for `npm@3` (and bring over the test, either way). I'll work on this now.

othiym23:
Good news! I landed this! I had to make some tweaks, though – a few tests were failing due to the change's assumption that `_resolved` would always be set by that point in the installation process (in the case of regular semver ranges or `dist-tag`ged packages, it might not be). I switched it to use `_from`, and added a guard that would prevent it from exploding in those cases. It still feels a little hacky, but that's fine, because the affected code is gonna die soon anyway. Landed as c618eeda3e321fd454d77c476b53a0330f2344cc and 39032e745ef6f610f489d429292df350d556ccc7. Thanks, Sam!

PR #6999, opened 2014-12-24
add smart `env` script if there is none present

gcb:
adding a "env":"env" run-script to my package every time i want to double check some environment var, and then removing to commit my changes, always feel awkward... but now that is a thing of the past! People from the future will be able to look at kids, raise their trembling canes, and say that in their time they had to add a 'env' rule to check their environment variables and that kids have it too easy today. I can't wait to have it too easy.

this little patch adds the ability to run `npm env` even if there is no such rule added to the package. and it also works across all platforms. (caveat, only tested on linux and windows7, but osx and others should also work with the linux rule)

PS: also removed a few `;` from the original code to conform to style guide.

### DIFF
--- a/lib/run-script.js
+++ b/lib/run-script.js
@@ -127,10 +127,12 @@ function run (pkg, wd, cmd, args, cb) {
       if (cmd === "test") {
         pkg.scripts.test = "echo \"Error: no test specified\""
       } else if (cmd === "env") {
-        log.verbose("run-script using default multiplatform env")
+        log.verbose("run-script using default multiplatform env:")
         if (process.platform === "win32") {
+          log.verbose("SET")
           pkg.scripts[cmd] = "SET"
         } else {
+          log.verbose("env")
           pkg.scripts[cmd] = "env"
         }
       } else {
--- a/lib/run-script.js
+++ b/lib/run-script.js
@@ -126,6 +126,9 @@ function run (pkg, wd, cmd, args, cb) {
     if (!pkg.scripts[cmd]) {
       if (cmd === "test") {
         pkg.scripts.test = "echo \"Error: no test specified\"";
+      } else if (cmd === "env") {
+        log.verbose("run-script adding default multiplaform env")
+        pkg.scripts[cmd] = "env" // TODO: add windows
       } else {
         return cb(new Error("missing script: " + cmd));
       }
--- a/lib/run-script.js
+++ b/lib/run-script.js
@@ -127,8 +127,12 @@ function run (pkg, wd, cmd, args, cb) {
       if (cmd === "test") {
         pkg.scripts.test = "echo \"Error: no test specified\""
       } else if (cmd === "env") {
-        log.verbose("run-script adding default multiplaform env")
-        pkg.scripts[cmd] = "env" // TODO: add windows
+        log.verbose("run-script using default multiplaform env")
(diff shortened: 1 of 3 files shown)


lib/run-script.js
```diff
@@ -125,9 +125,16 @@ function run (pkg, wd, cmd, args, cb) {
   } else {
     if (!pkg.scripts[cmd]) {
       if (cmd === "test") {
-        pkg.scripts.test = "echo \"Error: no test specified\"";
+        pkg.scripts.test = "echo \"Error: no test specified\""
+      } else if (cmd === "env") {
+        log.verbose("run-script using default multiplaform env")
```
othiym23: `s/multiplaform/multiplatform/`

lib/run-script.js
```diff
@@ -125,9 +125,16 @@ function run (pkg, wd, cmd, args, cb) {
   } else {
     if (!pkg.scripts[cmd]) {
       if (cmd === "test") {
-        pkg.scripts.test = "echo \"Error: no test specified\"";
+        pkg.scripts.test = "echo \"Error: no test specified\""
```
othiym23: thanks!

lib/run-script.js
```diff
@@ -125,9 +125,16 @@ function run (pkg, wd, cmd, args, cb) {
...
+      } else if (cmd === "env") {
+        log.verbose("run-script using default multiplaform env")
+        if (process.platform === "win32") {
+          pkg.scripts[cmd] = "SET"
+        } else {
+          pkg.scripts[cmd] = "env"
+        }
       } else {
-        return cb(new Error("missing script: " + cmd));
+        return cb(new Error("missing script: " + cmd))
```
othiym23: thanks again!

test/tap/run-script.js
```diff
@@ -87,6 +87,15 @@ test('npm run-script test', function (t) {
   })
 })
 
+test('npm run-script env', function (t) {
+  common.npm(['run-script', 'env'], opts, function (er, code, stdout, stderr) {
+    if (er)
+      throw er
```
othiym23: npm style is to replace this with `t.ifError(er, "script ran successfully")`

othiym23:
OK! I'm convinced! Address the nits I identified and add some docs for this and I'll merge it in. I should also probably fix `npm explore .` one of these days.

gcb:
where can i find docs for ifError? not familiar with that on other TAPs...though not a heavy tap user/lover. For now i followed what i understood from source and other examples.

so, fixed the typos you found, added docs. also added a very simple positive test (not just testing for errors anymore)

thanks!

PS: for `explore`, i don't use it often (hardly ever) but i always assumed it to be a way to spawn a subshell in a package path, nothing more.

othiym23:
Actually posting this to the right repo, heh heh:

> OK, I re-landed my tweaks, rebased, and landed this as acf287d2547c8a0a8871652c164019261b666d55. Congratulations! This means that the next version of npm will be `npm@2.2.0`, and it's due entirely to this feature! Thank you for your patience (and contribution)!

rmg:
:+1:

faiq:
good job getting this in!

gcb:
awesome! thank you for all the help and for removing this minor pain point on my daily job :)
cheers!

PR #7044, opened 2015-01-02
add documentation about .npmignore pattern rules

zeke:
I set about using npmignore for the first time recently (https://github.com/npm/npm-expansions/pull/201) and was a little unsure about the file format, but I got an answer (https://github.com/npm/docs.npmjs.com/issues/93).

This patch just adds some documentation about how npmignore files are pretty much that same as gitignore files.


othiym23:
I tweaked your patch to make sure we were using Markdown list nubbins™ consistently throughout the file, but otherwise landed without changes. Thanks a ton! Now I have something to put in a release this week! :megaman:

PR #7046, opened 2015-01-02
remove references to Node Package Manager

linclark:
As pointed out in https://twitter.com/trevordmiller/status/550051232253882368, there is some inconsistency as far as the name goes. This removes usages of the deprecated name.

### DIFF
--- a/doc/api/npm.md
+++ b/doc/api/npm.md
@@ -1,5 +1,5 @@
-npm(3) -- node package manager
-==============================
+npm(3) -- javascript package manager
+====================================
 
 ## SYNOPSIS
 
--- a/doc/cli/npm.md
+++ b/doc/cli/npm.md
@@ -1,5 +1,5 @@
-npm(1) -- node package manager
-==============================
+npm(1) -- javascript package manager
+====================================
 
 ## SYNOPSIS
 
--- a/html/index.html
+++ b/html/index.html
@@ -52,7 +52,7 @@ code { background:#fff ; outline: 1px solid #ccc; padding:0 2px; }
 }
 
 </style>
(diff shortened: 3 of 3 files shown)


othiym23:
Tweaked the commit message a little bit, rebased, and landed as 226a6776a1a9e28570485623b8adc2ec4b041335. Thanks, @linclark! One of these days we'll have completely rewritten histWE HAVE ALWAYS BEEN AT WAR WITH SEMICOLONS

UltCombo:
Shouldn't the [`package.json`'s description](https://github.com/npm/npm/blob/9d61e96fb1f48687a85c211e4e0cd44c7f95a38e/package.json#L4) be changed as well? https://www.npmjs.com/package/npm still reads "A package manager for node" due to the `package.json`'s description.

linclark:
Good catch! I've added a PR: #7217

othiym23:
I'm sorry, I should have mentioned that I landed this as fec9c851a1309a4e886fbf18907f27cc7930a28a on Saturday. How many more of these are lurking around?

KenanY:
@othiym23 [`doc/cli/npm.md`](https://github.com/npm/npm/blob/226a6776a1a9e28570485623b8adc2ec4b041335/doc/cli/npm.md) mentions node in a few places (ie "npm is the package manager for the Node JavaScript platform."). Should these removals apply to them as well?

othiym23:
Yes, please! I don't really have time to chase all that down (or, rather, it's much lower priority than many, many other things I have to do), but I'd really appreciate PRs that straighten out the language here. npm *is* still the package manager for Node, but that's not all it is, and that's what the language should reflect.

Issue #5267, opened 2014-05-13
Lifecycle scripts silently fail on win32 if cmd.exe is not on the PATH


edmorley:
Using node v0.10.28 and npm v1.4.9 on Windows 8.1 x64, inside an MSYS bash shell (where the path is overridden by my .bashrc for other reasons):
**package.json:**
```json
{
  "scripts": {
    "postinstall": "echo foo"
  }
}
```
**Output:**
```
[/c/npmtest]$ echo $PATH
.:/c/Python27:/c/Python27/Scripts:/mingw/bin:/bin:/c/Program Files/nodejs:/c/Users/Ed/AppData/Roaming/npm
[/c/npmtest]$ npm install
npm WARN package.json @ No description
npm WARN package.json @ No repository field.
npm WARN package.json @ No README data

> @ postinstall c:\npmtest
> echo foo


[/c/npmtest]$ export PATH="$PATH:/c/Windows/System32"
[/c/npmtest]$ npm install
npm WARN package.json @ No description
npm WARN package.json @ No repository field.
npm WARN package.json @ No README data

> @ postinstall c:\npmtest
> echo foo

foo
```

**npm-debug.log from the failed (first) |npm install| above:**
```
19 info postinstall @
20 verbose unsafe-perm in lifecycle true
21 error Error: spawn ENOENT
21 error     at errnoException (child_process.js:998:11)
21 error     at Process.ChildProcess._handle.onexit (child_process.js:789:34)
22 error If you need help, you may report this *entire* log,
22 error including the npm and node versions, at:
22 error     <http://github.com/npm/npm/issues>
23 error System Windows_NT 6.2.9200
24 error command "c:\\Program Files\\nodejs\\node.exe" "c:\\Program Files\\nodejs\\node_modules\\npm\\bin\\npm-cli.js" "install"
25 error cwd c:\npmtest
26 error node -v v0.10.28
27 error npm -v 1.4.9
28 error syscall spawn
29 error code ENOENT
30 error errno ENOENT
31 verbose exit [ 1, true ]
```

Similar to https://github.com/joyent/node/pull/7616, this is caused by using 'cmd' without an absolute path, and thus relying on it being present on the PATH. More frustratingly the |npm install| appears to succeed, meaning it took some time before I realised it hadn't and so looked at npm-debug.log

As such, I think:
1) The spawn ENOENT should be output using log.error(), rather than swallowed silently here: 
[...]

PR #7057, opened 2015-01-04
scripts: use full path for cmd.exe on Win32

edmorley:
Currently lifecycle.js assumes that Window's cmd.exe is on the PATH,
and fails with a spawn ENOENT error if it is not.

The Windows 'comspec' environment variable contains the full filepath
to the default command interpreter, eg "C:\Windows\System32\cmd.exe".
Should it not be set, we fall-back to using 'cmd' from PATH, as before.

Fixes https://github.com/npm/npm/issues/5267

### DIFF
--- a/lib/utils/lifecycle.js
+++ b/lib/utils/lifecycle.js
@@ -198,7 +198,7 @@ function runCmd_ (cmd, pkg, env, wd, stage, unsafe, uid, gid, cb_) {
   var shFlag = "-c"
 
   if (process.platform === "win32") {
-    sh = "cmd"
+    sh = process.env.comspec || "cmd"
     shFlag = "/c"
     conf.windowsVerbatimArguments = true
   }
--- a/test/tap/lifecycle-path.js
+++ b/test/tap/lifecycle-path.js
@@ -6,11 +6,13 @@ var pkg = path.resolve(__dirname, "lifecycle-path")
 var fs = require("fs")
 var link = path.resolve(pkg, "node-bin")
 
-// Without the path to the shell, nothing works usually.
 var PATH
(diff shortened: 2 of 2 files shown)


othiym23:
I like this! I would like it even better if it had a test! Even if that test only runs on Windows!

edmorley:
@othiym23 I've tweaked lifecycle-path.js to test for this, though I've not been able to get the tests to pass locally since they were broken on Windows (in several files, including lifecycle-path.js) before this change.

othiym23:
Landed as 5fce278a688a1cb79183e012bde40b089c2e97a4. I'm still working on the failing Windows tests, and one of these days I need to figure out how to get Windows-based CI up and running. Thanks!

edmorley:
Thanks! :-)

jden:
Great patch, and most of the way to where I'm trying to get, which is overriding `cmd.exe` as the command processor for npm run scripts on windows, since it's not compatible with basic shell-isms, like `&`, making it hard to write portable scripts even for basic tasks (like, say, running watchified sass and browserify at the same time) - as noted in http://substack.net/task_automation_with_npm_run#windows

I'm wanting to override `cmd.exe` with the msys `sh.exe` which ships with Git on Windows and works well - however I'd also need to override `shFlag` to be `-c` instead of `/c`.

My current work-around is to preface each of my run-scripts with `sh -c `, but this is crufty and results in nested shell processes.

othiym23:
@jden Please open a new issue for this feature request, and we can talk about the tradeoffs of using something like `win-spawn` or writing something that makes it possible to completely swap `cmd.exe` out on Windows, as long as it doesn't ever get triggered by accident.

Issue #6745, opened 2014-11-19
npm update --dev is not documented


mohsen1:
I just assumed `--dev` will update `devDependencies` and it worked. But [here](https://www.npmjs.org/doc/cli/npm-update.html) `--dev` is not mentioned

smikes:
PR #7066 is a candidate fix for this documentation bug.  Thanks for reporting it.

othiym23:
Fixed in 560c00945d4dec926cd29193e336f137c7f3f951. Thanks for pointing this out!

Issue #6247, opened 2014-09-20
npm update should only update top-level dependencies


othiym23:
This would also make `npm update -g` behave the way people mostly think it does already, and render [these scripts](https://gist.github.com/othiym23/4ac31155da23962afd0e) redundant.

williamboman:
I thought this would be fixed in `2.2.0`? Are [these scripts](https://gist.github.com/othiym23/4ac31155da23962afd0e) still relevant?

jokeyrhyme:
What if we had a way to specify recursive updates, making shallow updates the default? That way newbs like me have a safe `npm -g update` whilst experts can use `npm -g -R update` or something when they know that's what they want?

nicolasmccurdy:
@jokeyrhyme :+1: Sounds great to me.

To clarify, `npm update` (when not used globally) performs shallow updates by default, right?

othiym23:
> To clarify, npm update (when not used globally) performs shallow updates by default, right?

Nope! It updates all the way down the local dependency tree!

nicolasmccurdy:
Whoah, that's weird. Then I think the `-R`/`-r` idea should apply to `npm update` anywhere.

CaptSaltyJack:
How's this coming along?

othiym23:
[It's done!](https://github.com/npm/npm/commit/8b98f0e709d77a8616c944aebd48ab726f726f76)

CaptSaltyJack:
WOOT! This is fixed in 2.7.4, looks like?

othiym23:
`npm@>=2.6.1`.

jonnybarnes:
So to clarify, can I now simply run `npm update -g` instead of running one of these: https://gist.github.com/othiym23/4ac31155da23962afd0e

nicolasmccurdy:
Yes.

> This would also make `npm update -g` behave the way people mostly think it does already, and render [these scripts](https://gist.github.com/othiym23/4ac31155da23962afd0e) redundant.

jokeyrhyme:
Unless you're doing a fresh install of the latest node.js / io.js, you'll still start off with an old version of npm, which means you will still need to explicitly `npm -g install npm` before you `npm -g update`, at least for that first time. So what versions of node.js / io.js come with the fixed version of npm?

KenanY:
@jokeyrhyme:

  - Node.js probably v0.12.2
[...]

PR #7066, opened 2015-01-06
add warning and document --dev flag

smikes:
add warning re `npm update -g`
document --dev flag

Fix for #6745.  A mild warning about `npm update -g` was added, as well.

### DIFF
--- a/doc/cli/npm-update.md
+++ b/doc/cli/npm-update.md
@@ -10,13 +10,24 @@ npm-update(1) -- Update a package
 This command will update all the packages listed to the latest version
 (specified by the `tag` config).
 
-It will also install missing packages.
+It will also install missing packages. As with all commands that install
+packages, the `--dev` flag will cause `devDependencies` to be processed
+as well.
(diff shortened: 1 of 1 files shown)


doc/cli/npm-update.md
```diff
@@ -18,6 +20,15 @@ packages.
 If no package name is specified, all packages in the specified location (global
 or local) will be updated.
 
+## WARNING
```
othiym23: Split this out into a separate patch, plz -- at some point we're going to need to update the docs again to fix this, and it's probably going to be forgotten unless we have some way of tracking that on, say, #6247.

othiym23:
Split out the `npm -g update` warning into a separate PR and this is good to go. Thanks!

smikes:
will do.

othiym23:
Reworded the commit message and squashed into one commit, and then landed as 560c00945d4dec926cd29193e336f137c7f3f951. Once more, thanks!

Issue #4197, opened 2013-11-26
OS umask overrides npm umask


robnagler:
This is running on OS X 10.9, npm 1.3.14.  Also, I think npm get should display the umask in octal.

```
root# ls -l /usr/local/lib/node_modules/
total 0
drwxr-xr-x  15 nobody  wheel  510 Nov 26 08:05 ar-drone
drwxr-xr-x  21 nobody  wheel  714 Nov 26 08:05 espresso
drwxr-xr-x  13 nobody  wheel  442 Nov 26 08:17 express
drwxr-xr-x  20 24561   staff  680 Nov 24 13:28 npm
root# npm i -g async
npm http GET https://registry.npmjs.org/async
npm http 304 https://registry.npmjs.org/async
async@0.2.9 /usr/local/lib/node_modules/async
root# ls -l /usr/local/lib/node_modules/
total 0
drwxr-xr-x  15 nobody  wheel  510 Nov 26 08:05 ar-drone
drwxr-x---   7 nobody  wheel  238 Nov 26 08:23 async
drwxr-xr-x  21 nobody  wheel  714 Nov 26 08:05 espresso
drwxr-xr-x  13 nobody  wheel  442 Nov 26 08:17 express
drwxr-xr-x  20 24561   staff  680 Nov 24 13:28 npm
root# npm config get umask
18
root# umask
0027
root# npm --version
1.3.14
```

zkxs:
I'd expect `npm install -g` to make the installed files globally accessible.

    umask 077
    sudo -H npm install -g somepackage

The result of the preceding commands is that only root has access to run/read the installed package.  Is it expected that the shell's umask overrides npm's internal umask?

robnagler:
Yes, the shell's umask should not affect the install umask. The bug is that `mkdir` applies `umask`, and installers must override this behavior with an explicit `chmod`. Installers like RPM and the `install` command ignore umask.

Issue #6845, opened 2014-12-04
umask inside npm run [...]


syndr0m:
npm is loosing the umask inside npm run commands :
```
$> umask 0022
$> npm config get umask
18
$> cat package.json 
{
  "scripts": {
    "umask": "npm config get umask"
  }
}
$> npm run umask
 
> @ umask /home/marc/tmp/umask
> npm config get umask

1
$> npm --version
1.4.28
```

this leads to weird files/directories permissions when using npm install inside npm run [...]

smikes:
Hah!  I think I understand the problem here.  It's that the "022" internally gets turned into 18, which then turned into a string "18" which is then parsed again, using `parseInt(val, 8)` which silently terminates after the first char (because the `8` is an invalid octal digit).

May be fixed by my `umask` PR, #7068.

smikes:
Found and fixed ; candidate fix on #7068

othiym23:
Fixed by #7068, which is part of `npm@2.3.0`, which is currently `npm@next` and will be promoted to the stable version of npm this Thursday, barring something terrible.

PR #7068, opened 2015-01-06
report umask as zero-padded octal

smikes:
Fix for part of #4197, 

> `npm get` should display the umask in octal

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -18,6 +18,7 @@ var log = require("npmlog")
   , ini = require("ini")
   , editor = require("editor")
   , os = require("os")
+  , umask = require("./utils/umask")
 
 config.completion = function (opts, cb) {
   var argv = opts.conf.argv.remain
@@ -132,6 +133,7 @@ function set (key, val, cb) {
   val = val.trim()
   log.info("config", "set %j %j", key, val)
   var where = npm.config.get("global") ? "global" : "user"
+  if (key.match(/umask/)) val = umask.fromString(val)
   npm.config.set(key, val, where)
   npm.config.save(where, cb)
 }
@@ -141,7 +143,9 @@ function get (key, cb) {
   if (!public(key)) {
     return cb(new Error("---sekretz---"))
   }
-  console.log(npm.config.get(key))
+  var val = npm.config.get(key)
+  if (key.match(/umask/)) val = umask.toString(val)
+  console.log(val)
   cb()
 }
 
--- a/lib/utils/lifecycle.js
+++ b/lib/utils/lifecycle.js
@@ -11,6 +11,8 @@ var log = require("npmlog")
   , Stream = require("stream").Stream
   , PATH = "PATH"
   , uidNumber = require("uid-number")
+var umask = require("./umask")
+
 
 // windows calls it's path "Path" usually, but this is not guaranteed.
 if (process.platform === "win32") {
@@ -316,6 +318,7 @@ function makeEnv (data, prefix, env) {
     }
     var value = npm.config.get(i)
     if (value instanceof Stream || Array.isArray(value)) return
+    if (i.match(/umask/)) value = umask.toString(value)
     if (!value) value = ""
     else if (typeof value === "number") value = "" + value
     else if (typeof value !== "string") value = JSON.stringify(value)
--- a/lib/utils/umask.js
+++ b/lib/utils/umask.js
@@ -1,0 +1,18 @@
+// fromString with logging callback
+
+var umask = require("umask")
+var npmlog = require("npmlog")
+var _fromString = umask.fromString
+
+module.exports = umask
+
+module.exports.fromString = function (val, cb) {
+  _fromString(val, function (err, result) {
+    if (err) {
+      npmlog.warn("invalid umask", err.message)
+    }
+    val = result
+  })
+
+  return val
+}
--- a/doc/misc/npm-config.md
+++ b/doc/misc/npm-config.md
@@ -829,7 +829,7 @@ The location of user-level configuration settings.
 ### umask
 
 * Default: 022
-* Type: Octal numeric string
+* Type: Octal numeric string in range 0000..0777 (0..511)
 
 The "umask" value to use when setting the file creation mode on files
 and folders.
--- a/lib/config/core.js
+++ b/lib/config/core.js
@@ -8,8 +8,9 @@ var fs = require("fs")
 var path = require("path")
 var nopt = require("nopt")
 var ini = require("ini")
-var Octal = configDefs.Octal
+var Umask = configDefs.Umask
 var mkdirp = require("mkdirp")
+var umask = require("../utils/umask")
 
 exports.load = load
 exports.Conf = Conf
@@ -362,8 +363,8 @@ function parseField (f, k) {
   var isPath = -1 !== typeList.indexOf(path)
   var isBool = -1 !== typeList.indexOf(Boolean)
   var isString = -1 !== typeList.indexOf(String)
-  var isOctal = -1 !== typeList.indexOf(Octal)
-  var isNumber = isOctal || (-1 !== typeList.indexOf(Number))
+  var isUmask = -1 !== typeList.indexOf(Umask)
+  var isNumber = -1 !== typeList.indexOf(Number)
 
   f = (""+f).trim()
 
@@ -396,8 +397,11 @@ function parseField (f, k) {
     f = path.resolve(f)
   }
 
+  if (isUmask)
+    f = umask.fromString(f)
+
   if (isNumber && !isNaN(f))
-    f = isOctal ? parseInt(f, 8) : +f
+    f = +f
 
   return f
 }
--- a/lib/config/defaults.js
+++ b/lib/config/defaults.js
@@ -9,6 +9,7 @@ var path = require("path")
   , nopt = require("nopt")
   , os = require("os")
   , osenv = require("osenv")
+  , umask = require("../utils/umask")
 
 var log
 try {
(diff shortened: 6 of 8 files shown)


lib/config.js
```diff
@@ -141,10 +141,18 @@ function get (key, cb) {
...
-  console.log(npm.config.get(key))
+  var val = npm.config.get(key)
+  if (key.match(/umask/)) val = umask(val)
+  console.log(val)
   cb()
 }
 
+function umask (val) {
+  val = val.toString(8)
+  while (val.length < 3) val = "0" + val
```
othiym23: If you want this to look like an octal literal, it should always have a leading 0 and be 4 digits wide. Since it's a failure of ES5 strict mode to treat leading-0 literals as octals anyway, you _could_ future-proof this by adding a leading `0o`, as that's the way ES6 does it, but that's not really in keeping with the spirit of POSIX.

othiym23:
One suggestion, otherwise LGTM. Thanks, @smikes!

lib/config.js
smikes: s/3/4 then.  Thppt.  Bash user.

I think because (decades ago) I learned on tcsh, I internalized a 3 digit umask.

smikes:
No unit test; didn't know how to implement it.  There's already a unit test that checks that `config.set('umask', 022)` leads to `config.get('umask') === 022`

.. which is kind of interesting, now that I think about it, that it passes at all.  Maybe it's a "==" implicitly?

smikes:
Bash also tells me the highest legit umask is "0777".  Should I clamp umask to the range 0..0777 ?  or trust that config values will always be valid.  (kidding, answer is obvious..)

othiym23:
That unit test is there because of a regression I introduced when I first started at npm, Inc. trying to "fix" a bunch of eslint warnings about passing string values to the various mode-changing functions. The trick is that Node / JavaScript do the right thing when you're *not* running in strict mode, and that things only get hairy when you do the coercion to numeric values yourself before calling the relevant functions.

Clamping is better than no clamping, but emitting a warning that you've got a umask value that npm doesn't understand should go along with the clamp.

smikes:
bleah.  not sure if this is better now..

lib/utils/lifecycle.js
```diff
@@ -3,14 +3,16 @@ exports.cmd = cmd
...
 var log = require("npmlog")
-  , spawn = require("./spawn")
-  , npm = require("../npm.js")
-  , path = require("path")
-  , fs = require("graceful-fs")
-  , chain = require("slide").chain
-  , Stream = require("stream").Stream
-  , PATH = "PATH"
-  , uidNumber = require("uid-number")
+var spawn = require("./spawn")
```
othiym23: I'm probably going to semi-revert this hunk -- style changes will be done atomically, per @isaacs's recommendations, and until then, I don't want to noisy up the diff.

lib/utils/umask.js
```diff
@@ -0,0 +1,56 @@
+// commands for handling umask
```
othiym23: How would you feel about turning this into a separate module?

test/tap/umask.js
```diff
@@ -0,0 +1,64 @@
+var path = require("path")
+var test = require("tap").test
+var rimraf = require("rimraf")
+var mkdirp = require("mkdirp")
+var common = require("../common-tap.js")
+var npmlog = require("npmlog")
```
othiym23: This isn't used in the test, but you should be setting the log level to `silent` -- lotta warnings getting printed out right now.

test/tap/umask.js
```diff
@@ -0,0 +1,64 @@
+var path = require("path")
+var test = require("tap").test
+var rimraf = require("rimraf")
+var mkdirp = require("mkdirp")
+var common = require("../common-tap.js")
```
othiym23: Not used in the test.

test/tap/umask.js
```diff
@@ -0,0 +1,64 @@
...
+var npmlog = require("npmlog")
+var umask = require("../../lib/utils/umask")
+
+var npm = require("../../lib/npm.js")
```
othiym23: Not used in the test.

test/tap/umask.js
```diff
@@ -0,0 +1,64 @@
...
+
+test("setup", function (t) {
+  rimraf.sync(pkg)
+  mkdirp.sync(pkg)
+  t.end()
+})
+
+test("umask to string", function (t) {
+  t.equal("0000", umask.toString(0))
+  t.equal("0000", umask.toString(0000))
```
othiym23: `eslint` complains about every one of the octal literals in this file. ;)

othiym23:
Basically LGTM. There are a few tweaks I've suggested (to be clear: the tests don't need to be boilerplate; if a variable isn't used in the file, nuke it), so make those and it'll be ready to land.

test/tap/umask.js
rlidwka: `eslint --reset` + explicit list of all the rules you want to enable could help with that

lib/utils/lifecycle.js
smikes: OK

lib/utils/umask.js
smikes: Seems like overkill, but sure

test/tap/umask.js
smikes: arguably we should be checking that the warnings are printed.  Is there an easier way to do that than monkeypatching npmlog.warn ?

smikes:
Rebased and squashed; using a synchronous callback to test warning messages, no stderr noise anymore.

smikes:
Took your advice and published <https://npmjs.org/packages/umask> as a module.  That makes this change a lot lighter: mostly internal renaming (Octal => Umask) and then the test for #6845.

Rewriting the `umask` functions as a module allowed me to move the conversion unit tests out as well; also, I had a free hand on test runner and code style, so `umask` has 100% coverage and lints clean.  Thanks for making that suggestion.

I retained the file `lib/utils/umask.js` because I think it's cleaner for `npm` to use the `var x = umask.fromString(y)` syntax inline, but I also want to log any parse failures as    a warning.  The monkey-patch version of `umask.fromString` will log the error, if any, as `npmlog.warn()`.

I also added "legacy" support for decimal strings.  Before this patch lands, anyone who has set `umask` with `npm c set umask 022` will have the line `umask=18` in their .npmrc.  This fails validation by the `umask.validate` function as written, silently substituting a umask of `0022`.

I have **not** changed that behavior here, but I made `umask.fromString()` accept decimal strings as long as they have no leading 0 and are in the range 0..511

lib/utils/umask.js
```diff
@@ -0,0 +1,18 @@
+// fromString with logging callback
+
+var umask = require("umask")
+var npmlog = require("npmlog")
+var _fromString = umask.fromString
+
+module.exports = umask
+
+module.exports.fromString = function (val, cb) {
```
othiym23: This, uh, feels kinda gross, and like it's tempting the wrath of t̃ͮ̒͂͌ͨ̎̾̒̀͊ͮ̇͒ͥ́͞҉̵̤̩̣͔͎̖̰̼͓̖̤͞h̸̛̥̺͈͓̱̮̹͈͎͔̣̟̦͔̘͇̝͒̎̉ͮͭ̿̍ͧ̿͋̃͐̍̏̔̍ė͈͉̟̯̮͓͍̤̲͓̪̣̑͛̒̈́̃͋ͮ̔͐͢͟ ̈̆͌̌͜҉̴̺̜̺̠͕̜̭͓͔͘͟ͅḋ̨̛̹̰̯̫̖̭̘̰̗̙͎͆ͮͮ̋͘͜͜a̵̜͎̟̼͓̼̫̤̔̒̔ͪ̑̒̑ͩ̚͟͟͝r̵̯̫̬͍̻̪͖̬̫̼̍̀ͧ̎́̅̇ͫ͆̄̈́̃́̓̇̕͡ͅk̢̥̯͙̦̳͚̠̯̜̜̘͓̆͌̌ͯ̾ͧ́ͨ̎͊͌͋̃̊̍̇̋͘͠ ̸̦̮͔͕̮̠̺̲̙̯̳̠̣̯̋̍̿ͫ͑̆͛ͮͥ͠͠ͅp̸̱͎̲̗̙̫ͪͬ̀͆ͤͬ̉̔̿̊̋ͪ̀̚͜͟͡ȏ̡̨̬̠͎̪̣̯͔͙̞̜̭̲͉͋ͨͧ͒͌̓̋̽̇ͬ̆̆ͥ́͠ͅn̨̧̧̤̪̱̹̽͆͌ͬ̉̐͆̾̀y̥͉̘̯̥̗͕͖̙̜͙ͦ̓̌ͭ̍͌̆͂̌̚͟͡
̬̰͖̝͇̳̰͍͊̐͋͌͌͗̃̄͊̐̀͢ḥ̥̲̹̳͚̻̜̞̱͔̥̰̻̩̭̏̓ͪ̽̈́͂͒̒͂ͭͤ͛ͯͯ̉̆̅ͯ́̚e̐ͨͤ̄͝͞͞͏̭͎̗̯͎̼̜͉̙̗͢ ̛̮̯͍̖̹̠͈ͩ̍̏̎͘͞c̡̾ͫ́ͭ̀͆ͩ͋ͣ̊̽͡͏҉̨̙͉̳̺͇̦͈̮ͅo̡̖̤̦̫̣̞̳̰͕̠̝͚̐ͭ̏͋̊̀ͥ̆̿̂̽̇͘͟͞m̸̘͔̱̝̥̞͍̼̜̩̟̻͍͚̑̿͑ͤ̔̋̚̕͟ͅe̶̛̫̼̹̺̦͇ͪ̎͋̅ͦ͌̾ͩͨ̋ͯ͂ͦ̚͠s͉̫̞̣̝̮̤̩͂͌̇̐̿ͩͤ̎͂ͯ̌̍͛́

But I landed it anyway!

othiym23:
Landed as c6fb430e55672b3caf87d25cbd2aeeebc449e2f2 and 209713ebd4b77da11ce27d90c3346f78d760ba52, with some minor changes (the test was hardcoded to your machine), few tiny style tweaks. Thanks! This is a useful change!

PR #7077, opened 2015-01-07
default a script for `npm run env` like `npm test`

othiym23:
Paging @gcb -- I made some further tweaks to #6999 (mostly to fix issues on Windows), but in doing so, I noticed a few things that I decided needed further discussion before it landed. See below for discussion.

### DIFF
--- a/doc/cli/npm-run-script.md
+++ b/doc/cli/npm-run-script.md
@@ -11,10 +11,13 @@ npm-run-script(1) -- Run arbitrary package scripts
 This runs an arbitrary command from a package's `"scripts"` object.
 If no package name is provided, it will search for a `package.json`
 in the current folder and use its `"scripts"` object. If no `"command"`
-is provided, it will list the available top level scripts.
+is provided, it will list the available top level scripts. The `env` command
+can be used to list environment variables that will be available to the script
+at runtime. The `env` command can be overridden in your package but it is
+not recommended.
 
-It is used by the test, start, restart, and stop commands, but can be
-called directly, as well.
+`run[-script]` is used by the test, start, restart, and stop commands, but can
+be called directly, as well.
 
 As of [`npm@2.0.0`](http://blog.npmjs.org/post/98131109725/npm-2-0-0), you can
 use custom arguments when executing scripts. The special option `--` is used by
--- a/lib/run-script.js
+++ b/lib/run-script.js
@@ -7,7 +7,7 @@ var lifecycle = require("./utils/lifecycle.js")
   , log = require("npmlog")
   , chain = require("slide").chain
 
-runScript.usage = "npm run-script <command> [-- <args>]"
+runScript.usage = "npm run-script <command>|env [-- <args>]"
 
 runScript.completion = function (opts, cb) {
 
@@ -125,9 +125,16 @@ function run (pkg, wd, cmd, args, cb) {
   } else {
     if (!pkg.scripts[cmd]) {
       if (cmd === "test") {
-        pkg.scripts.test = "echo \"Error: no test specified\"";
+        pkg.scripts.test = "echo \"Error: no test specified\""
+      } else if (cmd === "env") {
+        log.verbose("run-script using default multiplatform env")
+        if (process.platform === "win32") {
+          pkg.scripts[cmd] = "SET"
+        } else {
+          pkg.scripts[cmd] = "env"
+        }
       } else {
-        return cb(new Error("missing script: " + cmd));
+        return cb(new Error("missing script: " + cmd))
       }
     }
     cmds = [cmd]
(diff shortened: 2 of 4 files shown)


othiym23:
First, and most notably, this is what I get when I run `npm run env` on Windows:

```
C:\Users\ogd\Documents\projects\npm\npm>npm run env
npm ERR! Windows_NT 6.2.9200
npm ERR! argv "C:\\Users\\ogd\\Documents\\projects\\nodist/v/0.10.35/node.exe" "C:\\Users\\ogd\\Documents\\projects\\nodist\\bin\\node_modules\\npm\\bin\\npm-cli.js" "run" "env"
npm ERR! node v0.10.35
npm ERR! npm  v2.1.17

npm ERR! missing script: env
npm ERR!
npm ERR!     <http://github.com/npm/npm/issues>

npm ERR!     C:\Users\ogd\Documents\projects\npm\npm\npm-debug.log
```

I'll look into why the test is passing but the command is failing.

othiym23:
Never mind! I wasn't running the latest code. `npm link` on Windows needs a little love when run against npm.

lib/run-script.js
```diff
@@ -7,7 +7,7 @@ var lifecycle = require("./utils/lifecycle.js")
   , log = require("npmlog")
   , chain = require("slide").chain
 
-runScript.usage = "npm run-script <command> [-- <args>]"
+runScript.usage = "npm run-script <command>|env [-- <args>]"
```
othiym23: I don't like special-casing this in the usage statement. We don't do the same for `npm run-script test`. Leave the complete documentation to the manpage.

lib/run-script.js
```diff
@@ -125,9 +125,16 @@ function run (pkg, wd, cmd, args, cb) {
   } else {
     if (!pkg.scripts[cmd]) {
       if (cmd === "test") {
-        pkg.scripts.test = "echo \"Error: no test specified\"";
+        pkg.scripts.test = "echo \"Error: no test specified\""
+      } else if (cmd === "env") {
+        log.verbose("run-script using default multiplatform env")
```
othiym23: I would prefer to replace this command with different statements for Windows and non-Windows saying what will actually be run.

doc/cli/npm-run-script.md
```diff
@@ -11,10 +11,13 @@ npm-run-script(1) -- Run arbitrary package scripts
 This runs an arbitrary command from a package's `"scripts"` object.
 If no package name is provided, it will search for a `package.json`
 in the current folder and use its `"scripts"` object. If no `"command"`
-is provided, it will list the available top level scripts.
+is provided, it will list the available top level scripts. The `env` command
+can be used to list environment variables that will be available to the script
+at runtime. The `env` command can be overridden in your package but it is
```
othiym23: "is not recommended" is pretty strongly-worded for a feature that we're just now adding. ;) Either drop this sentence or replace it with something that explains why it might be a bad idea.

lib/run-script.js
gcb: on the verbose message or the command alias? i think having the same name for both is a feature for people writing build scripts that actually checks any of those var.

lib/run-script.js
gcb: good point. i recall trying to find 'test' called out and failing... not sure why i left this here.

lib/run-script.js
othiym23: The problem is that `env` and `SET` are actually semantically quite distinct -- you can pass arguments to `env` and they'll be treated as a command invocation in that environment (which is why `env` is so useful on the shebang line in scripts). `SET` doesn't have this property, and it would be nice to be able to tell from the logs which you've gotten.

gcb:
@othiym23 i'm not sure how to move on this workflow, sorry :/

i committed the changes to my old fork. ...do i have write access to this branch?

lib/run-script.js
gcb: that is trick... `set` doesn't necessarily have arguments either. if the user is on solaris csh, it won't understand any of the options set from bash understands.
I'm honestly not sure what most use cases would expect.... for me, since all argument i see are for changing future behavior of the shell, i think nobody will have reasons to pass arguments on a build from this command. all i can see this being used is to probe the vars npm will add when running the script the use is writing/debugging.

othiym23: Ooops... little misunderstanding here -- I'm totally fine with Windows and Unix using different *commands*, I just think we should log which one we're *using*. Does that make more sense?

gcb: ah gotcha. that was my initial doubt :) sure, that makes total sense.

it is also on my old fork. if i should commit/merge it somewhere else, let me know. not familiar with the workflow.

othiym23:
OK, I re-landed my tweaks, rebased, and landed this as acf287d2547c8a0a8871652c164019261b666d55. Congratulations! This means that the next version of npm will be `npm@2.2.0`, and it's due entirely to this feature! Thank you for your patience (and contribution)!

Issue #7067, opened 2015-01-06
Password is no longer obfuscated with --sekretz--


smikes:
Noticed when looking at old case https://github.com/npm/npm/issues/5068

```
$ npm get //registry.npmjs.org/:_password
[base64 encoded data]
```

vs. the old way

```
$ npm get _auth
npm ERR! Darwin 14.0.0
npm ERR! argv "node" "/Users/smikes/.nvm/v0.10.35/bin/npm" "get" "_auth"
npm ERR! node v0.10.35
npm ERR! npm  v2.1.17

npm ERR! ---sekretz---
npm ERR! 
npm ERR! If you need help, you may report this error at:
npm ERR!     <http://github.com/npm/npm/issues>

npm ERR! Please include the following file with any support request:
npm ERR!     /Users/smikes/src/github/npm/lib/npm-debug.log
```

I propose that the test for secrecy (lib/config.js:142) be changed to use `public()` and the `public` function look for initial `"_"` OR internal `":_"`.  Thoughts?

othiym23:
Sounds good to me. I'd noticed this inconsistency in the middle of debugging other things, but not paused to consider why it might be happening.

PR #7098, opened 2015-01-08
#7067 omit private config

smikes:
Candidate fix for #7067, adds unit test


othiym23:
It LGTM, so I made a few (minor) stylistic tweaks, rewrote the commit message, and landed it as c76b8013bf1758587565822626171b76cb465c9e. Thanks!

Issue #7075, opened 2015-01-07
"invalid config Tag must not be a SemVer range" should be an error, not a WARN


othiym23:
In other words, it should cause an early failure rather than publishing without the tag. See https://github.com/npm/npm/issues/6837#issuecomment-68866411 for context.

/cc @smikes

KenanY:
Fixed by https://github.com/npm/npm/pull/7099?

othiym23:
Indeed, fixed by 4ef1412d0061239da2b1c4460ed6db37cc9ded27. Thanks, @KenanY!

PR #7099, opened 2015-01-08
change "Tag must not be a SemVer range" warning to error on publish

smikes:
This is the _start_ of a candidate fix for #7075. I adjusted one unit test, but something else odd is happening.

I say "it doesn't work" because I get a full `npm` usage message after my failure after the change:

```
$ node ./cli.js --publish --tag v1.2.3
npm WARN invalid config tag="v1.2.3"
npm ERR! invalid config Tag must not be a SemVer range

Usage: npm <command>

where <command> is one of:
    add-user, adduser, apihelp, author, bin, bugs, c, cache,
    completion, config, ddp, dedupe, deprecate, docs, edit,
    explore, faq, find, find-dupes, get, help, help-search,
    home, i, info, init, install, issues, la, link, list, ll,
    ln, login, ls, outdated, owner, pack, prefix, prune,
    publish, r, rb, rebuild, remove, repo, restart, rm, root,
    run-script, s, se, search, set, show, shrinkwrap, star,
    stars, start, stop, t, tag, test, tst, un, uninstall,
    unlink, unpublish, unstar, up, update, v, verison, version,
    view, whoami

npm <cmd> -h     quick help on <cmd>
npm -l           display full usage info
npm faq          commonly asked questions
npm help <term>  search for help on <term>
npm help npm     involved overview

Specify configs in the ini-formatted file:
    /Users/smikes/.npmrc
or on the command line via: npm <command> --key value
Config info can be viewed via: npm help config

npm@2.1.18 /Users/smikes/src/github/npm
```

Not sure what's causing **that**.

Testing is with `./node_modules/.bin/tap --timeout 120 test/tap/*semv*`


othiym23:
Let me know when you get this one sorted out. What's there right now looks very simple, so I'm surprised it's not working.

smikes:
I figured out part of my confusion.  My test command was `npm --publish --tag=v1.2.3` which has no subcommand (**--publish** instead of **publish**).  So that's why the usage message was coming out.

Now the only problem is that an error during config setup does not stop the `npm` process.  `npm publish --tag=v1.2.3` issues an error and then merrily goes on attempting to publish.

Tempting as it is to add a process.exit() immediately after the error message, I guess there is probably a better way.

smikes:
TL;DR: this is still not ready to patch.

I understand the problem better now.

Changing the config to catch invalid tags and report than as an error is not enough.  What happens when the tag validation fails is that we fall back to the default -- i.e., `--tag=latest`

This makes me question whether we want the semver checking to live in `config` at all, or pull it out into the various subcommands (note that `npm tag` already does its semver checking inline <https://github.com/npm/npm/blob/master/lib/tag.js#L23>

The command line arg checking in `npm tag` does not catch all ways of setting an invalid tag.  Examples below (but use a package you can't write to!):

This triggers the logic in `lib/tag.js` because the `tag` is supplied on the command line, not via arguments/config parsing:
```
Svarog:npm smikes$ node ./cli.js tag npm@2.2.0 v1.2.3
npm ERR! Darwin 14.0.0
npm ERR! argv "node" "/Users/smikes/src/github/npm/cli.js" "tag" "npm@2.2.0" "v1.2.3"
npm ERR! node v0.10.35
npm ERR! npm  v2.2.0

npm ERR! Tag name must not be a valid SemVer range: v1.2.3
```

When I pass the tag with `--tag v1.2.3` config validation fails, `tag` falls back to `latest`, and `npm` actually attempts to update the registry:
```
$ node ./cli.js tag npm@2.2.0 --tag v1.2.3
npm WARN invalid config tag="v1.2.3"
npm ERR! invalid config Tag must not be a SemVer range
npm http request PUT https://registry.npmjs.org/npm/latest
npm http 403 https://registry.npmjs.org/npm/latest
npm ERR! Darwin 14.0.0
npm ERR! argv "node" "/Users/smikes/src/github/npm/cli.js" "tag" "npm@2.2.0" "--tag" "v1.2.3"
npm ERR! node v0.10.35
npm ERR! npm  v2.2.0
npm ERR! code E403

npm ERR! forbidden user: smikes not authorized to modify npm
npm ERR! Changed: dist-tags.latest "2.1.18" -> "2.2.0"
npm ERR! Changed: time.modified "2015-01-10T07:29:05.122Z" -> "2015-01-15T13:30:57.243Z": npm,latest
```

smikes:
I think that this change might have been a mistake: https://github.com/npm/npm/commit/9bac6b860b674d24251bb7b8ba412fdb26cbc836

This makes the tag checking that was initially only for `lib/publish.js` be applied to all settings of the `tag` config variable.  But that means that it's no longer legal to **request** a bad (legacy) tag via configuration e.g.,

```
npm install jquery --tag 1.x
npm WARN invalid config tag="1.x"
npm WARN invalid config Tag must not be a SemVer range
... merrily install @latest
```

Now of course we can work around this with `npm install jquery@1.x`.  But in order to implement #7099 I think the fallback to `latest` has to be moved out of config.

So ironically, in order to make a better error in `npm publish`, I want to remove the failure from config here: https://github.com/npm/npm/commit/9bac6b860b674d24251bb7b8ba412fdb26cbc836#diff-149a2e39d5af7a8f5a3e1ee39767d039R45

Thoughts? /cc @othiym23 @isaacs

lib/publish.js
```diff
@@ -13,10 +13,12 @@ var npm = require("./npm.js")
   , cachedPackageRoot = require("./cache/cached-package-root.js")
   , createReadStream = require("graceful-fs").createReadStream
   , npa = require("npm-package-arg")
+  , semver = require('semver')
 
-publish.usage = "npm publish <tarball>"
-              + "\nnpm publish <folder>"
+publish.usage = "npm publish <tarball> [--tag <tagname>]"
```
othiym23: This is a good change, but seems unrelated to the rest. Separate PR?

smikes: OK.

othiym23:
Surprise! I made a small additional commit to remove the now-superfluous `validateTag` from `./lib/config/defaults.js` and merged this anyway. The change is small and looks good to me, and the tests all pass (although we really have to start tackling the tests soon, because they're getting so slooooooow...). Landed as 4ef1412d0061239da2b1c4460ed6db37cc9ded27. Thanks, Sam!

smikes:
Yay!

Issue #6995, opened 2014-12-23
globally installing more than one package using "npm install -g" causes "uid must be an unsigned int"


k-j-kleist:
This is Windows 7 / PowerShell 3.0
```
PS C:\projects> npm install -g foundation-cli bower gulp
npm ERR! Windows_NT 6.1.7601
npm ERR! argv "C:\\local\\nodist-0.4.8\\bin\\\\node.exe" "C:\\local\\nodist-0.4.8\\bin\\node_modules\\npm\\bin\\npm-cli.js" "install" "-g" "foundation-cli" "bower" "gulp"
npm ERR! node v0.11.14
npm ERR! npm  v2.1.14

npm ERR! uid must be an unsigned int
```
However, this works fine:
```
npm install -g foundation-cli
npm install -g bower
npm install -g gulp
```

othiym23:
Oh, hey! This is fixed in `npm@2.3.0` -- we made a bad change to how we interacted with the cache that was causing a lot of people to have issues when they cached something new. `npm@2.3.0` will become the newest stable version of npm on Thursday, but if you want to run it today and confirm that it fixes your issue, install `npm@next`.

Sorry about the mess!

smikes:
To upgrade your `npm` on Windows, do this:

```
cd C:\Program Files (x86)\nodejs
npm install npm@next
```

techelephant:
I can confirm that npm@2.3.0 solved the issue - at least in my case :-) I had the exact same problem as described above, including "npm config get user" returning 0 (zero). Upgrading npm to 2.3.0 (not globally but in the node installation directory) fixed the issue.

Thx othiym23 and smikes...

PR #7107, opened 2015-01-10
Handle cache creation on Windows

rmg:
On windows there is no `process.getuid`, causing `makeCacheDir()` to pass `{}` to its callback.

When this result makes its way to `saveToCache()`, it passed the `.uid === null` guard which results in `chown()` being called with `undefined` for the uid and gid, which triggers the error.

This behaviour was introduced in 34570414cd528debeb22943873440594d7f47abf which suggests the issue is present in every release from v2.1.7 onward for anyone running Windows.

I believe this is the root cause behind #6995, #6980, and probably [others](https://github.com/npm/npm/search?q=uid+must+be+an+unsigned+int&type=Issues).

@othiym23 the tests are just a quick hack'n'paste and I'm not sure they're really even appropriate for this sort of regression... but they're there if you want 'em!

### DIFF
--- a/lib/cache/add-local-tarball.js
+++ b/lib/cache/add-local-tarball.js
@@ -167,7 +167,7 @@ function addTmpTarball_ (tgz, data, shasum, cb) {
       if (er) return cb(er)
       var read = fs.createReadStream(tgz)
       var write = writeStream(target, { mode: npm.modes.file })
-      var fin = cs.uid && cs.gid ? chown : done
+      var fin = isNaN(cs.uid) || isNaN(cs.gid) ? done : chown
       read.on("error", cb).pipe(write).on("error", cb).on("close", fin)
     })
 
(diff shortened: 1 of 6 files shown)


rmg:
Looked a little further and the problem seems to be more pervasive than I first thought, suggesting it may have been around even longer.

othiym23:
This is completely a legit bug, but this patch just deals with consequences rather than the underlying cause. I poked at it some and put together ba2eb367c56d727a5f278128a1d96665f51bc051, which fixes the regression introduced by 734cc5c3de78fa2482863de08b1a4de1567bd7a3 (which means that this has been broken for all of `npm@2`'s life). I may steal your tests, because it's hard to have too much coverage around this particular issue, and I'm going to leave this issue open until builds for npm are running cleanly on Windows again.

Let me know what you think of my patch.

rmg:
@othiym23 thanks. It was late and I didn't think to look up what `chown` would do on Windows and whether uid/gid: 0 was a reasonable default there (certainly isn't on Linux).

rmg:
Oh, and feel free to cherry-pick the tests into your PR. They're not perfect at all, but they at least test the Windows/getuid-feature-test case (`process.getuid == null`).

othiym23:
Thanks for the review, @rmg. I called my solution good enough and merged it for this week's release. I think you're right that this will probably fix a lot of people's Windows issues (leaving only ℵ₀ more remaining!).

rmg:
Excellent :-)

PR #7132, opened 2015-01-14
support the whoami endpoint properly

othiym23:
Needs to be rolled out with npm/npm-registry-client#93 and npm/npm-registry-couchapp#219. This all feels sort of complicated to me.

Also: needs tests.

/cc @bcoe @isaacs


bcoe:
Thanks @othiym23, let's get the couch-app out tomorrow morning, @rockbot has some migration work she's doing with the couchapp as well ... we can have a CouchDB party \o/

Issue #6912, opened 2014-12-12
`npm install` should deduplicate its tree as part of the install process


iarna:
This is intended to solve a few larger project goals:

1. Installs should be smaller, without having to explicitly ask for it.
2. For Windows user's sake, we should do our best to avoid long path names. Many windows tools are limited to 256 characters.

As such, @othiym23 and I have decided this should be a maximally flat deduplication, which means if you have a dep tree of:

    A → B → C

Then if you install A, you'll get B and C next to it in your node modules. That is, it'll be installed as:

    A
    B
    C

But we'll hold onto metadata to know why B and C were installed such that removing A will remove B and C as it would now.

As for devDependencies, they shouldn't have their dependencies folded in as above. Instead each of them should essentially act as its own root. So if A above were a dev dependency then it would be installed as:

    A → B
        C

The following work remains to be done:

* [x] The core deduplicated install work needs to be completed. It is currently in progress.
* #7273 `npm outdated` needs to structure its output in a way that makes sense with a flat node_modules folder. 
* #7272 `npm ls` needs to structure its output in a way that makes sense with a flat node_modules folder. 

See also: #4761, #4037, #5465

timoxley:
Wholly +1 on this. 

Note that this is going to break a lot more package.json scripts incorrectly relying on executables being installed in `node_modules/.bin`. I'm wondering if perhaps it's worth putting an an explicit check for `node_modules/.bin` in scripts and shooting off a warning, because there's few valid reasons to reference `node_modules/.bin` directly in a package.json.

> there's a risk they'll come to depend on things that aren't actually part of the install. 

We could adapt some existing require detecting tooling to detect this pretty easily.

isaacs:
> I'm wondering if perhaps it's worth putting an an explicit check for node_modules/.bin in scripts and shooting off a warning, because there's few valid reasons to reference node_modules/.bin directly in package.json scripts, or anywhere for that matter.

That's really smart.  We should just strip `node_modules/.bin` from the start of them, I think, and maybe also warn.

iarna:
We actually *already do* strip `node_modules/.bin` already in `read-package-json`, here:

https://github.com/npm/read-package-json/commit/56c7f5de

I'm going to be moving this over to `normalize-package-data` and adding a warn.

piranna:
[...]

PR #7148, opened 2015-01-15
dedupe: decrease the log level of unavoidable conflicts

denis-sokolov:
Users running `npm dedupe` do not need to be informed about unavoidable conflicts, they are a part of regular process of dedupe, and are nothing of note.

In my personal opinion, this also aligns with semantic meaning. I was unable to find a detailed standard of log levels, thus I converged on an opinion that warnings require eventual attention, while infos are documenting regular behavior.

Feel free to reject this if your log level definition differs.
Consider that with #6912 the warnings will pop up for regular users during a simple install command.

### DIFF
--- a/lib/dedupe.js
+++ b/lib/dedupe.js
@@ -171,8 +171,8 @@ function installAndRetest (set, filter, dir, unavoidable, silent, cb) {
 
     // nothing to be done here.  oh well.  just a conflict.
     if (!locMatch && !regMatch) {
-      log.warn("unavoidable conflict", item[0], item[1])
-      log.warn("unavoidable conflict", "Not de-duplicating")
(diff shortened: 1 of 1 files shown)


othiym23:
IMO, conflicts deserve to be warnings. This helps people understand why parts of their trees have not been deduplicated, and is information that is useful to surface, so as to minimize people's surprise (and their likelihood to file issues based on that surprise). Also, "IMO" pretty much sums up the guidelines for when which log level is used -- it's kinda up to the developers' discretion which log level gets used where, and is likely to remain that way, just because this is mostly about creating a friendly (if occasionally noisy) user experience.

Thanks for putting this together anyway!

denis-sokolov:
Thank you for your feedback and reasoning!

PR #7149, opened 2015-01-15
Prevent people from overwriting their npm install

KevinSheedy:
When I first tried to install my app into the global bin folder, I used the code sample provided:
    { "bin" : { "npm" : "./cli.js" } }

Unfortunately, I forgot to replace npm with myapp. This replaced my installation of npm with myapp and I had to reinstall nodejs again scratch.

### DIFF
--- a/doc/files/package.json.md
+++ b/doc/files/package.json.md
@@ -154,12 +154,12 @@ command name to local file name. On install, npm will symlink that file into
 installs.
 
 
-For example, npm has this:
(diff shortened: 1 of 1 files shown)


othiym23:
Reworded the commit message and landed as 9eac0a14488c5979ebde4c17881c8cd74f395069. This is a small change, but I agree it will probably help keep some people out of trouble!

Issue #6850, opened 2014-12-04
npm link <path> broken


nigelzor:
```
git clone https://github.com/npm/fs-vacuum.git
mkdir example
cd example
npm link ../fs-vacuum
```
results in
```
npm ERR! Linux 3.16.0-24-generic
npm ERR! argv "node" "/home/neil/bin/npm" "link" "../fs-vacuum"
npm ERR! node v0.10.33
npm ERR! npm  v2.1.11

npm ERR! /home/neil/npm-repro/fs-vacuum/node_modules/mkdirp is not a child of /home/neil/npm-repro/example
npm ERR! 
npm ERR! If you need help, you may report this error at:
npm ERR!     <http://github.com/npm/npm/issues>
npm ERR! error rolling back Error: /home/neil/npm-repro/fs-vacuum/node_modules/mkdirp is not a child of /home/neil/npm-repro/example
npm ERR! error rolling back     at vacuum (/home/neil/lib/node_modules/npm/node_modules/fs-vacuum/vacuum.js:24:15)
npm ERR! error rolling back     at gentlyRm (/home/neil/lib/node_modules/npm/lib/utils/gently-rm.js:47:12)
npm ERR! error rolling back     at /home/neil/lib/node_modules/npm/lib/unbuild.js:30:20
npm ERR! error rolling back     at /home/neil/lib/node_modules/npm/node_modules/read-package-json/read-json.js:51:40
npm ERR! error rolling back     at /home/neil/lib/node_modules/npm/node_modules/read-package-json/read-json.js:97:49
npm ERR! error rolling back     at evalmachine.<anonymous>:207:20
npm ERR! error rolling back     at OpenReq.Req.done (/home/neil/lib/node_modules/npm/node_modules/graceful-fs/graceful-fs.js:141:5)
npm ERR! error rolling back     at OpenReq.done (/home/neil/lib/node_modules/npm/node_modules/graceful-fs/graceful-fs.js:61:22)
npm ERR! error rolling back     at Object.oncomplete (evalmachine.<anonymous>:107:15)
npm ERR! error rolling back  [Error: /home/neil/npm-repro/fs-vacuum/node_modules/mkdirp is not a child of /home/neil/npm-repro/example]
```

At https://github.com/npm/npm/blob/v2.1.9/lib/unbuild.js#L30 gentlyRm should be called with the link target instead of `npm.prefix`, but we no longer know it.

othiym23:
[...]

PR #7180, opened 2015-01-21
only specify npm.prefix as gentlyRm base when it is inside folder path

phated:
Possible fix for #6850 

Thanks @othiym23 for the guidance.

### DIFF
--- a/lib/unbuild.js
+++ b/lib/unbuild.js
@@ -5,6 +5,7 @@ var readJson = require("read-package-json")
   , gentlyRm = require("./utils/gently-rm.js")
   , npm = require("./npm.js")
   , path = require("path")
+  , isInside = require("path-is-inside")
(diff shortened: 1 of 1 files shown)


othiym23:
I need to review this one pretty carefully, and maybe write a couple more tests, but I'll try to land this for this week's release. @iarna, can you double-check @phated's logic here and make sure it makes sense to pass in the unbuilt folder like this?

othiym23:
Landed as 29a6ef38ef86ac318c5d9ea4bee28ce614672fa6 (a semicolon snuck in!). I still need to write a bunch of unhappy-path tests around this, but I looked it over and I think this is probably not a lot more dangerous than what we had before (and is a lot less dangerous than what we had before I rewrote `gently-rm.js` in the first place.

Thanks, @phated!

PR #7214, opened 2015-01-25
Add an NPMOPTS variable to toplevel install

aredridel:
This allows make install from a script to override userconfig in particular


othiym23:
Landed in eeff04da7979a0181becd36b8777d607e7aa1787. Thanks, @aredridel!

PR #7249, opened 2015-01-31
Make sure that the `etc` directory exists before attempting to write to it.

### LINKED ISSUE #7233: `npm config edit --global` errors when `globalconfig` path does not exist
The error message should convey more information about what the problem is.

Ref case #5881, a slightly different problem, but the one where I noticed this.

```
$ npm config edit --global
npm ERR! Darwin 14.1.0
npm ERR! argv "node" "/Users/smikes/.nvm/versions/io.js/v1.0.4/bin/npm" "config" "edit" "--global"
npm ERR! node v1.0.4
npm ERR! npm  v2.3.0
npm ERR! path /Users/smikes/.nvm/versions/io.js/v1.0.4/etc/npmrc.c33db61dc6e13a89a2ab11f8f1793b9e
npm ERR! code ENOENT
npm ERR! errno -2

npm ERR! enoent ENOENT, open '/Users/smikes/.nvm/versions/io.js/v1.0.4/etc/npmrc.c33db61dc6e13a89a2ab11f8f1793b9e'
npm ERR! enoent This is most likely not a problem with npm itself
npm ERR! enoent and is related to npm not being able to find a file.
npm ERR! enoent 

npm ERR!     /Users/smikes/src/github/npm-kludge-search/npm-debug.log
```

ljharb:
Fixes #7233.

### DIFF
--- a/lib/config/core.js
+++ b/lib/config/core.js
@@ -152,10 +152,17 @@ function load_(builtin, rc, cli, cb) {
     // annoying humans and their expectations!
     if (conf.get("prefix")) {
       var etc = path.resolve(conf.get("prefix"), "etc")
-      defaults.globalconfig = path.resolve(etc, "npmrc")
-      defaults.globalignorefile = path.resolve(etc, "npmignore")
+      mkdirp(etc, function (err) {
+        defaults.globalconfig = path.resolve(etc, "npmrc")
+        defaults.globalignorefile = path.resolve(etc, "npmignore")
+        afterUserContinuation()
+      })
+    } else {
+      afterUserContinuation()
     }
+  }
 
+  function afterUserContinuation() {
     conf.addFile(conf.get("globalconfig"), "global")
 
     // move the builtin into the conf stack now.
--- a/test/common-tap.js
+++ b/test/common-tap.js
@@ -18,7 +18,7 @@ exports.npm = function (cmd, opts, cb) {
   cmd = [bin].concat(cmd)
   opts = opts || {}
 
-  opts.env = opts.env ? opts.env : process.env
+  opts.env = opts.env || process.env
   if (!opts.env.npm_config_cache) {
     opts.env.npm_config_cache = npm_config_cache
   }
--- a/test/editor.js
+++ b/test/editor.js
@@ -1,0 +1,8 @@
+var fs = require("fs")
+if (fs.existsSync(process.argv[2])) {
+  console.log("success")
+} else {
+  console.log("error")
+  process.exit(1)
+}
+
--- a/test/editor.js
+++ b/test/editor.js
@@ -1,8 +1,0 @@
-var fs = require("fs")
-if (fs.existsSync(process.argv[2])) {
-  console.log("success")
-} else {
-  console.log("error")
-  process.exit(1)
-}
-
--- a/test/tap/config-edit.js
+++ b/test/tap/config-edit.js
@@ -1,0 +1,40 @@
+var fs = require("fs")
+var os = require("os")
+var mkdirp = require("mkdirp")
+var path = require("path")
+var child_process = require("child_process")
+var ini = require("ini")
+var test = require("tap").test
+var npmconf = require("../../lib/config/core.js")
+var common = require("./00-config-setup.js")
+var npm = require("../common-tap.js").npm
+
+test("saving configs", function (t) {
+  var tmp = '~/tmp/'//os.tmpdir()
+  var dirName = 'npm-global-edit-prefix-test-' + Date.now()
+  var prefixPath = path.join(tmp, dirName)
+  mkdirp(prefixPath, "0777", function (er, prefixPath) {
+    if (er)
+      throw er
+    var editorPath = path.resolve("../editor.js")
+    var command = "npm config --prefix=\"" + prefixPath + "\" edit --global"
+    var opts = { cwd: __dirname, env: { EDITOR: editorPath } }
+    npm(
+      [
+        "config",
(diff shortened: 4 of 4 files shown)


ljharb:
I can't figure out how to write a test for this. @othiym23, I'd love some guidance :-)

othiym23:
Create a directory tree that you can set as the `prefix` for an invocation of the CLI, point `EDITOR` at a simple Node script that verifies `dirname(argv[2])` exists and sets its exit code accordingly, and write out a config file pointing at that prefix. Sort of convoluted, but it shouldn't be that tough to write. Thanks, Jordan!

ljharb:
Interesting, ok thanks, I'll give it a shot.

othiym23:
Did you make any headway with the tests for this? I've now fixed the test suite for Node 0.11.16 / io.js 1.1.0 (as far as I know).

ljharb:
No, I haven't made any headway yet. However, I still get failures on latest master in io.js - also many of the tests still take forever. See https://gist.github.com/ljharb/a3366c4a21398d52f6b0 for the 22 failures I get.

othiym23:
Have you tried running `npm prune --prod && npm install`? You may not have picked up the new version of `npm-registry-mock`. I have green tests on both OS X and a pretty much pristine Ubuntu vm that's only been used for testing, not development work, from a fresh clone of the npm source tree.

ljharb:
@othiym23 After `rm -rf node_modules && npm install` it seems to be the same problem. I suspect it's because I have a non-default install path (installed with `nvm`), and I have the "~" prefix in my `npmrc`, and things like that? I'll update the gist with the second output tomorrow but so far the errors are identical to the first one.

ljharb:
@othiym23 Not sure if you did anything or if I unbroke something, but as of 7eecbd631b80fc080b31dbd7603c65fb9e2c1917 all tests pass on master on `io.js v1.1`. Yay! I'll see if I can get this PR's test written today.

ljharb:
I updated the PR with this test attempt - but I'm still getting EACCES errors even with the fix applied. Some more guidance, if you have it, would be lovely :-)

othiym23:
@ljharb I cleaned up the test a bunch in https://gist.github.com/othiym23/faa31f03736c3a747c5a. There's a lot going on in that diff, so I included the test. Here's a summary of the changes:

1. Didn't replace the `env` on the child process, because that was nuking `PATH` in the child npm process.
2. Gave the stub editor a shebang line (should be installing that as a shim for Windows support, but the tests are pretty lax about this right now and so is not a blocker for landing the test / patch).
3. Made the stub editor executable.
4. Pulled the stub editor out of `test/` and into an inline fixture in the script (using the neat multiline comment hack that @isaacs reminded me about).
5. Ran everything in a directory created and destroyed by the test case itself, following existing practice from the rest of the tests (BTW, if you want to use `$TMPDIR` in a way that will be portable across Node versions, use `osenv` instead of `os`.

I think that's probably good enough to land, so if you want, I can land that, or you can update the PR with whatever tweaks you want to make. Thanks for your patience!

ljharb:
Thanks, this is perfect! I added your commit to the PR and updated it.

othiym23:
Finally landed as 6fd0fbd, rebased and squashed, with a few small tweaks to the test to make it more robust. Thanks for the fix and your patience, Jordan!

ljharb:
Yay, thanks! :-D

Raynos:
@othiym23 @ljharb 

I think this change does not work as intended. If I run `npm install --prefix=.` it creates an `etc` folder in my CWD. If I run `npm install --prefix=./client` it will create an `etc` folder in my `client` folder.

We might have to only `mkdirp` the `etc` folder if the `global` flag is set.

Raynos:
See https://github.com/uber/npm-shrinkwrap/issues/74

ljharb:
Oops, that's clearly something I failed to consider. I'll take a crack at it today but please don't let that slow anyone else from fixing it sooner.

othiym23:
The change should probably be to move the creation of the containing `etc/` subdirectory to immediately prior to writing the config file to the directory. When I first saw that pattern used elsewhere in npm, it felt kinda janky to me, but now I see that situations like this are exactly why it exists.

Raynos:
@ljharb any progress on this?

ljharb:
None yet. I'll try to make time this week.

Raynos:
@ljharb Cool; I might look into it if I get a moment later this week; i'll let you know if I do.

matz3:
@ljharb @Raynos any updates on this?

atom0s:
Also seeing this issue when using the prefix argument. etc folders are flooding my directories anywhere an npm package gets installed. These folders are empty and harmless but they are kind of an eyesore to see everywhere an npm install is executed with a prefix set.

giowe:
I also see the issue when use --prefix on npm install... any news about that?

synthmeat:
Same here. Empty `path/etc` gets created with `--prefix path`. Maybe create it iff there's something to write in it?

finetimi:
Same here any fix on this? Appears when I run `npm install` with the `--prefix` flag from a different directory. But doesn't appear when I run `npm install` in the project root.

eyas-ranjous:
+1 @finetimi this is exactly what is happening with me right now.

justechn:
@othiym23 @Raynos @ljharb The etc folders are still an issue, can we get this reopened, or another opened in its place for a fix?

patrick-motard:
As justechn says.. 
```
npm i --prefix src . 
ls src 
```
yields:
`etc node_modules`

PR #7282, opened 2015-02-05
Document shortcut package.json/repository value

marcin-wosinek:
Looks that "npm/npm" is perfectly support as repository url, but it wasn't mentioned in documentation

### DIFF
--- a/doc/files/package.json.md
+++ b/doc/files/package.json.md
@@ -272,6 +272,10 @@ The URL should be a publicly available (perhaps read-only) url that can be hande
 directly to a VCS program without any modification.  It should not be a url to an
 html project page that you put in your browser.  It's for computers.
 
+For github repositories you can use shortcut syntax:
+
+    "repository": "npm/npm"
+
 ## scripts
 
 The "scripts" property is a dictionary containing script commands that are run


othiym23:
This is already documented in the [install docs](https://docs.npmjs.com/cli/install) [[source / context URL](https://github.com/npm/npm/blame/master/doc/cli/npm-install.md#L157)], and while it's not always a terrible idea to duplicate information, in this case it's pretty much redundant. Thanks anyway, though!

marcin-wosinek:
Thanks for the link. I would never connect those two pieces on my own

arthurvr:
Nah, might be worth it to document this there too. The page @othiym23 linked to are the docs for `npm install user/repo` while that isn't the same as the repo field in package.json.

othiym23:
Ohhhh... I misread! My apologies. I thought @marcin-wosinek was talking about repo shortcuts in dependencies. Yeah, the `org/repo` syntax in the repository field is not documented in the [`package.json` documentation](https://docs.npmjs.com/files/package.json#repository), like of the smaller "magic" affordances in the code base. I'd be happy to take a docs patch for that.

PR #7283, opened 2015-02-05
travis doesn’t need sudo

zeke:
Should make builds a lot faster, because http://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+sudo: false
 language: node_js
 node_js:
   - "0.11"


KenanY:
? https://github.com/npm/npm/pull/7283/files#diff-354f30a63fb0907d4ad57269548329e3R11

zeke:
oops.. yeah I kinda just threw this over the wall to get a conversation going, as we are updating the travis yamlz on lots of other repos. Closing..

Issue #7294, opened 2015-02-06
Unsupported URL Type: git+git+ssh when shrinkwrapping modules from private github repos


twhid:
Greetings,

shrinkwrap writes a bad protocol to npm-shrinkwrap if you originally installed from a private github repo. I.e. when the package.json dependency's protocol is `git+ssh:` then shrinkwrap generates a `git+git+ssh:` protocol.

I thought perhaps I could fix it here:
https://github.com/npm/npm/blob/6c47201a7d071e8bf091b36933daf4199cc98e80/lib/cache/add-remote-git.js#L185

But that code doesn't seem to run during shrinkwrap. I'm not familiar with the innards of npm, If anyone can point me in the right direction, I'd be happy to submit a pull request.

These steps to reproduce. You need a private github repo:

```
npm -v 
2.5.0
npm install my-github/my-module --save
```

Now, package.json:
```js
{
  "name": "npm-bug-test",
  "version": "1.0.0",
  "description": "test case to display git+git+ssh bug",
  "main": "index.js",
  "dependencies": {
    "my-module": "git+ssh://git@github.com:my-github/my-module"
  }
}
```
```
npm shrinkwrap
```

Now, npm-shrinkwrap.json
```js
{
  "name": "npm-bug-test",
  "version": "1.0.0",
  "dependencies": {
    "my-module": {
      "version": "1.0.0",
      "from": "git+ssh://git@github.com:my-github/my-module",
      "resolved": "git+git+ssh://git@github.com:my-github/my-module#a3208a093052b06b2d45f282822536ba3c65e3c3",
// etc
```

```
npm install 

0 info it worked if it ends with ok
1 verbose cli [ 'node', '/usr/local/bin/npm', 'install' ]
2 info using npm@2.5.0
3 info using node@v0.10.35
4 verbose node symlink /usr/local/bin/node
5 warn package.json npm-bug-test@1.0.0 No repository field.
6 warn package.json npm-bug-test@1.0.0 No README data
7 verbose readDependencies npm-shrinkwrap.json is overriding dependencies
8 info shrinkwrap file "/Users/me/projects/npm-bug-test/npm-shrinkwrap.json"
9 verbose readDependencies returned deps { my-module: 'git+git+ssh://git@github.com:my-github/my-module#a3208a093052b06b2d45f282822536ba3c65e3c3' }
10 verbose install where, deps [ '/Users/me/projects/npm-bug-test', [ 'my-module' ] ]
[...]

Issue #6422, opened 2014-10-07
installing modules with git slugs improperly resolves resource urls


dstokes:
```sh
$ npm install -g org/private-module
...

$ npm ls -g
npm ERR! Unsupported URL Type: git+git+ssh://git@github.com:org/private-module#hash
npm ERR!
npm ERR! If you need help, you may report this error at:
npm ERR!     <http://github.com/npm/npm/issues>

npm ERR! Please include the following file with any support request:
npm ERR!     /Users/stefansundin/npm-debug.log
```
note the `git+git+ssh://`protocol.

installing with the full git ssh url seems to fix the problem:
```sh
$ npm install -g git+ssh://git@github.com:org/private-module.git
```

here's the debug log
```log
0 info it worked if it ends with ok
1 verbose cli [ 'node', '/usr/local/bin/npm', 'ls', '-g' ]
2 info using npm@2.0.0
3 info using node@v0.10.32
4 verbose node symlink /usr/local/bin/node
6 verbose stack Error: Unsupported URL Type: git+git+ssh://git@github.com:org/private-module#hash
6 verbose stack     at parseUrl (/usr/local/lib/node_modules/npm/node_modules/npm-package-arg/npa.js:173:13)
6 verbose stack     at npa (/usr/local/lib/node_modules/npm/node_modules/npm-package-arg/npa.js:74:12)
6 verbose stack     at makeArchy_ (/usr/local/lib/node_modules/npm/lib/ls.js:281:9)
6 verbose stack     at /usr/local/lib/node_modules/npm/lib/ls.js:296:14
6 verbose stack     at Array.map (native)
6 verbose stack     at makeArchy_ (/usr/local/lib/node_modules/npm/lib/ls.js:295:22)
6 verbose stack     at makeArchy (/usr/local/lib/node_modules/npm/lib/ls.js:228:13)
6 verbose stack     at /usr/local/lib/node_modules/npm/lib/ls.js:65:13
6 verbose stack     at /usr/local/lib/node_modules/npm/node_modules/read-installed/read-installed.js:138:5
6 verbose stack     at /usr/local/lib/node_modules/npm/node_modules/read-installed/read-installed.js:251:14
7 verbose cwd /Users/user
8 error Darwin 13.4.0
9 error argv "node" "/usr/local/bin/npm" "ls" "-g"
10 error node v0.10.32
11 error npm  v2.0.0
12 error Unsupported URL Type: git+git+ssh://git@github.com:org/private-module#hash
[...]

PR #7296, opened 2015-02-08
fixes Unsupported URL Type: git+git+ssh bug

twhid:
More details here: https://github.com/npm/npm/issues/6422 and here: https://github.com/npm/npm/issues/7294.

### DIFF
--- a/lib/cache/add-remote-git.js
+++ b/lib/cache/add-remote-git.js
@@ -188,7 +188,9 @@ function resolveHead (p, u, co, origUrl, cb) {
       parsed.hash = stdout
       var resolved = url.format(parsed)
 
-      if (parsed.protocol !== "git:") resolved = "git+" + resolved
+      if (!/git(\+|:)/.test(parsed.protocol)) {
+        resolved = "git+" + resolved
+      }
 
       // https://github.com/npm/npm/issues/3224
       // node incorrectly sticks a / at the start of the path We know that the


othiym23:
Thank you! Is it too much to ask for a test?

twhid:
Should have a test, yes. I'd like to knock something up. 

However, when I `$ tap ./lib/cache/add-remote-git.js` 
```
      Error: npm.load() required
          at Object.npm.config.get (/Users/timwhidden/projects/npm/lib/npm.js:31:11)
          at Object.<anonymous> (/Users/timwhidden/projects/npm/lib/utils/git.js:18:18)
          at Module._compile (module.js:456:26)
          at Object.Module._extensions..js (module.js:474:10)
          at Module.load (module.js:356:32)
          at Function.Module._load (module.js:312:12)
          at Module.require (module.js:364:17)
          at require (module.js:380:17)
          at Object.<anonymous> (/Users/timwhidden/projects/npm/lib/cache/add-remote-git.js:3:11)
          at Module._compile (module.js:456:26)
    command: "/usr/local/Cellar/node/0.10.35_2/bin/node add-remote-git.js"
```

My inclination would be to refactor the code that generates `resolved` into it's own module and test that. I.e. everything from https://github.com/npm/npm/blob/master/lib/cache/add-remote-git.js#L188 to https://github.com/npm/npm/blob/master/lib/cache/add-remote-git.js#L201

Too invasive for this small change?

othiym23:
> Too invasive for this small change?

I think so. Tests should go into `test/tap` anyway. You can probably piggy-back on `add-remote-git` (or copy it into a new test), which already does the `npm.load()` dance for you. Sometime within the next few months, @iarna and I are going to start rewriting npm's test suite for consistency and simplicity, but until then, these tests are most useful as functional or integration tests, rather than unit tests.

Thanks for your patience!

twhid:
That's embarrassing. I ran the test on the wrong file. That's what I get for working on this after... :wine_glass: 

OK, I'll see about adding something to `/test/tap/add-remote-git`.

twhid:
Hiya,

The current test, [test/tap/add-remote-git.js](https://github.com/npm/npm/blob/master/test/tap/add-remote-git.js), tests that a mock package can install a mock dep against a real git server. To test a dependency that uses the `git+ssh` protocol would require adding a lot of setup to this test. Specifically, generating private and public keys for the current user and configuring the system properly (both the git server and the current user).

Not saying I'm not going to do it, but seems like a lot of rigmarole for a change that a simple unit test would cover. It would be a nice integration test to have, but tying it to this change seems somewhat arbitrary to me.

Best,

othiym23:
Since the underlying bug also affects every other kind of git dependency including `git+http:`, perhaps the test can do that instead of trying to do complicated mocking or trying to get an ssh server running in a test?

twhid:
Ah. Good point. I can try that.

On Fri, Feb 20, 2015 at 11:37 AM, Forrest L Norvell <
notifications@github.com> wrote:

> Since the underlying bug also affects every other kind of git dependency
> including git+http:, perhaps the test can do that instead of trying to do
> complicated mocking or trying to get an ssh server running in a test?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/npm/npm/pull/7296#issuecomment-75268925>.
>



-- 
*TIM WHID**DEN*
*Director, Front-End Engineering*
*1stdibs*
*p  *347.410.6760
*e  *tim@1stdibs.com <name@1stdibs.com>

*1stdibs <http://www.1stdibs.com/> -* *T**he Most Beautiful Things On
Earth.*
Follow us on Facebook <https://www.facebook.com/1stdibs>, Pinterest
<http://pinterest.com/1stdibs/> and Twitter <https://twitter.com/1stdibs>.

othiym23:
I went ahead and wrote a test and landed it as 533420752aba9b7c7e4abe7c5f256fc2a7ee5698. While doing so, I had to do some digging, and realized that there actually was a problem in the GitHub-specific caching code, which I fixed in 0f87f5ed28960d962f34977953561d22983da4f9. Finally, I tweaked the wording and rebased your patch and landed it as a2e04bd921feab8f9e40a27e180ca9308eb709d7.

Thanks for putting this together, and for your patience in working with me to get it merged!

twhid:
Thanks! Sorry I didn't find the time to write the test.

imsaar:
I was having the same issue of having tmp path in my "from" and I realized I was using npm 2.7.0.
I did a  `npm install npm -g` and now I got 2.11.0 and now npm shrinkwrap is generating the "from" correctly.

PR #7298, opened 2015-02-08
Two minor issues

isaacs:
The first cleans up some test folders at the start of the add-remote-git-fake-windows test.

The second handles an edge case where `outdated` returns an empty list.


othiym23:
Landed as 8a78653. Thank you for your valuable contribution to npm, @isaacs!

Issue #3405, opened 2013-05-03
Linking man pages silently fails


mathiasbynens:
`npm` doesn’t seem to install man pages for packages (using the `man` entry in `package.json`). I’ve checked using `npm install -g luamin --verbose`, and it does log the `linkMans` step, but `luamin.1` doesn’t show up in `/usr/local/share/man/man1` at all. Any instructions on how to debug this? Is there a better way to figure out where `npm` put the man pages, if anywhere?

Here’s an excerpt from the `npm install -g luamin --verbose` log:

```
[…]
npm info linkStuff luamin@0.2.5
npm verb linkBins luamin@0.2.5
npm verb link bins [ { luamin: 'bin/luamin' }, '/usr/local/bin', true ]
npm verb linkMans luamin@0.2.5
npm verb rebuildBundles luamin@0.2.5
npm verb rebuildBundles [ '.bin', 'luaparse' ]
[…]
```

([Here’s a link to luamin’s `package.json`](https://github.com/mathiasbynens/luamin/blob/master/package.json).)

Also, `man npm` etc. don’t work.

I’ve updated npm to v1.2.19 to try the latest fix for #109, but I’m still having the same issue.

busches:
@KenanY this appears to have been fixed by @bengl's PR #7299

Issue #6271, opened 2014-09-22
Missing manpages when installing a module globally


elieux:
When installing a module globally, manpages should be copied/linked to `$prefix/share/man/` as described in [package.json docs](https://www.npmjs.org/doc/files/package.json.html#man), but are not.

I discovered that the paths specified in the JSON are taken relatively to current working directory, so unless npm is called from the root directory of the module to be installed, `link.js:linkIfExists` stops on `ENOENT`.

Tested with modules jslint v0.6.4 and luamin v0.2.8.

Moreover, if I disable check for Windows in `build.js:linkMans`, it creates broken links. It seems there's a fallback to `junction` in `link.js:link` on Windows, but that feature is intended for directories and creates broken links when used on files. The Windows build I'm using is from MSYS2 project, downstream issue: Alexpux/MINGW-packages#217. I can make a separate ticket for the Windows issue, if needed.

mirabilos:
I just installed npm 3.3.4, and the manpages are missing – in `~/.npm-root/` there is not even a directory `share/`

mirabilos:
Can you please **fix** the issue instead of closing it without comment?

Issue #4768, opened 2014-02-24
Issue with man property in package.json


meandavejustice:
Following the docs [here](https://www.npmjs.org/doc/files/package.json.html#man)

I have generated my man file and pointed to it in my [package.json](https://github.com/meandavejustice/draft-dodger/blob/master/package.json#L20).

After globally installing my package I still do not have access to the man page when running `man {package-name}`.

busches:
@KenanY this appears to have been fixed by @bengl's PR #7299

PR #7299, opened 2015-02-08
install man pages correctly

bengl:
Near as I can tell, installing a package with man pages (from registry) globally was silently failing on linking the man pages all the time. It looks like #6271 has the best description of this.

The test I added is a copy of the `install-bad-man` test, except checking for a good install.

Probably fixes: #3405 #4768 #6271


bengl:
#109 also

othiym23:
Landed as a9aff38. Thanks for the patch, and thanks for the test case!

PR #7310, opened 2015-02-09
remove redundant backtick

rutsky:
Should fix examples formatting.


othiym23:
I will fix this, but unfortunately, to do so requires me to fix it in `node-semver` first, because `doc/misc/semver.md` is actually copied straight from the semver installation (and there's a test that ensures the two remain in sync. Thanks for bringing this to my attention!

othiym23:
Landed as 516688dd7072b185bf9173ff8e4c3e162d725c1e. Thanks!

Issue #3504, opened 2013-05-31
npm install {something thats in .dependencies}


jkroso:
imho should behave the same as `npm install {something thats in .devdependencies}` what that behaviour is I don't care about

smikes:
Confirmed that this bug is still present in `npm@next`, 2.5.1 as of today (2015-Feb-08).

/cc @iarna may be relevant to `multi-stage-install`

ehsalazar:
Fixed with `npm@3`. If you [update](https://github.com/npm/npm/wiki/Troubleshooting#try-the-latest-stable-version-of-npm) and continue to experience an issue, please let us know. We'll be happy to look into it.

PR #7313, opened 2015-02-09
NOT READY: add failing test for #3504

smikes:
This provides a failing test for issue #3504, `npm install foo` does
not honor the `semver` specified if `foo` is a `devDependency`.

Problems with this test:
- requires connectivity, does not use mock-registry
- does not demonstrate that `semver` is honored if `foo` is a `dependency`

### DIFF
--- a/test/tap/install-dependencies.js
+++ b/test/tap/install-dependencies.js
@@ -1,0 +1,73 @@
+var common = require('../common-tap.js')
+var test = require('tap').test
+var npm = require('../../')
+var path = require('path')
+var fs = require('fs')
+var rimraf = require('rimraf')
+var mkdirp = require('mkdirp')
+var mr = require('npm-registry-mock')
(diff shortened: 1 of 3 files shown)


smikes:
Rebased against master, squashed.

Style updated to `standard`.

I want to add a new dummy to npm-mock-registry that doesn't have any devDependencies; will create a new case over there and ref this one.

smikes:
Since this is a failing test, it is not a candidate for merging without a bugfix.  Marking as NOT READY in title.

smikes:
Blocking: Sam needs to create a PR on npm-registry-mock to implement https://github.com/npm/npm-registry-mock/issues/29

Blocking: needs a fix for the underlying bug

othiym23:
This has been around for 2 1/2 months now – can we either get some movement on this or close it for now?

smikes:
Let's kill it.

Issue #6855, opened 2014-12-05
/tmp files not cleaned up for shrinkwrapped packages


ianbamforth:
It appears that `npm install` cleans out /tmp only if the package you're installing is not shrinkwrapped.

To test, I created a very basic package (with a single dependency), publish it as 'unwrapped', then wrap (using `npm shrinkwrap`) and rename as wrapped, and publish again.  `npm install unwrapped` results in a clean /tmp folder. `npm install wrapped` results in a npm-<rand> folder being left in /tmp.

Either the code is wrong (i.e. the /tmp/npm-<rand> folder should be deleted) or the documentation needs updating to reflect the behaviour, preferably with an explanation as I can't think why they'd be treated differently!

avdd:
Just a thought: it's fair enough to defer to the user/OS regarding cleaning up tempfiles, but it would be polite to put them in a subdirectory (e.g. /tmp/npm-$USER) so they don't pollute the shared space.

ianbamforth:
avdd - not sure I agree, neither Windows nor Ubuntu will clean these up for you automatically, and in a successful install the tempfiles have no meaning / utility.  It's basically littering!

peterjmit:
To add a data point, we turned on `shrinkwrap.json` and within a couple of days our builds started failing. After some investigation we found __>15g__ of files from npm in our `/tmp` directory, and then we found this issue. 

I am not sure I agree with the labelling of this issue as a `feature-request` - on a reasonable sized app it takes 2/3 days for this to cause a failure, and there is no documentation for `shrinkwrap` or `install` that provides a warning or suggested resolution for this failure. As such I had no reasonable chance of avoiding this failure and nor will users in the future.

andiby:
Yes feature-request is wrong.
Either it is a bug that the Temp files are not deleted upon successful exit or the wrong documentation is a bug:
"Temp files are given a unique folder under this root for each run of the program, and are deleted upon successful exit." (npm-folders.md)
Already mentioned in #6853

othiym23:
[...]

PR #7324, opened 2015-02-10
candidate implementation of async tmpdir cleanup

smikes:
fixes issue #6855

### DIFF
--- a/lib/npm.js
+++ b/lib/npm.js
@@ -438,6 +438,15 @@ Object.defineProperty(npm, "tmp",
     }
   , enumerable : true
   })
+process.on("exit", function (code) {
+  if (!code && tmpFolder) {
+      require('child_process').spawn(path.resolve(__dirname, "..", "node_modules", ".bin", "rimraf"),
+                                     [npm.tmp],
+                                     { cwd: npm.config.get("tmp"),
+                                       detached: true,
+                                       stdio: "ignore" })
+  }
+})
 
 // the better to repl you with
 Object.getOwnPropertyNames(npm.commands).forEach(function (n) {
--- a/test/tap/add-remote-git.js
+++ b/test/tap/add-remote-git.js
@@ -36,6 +36,8 @@ test("install from repo", function (t) {
   npm.commands.install(".", [], function (er) {
     t.ifError(er, "npm installed via git")
 
+    t.deepEqual([], fs.readdirSync(resolve(pkg, "tmp")), "tmp directory cleaned")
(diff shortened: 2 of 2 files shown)


smikes:
/cc @othiym23 needs-review, candidate for #next-minor ?

Which is not to say that you should review it – perhaps one of your minions. :smirk_cat:

othiym23:
As far as I know, you *are* one of my minions if anyone is, @smikes. I mean, it's just me and @iarna on the core team, and if either of us is a minion, we both are (of each other).

My considered, logical reaction is: I dunno about this. It fulfills the requirement of cleaning up after npm without slowing down its perceived responsiveness, but I have questions:

1. What happens with this when something goes wrong with the removal, for whatever reason? Because I/O is going into the bit bucket, we'll have no way of signaling to the user that something untoward occurred.
2. Windows. I know that's not a question, but detached processes are kind of scary on Windows, and if, say, rimraf hangs, there will be a wedged Node process hanging around that users might not think to look for.
3. Isn't that test at least a little racy? It doesn't / can't wait for the cleanup to complete before checking to see if the temp directory is gone, and this is exactly the sort of race condition heavily-loaded CI servers with spinning disks love to expose.
4. Isn't this a lot of complexity (a detached process, magical behavior) to add for something that should arguably be handled by the operating system? As I'll comment over on #6855, @isaacs and I have talked it through, and neither of us are convinced that npm is being any more prolific in creating temporary directories now than it was in the past. Installations are growing ever more complex, it's true, but this is not a thing that is specific to npm – CI servers and the like should be running `tmpwatch` with a pretty tight timeout.

smikes:
I say kill it with one caveat: can we put in a thought about this wrt npm@3, such that git repository dependencies get sorted earlier in the process, and we add an async cleanup of tmpdir into the mix?  Realistically an async rimraf of a /tmp git repo could be fired off as soon as we know we don't need it anymore, i.e., as soon as the git-clone-to-cache completes.

othiym23:
> an async rimraf of a /tmp git repo could be fired off as soon as we know we don't need it anymore

This is a valuable insight, and one you should go add on to #6855. I don't think this needs to wait until `npm@3` to be implemented, either (although it might be convenient for @iarna to do that work against the `multi-stage` branch so she doesn't have to merge it in) – this is a self-contained feature that can land pretty much whenever.

Since we both agree in principle that this is no bueno, I'm going to close it now. Thanks for the time you put into it, Sam!

PR #7331, opened 2015-02-10
Fix npm link documentation typo.

marr:


othiym23:
Landed in 899e270. Thanks, @marr!

PR #7337, opened 2015-02-11
Improve npm verbose message to output correct runtime engine name, eg. iojs or node

bitinn:
Now that installing iojs or node bundle will replace each other, we should make it easier to figure out which engine npm is using under the hood.

```
// with iojs installed

$ node cli --verbose
npm verb cli [ 'node', '/Users/df/git/code/npm/cli', '--verbose' ]
npm info using npm@2.5.1
npm info using iojs@1.1.0
npm verb node symlink /usr/local/bin/node
```

### DIFF
--- a/bin/npm-cli.js
+++ b/bin/npm-cli.js
@@ -28,11 +28,19 @@ var path = require("path")
   , shorthands = configDefs.shorthands
   , types = configDefs.types
   , nopt = require("nopt")
+  , engine
 
 // if npm is called as "npmg" or "npm_g", then
 // run in global mode.
 if (path.basename(process.argv[1]).slice(-1)  === "g") {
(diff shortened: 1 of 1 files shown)


othiym23:
Reading from `process.execPath` feels like a kludge to me. More importantly, this is something I think is better handled in io.js / Node.js first. See iojs/io.js#269 and TooTallNate/node-gyp#564 for related discussions. For now, the tip-off is in the version number, and hopefully by the time that it's no longer unambiguous, all of `npm`, `node` / `iojs`, and `node-gyp` will have worked out a better solution.

bitinn:
Alright, I did open a ticket on io.js side before this pull request, maybe I should keep it open then? https://github.com/iojs/io.js/issues/780

PR #7338, opened 2015-02-11
Document shortcut package.json/repository value

marcin-wosinek:
#7282


othiym23:
Landed in c61aff3. Thanks for the clarification, @marcin-wosinek!

PR #7345, opened 2015-02-11
Candidate implementation of default depth=0 for outdated

smikes:
See https://github.com/npm/npm/issues/4471#issuecomment-56549227

### DIFF
--- a/doc/misc/npm-config.md
+++ b/doc/misc/npm-config.md
@@ -240,11 +240,16 @@ If true, then only prints color codes for tty file descriptors.
 
 ### depth
 
-* Default: Infinity
+* Default: -1
 * Type: Number
 
-The depth to go when recursing directories for `npm ls` and
-`npm cache ls`.
+The depth to go when recursing directories for `npm ls`, 
+`npm cache ls`, and `npm outdated`.
+
+If `depth` is the default (-1), the search depth will be `Infinity`
+for `npm ls` and `npm cache ls`; but for `npm outdated`, a
+setting of `-1` will be treated as `0` since that gives more
+useful information.
 
 ### description
 
--- a/lib/cache.js
+++ b/lib/cache.js
@@ -185,6 +185,10 @@ function ls (args, cb) {
   if (prefix.indexOf(process.env.HOME) === 0) {
     prefix = "~" + prefix.substr(process.env.HOME.length)
   }
+
+  // default depth for `ls` is Infinity (cf. `outdated`)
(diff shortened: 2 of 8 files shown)


doc/misc/npm-config.md
```diff
@@ -240,11 +240,16 @@ If true, then only prints color codes for tty file descriptors.
 
 ### depth
 
-* Default: Infinity
+* Default: -1
```
othiym23: Is there a way to do this without a magic number? This feels counterintuitive to me, and not really in keeping with the former range of `0..Infinity`.

othiym23:
See my comment above -- changing the default to `-1` makes for a pretty heavyweight diff, and I'm uncomfortable with assigning a special meaning to a number that's not in keeping with the rest of the parameter's domain.

Also, because this is a significant change to behavior, I'd like to slate this for inclusion in `npm@3`. Any objections to that?

smikes:
Second thing first: putting this for `npm@3` makes good sense to me, no rush.  I was poking around the edges of making `npm update -g` be sensible and saw that I could pick this one off easily.

First thing: using -1 as the default seemed the least bad option.  I considered all of these --

 * create a new config param just for outdated-depth
 * leave the default at Infinity and have outdated interpret Infinity as 0
 * leave the default at Infinity and have outdated use the depth "backward" (i.e. Infinity = 0, 0 = Infinity)
 * change the default to 0 for everything

All of those have ugly side effects, including breaking changes to other subcommands (`ls` and `cache ls`).  Setting the default to something outside the previously valid range of `0..Infinity` seemed like the right thing, and the nearest thing outside that range is `-1`.

Oh, I suppose another possibility would be to modify `nopt` to detect whether an option has the default value because it was **set** to default, or because it was never set.  

It's the meta-null problem, and it's insoluble in general, so I wrote a gross hack as a straw solution so we can lay all the bad options on the table and pick the least bad.

othiym23:
> leave the default at Infinity and have outdated interpret Infinity as 0

FWIW, this seems like the best solution, because `Infinity` rhymes with "default value" here. Outdated has a different default value, and really, I don't think it's super weird to require people to type e.g. `--depth=99999` or whatever in the rare cases they want to see outdated dependencies all the way down. @iarna, does that sound OK to you too, or am I taking the lazy way out here?

othiym23:
Landed as 8b98f0e, and had a #yolo moment and am including this as a patch-level change because it now only affects `npm outdated`.

douglasduteil:
+1

PR #7355, opened 2015-02-13
Candidate implementation of default depth=0 for outdated, preserving Infinity as default

smikes:
A new pR for a new day! Candidate fix for #4471, alternative to #7345


othiym23:
Landed as 8b98f0e, because I liked this one best. Thanks for the tests and docs changes, @smikes!

PR #7367, opened 2015-02-15
don't use $HOME/.npm as cache directory when running as root

tom95:
When installing packages globally using root we can't use $HOME/.npm as
$HOME is most often not changed by sudo and so it's still the user's own
home dir.  Now, if a local npm install is run it may fail since folders in
the cache dir are owned by root. In particular when the local npm install
tries to grab newer package versions. Instead we install to the temp
dir when run as root.

### DIFF
--- a/lib/config/defaults.js
+++ b/lib/config/defaults.js
@@ -95,7 +95,11 @@ if (home) process.env.HOME = home
 else home = path.resolve(temp, "npm-" + uidOrPid)
 
 var cacheExtra = process.platform === "win32" ? "npm-cache" : ".npm"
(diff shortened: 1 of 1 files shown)


othiym23:
While there have been issues in the past with cache permissions when running with elevated privileges, we've put a considerable amount of work into fixing those bugs. I _believe_ we've gotten them all (as of `npm@2.4` or so), but if we haven't, I'd rather see issues or patches from led against those holes than have separate caches. As such, we won't be taking this PR.

I appreciate you taking the time to put this together, though! Thank you!

tom95:
Thanks for the notice! I'll try that version.

PR #7383, opened 2015-02-18
fix broken andFinishTracker call

edef1c:
not sure if this is correct, but it definitely needs to pass _a_ tracker
there. cc @iarna @othiym23


othiym23:
As I noted on the commit, this has already been fixed and the fix rebased onto the branch upstream. Thanks for your time, Nathan!

PR #7389, opened 2015-02-19
Added an --if-present flag for not erroring on unspecified scripts.

### LINKED ISSUE #7354: npm run-script flag to exit with 0 when script doesn't exist
We have a CI setup with a generic template that triggers Travis-like hooks like `after_failure` using `npm run-script`. Previously this worked even when you had not specified the scripts (which was nice since you might not actually want to do anything on failure). 

Ever since #2959 was fixed with the introduction of 0aefae9bc2598a4b7a3ee7bb2306b42e3e12bb28 the builds that don't have these scripts specified started failing. It would be nice if there was some flag that made run-script return 0 when the script isn't available, e.g. `npm run --if-present after_success`.

jussi-kalliokoski:
Fixes #7354.

### DIFF
--- a/doc/misc/npm-config.md
+++ b/doc/misc/npm-config.md
@@ -385,6 +385,16 @@ A proxy to use for outgoing https requests. If the `HTTPS_PROXY` or
 `https_proxy` or `HTTP_PROXY` or `http_proxy` environment variables are set,
 proxy settings will be honored by the underlying `request` library.
 
+### if-present
+
+* Default: false
+* Type: Boolean
+
+If true, npm does not exit with non-zero when `run-script` is used for a script
+that is not defined in the `scripts` list of the package.json. This option can be
+used when it's desirable to optionally run a script when it's present and fail if
+the script fails, for example to provide a generic CI setup.
+
 ### ignore-scripts
 
 * Default: false
--- a/lib/config/defaults.js
+++ b/lib/config/defaults.js
@@ -160,6 +160,7 @@ Object.defineProperty(exports, "defaults", {get: function () {
     , group : process.platform === "win32" ? 0
             : process.env.SUDO_GID || (process.getgid && process.getgid())
     , heading: "npm"
+    , "if-present": false
     , "ignore-scripts": false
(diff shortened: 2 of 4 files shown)


doc/misc/npm-config.md
```diff
@@ -385,6 +385,14 @@ A proxy to use for outgoing https requests. If the `HTTPS_PROXY` or
 `https_proxy` or `HTTP_PROXY` or `http_proxy` environment variables are set,
 proxy settings will be honored by the underlying `request` library.
 
+### if-present
+
+* Default: false
+* Type: Boolean
+
+If true, npm does not exit with non-zero when `run-script` is used for a script
```
othiym23: I'd like to see this more explicitly describe why you might want to use this option.

jussi-kalliokoski: How about this?

jussi-kalliokoski:
@othiym23 it's a good thing to fight feature creep and poor wording in documentation, I appreciate it. ;)

I'm not sure how to spell the use case out in the docs, but let me try to explain and maybe we can reach something.

> Because there may be scripts that are present in the CI environment (like, say, a special-purpose server used for tests but not expected to be there in production?

Yes, so let's say that a certain build wants to do something to set up the environment in the CI, e.g. do some special cleanup before starting the build. All builds don't want to do this, so it wouldn't be nice to require them to have `"before_build": "true"` in the package.json just because of the few odd cases (just like Travis CI doesn't require you to specify `before_script`, but you can). You also want the build to fail if the pre-build cleanup fails (the reason not to use say `preinstall` would be that the clean up would be CI-specific and you might not want to do this cleanup on every install in local mode, or when running in production).

othiym23:
I did some wordsmithing and came up with something that's not too different from what you originally proposed, merged your changes into `test/tap/run-script.js` (there's been a lot of action around `run-script`s this week), and landed the whole thing as f33e8b8ff2de094071c5976be95e35110cf2ab1a.

Thanks so much for working with me on this! I think this will be useful to a relatively small number of organizations, but my gut tells me those will be large organizations that will find this very useful.

PR #7400, opened 2015-02-20
Fixing code fence

pdehaan:
Because it looks like using "javascript" was causing the rest of the page to be rendered inside a code block.

### DIFF
--- a/doc/misc/semver.md
+++ b/doc/misc/semver.md
@@ -106,7 +106,7 @@ similar risk on the *next* set of prerelease versions.
 The method `.inc` takes an additional `identifier` string argument that
 will append the value of the string as a prerelease identifier:
 
-```javascript
+```shell
 > semver.inc('1.2.3', 'pre', 'beta')
 '1.2.4-beta.0'
 ```


pdehaan:
From what I can tell, this failed Travis 'cause it doesn't match some output from the [semver](https://github.com/npm/node-semver/blob/master/README.md#prerelease-identifiers) module's docs.

Not sure what's going on...
https://github.com/npm/npm/blob/master/doc/misc/semver.md looks fine on GitHub.
https://github.com/npm/node-semver/blob/master/README.md looks fine on GitHub.
https://docs.npmjs.com/misc/semver looks objectively "poor".

othiym23:
`doc/misc/semver.md` is a clone of the `README.md` in `node-semver`, and there's a test to guarantee that the two stay in sync, so that explains that. The issue is that docs.npmjs.com is lagging a few versions behind www.npmjs.com in the version of [`marky-markdown`](http://npm.im/marky-markdown) it uses to render documentation, and the old versions aren't as good at dealing with code fences, as you've noticed. You could do us a favor by opening an issue (or leaving a comment on an existing issue) over on npm/docs.npmjs.com. Thanks!

Issue #7417, opened 2015-02-22
Add examples to 'npm update'


smikes:
npm update docs should have examples.  Ref #6549

smikes:
First cut at these docs on https://github.com/smikes/npm/tree/add-checkout-examples

othiym23:
Closing in favor of #7420.

PR #7420, opened 2015-02-22
first draft of update examples

smikes:
ref issue #7417

### DIFF
--- a/doc/cli/npm-update.md
+++ b/doc/cli/npm-update.md
@@ -8,7 +8,7 @@ npm-update(1) -- Update a package
 ## DESCRIPTION
 
 This command will update all the packages listed to the latest version
-(specified by the `tag` config).
+(specified by the `tag` config), respecting semver.
 
 It will also install missing packages. As with all commands that install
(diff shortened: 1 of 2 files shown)


othiym23:
These are fantastic. Thank you very much for putting them together.

Have you confirmed that all the scenarios you describe work that way in practice?

smikes:
How about a unit test?

smikes:
Note that the unit test depends on 7477, if missing it fails with

```
$ tap test/tap/update-examples.js
/Users/smikes/src/github/npm/node_modules/npm-registry-mock/node_modules/hock/lib/hock.js:385
          throw new Error('No Match For: ' + req.method + ' ' + req.url);
                ^
Error: No Match For: GET /abbrev
```

othiym23:
Tweaked a tiny bit (so tiny that you probably won't even spot it), rebased, and landed as 5bc70e6cfb3598da433806c6f447fc94c8e1d35d. This is excellent, Sam. Big thanks for putting it together.

Issue #6566, opened 2014-10-25
npm update does not honor shrinkwrap file


hueniverse:
Calling `npm install` and `npm update` produces different versions of dependencies.

smikes:
I created a PR that implements part of this functionality.  Rather than going the full way with having `npm update --save` ignore and then re-generate the shrinkwrap file, I just added code to make all `npm update` operations fail with an error message if `npm-shrinkwrap.json` is present in the package directory.

Comments and suggestions are very welcome, either here or on the PR, #7423

othiym23:
This has been moved to the [npm roadmap](https://github.com/npm/npm/wiki/Roadmap), which we're using instead of the confusing `next-*` labels now.

iarna:
This indeed was done with 3.x! And `npm@3` is `npm@latest` these days. =)

Issue #6048, opened 2014-08-27
difference between "install" & "update" git dependencies 


stemail23:
npm version 1.4.24

I have a dependency in my package.json of the format 

`"MODULE": "git+https://USERNAME:PASSWORD@github.com/ORG/MODULE#v1.0.0"`

When I try to execute 

`npm install MODULE`

I'm prompted to enter the github username and password, despite the fact that they're in the dependency path. Weirdly, if I run 

`npm update MODULE`

The dependency installs correctly without prompting for the github credentials!

PR #7423, opened 2015-02-22
update: presence of `npm-shrinkwrap.json` blocks update

smikes:
Candidate fix for issue #6566

Maybe also fix for #6048

This seemed hard: 

> The way @iarna and I want this to work is npm update --save will update both npm-shrinkwrap.json and package.json, but npm update alone in the presence of npm-shrinkwrap.json should be a no-op / WARN that shrinkwrapping is blocking updates and you should use npm update --save.

So I did the simpler thing first; this change just stops `update` from running in the presence of an `npm-shrinkwrap.json` file.  My proposed workflow for updating a shrinkwrap file, after this patch, is:

```
rm -f npm-shrinkwrap.json
npm update --save
npm shrinkwrap
```

This way `npm-shrinkwrap.json` functions as a lock, preventing `npm update` from running.  To remove the lock, you must remove the shrinkwrap file.  To restore the lock, you must run `npm shrinkwrap` again.

In #6566, the proposal was that `npm update --save` would remove the lock and reapply it.  I think it's better for the `npm` commands to run at a more granular level, and allow users to evolve scripts around them as needed, than to have a shrinkwrap-related workflow embedded in `npm update --save`.

Comments/bikeshedding/etc v welcome.

### DIFF
--- a/lib/update.js
+++ b/lib/update.js
@@ -12,6 +12,8 @@ update.usage = "npm update [pkg]"
 var npm = require("./npm.js")
   , asyncMap = require("slide").asyncMap
   , log = require("npmlog")
+  , fs = require("fs")
+  , path = require("path")
 
   // load these, just so that we know that they'll be available, in case
   // npm itself is getting overwritten.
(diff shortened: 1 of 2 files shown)


smikes:
> In #6566, the proposal was that npm update --save would remove the lock and reapply it. I think it's better for the npm commands to run at a more granular level, and allow users to evolve scripts around them as needed, than to have a shrinkwrap-related workflow embedded in npm update --save.

Since that's not the medium or long-term plan, this PR is not ready to patch yet.  I will mark it as such in the title.

smikes:
Reading this spec again: https://github.com/npm/npm/issues/6566#issuecomment-61334851

> The way @iarna and I want this to work is npm update --save will update both npm-shrinkwrap.json and package.json, but npm update alone in the presence of npm-shrinkwrap.json should be a no-op / WARN that shrinkwrapping is blocking updates and you should use npm update --save.

This PR implements the second half; `npm update` in the presence of a shrinkwrap file.  It does not document the shrinkwrap-updating behavior that `npm update --save` should give.

Now, this could go in as-is (with some documentation added), or it could be held over for a more complete solution.

Three ways I could take this:

1. polish this up, fixing formatting and adding documentation, for inclusion now, with the understanding that `npm@3` will implement the complete desired behavior ; 
2. close the PR, since `npm@3` is taking this in a different direction
3. actually solve the "update shrinkwrap.json" file, but not as elegantly as npm@3 will.  So, for example, move `npm-shrinkwrap.json` away, run the update command, run `npm shrinkwrap`, and then display the diff between old and new shrinkwrap files.

My preferences are towards option 1 or 2.  I will put this on the 'needs comment' tab.

smikes:
Withdrawing this for now.  Maybe revisit after `npm@3`

PR #7424, opened 2015-02-22
fix typo in test dir name: "build-alread-built" => "build-already-built"

smikes:


othiym23:
Rebased and landed as 53a47baf2ba674d2f70df99d7c527938de7c8dc4. Everything counts, even in small amounts! Take that, Martin Gore! Thanks for noticing and fixing this, Sam!

PR #7442, opened 2015-02-24
WIP: make npm code conform to `standard`

othiym23:
This time based against the correct branch.

This is probably too big a diff to effectively review, so once standard is passing (and the tests are passing multi-stage as well), we'll just have to cross our fingers and merge this. So far it's down from ~14,000 warnings to 1,569, most of which are comma-first clauses that need to be reformatted.

It doesn't look _that_ different.

r:@iarna

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@ npm-debug.log
 /node_modules/marked
 /node_modules/require-inject
 /node_modules/sprintf-js
+/node_modules/standard
 /html/api/
 /html/partial/
 /html/doc/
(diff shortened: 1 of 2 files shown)


othiym23:
After discussing this with @iarna, I decided the least risky way to make this happen was just to wait until we have the `multi-stage` tests in a good state (i.e. all passing) and then just redo this again from scratch. Yes, it's a fair amount of work, but it should go much faster the second time, and I'm not at all confident that my DVCS skills are up to the task of accurately rebasing this branch without completely breaking something in the process. Soon! :eyes:

PR #7465, opened 2015-02-26
Hooking into exec code for git, logging error if fails

### LINKED ISSUE #7392: Log exact command line when shelling out to `git
ref #5244 

It sounds like you have solved the problem you were experiencing.

> I think it would be worthwhile updating the error to include the exact Git command that npm is executing in order to aid people troubleshooting this.

This is a good feature request. I am not sure if it has been done already in the latest version of npm -- I don't use git-repo dependencies very often -- but I will forward this to Forrest since it's a feature request

/cc @othiym23 ; feature request: log exact command line passed to git; log has git clone git://github.com/michaelficarra/cscodegen.git which is the logical command line but does not include e.g., the target directory (C:\Users\DTuppeny\AppData\Roaming\npm-cache_git-remotes\git-github-com-michaelficarra-cscodegen-git-c4b567f3)

thriqon:
Printing out the actual arguments git was called with if it fails.

Tests are 'ok'.
2015-02-25-203652_1920x1080_scrot

fixes #7392

### DIFF
--- a/lib/utils/git.js
+++ b/lib/utils/git.js
@@ -19,7 +19,13 @@ function prefixGitArgs () {
 
 function execGit (args, options, cb) {
   log.info("git", args)
-  return exec(git, prefixGitArgs().concat(args || []), options, cb)
+  var fullArgs = prefixGitArgs().concat(args || []);
+  return exec(git, fullArgs, options, function (err) {
+    if (err) {
+      log.error("git ", fullArgs.join(' '));
+    }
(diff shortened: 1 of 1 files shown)


othiym23:
Rebased, tweaked, and landed as 44da66456b530c049ff50953f78368460df87461. Normally I ask for tests for these patches, but the difficulty of writing the test is out of proportion with the complexity of the feature, so it's fine. Thanks for your time and for putting this together!

PR #7474, opened 2015-02-26
Add heading to man-page for `npm version`

Zearin:


othiym23:
We've since cleaned up the documentation around `npm version` in such a way that this heading feels superfluous. Thanks for taking the time to submit this, though!

PR #7096, opened 2015-01-08
update warning from other PR

smikes:
This PR contains just the warning about `npm update -g`

PR #7476, opened 2015-02-26
re-apply #7096

kenany:
Looks like #7096 got clobbered(?) by the rebased-commit of #7066 (https://github.com/npm/npm/commit/560c00945d4dec926cd29193e336f137c7f3f951)... or something. So I think the solution would be to apply e309df642de33d10d6dffadaa8a5d214a924d0dc... again?

/cc @smikes

### DIFF
--- a/doc/cli/npm-update.md
+++ b/doc/cli/npm-update.md
@@ -20,6 +20,15 @@ packages.
 If no package name is specified, all packages in the specified location (global
 or local) will be updated.
 
+## WARNING
+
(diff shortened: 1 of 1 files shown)


smikes:
Oops!

Turns out my understanding of update was wrong, anyway.  Update **does** respect semver, it just updates too many things all at once.  Working on a different issue (https://github.com/npm/npm/pull/7420#issuecomment-76132669) and some unit tests to prove that.

KenanY:
Then this incident was meant to be!

PR #7477, opened 2015-02-26
npm: define properties of commands as configurable

smikes:
make npm commands configurable: true by default
allows npm.commands.foo to be replaced by mock fn in testing


othiym23:
Rebased and landed as bd2b2dd3787f28ce235661bf470261b7ff18f396. Thanks, Sam!

PR #7486, opened 2015-02-27
Adds function call to gitEnv in checkGitDir

functino:
In checkGitDir when executing `git config --get remote.origin.url` the gitEnv function is passed to git.whichAndExec instead of the result of that function.

The typo was introduced with this commit: https://github.com/npm/npm/commit/00ecb6101422984696929f602e14da186f9f669c#diff-949ff1cc9c69a73d643a809580064773R84

### DIFF
--- a/lib/cache/add-remote-git.js
+++ b/lib/cache/add-remote-git.js
@@ -94,7 +94,7 @@ function checkGitDir (p, u, co, origUrl, silent, cb) {
 
     git.whichAndExec(
       [ "config", "--get", "remote.origin.url" ],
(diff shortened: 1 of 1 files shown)


othiym23:
Aaaaaaa! Thank you for spotting this! I really wish there were tests that prevented this from going unnoticied, but the existing git tests were hard enough to write as it is. As such, I'm not going to insist on a test for this.

othiym23:
Rebased and landed as c8258f31365b045e5fcf15b865a363abbc3be616. Thank you very much!

Issue #7510, opened 2015-03-01
`npm update -g` should exclude `npm`


smikes:
Here's my thinking: even though `npm update -g` is now _less broken_, it's still partly broken in the following way:

run npm update -g
... several packages scheduled for update ...
... one package fails ...
... attempt rollback ...
npm is borked

My proposal (which would be a breaking change, maybe for npm@3 ?) is the following:

`npm upgrade` should be added as an alias for `npm install -g npm@latest`

on `npm update -g`, do the following:
- produce list of outdated global modules
- check if npm is in list, and remove it
- run update -g of global modules not including npm
- IF npm was out of date, write message to the effect of 'npm is out of date and can be updated by running npm install -g npm@latest'

In the case of `npm update -g foo bar npm`, I believe the same filtering should happen, but an additional warning message should be written: "you cannot update `npm` at the same time as another package"

The edge case is `npm update -g npm`, when `npm` is the only package explicitly named.  I think this should probably **not** be allowed, for consistency's sake.

This proposal inspired by yet another `npm update -g` broke everything post, this time relatively recent: https://github.com/joyent/node/issues/9291 (with npm 2.5.1 and node 0.12)

othiym23:
What @iarna and I have discussed is reprioritizing the upgrade (or downgrade, as the case may be) of npm to be either at the beginning or the end of the upgrade process. What would be _clever_, and kinda nice, would be to do it first, and then use the upgraded npm to do the rest of the upgrades. What would be _safest_ would be to do it last, so npm is left untouched if it decides to go ham on one of the other installs.

I'm not in favor of removing it from the update process altogether. Having lots of special-case logic for npm to deal with itself junks up the code base with special-case logic and makes npm more complicated to explain ("you gotta do this one thing unless npm is involved, in which case you gotta do this other thing, and npm will upgrade all the packages unless npm itself is involved", etc.).

smikes:
That sounds like a plan: I will put together a patch for update (separate from the PR re upgrade) that checks if npm is in the 'outdated' list and if so pulls it out to do separately.  Then whether it goes first or last is not terribly important, just reordering two lines. 




I still believe the "npm upgrade" issue is worthwhile, and there's a PR for that. I finally have the language to talk about why it matters: "npm upgrade" is needed when npm is not installed under 'prefix'.

othiym23:
[...]

Issue #6751, opened 2014-11-20
There should be an "npm upgrade" command


smikes:
Right now the instructions I give when I ask people to get a new `npm` are fractured by OS and distribution --

`npm i -g npm` works for many people, except for those who need
`sudo npm i -g npm`.

Then there are people who installed node via package manager `brew` or `apt-get`, where sometimes the target for global installs is distinct from the installation..

Are they on Windows?  Then we better just point them at the troubleshooting guide ( https://github.com/npm/npm/wiki/Troubleshooting#upgrading-on-windows ) because there's too many decisions to be made there...

So...

There should be an npm command `npm upgrade` that only does one thing: finds the currently-executing `npm`, determines if the current user has adequate privilege to upgrade it, gracefully asks for elevation if necessary, and gracefully attempts an upgrade of npm -- with failure resulting in restoration of the old npm (still in a usable state) and a clear message about how to get help.

Bikeshedding?

smikes:
Still worthwhile imo.

PR #7513, opened 2015-03-01
lib, doc: new command `npm upgrade`

smikes:
Ref issue #7510 

I thought about it a bit more and I realized that the desired behavior of `npm upgrade` is distinct from `npm install -g npm@latest`.  The latter only works to upgrade the currently-running `npm` when `npm` is installed under its own prefix (as is normal on OSX).

But on some systems, notably Windows, and with the official Red Hat and Ubuntu packages, `npm` is installed as a local binary (e.g. in /usr/local/bin) but npm's prefix is different (e.g., /usr/local/lib/nodejs).  This causes an `npm install -g npm@latest` to successfully install an updated npm in a different place, leaving the system npm at its existing version.  This leaves the system with two installed `npm` versions, generally causing problems.

So really what `npm upgrade` should want to do is figure out where the **currently running** npm is, and set up an install/update of **that** npm -- whether it's in `prefix` or not -- to `latest`.  If that directory not normally writable by `npm`, then the command will fail with EACCES and the user will be directed to use `sudo` or `Run As Administrator` (if on Windows).

This PR is a first cut at that.  As written the `npm upgrade` command is ugly, using `__dirname` to find the parent directory three levels above `upgrade.js`.  I set `global`, `prefix`, `tag` and `depth` explicitly in case the user manages to get weird stuff in their config file.  Possibly worth setting `registry` too?

No test yet, as this is a proof of concept, and also, I keep blowing away my changes when I try to test them :-P

This will require extensive testing on Windows and with default linux distribution.

However the upsides are:
- no more 'upgrading is a pain on Windows, follow these instructions'
- no more 'upgrading is a pain on Linux, follow these instructions'
- brew/apt-get/etc. users who expect a `foo upgrade` will now be happy

### DIFF
--- a/doc/cli/npm-upgrade.md
+++ b/doc/cli/npm-upgrade.md
@@ -1,0 +1,37 @@
+npm-upgrade(1) -- Update npm
+============================
+
+## SYNOPSIS
+
+    npm upgrade
+
+## DESCRIPTION
+
+This command updates the currently-running `npm` to the `latest`
+version that has been published on the `npm` registry.
+
+This command is similar to `npm install -g npm@latest` when `npm` is
+installed as one of the global `npm` modules.
+
+When `npm` is not installed in the location specified by `prefix`,
+this command attempts to upgrade the running `npm`; since `npm` may
+not have permission to write to its own installation directory, you
+may need to use `sudo` or a similar command to run with administrative
+permissions.
+
+## WARNINGS
+
+If this command fails, it may leave your `npm` in a broken state.  You
+can fix this on a Unix-like system by installing `npm` from scratch:
+
+```
+curl -L https://npmjs.org/install.sh | sudo sh
+```
+
+On Windows, it may be necessary to reinstall `node`.
+
+## SEE ALSO
+
+* npm-install(1)
+* npm-update(1)
+* npm-install(3)
--- a/lib/npm.js
+++ b/lib/npm.js
@@ -97,6 +97,7 @@ var commandCache = {}
               , "set"
               , "get"
               , "update"
+              , "upgrade"
               , "outdated"
               , "prune"
               , "pack"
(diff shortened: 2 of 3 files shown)


KenanY:
Also ref. https://github.com/npm/npm/issues/6751 (+ some closed issues asking for a command like this)

othiym23:
As much as people may ask for a command like this, out of symmetry with `homebrew` and `apt` and whatnot, npm is different from other systems in that it's pretty comfortable treating itself as just another global package.

What I would like to see is that we continue to work on making npm more transactional and otherwise robust, so that people can do things like run `npm upgrade -g` with confidence that it's not going to wreck their world. Thanks for putting this together, and some of the documentation can (and should?) be merged into `npm-update.md`, but I'm not going to land this.

smikes:
Okay.

Even though `npm` is comfortable treating itself as a global package, though, there will still be installations (Windows, various vendors) where it's done another way.

othiym23:
Vendored distributions are not the npm team's responsibility – I don't think anyone on the Debian _or_ the npm side wants there to be a tighter coupling between the two projects, for instance, and while I could probably write a book about all the interactions between the npm and homebrew teams, I think the safest thing to say is that it's best to leave the responsible parties to figure out how they want to work independently, and the npm team should concentrate on making it the best, most solid user experince for its direct consumers.

Re: Windows, there is still a ways to go to improve that experience, but we have made small but important improvements to making npm easier to upgrade on Windows, and I'm confident that we can get it as solid as it needs to be at least with io.js-distributed npm before the end of this year. My hope is that whatever work we're able to do to improve that experience in one node distribution will filter to the other(s), but I'll be happy if we can get it right in just one to start. I do feel pretty strongly that the correct way to do that is to make sure the `install` and `upgrade` paths are not doing surprising things when npm is the package being upgraded, and I also think `npm@3` is going to include some substantial improvements (as well as a few showstopper regressions, no doubt) in that regard.

PR #7523, opened 2015-03-03
Catch missing tmpdir exception

KarimGeiger:
When running npm with a node version below 0.8 os.tmpDir() is not defined and raises an exception. The then occurring error message is meaningless and leads to a time-consuming Google session. With this patch npm dies with error code 1 and tells that a newer version is required.

### DIFF
--- a/lib/config/defaults.js
+++ b/lib/config/defaults.js
@@ -86,7 +86,13 @@ else stableFamily = stableFamily.major + "." + stableFamily.minor
 
 var defaults
 
-var temp = osenv.tmpdir()
+try {
+  var temp = osenv.tmpdir()
+} catch (e) {
+  console.log("You need node v0.8 or higher to run this program.")
+  process.exit(1)
+}
+
 var home = osenv.home()
 
 var uidOrPid = process.getuid ? process.getuid() : process.pid


othiym23:
In theory, at least, npm still supports node 0.8. What we probably need to do is either pin the dependency on `osenv` to a version that still has the polyfill for `osenv.tmpdir()`, or publish a new version that restores that polyfill. Either way, bailing out isn't what npm should be doing here. Thanks for the contribution, though!

othiym23:
Actually, I'm not sure what you were encountering, because this isn't a problem on 0.8:

```
bauchelain% nave use 0.8.3 node
Already installed: 0.8.3
using 0.8.3
> var osenv = require('osenv')
undefined
> osenv.tmpdir()
'/var/folders/bf/1f70gl7x2_g0s1dchcrw97xm0000gn/T/'
```

How did you encounter this error?

othiym23:
Ohh... I misread entirely – you were talking about _below_ 0.8. My apologies! Either way, we've documented in a number of places that 0.6 is no longer a supported version of node to use with npm, but we try not to clutter up the code with version checks like this. If you're on 0.6, you're on your own.

KarimGeiger:
Yeah well I think a notification wouldn't be that worse since one more "if" won't hurt anyone. A missing check just lets newbies like me who think it's okay to install npm through the package manager on debian/ubuntu think: "okay, npm doesn't catch those easy-peasy errors.. I might not switch to Node.JS if the dependency manager is that broke". But of course, that's just my opinion... kind of. https://twitter.com/rdegges/status/572827517893144577

PR #7526, opened 2015-03-03
Handle scoped packages in the dedupe command

KidkArolis:
WIP.

Scoped packages now get deduped just like unscoped packages.
The fix is in the logic that scans node_modules looking for child packages.

I'm struggling to get a test going for this. As you can see in the PR, I was trying to reuse the `dedupe` and scope the `minimist` package to be `@scoped/minimist` in the setup. However, I think this won't work since dedupe is running `npm install` behind the scenes https://github.com/npm/npm/blob/448efd0eaa6f97af0889bf47efc543a1ea2f8d7e/lib/dedupe.js#L197. So the test is currently failing.

### DIFF
--- a/lib/dedupe.js
+++ b/lib/dedupe.js
@@ -314,11 +314,28 @@ function readInstalled (dir, counter, parent, cb) {
   })
 
   fs.readdir(path.resolve(dir, "node_modules"), function (er, c) {
-    children = c || [] // error is ok, just means no children.
-    children = children.filter(function (p) {
-      return !p.match(/^[\._-]/)
-    })
-    next()
+    children = children || [] // error is ok, just means no children.
+    // check if there are scoped packages.
+    asyncMap(c || [], function (child, cb) {
+      if (child.indexOf('@') === 0) {
+        fs.readdir(path.resolve(dir, "node_modules", child), function (er, scopedChildren) {
+          // error is ok, just means no children.
+          ;(scopedChildren || []).forEach(function (sc) {
+            children.push(child + "/" + sc)
+          })
+          cb()
+        })
+      } else {
+        children.push(child)
+        cb()
+      }
+    }, function (er) {
+      if (er) return cb(er)
+      children = children.filter(function (p) {
+        return !p.match(/^[\._-]/)
+      })
+      next();
+    });
   })
 
   function next () {
--- a/test/tap/dedupe-scoped.js
+++ b/test/tap/dedupe-scoped.js
@@ -1,0 +1,71 @@
+var test = require("tap").test
+  , fs = require("fs")
+  , path = require("path")
+  , existsSync = fs.existsSync || path.existsSync
+  , rimraf = require("rimraf")
+  , mr = require("npm-registry-mock")
+  , common = require("../common-tap.js")
+
+var EXEC_OPTS = {}
+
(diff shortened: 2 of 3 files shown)


iarna:
@smikes Thanks for the heads up!

Yes, scoped modules dedup in the multi-stage (npm@3) branch.

othiym23:
This probably needs to be there before private modules are available on the registry, so I'll see if I can help you get this one over the line, @KidkArolis.

KidkArolis:
1. I think there's still an edge case where deduping doesn't work, in one of my projects I see the following tree after dedupe  - notice 2 jquery's. It might be a problem with the dedupe itself - not necessarily related to this patch.

```
├── @qubit/css@0.1.6
├─┬ @qubit/emitter@1.0.0
│ └── @qubit/each@1.0.0
├── @qubit/event-kit@0.8.0
├─┬ @qubit/hover@1.0.10
│ └── @qubit/maxz@1.0.1
├── @qubit/inherits@1.0.0
├── @qubit/jquery@1.8.2
├── @qubit/jquery-ui@1.11.3
├── @qubit/underscore@1.5.2
└─┬ @qubit/vanish@1.0.0
  └── @qubit/jquery@1.8.2
```

2. Regarding the tests - I'm not sure what the approach should be. Should I somehow precreate a scoped package structure in the `dedupe-scoped` dir? Or should I do what `dedupe` does and somehow install them from the mock registry? Which case, how do I put some scoped packages into the fake registry.

othiym23:
I went ahead and pushed this through by rewriting the test to use the little-known alternate command `npm find-dupes`, which doesn't talk to the network* and just shows the duplicates it finds. Landed as c56cfcd79cd8ab4ccd06d2c03d7e04030d576683 for @KidkArolis's change, and 5faad0d8cddc2aab4fc096414c20017fa61c3d0d for my rewrite of the test. Thanks, @KidkArolis!

_* At some point we're going to need to get scoped packages in the mock registry, but it's not trivial, and this change is only meant to get us to `npm@3`, where dedupe should be in much better shape._

othiym23:
@KidkArolis If you figure out what's up with that edge case, don't be afraid to open another issue for it. Even if it just contains the output from running `npm dedupe`, that should help us figure out what's going on.

KidkArolis:
Awesome! I was using find-dupes to manually test the code while working on the patch, but didn't think of using it tests, neat!

The reason `@qubit/jquery@1.8.2` isn't being deduped in my case is because `@qubit/vanish@1.0.0` is a dev dependency. So this is the expected behaviour after all.

However, I wanted to question that a bit. In my case, `vanish` is a dev dependency since we only use it in tests, however, we do want jquery to be deduped (jquery is one of those problematic libraries when it's not deduped). I personally can't think of any reason to not dedupe dev dependencies. The following lines would have to be removed to achieve that:

```js
    // ignore devDependencies.  Just leave them where they are.
    children = children.filter(function (c) {
      return !pkg.devDependencies.hasOwnProperty(c)
    })
```

What's the thinking of this in the new multi stage code in `npm@3`?

KidkArolis:
Since this PR is now merged and closed, I've cross posted my question to an open issue: https://github.com/npm/npm/issues/6912#issuecomment-77681927

PR #7534, opened 2015-03-04
Update npm-install.md

olizilla:
Swap quote-mark for back-tick to fix inline code block that was otherwise unterminated.


othiym23:
Landed as 1226ca98d4d7650cc3ba16bf7ac62e44820f3bfa. Thanks, Oli!

Issue #7082, opened 2015-01-07
Short help (usage) is specified in multiple places, inconsistent


smikes:
`npm --help publish` yields

> npm publish &lt;tarball>
> npm publish &lt;folder>
>
> Publishes '.' if no argument supplied

from <https://github.com/npm/npm/blob/master/lib/publish.js#L16>

while the online docs and `npm help publish` give

>npm publish &lt;tarball> [--tag &lt;tag>]
> npm publish &lt;folder> [--tag &lt;tag>]

from <https://github.com/npm/npm/blob/master/doc/cli/npm-publish.md> (interesting github bug: can't link to line number in markdown document.)

There should be a single place where this is specified, so they don't have to be kept in sync.

I propose adding a script to the build process that extracts the `SYNOPSIS` section of the docs into a file containing a json object, and modifying all subcommands to draw their usage information from this json object.

Edited to add - in the process of doing this I would reconcile the current `usage` with the markdown `SYNOPSIS` section.

/bikeshed should be green

/cc  @linclark @othiym23 (others?)

linclark:
I'm not sure whether the difference is in there for a reason, but if not, I do like the idea of consolidating. 

Would it make sense for it to go the other direction... for the module file to be the source of truth? That way, we could take the usage property which is already exported for each and split it on `\n`, then inject it as the synopsis. I think this might be easier to implement, and it would also keep the synopsis closer to the code so it's more obvious that it needs to be updated if code changes.

smikes:
I agree source of truth should be closer to code .. I am surprised that the docs were updated to include the --tag argument but not the usage, I would have expected it to be the other way.

It's easier for me to imagine writing code that parses markdown and generates JSON than the other way, but I have no commitment either way. :cat2:

othiym23:
I agree with @linclark that keeping the source of truth closer to the code is likelier to yield up-to-date usage docs, and I agree with @smikes that coming up with a way to generate that stuff automatically is a lot easier coming from the `.md` files than the other way around. I don't think those two goals are necessarily in conflict, given that lots of things that npm does aren't particularly easy. We should have a (quick) discussion about how the docs-building process works that also involves npm's web site team (@zeke seems to have the most interest in this) so we can come up with something that is relatively painless for everyone to use once we've built it out. Maybe exporting `.usage` on the exported function for each command offering it?

smikes:
[...]

PR #7535, opened 2015-03-04
usage/doc synchronization

smikes:
As a first approach to the problem laid out in #7082, this PR aims to bring the short help displayed on `npm <subcommand> --help` and the `SYNOPSIS` section of the `cli/npm-<subcommand>.md` into harmony.

There is a list of all current published `npm` subcommands, and both the `--help` usage message and `SYNOPSIS` text, on tab `UsageSync` of my spreadsheet, here - https://docs.google.com/spreadsheets/d/1r-Gn2mEqPDwDVwIWYxP-ChxFQFVtrju5HEjruEa0tyQ/edit#gid=1285621165

There are a number of variations which need to be handled -- e.g., some subcommands `--help` launches a page in the pager/browser; etc. I'm opening this PR now so discussion about those things can happen.

This PR is based against `npm/multi-stage` and will be targeted for `npm@3` or later.

Initial changes:

> whoami: add --registry to usage
> remove npm-rm.md, redundant to npm-uninstall.md


smikes:
As the initial change, I noted that `npm-rm.md` and `npm-uninstall.md` document the same command.  See  <https://docs.npmjs.com/cli/rm> and <https://docs.npmjs.com/cli/uninstall>.

Since `uninstall` is longer, I have kept it and removed `npm-rm.md`.  An alias or redirect should be created on the docs website.

othiym23:
We should probably figure out how to deal with those redirects on the client side, @smikes. I need to read up on how `mandb` deals with multiple commands being pointed at the same page, but npm should do something to be more consistent. I think for now, nuking / merging duplicate entries is a sensible plan. @isaacs might have had a reason once for having `rm` and `uninstall` as separate pages, but I no longer know what it is.

othiym23:
See also #7600 / #7601. What are next steps here? Having PRs sitting around the queue with `NOT READY` in the name makes me twitchy. ;)

smikes:
How about I split off the commit for combining rm and uninstall into its own PR;

and then tomorrow morning I can finish the first pass on synchronizing the --usage messages with the SYNOPSIS sections and then you can review it.

smikes:
I am ready for review of this

doc/cli/npm-completion.md
```diff
@@ -3,7 +3,10 @@ npm-completion(1) -- Tab Completion for npm
 
 ## SYNOPSIS
 
-    . <(npm completion)
+    npm completion >> ~/.bashrc
+    npm completion >> ~/.zshrc
+
+    source <(npm completion)
```
othiym23: The way this is written is a little confusing. Maybe move the `npm completion >> ~ /.{z,ba}shrc` stuff into a section down below?

doc/cli/npm-config.md
```diff
@@ -8,7 +8,6 @@ npm-config(1) -- Manage the npm configuration files
     npm config delete <key>
     npm config list
     npm config edit
-    npm c [set|get|delete|list]
```
othiym23: good call.

doc/cli/npm-edit.md
```diff
@@ -3,7 +3,7 @@ npm-edit(1) -- Edit an installed package
 
 ## SYNOPSIS
 
-    npm edit <name>[@<version>]
+    npm edit <pkg>
```
othiym23: For this command, and a few others, making clear that you can fetch a specific version of the package in the command is useful, inasmuch as most people wouldn't even think that was a possibility if they came to the manpage cold.

doc/cli/npm-install.md
```diff
@@ -3,15 +3,23 @@ npm-install(1) -- Install a package
...
+    npm install <tarball file>
+    npm install <tarball url>
+    npm install <folder>
+
+    alias: npm i
+    common options: [--save|--save-dev|--save-optional] [--save-exact]
+
+    Can specify one or more: npm install ./foo.tgz bar@stable /some/folder
+
+    If no argument is supplied and ./npm-shrinkwrap.json is present,
```
othiym23: This section feels out of place in the `SYNOPSIS`. It's OK to draw people's attention to it with a callout, but it should be in the body of the `DESCRIPTION` with the rest of the docs.

doc/cli/npm-logout.md
```diff
@@ -3,7 +3,7 @@ npm-logout(1) -- Log out of the registry
 
 ## SYNOPSIS
 
-    npm logout [--registry=url] [--scope=@orgname]
+    npm logout [--registry=<url>] [--scope=@<scope>]
```
othiym23: `--scope=<@scope>` – this is a point of some contention within the npm team, as @zeke can attest, but my feeling is that the `@` is part of the scope name.

doc/cli/npm-outdated.md
```diff
@@ -3,7 +3,7 @@ npm-outdated(1) -- Check for outdated packages
 
 ## SYNOPSIS
 
-    npm outdated [<name> [<name> ...]]
+    npm outdated [[@<scope>/]<pkg> ...]
```
othiym23: Same point about `<@scope>` – I think this one is probably on me originally, in that I've changed my mind about what makes the most sense. So I can make this change if it seems too exasperating to you.

doc/cli/npm-install.md
smikes: OK, I can pull it.  That means pulling it from the `usage` too, though -- is that desirable?

doc/cli/npm-outdated.md
smikes: perl one-liner. already done.

smikes:
ready for additional review

doc/cli/npm-completion.md
```diff
@@ -3,7 +3,7 @@ npm-completion(1) -- Tab Completion for npm
 
 ## SYNOPSIS
 
-    . <(npm completion)
+    source <(npm completion)
```
zeke: yes!

doc/cli/npm-ls.md
```diff
@@ -3,10 +3,9 @@ npm-ls(1) -- List installed packages
 
 ## SYNOPSIS
 
-    npm list [[@<scope>/]<pkg> ...]
-    npm ls [[@<scope>/]<pkg> ...]
-    npm la [[@<scope>/]<pkg> ...]
-    npm ll [[@<scope>/]<pkg> ...]
+    npm ls [[<@scope>/]<pkg> ...]
+
+    aliases: list, la, ll
```
othiym23: It's worth noting at some point that `npm ll` and `npm la` do something very different from `npm ls` (try them!), but that doesn't need to be part of this commit.

othiym23:
This LGTM, but I leave it to @iarna to merge changes into `multi-stage` when she thinks they're ready.

iarna:
Landed as c20e798 in https://github.com/npm/npm/tree/multi-stage

PR #7538, opened 2015-03-04
Add note about package name

RichardLitt:
In case anyone doesn't know that `node-redis` is called `redis`. Also added a period that was missing.


othiym23:
This would all have been a lot simpler if Matt hadn't changed the name from `node-redis` to `redis`. So confusing! Landed as 903bdd105b205d6e45d3a2ab83eea8e4071e9aeb. Thanks, @RichardLitt!

RichardLitt:
Thanks @othiym23

PR #7539, opened 2015-03-05
WIP: restore order to caching git remotes

othiym23:
Also, update code coments to be accurate, and extirpate single-letter variable names. And introduce a params object because those argument lists were completely ridiculous.

**r:** @isaacs 
**r:** @iarna

### DIFF
--- a/lib/cache/add-remote-git.js
+++ b/lib/cache/add-remote-git.js
@@ -1,69 +1,313 @@
-var mkdir = require("mkdirp")
-  , assert = require("assert")
-  , git = require("../utils/git.js")
-  , fs = require("graceful-fs")
-  , log = require("npmlog")
-  , path = require("path")
-  , url = require("url")
-  , chownr = require("chownr")
-  , crypto = require("crypto")
-  , npm = require("../npm.js")
-  , rm = require("../utils/gently-rm.js")
-  , inflight = require("inflight")
-  , getCacheStat = require("./get-stat.js")
-  , addLocal = require("./add-local.js")
-  , realizePackageSpecifier = require("realize-package-specifier")
-  , normalizeGitUrl = require("normalize-git-url")
-  , randomBytes = require("crypto").pseudoRandomBytes // only need uniqueness
+var mkdir = require('mkdirp')
+var assert = require('assert')
+var git = require('../utils/git.js')
+var fs = require('graceful-fs')
+var log = require('npmlog')
+var path = require('path')
+var url = require('url')
+var chownr = require('chownr')
+var crypto = require('crypto')
+var npm = require('../npm.js')
+var rm = require('../utils/gently-rm.js')
+var inflight = require('inflight')
+var getCacheStat = require('./get-stat.js')
+var addLocal = require('./add-local.js')
+var realizePackageSpecifier = require('realize-package-specifier')
+var normalizeGitUrl = require('normalize-git-url')
+var randomBytes = require('crypto').pseudoRandomBytes // only need uniqueness
 
-var remotes = path.resolve(npm.config.get("cache"), "_git-remotes")
-var templates = path.join(remotes, "_templates")
+var remotes = path.resolve(npm.config.get('cache'), '_git-remotes')
+var templates = path.join(remotes, '_templates')
 
 var VALID_VARIABLES = [
-  "GIT_SSH",
-  "GIT_SSL_NO_VERIFY",
-  "GIT_PROXY_COMMAND",
-  "GIT_SSL_CAINFO"
+  'GIT_SSH',
+  'GIT_SSL_NO_VERIFY',
+  'GIT_PROXY_COMMAND',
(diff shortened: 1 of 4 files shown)


lib/cache/add-remote-git.js
```diff
@@ -26,146 +26,149 @@ var VALID_VARIABLES = [
...
+module.exports = function addRemoteGit (gitURL, silent, cb) {
+  assert(typeof gitURL === 'string', 'must have git URL')
+  assert(typeof cb === 'function', 'must have callback')
+
+  var normalized = normalizeGitUrl(gitURL)
+  var strippedURL = normalized.url
+  var treeish = normalized.branch
+  var tmpPath = strippedURL.replace(/[^a-zA-Z0-9]+/g, '-') + '-' +
+    crypto.createHash('sha1').update(strippedURL).digest('hex').slice(0, 8)
+  var checkoutPath = path.join(remotes, tmpPath)
```
iarna: actual variable names!

lib/cache/add-remote-git.js
```diff
@@ -26,146 +26,149 @@ var VALID_VARIABLES = [
...
+      ['config', '--get', 'remote.origin.url'],
+      { cwd: params.checkoutPath, env: gitEnv() },
       function (er, stdout, stderr) {
-        var stdoutTrimmed = (stdout + "\n" + stderr).trim()
-        if (er || u !== stdout.trim()) {
-          log.warn( "`git config --get remote.origin.url` returned "
-                  + "wrong result ("+u+")", stdoutTrimmed )
-          return rm(p, function (er){
+        var originURL = (stdout + '\n' + stderr).trim()
+        if (er || params.gitURL !== stdout.trim()) {
```
iarna: What's going on with originURL having stderr in it, and then not using that when validating the result?

lib/cache/add-remote-git.js
```diff
@@ -26,146 +26,149 @@ var VALID_VARIABLES = [
...
+          log.warn(
```
iarna: Should the first argument be a log section eg `checkGitDir`?

lib/cache/add-remote-git.js
```diff
@@ -173,97 +176,124 @@ function fetchRemote (p, u, co, origUrl, cb) {
...
+function resolveHead (params, cb) {
+  log.verbose('resolveHead', 'original treeish', params.treeish)
   git.whichAndExec(
-    [ "rev-list", "-n1", co ],
-    { cwd : p, env : gitEnv() },
+    ['rev-list', '-n1', params.treeish],
+    { cwd: params.checkoutPath, env: gitEnv() },
     function (er, stdout, stderr) {
-      stdout = (stdout + "\n" + stderr).trim()
+      params.exactTreeish = (stdout + '\n' + stderr).trim()
```
iarna: Mutating params feels icky. And again, I'm left wondering why stderr is glommed on here?

lib/cache/add-remote-git.js
```diff
@@ -173,97 +176,124 @@ function fetchRemote (p, u, co, origUrl, cb) {
...
-
-    mkdir(tmp, function (er) {
-      if (er) return cb(er)
-
-      git.whichAndExec(["clone", p, tmp], { cwd : p, env : gitEnv() }, clone)
-    })
-
-    function clone (er, stdout, stderr) {
-      stdout = (stdout + "\n" + stderr).trim()
+function clone (params, cb) {
```
iarna: The term `clone` feels a little overloaded. `gitClone` would be clearer to me.

iarna:
Looks great! Much easier to read. The variable name changes alone!

lib/cache/add-remote-git.js
othiym23: It does, but I wonder how much that is a real issue, and how much is an urge towards ideological purity, given that this particular `params` never escapes the context of this file. I have an alternative approach that I'm going to try, though, which is pulling all the non-utility parts of the caching flow back into the main function, and setting the individual variables within the scope of the closure. That's how it was originally written, and now that the control flow has been cleaned up a lot, it makes sense to bring it back to that style again.

lib/cache/add-remote-git.js
othiym23: That was how it was written originally, which probably means that it's a byproduct of how @isaacs originally hacked it together, and yes, I think its use is purely informational. This particular check is just verifying that the repo we've been handled is for the source we're expecting. I'll try cleaning this up and see if anything breaks.

lib/cache/add-remote-git.js
othiym23: I'll brainstorm this as I rewrite. More descriptive function names would probably be a good complement to the variable names.

othiym23:
I ended up gitting some more work done on this, with more comments, gitting rid of the params object, and hopefully improving this git workflow to make it easier to git. Everyone is going to be so happy to git this change.

PR #7548, opened 2015-03-05
update: only update when current !== wanted

### LINKED ISSUE #6441: Calling npm update multiple times keep grabbing same "updates"
```bash
npm install hapi
cd node_modules/hapi
npm install
npm update
npm update
```

Why are the last two `npm update`s spitting out anything? Nothing has changed in between those updates in the registry but it looks like it is finding differences.

othiym23:
Fixes #6441.

**r**: @smikes (I added a case to `test/tap/update-examples.js`)
**r**: @iarna


iarna:
:shipit:

PR #7597, opened 2015-03-11
doc: use perl in `scripts/update-authors.sh` instead of awk

kenany:
As a consequence of the [various implementations](https://en.wikipedia.org/wiki/AWK#Versions_and_implementations) of AWK across platforms, the use of AWK in `scripts/update-authors.sh` might result in different sorting behavior on different platforms (i.e. BSD awk on OS X does not maintain the desired sort-by-first-contribution). So, let's use perl instead.

### DIFF
--- a/scripts/update-authors.sh
+++ b/scripts/update-authors.sh
@@ -1,20 +1,18 @@
 #!/bin/sh
 
-git log --reverse --format='%aN <%aE>' | awk '
+git log --reverse --format='%aN <%aE>' | perl -we '
+
 BEGIN {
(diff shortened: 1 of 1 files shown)


othiym23:
Landed as fcd92476f3a9092f6f8c83a19a24fe63b206edcd. I also read up on the format of `.mailmap` and made a few tweaks to it in 7bc4cf3dbc0cdbf1de5287ddd5468a53089c6a66. Regenerated `AUTHORS` and it looks muuuuch better. Thank you, Kenan!

PR #7598, opened 2015-03-11
Removed old comment

RichardLitt:
This is obviously no longer experimental, and I'm not sure this comment
belongs here anymore.


othiym23:
Landed in 721b17a31690bec074eb8763d823d6de63406005. I use `-S` like every day, so you're right, it really shouldn't be considered experimental. Thanks, @RichardLitt!

PR #7601, opened 2015-03-11
Added shorthands docs file

RichardLitt:
There's got to be more to do than just add this file. I'm also not sure this is the best way to do it, as now this is both here and in config. Thoughts?

### DIFF
--- a/doc/misc/npm-shorthands.md
+++ b/doc/misc/npm-shorthands.md
@@ -1,0 +1,30 @@
+npm-shorthands(7) -- Shorthands and Other CLI Niceties
+================================================
+
+## Synopsis
+
+The following shorthands are parsed on the command-line:
+
+* `-v`: `--version`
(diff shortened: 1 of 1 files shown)


othiym23:
It's really hard to determine the appropriate hierarchy of information in the npm docs, because what's most important will vary depending on what you're trying to do. I don't think this is the best solution to that problem, though, because it's just a simple list of shortcuts bereft of context.

It makes more sense to me to follow Git's model here and start pushing more of the shortcuts out to the documentation for the subcommands that use the shortcuts, which is something that people have been doing in fits and starts over the last couple months. I also think it would be nice to have Git-style command-line help tailored per subcommand.

Thanks for the time you spent putting this together! Let me know what you think of those ideas!

RichardLitt:
Agreed. I don't think this is the best solution, either. I think we should start farming out the shortcuts. I can look into git-style command line help, too. Thanks @othiym23

PR #7605, opened 2015-03-11
clean up lib/utils/gently-rm.js

othiym23:
It was too confusing understanding what was going on with it before, so I did a bunch of cleanup of the logging and how the control flow was nested, and commented the crap out of it to make plainer why it works the way it does. The tests all pass either way, but I want to make sure that other humans put their eyeballs on it before I commit it to master.

**r**: @iarna 
**r**: @isaacs (if you feel like it)


iarna:
Looks good! I don't envy your tracking this down, but the changes seem straightforward.

Issue #5214, opened 2014-05-06
npm report


isaacs:
If there's an `npm-debug.log` file, then `npm report` should upload that as an anonymous gist, and then take the user to a page like https://github.com/npm/npm/issues/new?title=error%20title%20report&body=help%20me%20please%20blah%20blah.

Once that's done, we should do #5213.

watilde:
Wow I thought same thing with this so I developed and published as a package:
https://www.npmjs.com/package/npm-report

Well, this feature seems happy if it's in the npm.

othiym23:
@watilde This is a good start. Do you know about npm/support-cli? That's where we want to start sending support tickets / `npm-debug.log` links. Also, if `npm-report` would use `gist` or something similar to automatically upload `npm-debug.log` to gist.github.com, that would make it very useful. If you want to keep working on it, it could very easily become something we eventually incorporate into npm itself. Thanks for getting the ball rolling!

watilde:
@othiym23 Thanks for your reply with all your kindness! I've changed endpoint creating issue to `npm/support-cli` in `npm-report`. Well using gist sounds really useful, but I think it might makes npm a little bit complex. We have a 4 way for uploading gist:

|#|Post as who?|share with|
|---|---|---|
|1|Anonymous|Public|
|2|Anonymous|Private|
|3|User|Public|
|4|User|Private|

These way have two problems
+ Deleting an anonymous gist need a human power:
  + https://help.github.com/articles/deleting-an-anonymous-gist/
+ Posting as a user need authentication:
  + https://www.npmjs.com/package/gistjs#login

So, I like to use a issue's body like this:
  + https://github.com/npm/support-cli/issues/new?body=blahblah

Could you please any thought on this please?

othiym23:
This has been moved to the [npm roadmap](https://github.com/npm/npm/wiki/Roadmap), which we're using instead of the confusing `next-*` labels now.

othiym23:
@watilde Where did you leave `npm-report`? The CLI team is pretty heads down with bugs and high-priority feature work right now, but it would be useful to see if this project is in a place where we can turn the crank on it a bit. The issue template has helped somewhat, at least in terms of soliciting better issues from users, but there's still a lot of work to be done around making it easier to triage those issues, and also, you know, to handle them.

watilde:
@othiym23 I've released it as a package and you can see what does it bring:
```
$ npm i -g npm-report
[...]

Issue #6350, opened 2014-09-29
npm-debug.log should always be written when an error occurs


dylang:
### Current situation

Currently the `npm-debug.log` is stored in memory, and then dumped to disk when certain problems occur. 
### Problems
1. Some errors don't cause a log to be generated, such as:
   
   ```
   npm ERR! cb() never called!
   npm ERR! not ok code 0
   ```
2. Sometimes npm just hangs and there's no way to know what it was waiting for.
3. When it fails it only writes out the last 10,000 lines because `npmlog` has a `maxRecordSize` of 10000. If it was streaming to disk this wouldn't be an issue. _Note: I'm not a consumer of the logs so I'm not sure if the full log is necessary._
### Proposed Solutions

Here's some ideas. Any of these might be enough to close this ticket.
1. Fix the logic causing no long to be written with the above example, and auto-write a log file after some amount of time with no output to stdout to have a debug log for hang events.
2. Write to the log file immediately (or asynchronously as long as an error won't prevent the last bits from being written.) On success unlink the log file.
3. Introduce an option to force log file to be written no matter what. 
## 

Semi-related issue: #5214 `npm report`

jsdevel:
It would be really nice if the creation of `npm-debug.log` could be entirely disabled in `~/.npmrc`.

mike-spainhower:
@maxogden @jsdevel Feel free to move off-topic discussion to https://github.com/npm/npm/issues/7648 :grin:

othiym23:
This has been moved to the [npm roadmap](https://github.com/npm/npm/wiki/Roadmap), which we're using instead of the confusing `next-*` labels now.

othiym23:
Whoopsie, didn't mean to actually close this!

legodude17:
I would suggest that npm always writes to log. Then deletes it if there was no error and no --always-write or there was a --never-write. Maybe the log could start as a tmp file and then get moved to the dir if needed.

othiym23:
We’ve finally returned to this issue after some time, and there are a number of related issues devoted to dealing with the issue of _always_ getting debugging logs, and putting them in a more convenient place. See also:

- #1548
- #1584
- #5252
- #6181
- #6744
- #6766
- #7041
- #7614
- #7648
- #11439

[...]

PR #7611, opened 2015-03-12
report: add new command

watilde:
Hi, I added report command for fixing #5214. That proposal might make support useful, people become needn't bring manually a npm-debug.log to the issue description in GitHub.

Well, I'd like to use `gist` but I think it might makes npm a little bit complex so I've built in this way.
See also: https://github.com/npm/npm/issues/5214#issuecomment-77362757 

Thanks.


rlidwka:
> Well, I'd like to use gist but I think it might makes npm a little bit complex so I've built in this way.

Just post it to some kind of pastebin with 1 month expiration time imho.

othiym23:
I like this! I do think it's going to take a little effort to land this, but now is probably as good a time as ever to get this moving along. This is an actual honest-to-gosh feature in npm, and as such needs a little extra effort to get right:

- [ ] document the command as it currently stands
- [ ] land #6744
- [ ] land #7614
- [ ] figure out how to hang on to debug logs (putting them somewhere ephemeral prevents us from being able to refer back to specific support cases later, and also limits our ability to do data mining and learn from our support cases, so is a no go – maybe npm will write a microservice to handle this)

It would be fantastic to actually implement #5214, as we've been wanting this for nearly a year, so I'm willing to budget some of npm's time towards making sure this keeps moving forward. Thanks, @watilde!

smikes:
What happens on issues/new if the user is not logged in to github?

watilde:
Sorry for my late reply,

@rlidwka Thanks for your information! I'll use it a little, but maybe I can't decide to include that to this patch in this time..

@othiym23 Sounds great! I'm going to added a document in `doc/api` | `doc/cli` to this patch.

@smikes Thanks for your comment, Redirect to [this URL](https://github.com/login?return_to=https%3A%2F%2Fgithub.com%2Fnpm%2Fsupport-cli%2Fissues%2Fnew%3Fbody%3Dblahblah), after logged in report process will go well.

othiym23:
1 down, 3 to go! Thanks for the docs!

By the way, I think this will probably be something that lands after the release of `npm@3`, as I'm trying to hold off on landing any more significantly-sized features to the `npm@2` code base that aren't directly related to support for the new npm registry infrastructure. So you may want to rebase against `multi-stage` at some point, and also be prepared to wait a little while for this one to get merged. But getting this in will be great!

watilde:
@othiym23 I agreed to be something that lands after the release of npm@3. I'll git rebase `multi-stage` and wait as a `next-major`! I hope to make it helps npm supports, thanks.

othiym23:
This has been moved to the [npm roadmap](https://github.com/npm/npm/wiki/Roadmap), which we're using instead of the confusing `next-*` labels now. Also, because this PR is pretty old and would need some work to land against master, I'm going to close this particular PR.

@watilde -san, once `npm@3` has landed, would you be interested in working with us to get a new version of this put together that includes #7611 and #6350? npm, Inc. now has an actual (and fantastic) support team to help out with the issues on this repo, but it would be great if we could channel front-line support into a more useful place.

lib/report.js
```diff
@@ -0,0 +1,32 @@
...
+  var body = ""
+  var exist = fs.existsSync(npmDebugLog)
+  var opt = ""
+
+  if (exist || title) opt += "?"
+  if (exist) {
+    body = "npm-debug.log:\n"
+    body += "```\n"
+    body += fs.readFileSync(npmDebugLog)
+    body += "```\n"
```
legodude17: I think that it should ask user for input like:
`What were you trying to do?`
`Any other comments?`
Stuff like that. Also, maybe write to a file and then let the user edit it before sending it off?

PR #7627, opened 2015-03-14
run `prepublish` for git url packages

### LINKED ISSUE #3055: run `prepublish` for git url packages
After installing a git url package, perhaps the `prepublish` script should be run, since git repos will typically not keep around any built artifacts that the prepublish script generates.

However, `prepublish` scripts also typically require the use of `devDependencies` packages, so it might not be possible to do this in a good way.  (In which case, it'll just be on the users of git url packages to run whatever scripts are necessary to make them ready to use.)

valeriangalliat:
> When there is a `prepublish` script, the `install` command is run to fetch all dependencies (including development ones) and run the `prepublish` script.
> 
> Fixes #3055

I'd really like feedback on this, I never worked with npm core before and I'm not sure of what I'm doing. But at least it works for the package I tried:

``` json
{
  "dependencies": {
    "es6-denodeify": "valeriangalliat/es6-denodeify"
  }
}
```

Without the fix, I get a `node_modules/es6-denodeify` without `index.js` because it was not compiled (and thus it's not usable), but with this PR, the `index.js` is properly compiled with the `prepublish` script. However it takes longer to install since it fetches all the dependencies (including development ones) in order to run the `prepublish` script (basically it's doing a `npm install` in the cloned directory before packaging it).

Also this step is skipped if there is no `prepublish` script (because it would otherwise uselessly fetch dependencies).

Here's a `diff -w` (much less lines than the GitHub diff):

``` diff
diff --git a/lib/cache/add-remote-git.js b/lib/cache/add-remote-git.js
index 9eaf6b1..0435797 100644
--- a/lib/cache/add-remote-git.js
+++ b/lib/cache/add-remote-git.js
@@ -15,6 +15,7 @@ var addLocal = require('./add-local.js')
 var realizePackageSpecifier = require('realize-package-specifier')
 var normalizeGitUrl = require('normalize-git-url')
 var randomBytes = require('crypto').pseudoRandomBytes // only need uniqueness
+var readJson = require('read-package-json')

 var remotes = path.resolve(npm.config.get('cache'), '_git-remotes')
 var templates = path.join(remotes, '_templates')
@@ -257,6 +258,16 @@ module.exports = function addRemoteGit (uri, silent, cb) {
     )
   }

+  // install dependencies and run prepublish script to get dist files
+  // https://github.com/npm/npm/issues/3055
+  function buildPackage (cb) {
+    readJson(path.join(tmpdir, 'package.json'), false, function (er, data) {
+      if (er) return cb(er)
+      if (!data.scripts.prepublish) return cb()
+      npm.commands.install(tmpdir, [], cb)
+    })
+  }
+
   // there is no safe way to do a one-step clone to a treeish that isn't
   // guaranteed to be a branch, so explicitly check out the treeish once it's
   // cloned
@@ -273,6 +284,9 @@ module.exports = function addRemoteGit (uri, silent, cb) {
         }
         log.verbose('addRemoteGit', 'checkout', stdout)

+        buildPackage(function (er) {
+          if (er) return cb(er)
+
           // convince addLocal that the checkout is a local dependency
           realizePackageSpecifier(tmpdir, function (er, spec) {
             if (er) {
@@ -301,6 +315,7 @@ module.exports = function addRemoteGit (uri, silent, cb) {
               cb(er, data)
             })
           })
+        })
       }
     )
   }
```

### DIFF
--- a/lib/cache/add-remote-git.js
+++ b/lib/cache/add-remote-git.js
@@ -15,6 +15,7 @@ var addLocal = require('./add-local.js')
 var realizePackageSpecifier = require('realize-package-specifier')
 var normalizeGitUrl = require('normalize-git-url')
 var randomBytes = require('crypto').pseudoRandomBytes // only need uniqueness
+var readJson = require('read-package-json')
 
 var remotes = path.resolve(npm.config.get('cache'), '_git-remotes')
 var templates = path.join(remotes, '_templates')
@@ -257,6 +258,16 @@ module.exports = function addRemoteGit (uri, silent, cb) {
     )
   }
 
+  // install dependencies and run prepublish script to get dist files
+  // https://github.com/npm/npm/issues/3055
+  function buildPackage (cb) {
+    readJson(path.join(tmpdir, 'package.json'), false, function (er, data) {
+      if (er) return cb(er)
+      if (!data.scripts.prepublish) return cb()
+      npm.commands.install(tmpdir, [], cb)
+    })
+  }
+
   // there is no safe way to do a one-step clone to a treeish that isn't
   // guaranteed to be a branch, so explicitly check out the treeish once it's
   // cloned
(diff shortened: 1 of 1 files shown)


othiym23:
This is pretty nice in that it solves the problem So, looking good there! At the same time, it poses kind of a thorny architectural problem, because ideally all the cache is aware of is its own little chunk of filesystem, and knowing how to ask its host application (i.e. the rest of the CLI) for data from the outside world. It really shouldn't be calling the installer directly. I don't really see an easy way to get around that in this case, though, so I'm going to need to spend a little time thinking about how to make everything line up correctly. Until then, thanks a bunch for your time and work in putting this together.

Dakuan:
This is real issue for us, we have lots of react components that all live in their own private github repo. It was really confusing to find that the behaviour of github based packages was different from npm installed ones.

rom1504:
Getting a dependency directly from a git repo means the files that should be generated before inclusion into an npm package haven't been generated.
I think that means that, for git dependencies, prepublish should be executed and only then can the files be put inside the cache (with the correct git url metadata) and in node_modules.
If storing and handling the git url metadata inside the cache is too complicated, maybe a first step would be disabling the cache for git dependencies.

Anyway I think this issue is important because it makes it impossible to use git dependencies of projects that need to generate files before publishing (which is useful for example during the es6 transition)

othiym23:
The blocker on this isn't the behavior in the patch (which is more or less correct), but how to factor things such that the internals of npm aren't conflating concerns unnecessarily. This is on the list to get addressed eventually, but it might be a little while.

iarna:
I've been thinking about this a lot lately…

I think the problem here is that the cache is trying to create packages on its own. It should get out of that business and cede that to something that accepts a directory and produces a tarball. Ultimately this would have to call the installer, but it would keep it at arms length. The underlying system should be using the same mechanism as `npm pack`.

othiym23:
One of the major near-term projects for the npm CLI team is to make it treat registry-hosted and git-hosted packages more consistently, and so we really want this behavior. However, it's going to take some significant redesign of the cache, packing, and lifecycle bits of the CLI to do it in the way that will be sustainable over the long run.

I _really_ appreciate the work you put into this, @valeriangalliat! However, I'm going to close this PR because of the above. Getting this functionality into the CLI is something that should happen soon, even if we don't land this particular patch. Keep an eye on [the road map](https://github.com/npm/npm/wiki/Roadmap) for further develoments.

PR #7629, opened 2015-03-14
tildify the path in the `npm ls` output

sindresorhus:
for a bit nicer output

from `npm@2.7.2 /Users/sindresorhus/dev/npm` to `npm@2.7.2 ~/dev/npm`
## 

If you like this there are more places `tildify` could be used to improve the output.


stephenplusplus:
+1

kevva:
Yes, please!

othiym23:
I use `tildify` and `untildify` in some of my non-npm projects, but those projects aren't as intimately tied to the platform as npm is.

Why is this important? Windows. You _can_ use tilde-prefixed paths in Windows output (and many things do, including Git Bash / mingw), but if you're going to do that, but `cmd.exe` doesn't know what to do with tilde-prefixed paths, so if we were going to do this, we'd have to put `untildify` everywhere we're accepting paths as command-line parameters, and that's a lot of additional code (and new tests) for something that's not (in my judgment) worth the burden.

If you want to take on the job of making _everything_ handle tilde in user-provided paths, including on Windows, I'm willing to work with you to get that in. But doing it piecemeal? Sorry, I don't think that's a good idea.

sindresorhus:
This is only about improving some of the visuals in the output, not trying to add support for tilde shortened paths in any inputs, but ok.

othiym23:
I don't think it's unreasonable that npm should accept whatever kinds of things it outputs, and experience has taught me that when that sort of symmetry is violated, we get confused users in the issue tracker complaining about things not working. `~` is just not a convention in the Windows world the way it is in the Unixy / bash-y world.

Issue #7636, opened 2015-03-15
readme.md is wrong for windows


rickdog:
readme.md says:
[quote]
You can download a zip file from <https://npmjs.org/dist/>, and unpack it in the same folder where node.exe lives.
[/quote]

Not true, must unpack to subfolder node_modules\npm where node.exe lives.
(npm 1.4.28, probably others too)

msikma:
I think the reference to 1.4.28 in the readme is not needed anymore, anyway—unless I'm missing something, all versions still come in a zip file with binary. But the note about the path seems to be correct. On a side note, you seem to have an old version, since the readme now links to https://github.com/npm/npm/releases

othiym23:
I just landed @msikma's tweak to the wording as 402f52ab201efa348feb87cad753fc4b91e8a3fb, which is close enough to call this resolved. As @msikma said, if you're seeing a message about `npmjs.org/dist`, you have a version of npm old enough that you really ought to think about upgrading anyway. There's not a lot we can do about changing the wording in the documentation in old versions.

PR #7637, opened 2015-03-15
Modify Windows install/upgrade instructions

msikma:
Add a reference to the `node_modules\npm` directory (which is where npm's files end up after installing node), and remove the reference to v1.4.28 since all latest releases also come in a zip file. See #7636.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,11 +64,11 @@ for testing, or running stuff without actually installing npm itself.)
 
 ## Windows Install or Upgrade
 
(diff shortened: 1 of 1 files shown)


othiym23:
Tweaked the commit message slightly, rebased, and landed as 402f52ab201efa348feb87cad753fc4b91e8a3fb. Thanks for the tweak! It's good to mostly have people on `npm@2` right as we get ready to roll out `npm@3`.

Issue #7672, opened 2015-03-18
Using the npm client with registry behind client certificates fails


atamon:
I'm trying to provide the NPM client with client certificates using the `key` and `cert` options.

This because I need to be able to connect to a local registry running sinopia, and the server has client certificate verification set up as outer protection.

However, I get this error even before NPM can make a connection to the sinopia server 

```
npm verb stack Error: error:0906D06C:PEM routines:PEM_read_bio:no start line
npm verb stack     at Object.exports.createCredentials (crypto.js:100:17)
npm verb stack     at Object.exports.connect (tls.js:1331:27)
npm verb stack     at Agent.createConnection (https.js:79:14)
npm verb stack     at Agent.createSocket (http.js:1297:16)
npm verb stack     at Agent.addRequest (http.js:1273:23)
npm verb stack     at new ClientRequest (http.js:1420:16)
npm verb stack     at Object.exports.request (https.js:123:10)
npm verb stack     at Request.start (/Users/atamon/code/npm/node_modules/request/request.js:906:30)
npm verb stack     at Request.end (/Users/atamon/code/npm/node_modules/request/request.js:1637:10)
npm verb stack     at end (/Users/atamon/code/npm/node_modules/request/request.js:676:14)
```

Where my `.npmrc` config looks like this at the moment (I have also tried using the paths as well as different ways of inlining the certificates)

```
registry=https://npm.domain.com
cert=/path/to/correct/cert.pem
key=/path/to/correct/key.pem
```

I know that both cert and key are correct as I can connect to our server with them using stunnel or openssl.

I downloaded the npm repo to try to debug this. And after much digging I noticed that I can get this to work with a small change in the registry client. See pull request attached for details

Is this the correct way of handling this issue? Maybe this isn't even an issue and I just missed completely on how client certificates should be provided?

Cheers

othiym23:
The configuration documentation is maybe not as explicit about this as it should be, but the way it works right now is how it's documented and intended to work – `ca`, `cert`, and `key` are all meant to be used with the actual _contents_ of the PEM-formatted files put into the relevant configuration properties, with something like `npm config set cert "$(cat certfile.pem)"`.

[...]

PR #7673, opened 2015-03-18
Fix client certificate options being misread

atamon:
This pull request is to be able to use client certificates properly with npm, it is a possible fix for https://github.com/npm/npm/issues/7672.


othiym23:
You've misread the configuration documentation for how `cert` and `key` configuration works. See #7672 for details.

Issue #7635, opened 2015-03-15
"undefined is not a function" with iojs 1.5.1/npm 2.7.0 (Linux)


mscdex:
Steps to reproduce:

* Create `package.json`:
```javascript
{
  "dependencies": {
    "socksv5": "git+https://github.com/mscdex/socksv5.git"
  }
}
```
* Then within same directory: `npm install`

It *does* work with node v0.10.30/npm 1.4.21

mscdex:
[`npm-debug.log`](https://gist.github.com/mscdex/a3f82cdc679c57ed6b57)

KenanY:
@mscdex I believe https://github.com/mscdex/socksv5/pull/5 will fix your issue. `bundledDependencies` should be an array, not a hash https://docs.npmjs.com/files/package.json#bundleddependencies

Works for me:

``` json
{
  "dependencies": {
    "socksv5": "git+https://github.com/KenanY/socksv5.git#bundled-array"
  }
}
```

mscdex:
You're right, my mistake. However, there should be a better error message than that...

KenanY:
@mscdex I agree. I'm sure all that's needed is an `isArray` check somewhere. Maybe I'll take a look tomorrow.

KenanY:
Relevant: so apparently there's a warning for this kind of thing; it is [emitted](https://github.com/npm/normalize-package-data/blob/b73bbb7fe07414f25caf92f7a49509688e7a89e4/test/dependencies.js#L28) by [`npm/normalize-package-data`](https://github.com/npm/normalize-package-data) when one runs `npm install` in a package whose `bundledDependencies` is not an _Array_. Error message can still be better  though.

KenanY:
Fixed in [`npm@2.7.4`](https://github.com/npm/npm/releases/tag/v2.7.4) :)

```
$ npm i mscdex/socksv5#0d0acea
npm ERR! Linux 3.18.6-1-ARCH
npm ERR! argv "/home/kenan/local/bin/iojs" "/home/kenan/.npm-packages/bin/npm" "i" "mscdex/socksv5#0d0acea"
npm ERR! node v1.6.1
npm ERR! npm  v2.7.4

npm ERR! socksv5's `bundledDependencies` should be an array
npm ERR! 
npm ERR! If you need help, you may report this error at:
npm ERR!     <https://github.com/npm/npm/issues>

npm ERR! Please include the following file with any support request:
[...]

PR #7682, opened 2015-03-19
test: add test for non-array bundledDependencies

kenany:
A test for my `fstream-npm` patch: https://github.com/npm/fstream-npm/pull/11

Ref. https://github.com/npm/npm/issues/7635


othiym23:
Landed as 3ce3cc242fc345bca6820185a4f5a013c5bc1944. I combined your test with some tweaks that I made to it (to make it match `standard` style, mostly) and the update to `fstream-npm` itself into a single commit. Thanks as always, Kenan!

PR #7685, opened 2015-03-20
cache: use new npm-package-arg, realize-package-specifier, and hosted-git-info

othiym23:
This will ensure that all GitHub URLs are run through a consistent process for cloning (or, at any rate, a more consistent process than they were run through before). For now, doesn't support the other shortcut syntaxes.

**r**: @iarna


iarna:
lgtm

othiym23:
Merged! I wonder what will break!?

PR #7686, opened 2015-03-20
ensure that --registry takes precedence over scoped configuration

### LINKED ISSUE #7661: npm login with specified port different from saved port failed
Interesting bug, reproducible in npm@2.7.1. Requires an npm Enterprise or other registry operating on a non-standard port. First, login to the registry:

`npm login --registry=http://some.registry:8080 --scope=@somescope`

Now edit the lines in your .npmrc:

```
@somescope:registry=http://some.registry:1234/
//some.registry:1234/:_authToken=XXXX
```

Attempt to login again:

`npm login --registry=http://some.registry:8080 --scope=@somescope`

The login will time out, because npm will try port 1234 even though you've specified the port in the login.

(This was discovered by an npm Enterprise customer who changed ports post-install and then couldn't login anymore)

othiym23:
Fixes #7661.

**r**: @iarna

### DIFF
--- a/lib/adduser.js
+++ b/lib/adduser.js
@@ -135,7 +135,8 @@ function save (c, u, cb) {
     if (scope.charAt(0) !== "@") scope = "@" + scope
 
     var scopedRegistry = npm.config.get(scope + ":registry")
-    if (scopedRegistry) uri = scopedRegistry
+    var cliRegistry = npm.config.get("registry", "cli")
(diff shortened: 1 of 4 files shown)


iarna:
Oooh, so, `config-chain` let's us determine specifically where a config comes from! It's good to have gotten to know that corner of the code base a bit.

lgtm!

PR #7696, opened 2015-03-20
Incorrect date format in doc-build.sh

MeddahJ:
In `scripts/doc-build.sh`, the date format simply uses minutes and months at the wrong places.

(credits due to @twillouer for detecting it, I'm just a messenger)


othiym23:
Heh heh whoops. :tada: Rebased and landed in f87c728f8732c9e977c0dc2060c0610649e79155. Thanks for the fix! And thanks to @twillouer for pointing it out!

PR #7710, opened 2015-03-23
fix npa test in windows

Bacra:
npa_test_windows_err


KenanY:
@Bacra Surely you should send this upstream to [`npm/npm-package-arg`](https://github.com/npm/npm-package-arg) directly? Otherwise this change will be overwritten next time npa is updated.

Bacra:
@KenanY thx

PR #7725, opened 2015-03-24
link: add --save to save a package when installing as a link

watilde:
Fixes: https://github.com/npm/npm/issues/1166
// I'll write a test.

### DIFF
--- a/doc/cli/npm-link.md
+++ b/doc/cli/npm-link.md
@@ -4,7 +4,7 @@ npm-link(1) -- Symlink a package folder
 ## SYNOPSIS
 
     npm link (in package folder)
-    npm link [@<scope>/]<pkgname>
+    npm link [@<scope>/]<pkgname> [--save|--save-dev|--save-optional] [--save-exact]
     npm ln (with any of the previous argument usage)
 
 ## DESCRIPTION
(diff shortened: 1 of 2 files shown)


othiym23:
This is yet another great little usability tweak! Thanks for putting it together! I have concerns, though, about how much of the save code is duplicated from `lib/install.js`. That code should really all be in one place (especially the big chunk of it devoted to handling `bundledDependencies` so that we don't have two very similar pieces of code to maintain.

Also, I'm trying to avoid adding any more features not directly related to the rollout of the new registry infrastructure to `npm@2`, so features like this should be done against the `multi-stage` branch. The silver lining to this cloud is that it's possible that it would be easier to write this code without duplication there. Give it a shot?

watilde:
@othiym23 That sounds absolutely great! I just read `install specific libraries` in `multi-stage`, actually this makes it easier to write this code. I’d like to step up to the stage so I’ll create a patch to multi-stage. 

Thanks!

watilde:
@othiym23 I read inside of the `multi-stage` and tried to create a patch, but it seemed hard to create a tree Object for executing the method called `saveToDependencies`... I'll try reread when I get a time. Also I'm happy if there is any good tips. Thanks.

Issue #7713, opened 2015-03-23
npm link error


watilde:
Hi
`npm link` seems failed when use a node version manager has different depth of global node_modules, see also: https://github.com/hokaccha/nodebrew/issues/30

I thought this fix might be good in [util/link.js](https://github.com/npm/npm/blob/master/lib/utils/link.js#L27):
```diff
-    target = path.relative(path.dirname(to), from)
+    target = path.relative(to, from)
```

Also here are logs: https://gist.github.com/watilde/1cf414a13506812a0cc5

Thanks.

watilde:
Here is minimum code to out result what happened in `util/link.js`:
```js
var path = require("path");

var from = "/path/to/development/npm/foo";
var to = "/path/to/global/npm/dir";

var target1 = path.resolve(from); // original from
var target2 = path.relative(path.dirname(to), from); // current case
var target3 = path.relative(to, from); // fixed case

console.log(path.resolve(target1));
// => /path/to/development/npm/foo

console.log(path.join(from, target2));
// => /path/to/development/development/npm/foo

console.log(path.join(from, target3));
// => /path/to/development/npm/foo
```

Additional information, `utils/link.js` has used only in 2 files and it seems to hasn't used in `lib/build.js`.
```
$ pwd
/path/to/npm

$ grep -irl utils\/link .
./build.js
./link.js
```
so it's ok to check a behavior only in lib/link.js.

watilde:
oh, I found the truly reason of this issue when I'm writing a test for link...
nodebrew used a symlink to switch node version, but the pathes are different between execPath and symlink. So I'll fix nodebrew.

Also I'll create a pull request to add test link and link-install. Thanks.

PR #7730, opened 2015-03-24
test: add link and link-install test

watilde:
refs: https://github.com/npm/npm/issues/7713#issuecomment-85591249

### DIFF
--- a/test/tap/link-install/package.json
+++ b/test/tap/link-install/package.json
@@ -1,0 +1,11 @@
+{
+  "name": "bar",
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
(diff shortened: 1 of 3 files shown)


othiym23:
I made a few tweaks to the tests (visible in 8b2ada82eed0d37292c71a1dfad4c30806ae68ce) to make it so that tests no longer depend on having write access to the tester's global prefix, and then rebased and landed this as 5811468e104ccb6b26b8715dff390d68daa10066. Thanks as always, @watilde!

watilde:
I didn't know how to not use global prefix in this case so your commit was really useful for me! Thanks as always too, @othiym23

Issue #7757, opened 2015-03-26
npm view prints a string instead of a list when there is only one version


rightfold:
For packages with many versions, `npm view --json $package versions` prints a JSON array:

    $ npm view --json coffee-script versions | head                                                                   
    
    [
      "0.7.0",
      "0.7.1",
      "0.7.2",
      "0.9.0",
      "0.9.1",
      "0.9.2",
      "0.9.3",
      "0.9.4",

However, for packages with only a single version, `npm view --json $package versions` prints a JSON string:

    $ npm view --json node-test versions
    "0.0.1"

This special case (which is invented; single-version packages aren't special in any way) makes it difficult to parse the output of `npm view`.

KenanY:
Consistency is nice. This is apparently an intentional thing, for neatness, perhaps. It extends beyond `versions`: any _Array_ whose length is `1` will be turned into a _String_, there's even a test for it (`maintainers` in this case):

``` js
t.similar(out, {
  maintainers: "jashkenas <jashkenas@gmail.com>"
}, "should have the same maintainer")
```

KenanY:
@rightfold This is now (https://github.com/npm/npm/commit/7d55ee5522dc51fa8f205c822485d826abd18130) fixed in [npm/npm#multi-stage](https://github.com/npm/npm/tree/multi-stage) so you'll see this in `npm@3` :)

ehsalazar:
We haven’t heard back and we’re trying to clean up some old issues.  If this is still a problem, can you please reply and let us know?  We’ll be happy to reopen if necessary.

PR #7758, opened 2015-03-26
view: always use arrays for versions, maintainers, etc.

kenany:
For consistency.

Fixes https://github.com/npm/npm/issues/7757

### DIFF
--- a/lib/view.js
+++ b/lib/view.js
@@ -259,11 +259,7 @@ function printData (data, name, cb) {
 }
 function cleanup (data) {
   if (Array.isArray(data)) {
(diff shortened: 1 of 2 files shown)


othiym23:
Could you rebase this change against `multi-stage`? Somebody, somewhere, could be relying on the output of `npm view` being in the old way, so this is (very technically) a semver-major change. I'm fine with landing it against `npm@3`, though.

KenanY:
@othiym23 Sure thing :) I've created a new PR since GitHub doesn't let you change the target branch https://github.com/npm/npm/pull/7770

PR #7766, opened 2015-03-26
github: fix the ssh fallback for private github modules

iarna:
This only impacts github modules installed via github shortcuts, eg

```
npm install project/module
```

When the module is marked as private.


othiym23:
:tada:

PR #7784, opened 2015-03-29
fix for 'npm: command not found'

giufus:
npm and npm-cli.js are in the same directory

### DIFF
--- a/bin/npm
+++ b/bin/npm
@@ -10,5 +10,5 @@ esac
 if [ -x "$basedir/node.exe" ]; then
   "$basedir/node.exe" "$basedir/node_modules/npm/bin/npm-cli.js" "$@"
 else
-  node "$basedir/node_modules/npm/bin/npm-cli.js" "$@"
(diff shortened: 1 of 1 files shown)


othiym23:
This script is meant to be used in certain very specific circumstances when bundled directly with Node, in which case it may be nowhere near `npm-cli.js`. Do you have a specific instance where you needed this? Thanks for taking the time to put this together, but I can't take this patch.

giufus:
I just remember that I could not launch npm after node installation on OS X 10.10, and the npm in my path was looking for that npm-cli.js

othiym23:
That's a bad install, then, because `npm -g install npm` and / or the installer should create a symbolic link from the global prefix's bin directory (typically `/usr/local/bin`) directly to `/usr/local/lib/node_modules/npm/bin/npm-cli.js`. If that isn't happening, there's either a bug in the Node installer / nvm / whatever you're using to install Node, or in npm itself.

PR #7799, opened 2015-03-30
fix: always quote args

### LINKED ISSUE #7743: npm run removes quotes from given parameters
I'm [using npm as a build tool](http://blog.keithcirkel.co.uk/how-to-use-npm-as-a-build-tool/) and ran into the following issue with `npm run` and quoted parameters.

This is the interesting part of the "scripts" section in my package.json:

```
"scripts": {
  "test": "mochify",
  "cover": "npm run test -- --plugin [ mochify-istanbul --report cobertura --exclude '**/*.mustache' ]"
}
```

Running `npm run cover` prints this:

```
> npm run test -- --plugin [ mochify-istanbul --report cobertura --exclude '**/*.mustache' ]
...
> mochify --plugin [ mochify-istanbul --report cobertura --exclude **/*.mustache ]
```

You can see that  `**/*.mustache` is quoted on the first line, but the quotes are removed on the second line. This causes the script to fail because I have to pass the glob expression as a plain string.

If I change the cover script to this:

```
"cover": "mochify --plugin [ mochify-istanbul --report cobertura --exclude '**/*.mustache' ]"
```

... calling `npm run cover` works fine and npm prints the quotes as expected:

```
> mochify --plugin [ mochify-istanbul --report cobertura --exclude '**/*.mustache' ]
```

Thanks for looking into it!

mantoni:
This fixes #7743 by simply always quoting all passed arguments after `--`.

I struggled to write an automated test case, although it's actually quite easy to reproduce.
Put a `package.json` with this content into an empty directory:

```
{
  "name": "test",
  "version": "1.0.0",
  "scripts": {
    "test": "npm run echo -- '*.json'",
    "echo": "echo"
  }
}
```

Without the fix `npm t` will print

```
> test@1.0.0 test /Users/max/projects/npm-test
> npm run echo -- '*.json'

> test@1.0.0 echo /Users/max/projects/npm-test
> echo *.json

package.json
```

With the patch applied, the output is

```
> test@1.0.0 test /Users/max/projects/npm-test
> npm run echo -- '*.json'

> test@1.0.0 echo /Users/max/projects/npm-test
> echo "*.json"

*.json
```

The added test case only asserts that args with `'` get passed on correctly. This already worked without this fix, but it wasn't tested yet.

### DIFF
--- a/lib/run-script.js
+++ b/lib/run-script.js
@@ -189,8 +189,7 @@ function run (pkg, wd, cmd, args, cb) {
 function joinArgs (args) {
   var joinedArgs = ""
   args.forEach(function(arg) {
-    if (arg.match(/[ '"]/)) arg = '"' + arg.replace(/"/g, '\\"') + '"'
-    joinedArgs += " " + arg
+    joinedArgs += ' "' + arg.replace(/"/g, '\\"') + '"'
   })
   return joinedArgs
 }
--- a/test/tap/run-script.js
+++ b/test/tap/run-script.js
@@ -111,6 +111,10 @@ test('npm run-script with args that contain double quotes', function (t) {
   common.npm(['run-script', 'start', '--', 'what"s "up"?'], opts, testOutput.bind(null, t, 'what"s "up"?'))
 })
 
+test('npm run-script with args that contain ticks', function (t) {
+  common.npm(['run-script', 'start', '--', 'what\'s \'up\'?'], opts, testOutput.bind(null, t, 'what\'s \'up\'?'))
+})
+
 test('npm run-script with pre script', function (t) {
   common.npm(['run-script', 'with-post'], opts, testOutput.bind(null, t, 'main;post'))
 })


mantoni:
@smikes @KenanY Thoughts?

KenanY:
@mantoni Looks good, especially since the `run-script` tests still pass. Trying to think of a test for this.

smikes:
This seems like an improvement, but I am not sure what level of change it is.  I still think I could construct pathological filenames or argument values that defeat this added quoting (bash quoting is, at least, a context-free language... possibly worse).  I think this has to be a next-major change.

mantoni:
@smikes The behavior would change for people that assume bash expansion for `*.js` is applied and with this patch it isn't anymore. I don't have any strong feelings about this. Feel free to go with the safe option.

othiym23:
This is a potentially far-reaching change that is pretty hard to test with my current matrix of environments, so I may have to revert it / push it to `npm@3` as a breaking change if I get feedback that it's causing problems. Thanks for taking this on, though! Rebased, tested as best I was able, and landed as 78005ebb6f4103c20f077669c3929b7ea46a4c0d.

mantoni:
:+1:

mantoni:
Just came accross [shell-quote](https://www.npmjs.com/package/shell-quote). This seems to handle a lot of edge cases and might be worth a look to have properly quoted shell parameters in npm run-script.

Issue #7798, opened 2015-03-30
document logging when using npm programmatically


lookfirst:
https://github.com/npm/npm#using-npm-programmatically

```
  npm.registry.log.on("log", function (message) { .... })
```

It seems that `npm.registry` no longer exists, what is the replacement?

I'd also like to see some documentation on how to insert my own logger or at least better control the npm one. I use winston and would like to log through that instead.

othiym23:
`npm.registry` is the registry client, and it absolutely should be there after the call to `npm.load`, which bootstraps npm's internals. `npm.registry.log` is passed into the `npm-registry-client` instance at `npm.load()` time, when the registry client is instantiated, and is the same thing you'd get if you called `require('npmlog')` from inside the code using the npm API. If you _have_ called `npm.load()` and `npm registry.log` isn't there, that's a bug. That code example should work as-is.

There is no reasonable way to swap out `npmlog` for another logging framework right now. As the name indicates, it's pretty much baked into npm at a very low level. And, as I mentioned on #7800, the API is kind of the Wild West. If you want documentation on how to use `npmlog` as a passthrough for another logger (which should be possible), you'll need to read [its documentation](https://www.npmjs.com/package/npmlog) for the details.

lookfirst:
I stand corrected, I just tested the code again and it was fine. Sorry about that.

PR #7800, opened 2015-03-30
use built in logging package instead of console.log

lookfirst:


othiym23:
This might not be apparent from looking at the code, but everywhere where `console.log` is used in lieu of `npmlog` is a place where the output is part of the contract of the npm subcommand being run. The decision to use `console.log` is intentional, and isn't likely to change without a specific case being made for the change in each case. Thanks for putting this together, though!

lookfirst:
I'm using npm programmatically and I'd like more control over the output of npm. It is kind of a mess to not only configure npmlog during the npm.load(), but also to have to monkey patch console.log.

If you have a better suggestion, I'm all ears, but it feels like the programmatic case isn't being considered here.

```
	install(dependencies) {
		new Promise((resolve, reject) => {
			let stream = winstonStream(log, 'trace'); // make npm a bit more silent. sigh.

			npm.load({'logstream': stream}, (err) => {
				if (err) return reject(err);

				let log = console.log;
				console.log = function() {};  // make npm a bit more silent. sigh.

				npm.commands.install(dependencies, (err, data) => {
					console.log = log;

					if (err) return reject(err);
					resolve(data);
				});
			})
		});
	}
```

lookfirst:
#7798 related

lookfirst:
I think it would make more sense to abstract out the call to `console.log` to some logging facility that by default writes to `console.log`, but at least it can be configured to be some other output by anyone wanting to have more control over things, without having to monkey patch `console.log`. 

It would also make it more clear in the code the reasoning that you've suggested because then that facility could have documentation for **why** it is being used instead of `npmlog`.

Is that a PR you'd accept?

othiym23:
You're right! That the API is currently exposed and documented is more or less a consequence of various community contributions, and is not really a statement about the API's fitness for use outside the project. It's not really meant for outside use – breaking changes to the API don't count towards the CLI's semver bump. The API is absolutely a second-class citizen within the npm universe.

Not to be presumptuous, but I'm pretty sure that's not the response you were looking for. It is [absolutely part of the long-term roadmap for npm](http://blog.npmjs.org/post/91303926460/npm-cli-roadmap-a-periodic-update) that we make it easier to consume the various pieces of npm as APIs and built your own CLIs out of npm's components. However, we're still a substantial distance from that goal. Until then, the main thing that keeps us from making changes to the API is mostly a desire to avoid having to fix up all the tests.

My personal opinion is that until we get to the point where npm does have an official set of APIs with their own semvers and documentation, the safest / best way to use npm is as a child process, rather than trying to consume the current API.

othiym23:
> Is that a PR you'd accept?

No. It really does not make sense to me to have two independent frameworks controlling I/O from the CLI. If you _really_ don't want the standard output from npm being displayed, a child process with stdout redirected or suppressed is the right way to go.

lookfirst:
Technically, you already have two independent frameworks controlling I/O from the CLI (npmlog and console.log). I agree, adding a third doesn't make much sense, so maybe it would belong in npmlog. Maybe even just a `log.console` method. Then at least I don't need to monkey patch `console.log`.

The documentation should clearly state what you just said above, that isn't clear at all right now.

https://docs.npmjs.com/api/npm
https://github.com/npm/npm#using-npm-programmatically

As far as shelling out to the `npm` cli... well, I can only say I'm disappointed, but thanks for your responses. :cry:

othiym23:
Which would you use: an API that isn't really tested as such, with a bunch of known warts and issues, that wasn't really designed to be consumed as an API, or an API that every user of the product uses all day every day, where issues are fixed as soon as is practicable after they're noticed, and where considerable thought has gone into its ergonomics and usability? I get that it seems hacky and uses more resources to shell out to a child process, but it's the way npm is meant to be used – just look at the tests, which primarily shell out to child processes themselves.

lookfirst:
Maybe instead of documenting the API that is there, it should be removed from the site and not exposed at all. Heck, maybe even provide a nice wrapper around shelling out to the npm child process and call that the 'api'.

othiym23:
> Maybe instead of documenting the API that is there, it should be removed from the site and not exposed at all.

I would not be opposed to this on principle, but I think enough people are using the API, even with all its problems and caveats, that nuking the existing documentation would be needlessly hostile. I'm kinda sorta the boss of what goes into the CLI code base, but I'm not the boss of the npm user community.

iarna:
On the other hand, I think we could get behind a patch to the API docs that notes their intended audience and caveats about their usage.

And personally I'd love to see a wrapper around npm to make the command line version easy to call as an API, but I'd like to see that developed and worked out as an independent module. That is to say, you don't need our permission to make that! And if it works out well, I'd be for linking to it from API docs as an alternative.

PR #7804, opened 2015-03-31
add support for nw.js

xueeinstein:
details please visit:
https://github.com/xueeinstein/nw-npm/blob/master/README.md

### DIFF
--- a/lib/utils/lifecycle.js
+++ b/lib/utils/lifecycle.js
@@ -89,19 +89,19 @@ function lifecycle_ (pkg, stage, wd, env, unsafe, failOk, cb) {
   if (packageLifecycle) {
     // define this here so it's available to all scripts.
     env.npm_lifecycle_script = pkg.scripts[stage]
-	var projNPMrc = null
-	if (npm.config.sources.project && npm.config.sources.project.data) {
-		projNPMrc = npm.config.sources.project.data;
-	}
+    var projNPMrc = null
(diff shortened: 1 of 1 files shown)


KenanY:
Indentation seems off.

othiym23:
What are you trying to do with this change? Why do these specific changes need to be made to npm itself, and within `nw-gyp`?

xueeinstein:
This change is for nw.js to install node.js native modules. `nw-gyp` is forked and developed from `node-gyp`. It'll download our customized node.js header files for specified nw.js version, in this way the built binary file could be compatible with nw.js ABI. 

PS, this code snippet only execute when `nw_version` is set up in project `.npmrc`. __No other side effect if `nw_version` isn't set up__. (`nw.js` is renamed from `node-webkit`)

xueeinstein:
@KenanY , sorry, my vim indentation setup doesn't follow npm coding principle. I have corrected it.

othiym23:
I think this is too special-purpose to go directly into npm. I'd be amenable to extending the support we've added for [`--node-gyp`](https://github.com/npm/npm/commit/145af6587f45de135cc876be2027ed818ed4ca6a) to work with `nw-gyp` (if it's even necessary – that may work as a drop-in for the build scripts that `nw.js` projects use with a little futzing. But this isn't a general-purpose enough change to go into npm proper. Thanks for the time you spent putting it together and explaining your use case, though!

PR #7820, opened 2015-04-01
Update npmlog’s dependency are-we-there-yet to v1.0.4

sonicdoe:
are-we-there-yet v1.0.4 is [now compatible with Node.js v0.8](https://github.com/iarna/are-we-there-yet/pull/2).


othiym23:
Tweaked a little, rebased, and landed as 1f380f66c1e944b8ffbf096fa94d09e931626e12. Thanks for helping npm keep working under Node 0.8!

Issue #7842, opened 2015-04-03
tests fail against Node 0.8.28


othiym23:
* [ ] `test/tap/dedupe-scoped.js` (2/5)
* [ ] `test/tap/ls-depth-unmet.js` (10/20)
* [ ] `test/tap/peer-deps-invalid.js` (2/5)
* [ ] `test/tap/peer-deps-toplevel.js` (2/5)
* [ ] `test/tap/peer-deps.js` (0/2)
* [ ] `test/tap/shrinkwrap-scoped-auth.js` (3/6)
* [ ] `test/tap/sorted-package-json.js` (3/7)
* [x] `test/tap/spawn-enoent-help.js` (1/3)
* [x] `test/tap/umask-lifecycle.js` (0/3)
* [ ] `test/tap/update-save.js` (24/27)
* [ ] `test/tap/url-dependencies.js` (7/9)

## nock requires `setImmediate`

* [x] `test/tap/publish-access-scoped.js` (3/4)
* [x] `test/tap/publish-access-unscoped.js` (3/4)
* [x] `test/tap/publish-scoped.js` (2/3)

## `npm-registry-couchapp` tests require Node.js 0.10+ / io.js

* [x] ~~`test/tap/registry.js`~~

smikes:
These are now fixed in master: 
 test/tap/dedupe-scoped.js (2/5)
 test/tap/peer-deps-toplevel.js (2/5)
 test/tap/peer-deps.js (0/2)
test/tap/url-dependencies.js (7/9)
test/tap/update-save.js (24/27)

#7850 contains a test for spawn-enoent-help (thanks @SonicHedgehog, I introduced that one).

Only ones still failing are the publish-access tests.  I will look at replacing `nock` with `mockRegistry`

othiym23:
I think there may be an issue with newish versions of tap on Node 0.8, at least when used as the test runner. @isaacs, does this ring any bells with you? In the output below, I've marked the tests that work when run directly from the command line.

Current failure output on my machine when tests are run via `nave use 0.8.28 npm run-script test-all`:

* [`test/tap/dedupe-scoped.js`](https://gist.github.com/othiym23/eab27c1bff197b0442a7#file-npm-run-script-test-all-node-0-8-28-log-L2451) (works when run directly)
[...]

PR #7847, opened 2015-04-03
Fix failing umask-lifecycle tests under Node.js v0.8

sonicdoe:
Under Node.js v0.8, [`node -p` alone does nothing](https://github.com/joyent/node/issues/3938), so I have added `-e`.

This checks off one item on #7842.


othiym23:
Landed in 54cd8956ea783f96749e46597d8c2cb9397c5d5f. Thanks for helping out!

PR #7873, opened 2015-04-07
Update hosted-git-info and the many related modules

iarna:
All so that hosted git normalization doesn't change protocols on us.


othiym23:
Closing in favor of #7907, which has all my changes to take advantage of these updates and a bunch more context about why the updates happened. Thanks a ton for doing all this work, Rebecca!

PR #7893, opened 2015-04-09
test/tap-slow: pull out slow tests into own dir

smikes:
create slow test dir 
run builtin-config in tmp dir 
move builtin-config and registry to tap-slow


othiym23:
"Slow vs test" isn't a great partitioning of the tests – better to increase the overall timeout so that the tests have enough time to run, and then partition the tests into "unit" and "functional", which have no external dependencies (network or programs), and "integration" (which depend on external resources). Thanks for putting this together, though!

PR #7907, opened 2015-04-10
Handle git remotes more consistently.

### LINKED ISSUE #7867: handle GitHub shorthand syntax consistently
Here are the apparently not-so-simple requirements:

1. GitHub shorthand syntax (`org/repo`) should be handled the same way by the CLI, regardless of whether it's passed in via the command line, or edited into `package.json`.
2. `--save`, `--save-dev`, and `--save-optional` should save GitHub dependencies _as shorthand syntax_, and not normalized to `git+whatever:` URLs, so that the symmetry of using them is preserved between the first and subsequent times you install them (shrinkwrapped GitHub dependencies should continue to resolve to `git+x:` URLs + resolved commit hashes).
3. Explicit Git URLs pointing to GitHub repos should preserve the protocol that was provided either on the command line or in `package.json`.
4. Explicit Git URLs pointing elsewhere should also be left alone, whether passed in on the command line, or via `package.json`.

We've made a few passes at this, which can be seen in #2556, #6978, #7121, #7486, #7630, #7766, and #7829. It's time to nail this down once and for all, and to get the right set of tests around it to prevent it from regressing in the future (although that is proving to be hard to do).

othiym23:
- Save the lightest-possibly normalized URL into `package.json` and `_from` fields in cached `package.json` files, because `read-package-json`, `read-installed`, and the cache all know how to handle them now.
- Add support for `github:`, `gist:`, `bitbucket:`, and `gitlab:` shortcut syntax. GitHub shortcuts will continue to be normalized to `org/repo` instead of being saved with a `github:` prefix (for now).
- If presented with shortcuts, try cloning via git, SSH, and HTTPS (in that order, skipping any methods that aren't supported by a given hosting provider).
- No longer prompt for credentials -- it didn't work right with the spinner, and wasn't guaranteed to work anyway. We may experiment with doing this a better way in the future. Users can override this by setting `GIT_ASKPASS` in their environment if they want to experiment with interactive cloning, but should also set `--no-spin` on the npm command line (or run `npm config set spin=false`).

The code changes to `lib/cache/add-remote-git.js` are less extreme than they look at first, because I decided to go back to explicit parameter passing because with the new logic it was less confusing that way. I'd like feedback on whether the code comments are adequate to figure out what's going on, and if any of them are clutter at this point.

Fixes #7867.

**r**: @iarna


WORMSS:
Is it possible to do --no-spin --git-askpass
I am trying to write a bat file to install some development tools. In pre 2.8.0 this was working fine. Until someone today installed the latest node.js which had npm v2.11.something.. After a lot of headaches I found that you guys decided to stop asking for usernames and passwords for https.. 

I need to be able to set --git-askpass during the install process.
So I can do something like the below in a .bat file.

````shell
call npm install -g --no-spin --git-askpass git+https://github.com/*******/*******.git
````

We are not npm experts, I would have a lot of blank looks if I asked them to set GIT_ASKPASS in their npm global environment. And to be honest, I don't know how to do that either without having to look it up.

PR #7920, opened 2015-04-11
Fix target folders for peerDependencies of scoped packages.

### LINKED ISSUE #7454: Peer dependencies of namespaced packages installed wrong
When installing a package with a namespace, e.g. @foo/mymodule, if it has a peer dependency that doesn't exist, that dependency is installed automatically. However, it is installed into ./node_modules/node_modules/package instead of ./node_modules/package, giving a directory tree like this:

    package.json
    node_modules
        @foo
            mymodule
                package.json
        node_modules
            peer
                package.json

ewie:
Package scopes cause an additional level in the tree structure which
must be considered when resolving the target folders of a package's
peerDependencies.

Fixes #7454.

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -1128,7 +1128,12 @@ function write (target, targetFolder, context, cb_) {
               "in npm 3+. Your application will need to depend on it explicitly."
             ], pd+","+data.name)
         })
-        var pdTargetFolder = path.resolve(targetFolder, "..", "..")
+
+        // Package scopes cause an addditional tree level which needs to be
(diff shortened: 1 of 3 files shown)


othiym23:
Rebased and landed as b0273190c71eba14395ddfdd1d9f7ba625297523 (with a few tweaks as 649e31ae4fd02568bae5dc6b4ea783431ce3d63e – I'm trying to future-proof how npm deals with scoped packages a little by standardizing how we test for scopedness). This is a great find, as we're very close to rolling out support for scoped packages on the registry, so thank you very much for putting this together!

PR #7923, opened 2015-04-11
Tweaks to npm test-all

watilde:
Updates:
- Sync timeout time of test-all to test and tap in scripts.
- Use an alias of `scripts` and `run-scripts` in `test-all`.

Thanks.

### DIFF
--- a/package.json
+++ b/package.json
@@ -184,7 +184,7 @@
     "test-legacy": "node ./test/run.js",
     "test": "tap --timeout 240 test/tap/*.js",
     "tap": "tap --timeout 240 test/tap/*.js",
-    "test-all": "node ./test/run.js && tap --timeout 240 test/tap/*.js",
+    "test-all": "npm run test-legacy && npm test",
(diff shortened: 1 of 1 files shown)


othiym23:
Perfect! I don't know why we didn't think to do this ourselves sooner, since most of my personal projects use recursive npm scripts like this. Thank you!

watilde:
Thank always!

Well, I thought when I made this PR, `npm run tap` and `npm test` looks the same. Has npm run tap used somewhere...?

othiym23:
I don't know, which is the same as saying we can't take it out, because maybe somebody's CI relies on that task and that's not a good reason to break it. I thought about having one of `npm test` and `npm run tap` call the other, but I decided that was just too much. The problem with npm recursively invoking npm is that it makes it much more difficult for people running on hardware-constrained environments (like Beaglebone Blacks and Raspberry Pis) to set up CI hosts, because they're memory constrained.

Anyway, it's a good question, and keep asking them!

PR #7924, opened 2015-04-11
remove child-process-close

robertkowalski:
we don't support node 0.6 any more


othiym23:
I combined the two commits into one (since they're basically doing the same thing), rebased, and landed this as d90d0b992acbf62fd5d68debf9d1dbd6cfa20804. This is a good bit of cruft to remove – at some point we may even want to add check to npm to prevent it from starting when run under unsupported Node versions. It's a tiny bit user-hostile for people who are running on embedded platforms (or otherwise stuck on old versions), but they're probably better off using the version of npm that came with their platform, with all its limitations than trying to use a tool that has many parts that aren't suited for it. Thanks for putting this together, Robert!

PR #7934, opened 2015-04-13
outdated/update: support for private local modules

### LINKED ISSUE #7426: Update of local dependencies does not work
I have some local dependencies on sub-modules in the same git repo like this:

```
"dependencies": {
  "sub-mod-0": "./sub-mod-0",
  "sub-mod-1": "./sub-mod-1"
}
```

With this I can do `require('sub-mod-0')` without having to use something like `require('../../sub-mod-0')`

Installation via `npm install` works fine and populates `node_modules/sub-mod-0` and `node_modules/sub-mod-1` but updating them is not possible without manually deleting `node_modules/sub-mod-*` beforehand. Neither `npm install` nor `npm update` nor `npm update sub-module-*` have an effect even if the version of the sub-module is incremented.

**Environment:**
npm v2.6.0
node v0.10.25
Ubuntu 14.04.1 LTS

ArnaudRinquin:
Do not ignore them and read the `wanted`+`latest` version from their local `package.json`

This fixes #7426

### DIFF
--- a/lib/outdated.js
+++ b/lib/outdated.js
@@ -301,7 +301,7 @@ function shouldUpdate (args, dir, dep, has, req, depth, cb, type) {
   }
 
   if (args.length && args.indexOf(dep) === -1) return skip()
-  var parsed = npa(req)
+  var parsed = npa(dep + '@' + req)
   if (parsed.type === "git" || (parsed.hosted && parsed.hosted.type === "github")) {
     return doIt("git", "git")
   }
@@ -313,8 +313,34 @@ function shouldUpdate (args, dir, dep, has, req, depth, cb, type) {
     npm.registry.get(uri, { auth : auth }, updateDeps)
   })
 
+  function updateLocalDeps (latestRegistryVersion) {
+    readJson(path.resolve(parsed.spec, "package.json"), function(er, localD){
+      if(er) return cb()
+      var wanted = localD.version;
+      var latest = localD.version;
+
+      if (latestRegistryVersion) {
+        latest = latestRegistryVersion;
+        if (semver.lt(wanted, latestRegistryVersion)) {
+          wanted = latestRegistryVersion;
+          req = dep + '@' + latest;
+        }
+      }
+
+      if(curr.version !== wanted){
+        doIt(wanted, latest);
+      } else {
+        skip()
+      };
+    });
+  }
+
   function updateDeps (er, d) {
-    if (er) return cb()
+    if (er) {
+      if(parsed.type !== "local") return cb()
+      return updateLocalDeps();
+    }
+
     if (!d || !d["dist-tags"] || !d.versions) return cb()
     var l = d.versions[d["dist-tags"].latest]
     if (!l) return cb()
@@ -355,6 +381,9 @@ function shouldUpdate (args, dir, dep, has, req, depth, cb, type) {
       if (!curr || dFromUrl && cFromUrl && d._from !== curr.from
           || d.version !== curr.version
           || d.version !== l.version) {
+        if (parsed.type === "local") {
+          return updateLocalDeps(l.version);
+        }
         doIt(d.version, l.version)
       }
       else {
--- a/test/tap/outdated-local.js
+++ b/test/tap/outdated-local.js
@@ -1,0 +1,167 @@
+var common = require("../common-tap.js")
+var test = require("tap").test
+var npm = require("../../")
+var rimraf = require("rimraf")
+var path = require("path")
+var mr = require("npm-registry-mock")
+var osenv = require("osenv")
+var mkdirp = require("mkdirp")
+var fs = require("graceful-fs")
+
+var pkg = path.resolve(__dirname, "outdated-local")
+var pkgLocal = path.resolve(pkg, "local-module")
+var pkgScopedLocal = path.resolve(pkg, "another-local-module")
+var pkgLocalUnderscore = path.resolve(pkg, "underscore")
+var pkgLocalOptimist = path.resolve(pkg, "optimist")
+
+function mocks (server) {
+  server.get("/local-module")
+    .reply(404)
+  server.get("/@scoped%2fanother-local-module")
+    .reply(404)
+}
+
+test("setup", function (t) {
+  bootstrap()
+  t.end()
+})
+
+test("outdated support local modules", function (t) {
+  t.plan(3)
+  process.chdir(pkg)
+  mr({ port : common.port, plugin: mocks }, function (err, s) {
+    npm.load(
+      {
+        loglevel  : "silent",
+        parseable : true,
+        registry  : common.registry
+      },
+      function () {
+        npm.install(".", function (err) {
+          t.ifError(err, "install success")
+          bumpLocalModules()
+          npm.outdated(function (er, d) {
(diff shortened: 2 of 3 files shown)


ArnaudRinquin:
Test are expected to be broken as they haven't been updated yet, especially `test/tap/outdated-private.js`.

ArnaudRinquin:
Tests are now all passing except for the `node 0.8.x` target which is [expected at the moment](https://github.com/npm/npm/issues/7842).

As previously commented, their is a breaking change in `npm outdated` behaviour towards locally _cloned_ modules that also exist on repository (i.e: `file:underscore`): 

Previously: newer version on the repository would have have shown on `outdated` and triggered an updated on `update`.
Now: if the module is local (i.e. `file:`), we completely ignore / skip the repository request and do the update if locale `package.json` version is different.

ArnaudRinquin:
I just rebased to squash the commits into a single one matching the commit style.

othiym23:
Having `npm outdated` handle local packages more consistently would be very nice. However, I don't think that the presence of a local module in the tree should short-circuit checking the versions in the registry altogether, which is what the current patch does. Some developers are using local packages as a way to work on a number of packages simultaneously, and while they may not want to use the versions from a registry, it's useful to at least tell them those versions are available. Semver should apply either way, but `outdated` should still check the registry either way.

Also, the current tests conflate `local` modules and `private` modules. Something installed from a local path isn't necessarily private – only a package with `"private": true` in `package.json` is truly private. The current test code is confusing because of the way that it suggests that a local package is private.

If you want to use the `private` flag to short-circuit registry checking for versions, that would make sense. In that case, though, you should probably rebase this against the `multi-stage` (what will be `npm@3`) branch (and create a new PR, because GitHub doesn't support changing the base branch of existing PRs). It's a very slight breaking change, but it's still breaking for the (probably theoretical) case in which developers are shadowing a registry dependency with a private one and actually want the registry version to override their locally-installed version.

If you can get those two things addressed, I'll be happy to land this. Thanks for putting this together, and thanks for your enthusiasm!

KenanY:
> use the `private` flag to short-circuit registry checking for versions

This already happens since https://github.com/npm/npm/pull/6862?

othiym23:
Oh hey, look at that! Thanks for reminding me, @KenanY! That means that the scope of this patch should be very narrow: add local checking of outdated versions without removing checking the registry.

ArnaudRinquin:
It is indeed. I can restore the registry lookup easily. However, I am not sure what to do with the result.

Is this correct:

| node_modules | local_modules | registry | wanted | latest | update?             |
|---------|--------------------|----------|--------|--------|---------------------|
| 1.0.0   | 1.0.0              | 1.0.0    | 1.0.0  | 1.0.0  | no                  |
| 1.0.0   | 1.1.0              | 1.0.0    | 1.1.0  | 1.1.0  | yes                 |
| 1.0.0   | 1.0.0              | 1.1.0    | 1.0.0  | 1.1.1  | no                  |
| 1.0.0   | 1.1.0              | 1.2.0    | 1.1.0  | 1.2.0  | yes, to local 1.1.0 |
| 1.0.0   | 1.1.0              | none     | 1.1.0  | 1.1.0  | yes                 |

ArnaudRinquin:
This array boils to a few rules:

In the case of a local module:
* `wanted` is the local module version, no matter what.
* `latest` is the higher Semver between local and registry.
* update only if `wanted` is different that `current`, using local module.

ArnaudRinquin:
On the two latest commits, I:

* reverted the private package handling to the way it was before (which means no `package.json` check modules that are both local and `private`)
* I have updated the behaviour to match the array and rules described in the previous commit.
* I have added tests for local modules. These are broken at the moment.

The new tests are broken because the `npm-registery-mock` component doesn't return a `404`, or a regular http error when requested package doesn't exist but throws an error instead.

I am not sure how to fix that, some help would be welcome.

ArnaudRinquin:
@othiym23 Any hint on handling `404` on `npm-registery-mock`?

othiym23:
> Is this correct:

No. It shouldn't matter if it's the local or registry version that's newer, only what's the newest version that matches the semver range for the dependency in `package.json`. To do anything else would be confusing to anyone who wasn't paying attention when this decision was made. `package.json` is the source of truth, not the source of installation of the package.

So:

| current (node_modules) | local_modules | registry | wanted | latest | update?             |
|---------|--------------------|----------|--------|--------|---------------------|
| 1.0.0   | 1.0.0              | 1.0.0    | 1.0.0  | 1.0.0  | no                  |
| 1.0.0   | 1.1.0              | 1.0.0    | 1.1.0  | 1.0.0  | yes                 |
| 1.0.0   | 1.0.0              | 1.1.0    | 1.1.0  | 1.1.0  | no                  |
| 1.0.0   | 1.1.0              | 1.2.0    | 1.2.0  | 1.2.0  | yes, to registry 1.2.0 |
| 1.0.0   | 1.1.0              | none     | 1.1.0  | 1.1.0  | yes                 |

> Any hint on handling `404` on `npm-registry-mock`?

Take a look at [this chunk](https://github.com/npm/npm/blob/master/test/tap/404-parent.js#L43-L46) of an existing test. You'll need to write a similar plugin. Also take a look at the new [code review process](https://github.com/npm/npm/wiki/code-review-process) for some tips on what the code should look like to be easiest to land.

Thanks for your patience! This will be a nice little improvement once it's whipped into shape.

ArnaudRinquin:
I just pushed a new version that matches what you described.

However, I find it a bit weird in this exemple scenario:

1. Clone `underscore 1.0.0` into a local module
1. `npm install` locally (`saved in package.json as "underscore": "file:local_modules/underscore"`)
1. Modify it, bump local version to `1.1.0`
1. Independently, registry version of `underscore` is bumped to `1.2.0`
1. You `npm update` to use your latest local version: registry `1.2.0` version is installed and your `package.json` dependency becomes `underscore@1.2.0`, overwriting your local module setup.

othiym23:
That's exactly the behavior I want – the fact that the package is local shouldn't take precedence over the fact that a newer version of the package that matches the dependency's semver range is available. If you're concerned about upstream changes overwriting your local modules setup, you probably shouldn't be using `npm update`.

ArnaudRinquin:
I understand, but `npm update` is the only way to apply your local module changes to its `node_module` copy. That is what motivated me to fix the local `update` in the first place.

othiym23:
How do you expect this to become a problem in practice? It seems much more confusing to me that where the dependency is sourced would imply a precedence that supersedes semver.

ArnaudRinquin:
Well, in `package.json` we declare local module specifically, using `file:path` and no semver. This has a strong meaning. I understand it as "whatever exists, just use what's in that folder" and semver is just here for `outdated` to know when to update local copy.

othiym23:
Rebased and landed in 279932298ce5b589c5eea9439ac40b88b99c6a4a (with some minor tweaking, mostly for style). Thanks for putting this together, and also for your patience! Let's see how this works, and maybe once `npm@3` is out we can tweak it to incorporate behavior like `linklocal`'s.

ArnaudRinquin:
Great! That's a good one. I think it covers pretty much all the local modules use cases and opens ways of using npm.

:wine_glass:

PR #7946, opened 2015-04-14
Make npm init text a little more friendly

sandfox:
I'm not sure we should use the word "sane" to refer settings etc, it seems a little ableist and there are nicer and more accurate words we could use instead.

_Personally it gets me a little down every time I make a new package._

### DIFF
--- a/lib/init.js
+++ b/lib/init.js
@@ -17,7 +17,7 @@ function init (args, cb) {
   if (!initJson.yes(npm.config)) {
     console.log(
       ["This utility will walk you through creating a package.json file."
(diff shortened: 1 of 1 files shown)


othiym23:
:+1: good looking out, thank you. I may tweak the wording a little bit, but I'll try to get to this today.

othiym23:
This is landed as 1fe584089f5bef133de5518aa26eaf6064be2bf7. Thanks! Please do submit more of these kinds of tweaks, or just file issues as you notice them! I really appreciate this kind of feedback, and I'm sure the rest of the team does as well.

sandfox:
thanks @othiym23 - means a lot :-)

Issue #7961, opened 2015-04-15
2.8.2 git shrinkwrap contains git+null


KidkArolis:
With 2.8.2, we've run `npm install qubitdigital/ambassador --save && npm shrinkwrap` and the shrinkwrap was updated to:

```
"ambassador": {
       "version": "1.0.0",
       "from": "qubitdigital/ambassador",
       "resolved": "git+null:git@github.com:qubitdigital/ambassador.git#a37f7f62fb1636cb0fe08fc65a233218a0fab334"
    }
```

Notice the `git+null` - that looks like it's incorrect.

othiym23:
That is definitely incorrect. It looks like `hosted-git-info` may be returning bogus SSH URLs in some cases. I'll take a look. Thanks!

iarna:
wow,
`git+null:git@github.com:qubitdigital/ambassador.git#a37f7f62fb1636cb0fe08fc65a233218a0fab334`
is extra invalid.

It looks like `'git+' + protocol + ':' + info.ssh()` and not 'info.sshurl()'

An actual valid URL would look like:

```
git+ssh:git@github.com/qubitdigital/ambassador.git#a37f7f62fb1636cb0fe08fc65a233218a0fab334
```

iarna:
Ah, yup:

https://github.com/npm/npm/blob/master/lib/cache/add-remote-git.js#L87

Should be sshurl()

Which is then further mangled by:

https://github.com/npm/npm/blob/master/lib/cache/add-remote-git.js#L415-L419

To result in the confusing bit above.

othiym23:
Thanks for tracking that down, @iarna. For cloning purposes, npm needs `.ssh()`, but for saving purposes, is needs `.sshurl()` – it may be necessary to pass the parsed `hostInfo` all the way through, at least as far as `getResolved`. I'll investigate.

othiym23:
Fixed in 387f889c0e8fb617d9cc9a42ed0a3ec49424ab5d.  Thanks for bringing this up! This will be part of the next patch release, going out very soon (like, less than an hour).

othiym23:
...and indeed, this was included as part of `npm@2.8.3`.

iarna:
For the record, just using `sshurl` proved to be insufficient, as `sshurl` now returns a `git+ssh` protocol and git only actually supports the `ssh` protocol and scp-style non-urls.

PR #7965, opened 2015-04-15
git: resolved git URLs should be in normal form

othiym23:
Candidate fix for #7961.

### DIFF
--- a/test/tap/add-remote-git-get-resolved.js
+++ b/test/tap/add-remote-git-get-resolved.js
@@ -1,0 +1,101 @@
+'use strict'
+var test = require('tap').test
+
+var npm = require('../../lib/npm.js')
+var common = require('../common-tap.js')
+
+test('setup', function (t) {
+  var opts = {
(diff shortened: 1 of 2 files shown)


test/tap/add-remote-git-get-resolved.js
```diff
@@ -0,0 +1,96 @@
...
+  verify('git:github.com/foo/repo')
+  verify('git:github.com/foo/repo.git')
+  verify('git://github.com/foo/repo#decadacefadabade')
+  verify('git://github.com/foo/repo.git#decadacefadabade')
+
+  function verify (uri) {
+    t.equal(
+      getResolved(uri, 'decadacefadabade'),
+      'git://github.com/foo/repo.git#decadacefadabade'
+    )
```
iarna: I'd really like there to be a label on this assert, even if it's just `uri`

iarna:
w/ assert labels, LGTM

PR #8007, opened 2015-04-19
update npm-scope to note npm private modules

smikes:


smikes:
on reflection I think the dates should be removed.  I'll let this sit to see if I get any more comments, and then rewrite it.

doc/misc/npm-scope.md
```diff
@@ -13,9 +13,10 @@ followed by a slash, e.g.
 Scopes are a way of grouping related packages together, and also affect a few
 things about the way npm treats the package.
 
-**As of 2014-09-03, scoped packages are not supported by the public npm registry**.
-However, the npm client is backwards-compatible with un-scoped registries, so
-it can be used to work with scoped and un-scoped registries at the same time.
+As of 2015-04-19, scoped packages **are** not supported by the public
```
KenanY: So remove the "not" here?

smikes: Nice catch Thx

othiym23:
Squashed, reworded commit message, and landed in 546626059909dca1906454e820ca4e315c1795bd. Thanks, Sam!

StephanBijzitter:
The documentation is good and this PR is definitely a good one, but the fact that `private: true` does not allow you to publish privately is quite stupid.

Issue #8011, opened 2015-04-20
Use HTTP_PROXY and HTTPS_PROXY for proxy config


mdamien:
Looks like npm is ignoring the HTTP_PROXY and HTTPS_PROXY env var.

Looks like the problem is here: https://github.com/npm/npm/blob/448efd0eaa6f97af0889bf47efc543a1ea2f8d7e/lib/cache/caching-client.js#L196-L197

Would be more convenient if npm would just follow the env vars if nothing is set.

The code to modify is here: https://github.com/npm/npm/blob/91f5cb1fb91520fbe25a4da5b80848ed540b9ad3/lib/config/core.js#L348-L365

Will do a PR when I have the time

smikes:
Thanks for the bug report.

`npm` doesn't use the http_proxy or https_proxy environment variables directly.  Instead, `npm`, uses the `request` module (see https://github.com/request/request), which is expected to honor the variables.

Can you tell us a bit more about the context of the error you are seeing?

mdamien:
In fact, I made the PR a bit too fast. You're using the HTTP_PROXY env var. Sorry :sweat_smile:

 You're just having a problem with https proxies. I saw something similar with python `urllib` and I fixed it by using `urlib2` (or `requests`). Now it's more a bug for `request`.

Here is a trace for history:
        $ export https_proxy="https://myproxy:3128"
        $ npm install jasmine
        npm http GET https://registry.npmjs.org/jasmine
        npm http GET https://registry.npmjs.org/jasmine
        npm http GET https://registry.npmjs.org/jasmine
        npm ERR! network tunneling socket could not be established, cause=139996955195264:error:140770FC:SSL routines:SSL23_GET_SERVER_HELLO:unknown protocol:s23_clnt.c:795:
        npm ERR! network 
        npm ERR! network This is most likely not a problem with npm itself
        npm ERR! network and is related to network connectivity.
        npm ERR! network In most cases you are behind a proxy or have bad network settings.
        npm ERR! network 
        npm ERR! network If you are behind a proxy, please make sure that the
        npm ERR! network 'proxy' config is set properly.  See: 'npm help config'

        npm ERR! System Linux 3.13.0-49-generic
        npm ERR! command "/usr/bin/nodejs" "/usr/bin/npm" "install" "jasmine"
        npm ERR! cwd /home/bob/prog/__/ntest
        npm ERR! node -v v0.10.25
        npm ERR! npm -v 1.3.10
[...]

PR #8012, opened 2015-04-20
Use HTTP_PROXY and HTTPS_PROXY env vars if available

mdamien:
- rename `http_proxy` to `proxy` to keep npm vocabulary
- don't overwrite the npm_config_(https)_proxy if set

Fix https://github.com/npm/npm/issues/8011

Also, I had some idea to improve the PR : (I decided to do a minimal commit first)
- add a test
- https_proxy inherit from http_proxy by default
- handle it in a separate function (get_http_proxy)

### DIFF
--- a/lib/config/core.js
+++ b/lib/config/core.js
@@ -349,7 +349,8 @@ Conf.prototype.addEnv = function (env) {
   env = env || process.env
   var conf = {}
   Object.keys(env)
-    .filter(function (k) { return k.match(/^npm_config_/i) })
+    .filter(function (k) { return k.match(/^npm_config_/i)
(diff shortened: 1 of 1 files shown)


othiym23:
See @smikes's comment on #8011. This is a breaking change as-is, particularly because `HTTP_PROXY`, `HTTPS_PROXY`, and `NO_PROXY` are all handled by `request` – `npm-registry-client` will set the proxy variables if they're present in the config passed from npm itself, but the omission of environment variable handling for them is entirely intentional. Also, this is probably the most fragile code path in npm, so I'm not going to accept changes at this scale without a lengthy rationale and extensive tests. Thanks for putting this together, though!

Issue #3032, opened 2013-01-03
`npm unpublish` does not respect publishConfig setting for registry


jden:
Running `npm publish` in a package's directory will get registry settings by reading the publishConfig section of package.json. However, running `npm unpublish` does not look in `package.json` and instead uses the default global registry.

Would you accept a patch to make `npm unpublish`'s registry configuration registration behave like `npm publish`?

dweidenfeld:
Is there some update on this?

jden:
I no longer have the multiregistry use case, nor the time to put together this PR, although I am generally in favor of the npm client being multi-registry friendly and being able to control this via package.json.

watilde:
I just tried to create a patch at https://github.com/npm/npm/pull/8013.

othiym23:
@watilde's patch was a good one, and so is landed as 18ce0ecd2d94ad3af01e997f1396515892dd363c. I still think that for most per-project configuration, `.npmrc` is a better place to put this configuration, but now `npm publish` and `npm unpublish` use the exact same logic to set the configuration from `package.json`'s `publishConfig` stanza.

PR #8013, opened 2015-04-20
unpublish: check the publishConfig field as same behavior as npm publish

watilde:
I updated unpublish to check the publishConfig field as same behavior as npm publish. Also, this's for fixing https://github.com/npm/npm/issues/3032. If this's all right, I'll add a test.

Thanks!

### DIFF
--- a/lib/unpublish.js
+++ b/lib/unpublish.js
@@ -7,6 +7,8 @@ var log = require("npmlog")
   , path = require("path")
   , mapToRegistry = require("./utils/map-to-registry.js")
   , npa = require("npm-package-arg")
+  , Conf = require("./config/core.js").Conf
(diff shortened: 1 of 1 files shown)


othiym23:
Looks good! Rebased and landed as 18ce0ecd2d94ad3af01e997f1396515892dd363c, and I went ahead and did a little refactoring and added a test as 9c91040d7bcbf138cf98fe8800d1c5d7aed7098f, just so I could put it in this week's release. Thanks a bunch, @watildeさん!

watilde:
Wow thanks for your help. I actually learned a lot from your commit(e.g. split common feature getting publish config to util and write a test for npm unpublish). Thanks always again, @othiym23 san!

PR #8014, opened 2015-04-20
Version prefix should be a config param rather than a magic string

nivthefox:
My workplace does not use the "v" prefix for our versioning, which means I cannot use npm version.  While this is ultimately a mere inconvenience, it'd be nice if npm was not opinionated on this subject.

### DIFF
--- a/doc/misc/npm-config.md
+++ b/doc/misc/npm-config.md
@@ -804,6 +804,17 @@ it will install the specified tag.
 Also the tag that is added to the package@version specified by the `npm
 tag` command, if no explicit tag is given.
 
+### tag-version-prefix
+
+* Default: `"v"`
+* Type: string
+
+If set, alters the prefix used when tagging a new version when performing a
+version increment using  `npm-version`.
+
+Only use this property if it is absolutely necessary.  Use of this property
+can break unrelated features, and should be avoided if possible.
+
 ### tmp
 
 * Default: TMPDIR environment variable, or "/tmp"
--- a/lib/config/defaults.js
+++ b/lib/config/defaults.js
@@ -196,6 +196,7 @@ Object.defineProperty(exports, "defaults", {get: function () {
     , spin: true
     , "strict-ssl": true
     , tag : "latest"
+    , "tag-version-prefix" : "v"
     , tmp : temp
     , unicode : true
     , "unsafe-perm" : process.platform === "win32"
@@ -308,6 +309,7 @@ exports.types =
   , userconfig : path
   , umask: Umask
   , version : Boolean
+  , "tag-version-prefix" : String
   , versions : Boolean
   , viewer: String
   , _exit : Boolean
--- a/lib/version.js
+++ b/lib/version.js
@@ -155,7 +155,7 @@ function commit (version, hasShrinkwrap, cb) {
       git.chainableExec([ "add", "package.json" ], options),
       hasShrinkwrap && git.chainableExec([ "add", "npm-shrinkwrap.json" ] , options),
       git.chainableExec([ "commit", "-m", message ], options),
-      git.chainableExec([ "tag", "v" + version, flag, message ], options)
+      git.chainableExec([ "tag", npm.config.get("tag-version-prefix") + version, flag, message ], options)
     ],
     cb
   )
--- a/test/tap/tag-version-prefix.js
+++ b/test/tap/tag-version-prefix.js
@@ -1,0 +1,88 @@
+var common = require('../common-tap.js')
+var fs = require('fs')
+var path = require('path')
+
+var mkdirp = require('mkdirp')
+var osenv = require('osenv')
+var rimraf = require('rimraf')
+var test = require('tap').test
+
+var npm = require('../../lib/npm.js')
+
+var pkg = path.resolve(__dirname, 'version-message-config')
+var cache = path.resolve(pkg, 'cache')
+var npmrc = path.resolve(pkg, '.npmrc')
+var packagePath = path.resolve(pkg, 'package.json')
+
+var json = { name: 'blah', version: '0.1.2' }
+
+var configContents = 'sign-git-tag=false\nmessage=":bookmark: %s"\n'
+
+test('npm version <semver> with message config', function (t) {
+    setup()
+
+    npm.load({ prefix: pkg, userconfig: npmrc }, function () {
+        var git = require('../../lib/utils/git.js')
+
+        common.makeGitRepo({ path: pkg }, function (er) {
+            t.ifErr(er, 'git bootstrap ran without error')
+
+            common.npm([
+                'config',
+                'set',
+                'tag-version-prefix',
+                'q'
+            ], { cwd: pkg, env: { PATH: process.env.PATH } },
+            function (err, code, stdout, stderr) {
+                t.ifError(err, 'npm config ran without issue')
+                t.notOk(code, 'exited with a non-error code')
+                t.notOk(stderr, 'no error output')
+
+                common.npm(
+                    [
+                        'version',
+                        'patch',
+                        '--loglevel', 'silent'
+                        // package config is picked up from env
+                    ],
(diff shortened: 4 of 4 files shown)


othiym23:
This is a call for @isaacs, as I have no opinion either way, and he might know something about the rationale for the `v` that I don't.

kkragenbrink:
Either way it shouldn't really matter, since the implementation I've suggested preserves the "v" by default.  It just can be turned off with a configuration setting, now. :)

othiym23:
There might be a good reason to make it non-configurable, which is why I'm asking @isaacs to weigh in. Also, adding new configuration does need to clear a usefulness bar, because npm has a confusingly large amount of it already.

othiym23:
Also, this is going to need some tests and documentation before it will pass CI, much less be in a mergeable state.

kkragenbrink:
I'll work on adding those, today.  What possible reason could there be to make it non-configurable, though?  Regardless of whether there is a confusingly large amount of configuration, I have an active use for this change immediately, and I doubt I'm the only one whose business practices require our versions to lack the preceding "v".   Obviously if you're going to check it into npmjs.org, you need the "v", but it shouldn't be required for all packages ever; especially not private repositories.

kkragenbrink:
Tests and documentation added. Should pass ci, now.

doc/misc/npm-config.md
```diff
@@ -879,6 +879,14 @@ If true, output the npm version and exit successfully.
 
 Only relevant when specified explicitly on the command line.
 
+### version-prefix
+
+* Default: `"v"`
+* Type: string
+
+If set, alters the prefix used when tagging a new version when performing a
+version increment using  `npm-version`.
```
othiym23: This needs a note saying that you should _only_ use this property if you _absolutely need to_, due to the hazards of it breaking something completely unrelated elsewhere.

doc/misc/npm-config.md
```diff
@@ -879,6 +879,14 @@ If true, output the npm version and exit successfully.
 
 Only relevant when specified explicitly on the command line.
 
+### version-prefix
```
othiym23: This is ambiguous, and should be either `tag-prefix` or (preferably) `tag-version-prefix`.

test/tap/version-config-prefix.js
```diff
@@ -0,0 +1,96 @@
...
+                        '--loglevel', 'silent'
+                        // package config is picked up from env
+                    ],
+                    { cwd: pkg, env: { PATH: process.env.PATH } },
+                    function (err, code, stdout, stderr) {
+                        t.ifError(err, 'npm version ran without issue')
+                        t.notOk(code, 'exited with a non-error code')
+                        t.notOk(stderr, 'no error output')
+
+                        git.whichAndExec(
```
othiym23: This is extraneous for this particular test, and should be removed.

othiym23:
> What possible reason could there be to make it non-configurable, though?

For starters, a previous version of [semver.org standard](http://semver.org/spec/v1.0.0.html):

> ## Tagging Specification (SemVerTag)
> 
> This sub-specification SHOULD be used if you use a version control system (Git, Mercurial, SVN, etc) to store your code. Using this system allows automated tools to inspect your package and determine SemVer compliance and released versions.
> 
> 1. When tagging releases in a version control system, the tag for a version MUST be "vX.Y.Z" e.g. "v3.1.0".
> 
> 2. The first revision that introduces SemVer compliance SHOULD be tagged "semver". This allows pre-existing projects to assume compliance at any arbitrary point and for automated tools to discover this fact.

This a section was removed in v2.0.0 (probably because it's overly prescriptive), but this is the sort of thing I was looking for – semver is a somewhat odd mixture of specification and conventions, and it's often hard to tell what's going to fall afoul of it. @isaacs doesn't have strong opposition to this (I asked him about it IRL), and since you need it, we can land it, with some reservations:

* changing the prefix can break interoperability with other tools that expect npm packages to be using the standard format (which is an issue for people who want to depend on packages hosted in GitHub repos – the registry uses the version declared in `package.json`, and that format must be a simple semver triplet)
* this is likely to be confusing for most people who browse by package version tags, as up until now `vX.Y.Z` has been the only way tags are named
* other stuff like this is likely to break, as at this point it's nearly impossible to make changes like this without breaking _something_

As such, I've requested that the documentation make clear that is a feature that should only be used when absolutely required, and preferably only for packages that are not going to be shared out into the wider npm ecosystem.

> What possible reason could there be to make it non-configurable, though?

Every bit of conditional logic you add to a product requires support and maintenance. It's nearly impossible for those of us working on the npm CLI to remember all of the stuff it can do, and as such we give people incorrect answers when supporting them, and sometimes inadvertently break features when landing "small" changes like this because it's so difficult to get adequate test coverage over such a complicated CLI tool. (See also: this issue tracker.) Thanks for your patience, and I hope I'm not being too authoritarian, but experience has taught me that it's exactly seemingly simple changes like this that end up causing major, frequently unexpected, complications later on.

kkragenbrink:
I will make the changes you've requested later today.  Also, I really appreciate you taking the time to explain why you are hesitant to add this feature.   You are not coming across as authoritarian; only as a concerned admin. In a similar vein, I hope my questions aren't coming across as too challenging of your authority!  I am only trying to understand your perspective so I can try and find the best way to approach this patch.

Whatever the case, I will adjust my pull request shortly to accommodate the changes you have requested.  Do you prefer concise pull requests (e.g. should I use rebase to keep it to the 3 commits I already have), or do you prefer a fully preserved history (e.g. should I just add an additional commit for these new changes)?

kkragenbrink:
You know, before I commit this, would it be preferable to have this as a property of the package.json instead of the npm configuration? Then it could be configured per package, and potentially available to npmjs.org and other resources, rather than hidden away.  I'm not sure HOW to put it in the package.json, but I can certainly read the code and figure it out if that would be preferred.

test/tap/version-config-prefix.js
kkragenbrink: Is it not important to test whether or not the tag was properly created? If I remove this section then all I am testing is that the configuration setting was created; it won't test that it gets USED properly.

othiym23:
> Do you prefer concise pull requests (e.g. should I use rebase to keep it to the 3 commits I already have), or do you prefer a fully preserved history (e.g. should I just add an additional commit for these new changes)?

I generally rebase, reword commit messages, and squash before merge anyway, so it's completely up to you.

> would it be preferable to have this as a property of the package.json instead of the npm configuration?

Config in `package.json` should be restricted for stuff that is going to be specific to a given registry – everything else belongs in a per-project `.npmrc` (or, you know, a global `.npmrc`, if that's how you roll). There's probably too much that's configurable via `package.json` as it stands. Please keep this in `.npmrc` (but remember that each project can have its own `.npmrc` that will be obeyed by all the relevant npm commands).

test/tap/version-config-prefix.js
othiym23: Testing for the presence of the tag is important, but scanning the log is redundant in this case. I presume you copied the test for commit messages, where it's important to verify that the message formatting made it through.

kkragenbrink:
And thus I learn something new in the process of learning something new. Thanks for the tip about per-project `.npmrc` files. I'll make sure to use them with this change.

kkragenbrink:
Changes submitted per your request. Should pass CI.

kkragenbrink:
I am not entirely sure why this build failed on Travis.  It doesn't look like it failed due to my changes.

othiym23:
The tests are unfortunately somewhat flappy on Node 0.8, probably due to something in `tap` that we're working on tracking down. Since the tests are fine on 0.10 / 0.12 / io.js, don't worry about it.

sindresorhus:
:+1: I also don't use the `v` prefix, which means I can't currently use `npm version`.

othiym23:
Merged, minor tweaks made to the docs, and landed as 21148620fa03a582f4ec436bb16bd472664f2737. Thanks for your patience, and thanks for this work!

PR #8018, opened 2015-04-20
candidate fix for 8008

smikes:
fallback- check for desired version in devDeps if not in deps
attempt to reproduce pre-2.7.5 behavior when readInstalled
used to modify cached version of package.json

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -696,10 +696,15 @@ function installMany (what, where, context, cb) {
     // there's no harm in using that.)
     if (context.explicit) wrap = null
 
+    var getVersion = function (name) {
+        return d[name] || (data.devDependencies && data.devDependencies[name]) || "*"
+    }
+
+
     // what is a list of things.
     // resolve each one.
     asyncMap( what
-            , targetResolver(where, context, d)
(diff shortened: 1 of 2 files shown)


smikes:
Added unit test, rebased and squashed, passes `standard`.

othiym23:
Tests that are consistently failing with this PR applied:
```
test/tap/install-scoped-already-installed.js
test/tap/install-scoped-with-peer-dependency.js
test/tap/peer-deps-invalid.js
test/tap/peer-deps-toplevel.js
test/tap/peer-deps-without-package-json.js
```

I think it probably needs a guard to ensure that there are `devDependencies` available (or a blank object) before `lib/install.js:810`,

smikes:
ah, good. I ran into this from another direction.  Will send you a new one

On Wed, Apr 22, 2015 at 3:15 PM, Forrest L Norvell <notifications@github.com

othiym23:
Landed the fix in 9ad210042242e51d52b2a8b633d8e59248f5faa4 and simplified it a little in f2fe82887a8ee8764dff5dab70e9e744849fd264. Thanks, Sam!

Issue #8020, opened 2015-04-21
read-package-tree assumes the path it's passed exists


othiym23:
This is the proximate cause for `node test/run.js` exploding. I'm still puzzling out what the root cause is, and how to fix this, but here's a snapshot of what I've figured out so far:

* `test/run.js` [sets `global` to true](https://github.com/npm/npm/blob/master/test/run.js#L46) via the environment
* the `globalDir` accessor returns a value of `prefix` + `/lib` as the [location used by install](https://github.com/npm/npm/blob/master/test/run.js#L46). At this point, only `prefix` and `prefix` + `/etc` exist.
* `readLocalPackageData()` [calls out to `readPackageTree()`] with `globalDir` as the value for `where`, which doesn't yet exist, and `readPackageTree()` blows up

I suspect, but am not yet entirely sure, that instead of using `iferr` there, we should be testing for `ENOENT` and calling some kind of generic utility function that creates the base directory if it doesn't yet exist. Either that, or there should be a previous step that ensures that the baseline conditions for a successful install are in place. It's rare that we're going to be using npm to install to an entirely new (global) prefix, but it's definitely something that should be supported.

iarna:
So these are the errors I get:
```
🕠  rebecca@Caldina:~/code/npm$ [multi-stage *] node test/run.js
# testing in /var/folders/56/cm7r6x9n3h78yb6k6md53k940000gn/T/npm-test-49741
# global prefix = /var/folders/56/cm7r6x9n3h78yb6k6md53k940000gn/T/npm-test-49741/root

+node "/Users/rebecca/code/npm/bin/npm-cli.js" install "/Users/rebecca/code/npm"
[...]

PR #8028, opened 2015-04-21
Creating `lib` directory before installing

thefourtheye:
The problem was identified to be the missing `lib` directory when the installation of npm begins. So, we manually create the directory before proceeding with the installation.

This has been introduced as a separate function because, [once we figure out all the baseline conditions](https://github.com/npm/npm/issues/8020#issuecomment-94860080), we can satisfy all the conditions in this function itself.

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -193,19 +193,8 @@ function Installer (where, dryrun, args) {
 }
 Installer.prototype = {}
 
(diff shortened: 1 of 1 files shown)


othiym23:
This looks good as far as it goes, but I'm running into errors further along the bootstrap install that I'm going to keep banging on, and see if there's a more general way to solve this problem for all of the potentially missing prefixes.

PR #8067, opened 2015-04-24
run-script: completion only prints local scripts

### LINKED ISSUE #7973: npm completion script for `npm run` incorrectly lists dependencies
`npm run` prints only scripts listed in package.json
`npm run <tab>` prints scripts from package.json *and* all dependecies/devDependencies

At first I assumed the addition completion options were being fulfilled from the local node_modules binaries. However, attempting to execute `npm run jshint` for example failed with "missing script: jshint". (As it should.) Then I noticed that not just local binaries were listed, but all the package deps entirely.

`npm run <tab>` should only be listing scripts from package.json

(using npm v2.7.6)

mantoni:
Remove the `installed-shallow` logic so that only scripts of the pwd package.json are considered for completion. This might fix #7973.

I'm not sure if I'm overlooking a use case here. It's also quite a lot of work to bootstrap test cases for completion since there's nothing there yet. This make me feel uncomfortable deleting a whole code branch, but currently the run-script completions produce a lot of garbage on my machine.

### DIFF
--- a/lib/run-script.js
+++ b/lib/run-script.js
@@ -13,7 +13,6 @@ runScript.completion = function (opts, cb) {
 
   // see if there's already a package specified.
   var argv = opts.conf.argv.remain
-    , installedShallow = require("./utils/completion/installed-shallow.js")
 
   if (argv.length >= 4) return cb()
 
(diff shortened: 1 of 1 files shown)


othiym23:
The absence of tests for completion definitely makes it harder to make these changes with confidence, but that code has remained unchanged since 2011, for the most part, and much has changed with `npm run-script` and its expected behavior since then. Since this makes `npm run` completion far more useful for the common case, I'm fine with the amount of code deleted. The npm user base is generally pretty quick to let us know when we've removed something they depend upon, and it'll be easier to add this back in to the now much simpler code.

We really do need to write more comprehensive completion tests (or _any_ per-command completion tests, really). But that can probably be handled out of band.

Rebased and landed as 5abfc9c9017da714e47a3aece750836b4f9af6a9. Thanks for putting this together!

PR #8069, opened 2015-04-25
remove google group email

robertkowalski:


othiym23:
Landed, with a very minor change to the commit message (indicating what was affected by the change), as 873e46757e1986761b15353f94580a071adcb383. Thanks for this, Robert! This was a good catch!

package.json
```diff
@@ -19,7 +19,6 @@
     "url": "https://github.com/npm/npm"
   },
   "bugs": {
```
rickzwerld0470: fix bugs in gmail app

Issue #8094, opened 2015-04-27
Wrap os.networkInterfaces() in try/catch in getLocalAddresses()


mhart:
Related to #8093 – if the function is to be kept, can we just wrap the `os.networkInterfaces()` call in `getLocalAddresses` in a try/catch statement for systems that don't support it or don't have permissions for it? Considering the function never actually returned anything I can't imagine this would affect anything.

mhart:
Ah, sorry, code in question is here: https://github.com/npm/npm/blob/21148620fa03a582f4ec436bb16bd472664f2737/lib/config/defaults.js#L318-L329

mhart:
Looks like someone else has needed to do the same thing (not sure what for though): https://api.merproject.org/package/view_file/home:sfietkonstantin:sailfish:experiments/nodejs/npm-network-interfaces.patch?expand=1

othiym23:
> for systems that don't support it

Such as?

> or don't have permissions for it

Likewise – what needs elevated permissions to run `os.networkInterfaces()`?

mhart:
Well – I already linked to one example – that person needed to patch it to run on Sailfish OS.

AWS Lambda is a sandboxed Node.js environment where the user doesn't have permissions to make this call – this is where I ran into it.

Here's another example where EAFNOSUPPORT is thrown with QEMU: https://github.com/joyent/node/issues/8616

If the call's not actually needed (unless there's some rare proxy environment?) – then it should be fine to continue if it fails. Or not?

mhart:
(You get an EPERM on AWS Lambda FWIW)

mhart:
(and once wrapped in a try/catch, npm works fine on AWS Lambda – at least in the testing I've done)

othiym23:
I would like to have a test around this, but it will probably involve some epic Docker jiggery-pokery, so I've just contented myself with verifying that the existing tests pass. Fixed in a4b62387b41848818973eeed056fd5c6570274f3. Thanks for bringing this to our attention!

mhart:
Great! Thanks for getting that in :+1:

Issue #8093, opened 2015-04-27
getLocalAddresses has no return statement (ie, always returns undefined) – is it even needed?


mhart:
See here: https://github.com/npm/npm/blob/21148620fa03a582f4ec436bb16bd472664f2737/lib/config/defaults.js#L318-L329

I discovered this when on a system that throws EPERM when calling `os.networkInterfaces()` – but it raises the question: can this function just be removed? It would certainly make life easier for systems that have issues with network interface access (Googling revealed a few)

othiym23:
See #8096 for the very simple fix of this. We'd have to take a further look into the history of npm/npmconf, which is where this functionality originated, to understand why `getLocalAddresses` existed in the first place, but my understanding is that it is indeed necessary when dealing with some complex network infrastructures (i.e. large enterprise intranets, particularly those running dual-stack IPv4 and IPv6).

PR #8096, opened 2015-04-27
config: fix broken getLocalAddresses

othiym23:
see also #8093 / #8094 for further discussion about the advisabilaty of this patch. I do believe that it's necessary in certain dual-stack IPv4 / IPv6 environments and a few other intrenetty situations.

/cc @mhart


othiym23:
(The tests all pass for me on my OS X 10.10.3 laptop running io.js 1.8.1, FWIW.)

PR #8097, opened 2015-04-27
Fix #7202 by cloning separate committishes in the same repo separately

### LINKED ISSUE #7202: Git-based npm modules not installing correct versions
I've found that since commit 1a90ec2f2cfbefc8becc6ef0c480e5edacc8a4cb, when npm dependencies are specified by using their git URLs, then child modules don't install the correct versions.

I've created a repo here that demonstrates this:https://github.com/BryanDonovan/npm-git-test

If you clone that repo and run `npm install` with a commit prior to 1a90ec2f2cfbefc8becc6ef0c480e5edacc8a4cb, (e.g. npm@2.0.0 works) the child dependencies are installed correctly.  E.g.,:

```
 $ npm list
npm-git-test@1.0.0 /Users/bdonovan/code/node/npm-git-test
├─┬ dummy-npm-bar@4.0.0 (git+ssh://git@github.com:BryanDonovan/dummy-npm-bar.git#575b86626745cda65c05da620ae1c64e8a34dee7)
│ ├── dummy-npm-buzz@3.0.0 (git+ssh://git@github.com:BryanDonovan/dummy-npm-buzz.git#fc6c1ebffccd5ef8755c833072c41be51448f2d6)
│ └── dummy-npm-foo@3.0.0 (git+ssh://git@github.com:BryanDonovan/dummy-npm-foo.git#8524863b7db0ab64eedf7a38c8af670d04865839)
├── dummy-npm-buzz@3.0.0 (git+ssh://git@github.com:BryanDonovan/dummy-npm-buzz.git#fc6c1ebffccd5ef8755c833072c41be51448f2d6)
└─┬ dummy-npm-foo@4.0.0 (git+ssh://git@github.com:BryanDonovan/dummy-npm-foo.git#295739950f3366d9ce668082748e80674670d605)
  └── dummy-npm-buzz@2.0.0 (git+ssh://git@github.com:BryanDonovan/dummy-npm-buzz.git#21673508ffed3584acb800421be7c4a337f4fe78)
```

But if we run npm install with a newer version of npm, the `dummy-npm-bar` dependency has the wrong version of `dummy-npm-buzz` installed (it should be 3.0.0, but it's 2.0.0 instead):

```
$ npm list
npm-git-test@1.0.0 /Users/bdonovan/code/node/npm-git-test
├─┬ dummy-npm-bar@4.0.0 (git+ssh://git@github.com:BryanDonovan/dummy-npm-bar.git#575b86626745cda65c05da620ae1c64e8a34dee7)
│ ├── dummy-npm-buzz@2.0.0 (git+ssh://git@github.com:BryanDonovan/dummy-npm-buzz.git#21673508ffed3584acb800421be7c4a337f4fe78)
│ └── dummy-npm-foo@3.0.0 (git+ssh://git@github.com:BryanDonovan/dummy-npm-foo.git#8524863b7db0ab64eedf7a38c8af670d04865839)
├── dummy-npm-buzz@3.0.0 (git+ssh://git@github.com:BryanDonovan/dummy-npm-buzz.git#fc6c1ebffccd5ef8755c833072c41be51448f2d6)
└─┬ dummy-npm-foo@4.0.0 (git+ssh://git@github.com:BryanDonovan/dummy-npm-foo.git#295739950f3366d9ce668082748e80674670d605)
  └── dummy-npm-buzz@2.0.0 (git+ssh://git@github.com:BryanDonovan/dummy-npm-buzz.git#21673508ffed3584acb800421be7c4a337f4fe78)
```

That is, `dummy-npm-bar@4.0.0` lists `dummy-npm-buzz@3.0.0` as a dependency, but `dummy-npm-buzz@2.0.0` is installed.  

The `dummy-npm-foo` npm lists `dummy-npm-buzz@2.0.0` as a dependency, so somehow this is getting mixed up with the `dummy-npm-bar` dependency of `dummy-npm-buzz@3.0.0`.

iarna:
Testing this is… difficult, like all race conditions. 😕


othiym23:
The easiest way to test this would be to write a test case that uses a local cache and then inspect the remotes in the cache to make sure there are a number that match the number of branches being used from the repository. `test/tap/add-remote-git.js` can be used as a template here. I agree that it's a giant pain in the butt, but it seems very difficult to mock the relevant code while still having confidence that the underlying issue has been addressed.

Also, I'm glad that my gut guess about what was going on here was correct, because the change you've made is exactly what I thought the bug was. Yet another case of a test that's way more complicated than the change it's testing! Sorry! 🙇

iarna:
Yeah, there are multiple approaches I can take that would fail with the old code if the race fails. That's not so bad. 

The thing I find concerning... the _hard_ part is ensuring the race fails.

Now, we use CI so even an intermittently failing test will get caught eventually, if we regressed and reintroduced the race, so having a test that can't guarantee a failure w/ bad code isn't _so_ bad. But it still puts any test below the quality bar I'd like to have.

iarna:
When the network is removed from the equation, this seems to happen very consistently. So I'm feeling a lot better about this test now. 😄

test/tap/git-races.js
```diff
@@ -0,0 +1,171 @@
...
+}
+
+test('setup', function (t) {
+  setup(function (er) {
+    t.ifError(er, 'setup ran OK')
+    t.end()
+  })
+})
+
+test('off-to-the-races', function (t) {
```
othiym23: Could you include the bug number and / or a brief description of what the test is testing here? I fear this is going to be a bit enigmatic in a few months / weeks / days.

iarna: Yeah, that's a good idea: I've added it w/ https://github.com/npm/npm/commit/cf703cbcaf61f6aee549c0bc819d4fe86824fddd

othiym23:
Aside from the point about the test description (I don't mean to be a killjoy, but it saves a _lot_ of time down the road if test descriptions and assertion descriptions are descriptive), this looks great.

I'm sorry the testing code turned out to be such a bear. At some point in the future it would be great to take the git stuff out of here and replace the common.git stuff with it, so more tests don't have to use `npm.load()`. Do you want to merge, or shall I?

PR #8105, opened 2015-04-28
Add missing etc/ to $PREFIX-path in npmrc.md

anttti:
Added missing etc/ to the $PREFIX line at the top of the file.


anttti:
(No idea why the build failed, does not look like it has anything to do with the markdown file I edited.)

othiym23:
Good catch, and thanks for the fix! Tweaked the commit message, rebased, and landed as 1ff636e2db3852a53e38c866fed7eafdacd307fc. Thank you!

PR #8140, opened 2015-04-30
Update npm.js - style fix

jub3i:
match up colon style to rest of property list


othiym23:
Sorry, we're not taking style changes to npm right now, because as soon as we have all the tests passing in the `multi-stage` branch (which will also become `npm@3` shortly thereafter), we're going to be switching the entire code base over to follow the style rules set by [`standard`](http://npm.im/standard).

Thanks for taking the time to put this together, though!

jub3i:
hey no problem, TIL about `standard`, looks nice

PR #8144, opened 2015-05-01
docs: add package name restrictions

### LINKED ISSUE #8137: package name restrictions aren't documented
This is what the documentation currently has to say about the `name` field in `package.json`:

> ## name
> 
> The *most* important things in your package.json are the name and version fields.
> Those are actually required, and your package won't install without
> them.  The name and version together form an identifier that is assumed
> to be completely unique.  Changes to the package should come along with
> changes to the version.
> 
> The name is what your thing is called.  Some tips:
> 
> * Don't put "js" or "node" in the name.  It's assumed that it's js, since you're
>   writing a package.json file, and you can specify the engine using the "engines"
>   field.  (See below.)
> * The name ends up being part of a URL, an argument on the command line, and a
>   folder name. Any name with non-url-safe characters will be rejected.
>   Also, it can't start with a dot or an underscore.
> * The name will probably be passed as an argument to require(), so it should
>   be something short, but also reasonably descriptive.
> * You may want to check the npm registry to see if there's something by that name
>   already, before you get too attached to it. <https://www.npmjs.com/>
> 
> A name can be optionally prefixed by a scope, e.g. `@myorg/mypackage`. See
> `npm-scope(7)` for more detail.

Missing from this description:
* length restrictions (currently a max of 214 characters, including the scope, for hysterical raisins)
* the fact that new packages must not have uppercase characters in the name
* the fact that published packages probably shouldn't shadow built-in packages (`browserify` shims complicate this a little, but we should at least document that fact)

 We should square this up soonish, along with fixing #4665, #7260, and #7269.

Also, right now the registry doesn't really return helpful error messages for these failures, but that's a known issue and being tracked elsewhere.

linclark:
Fixes #8137

### DIFF
--- a/doc/files/package.json.md
+++ b/doc/files/package.json.md
@@ -17,14 +17,23 @@ them.  The name and version together form an identifier that is assumed
 to be completely unique.  Changes to the package should come along with
 changes to the version.
 
-The name is what your thing is called.  Some tips:
+The name is what your thing is called.
 
+Some rules:
+
+* The name must be shorter than 214 characters. This includes the scope for
+  scoped packages.
+* The name can't start with a dot or an underscore.
+* New packages must not have uppercase letters in the name.
+* The name ends up being part of a URL, an argument on the command line, and a
+  folder name. Therefore, the name can't contain any non-URL-safe characters.
+
+Some tips:
+
+* Don't use the same name as a core Node module.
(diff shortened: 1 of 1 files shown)


othiym23:
Thanks! I tweaked this a little bit, because a couple of the existing "tips" were actually "rules" wearing a funny hat, and were also smooshed into one rule. This one snuck in just under the wire! :racehorse:

Issue #4473, opened 2014-01-13
Strongly encourage setting "license" and "licenses" in package.json


garthk:
Specifying [license] in `package.json` makes it a lot easier for users to find and comply with [GNU Affero] licenses. 

Also giving `licenses` (an array of objects with `type` and `url` keys) helps  users whose legal departments request both the license name *and the text*. See this example from [async]:

      "licenses": [{
        "type": "MIT",
        "url": "https://github.com/caolan/async/raw/master/LICENSE"
      }]

If npm \[had] `WARN`-ed about missing `license` and `licenses` like it did missing `description`, `repository`, and `README`, my current task would be a lot easier. 

Solid `license` and `licenses` adoption might also help Node adoption by large companies. If I find any problematic licenses during this mostly manual search — especially buried amongst the dependencies of framework packages like `express`, `connect`, or `restify` — it'll cause the kind of chaos that \[unfairly] entrenches hostility against new platforms. 

Though `licenses` isn't specified in the [npm package.json spec][license], it *is* supported by the NPM site, as you'll see if you check [async's NPM page][async]. So, I guess we've also got a doc-bug. 

[GNU Affero]: http://www.gnu.org/licenses/agpl-3.0.html
[async]: https://npmjs.org/package/async
[asyncpj]: https://github.com/caolan/async/blob/master/package.json
[license]: https://npmjs.org/doc/json.html#license

kemitchell:
I am working on https://github.com/kemitchell/spdx and https://github.com/kemitchell/spdx-correct. Issues, patches, and comments are welcome.

`spdx` aims to support the new "license expression syntax", including`WITH` and `AND` expressions. `spdx-correct` could helpfully suggest corrections when module authors _nearly_ specify valid SPDX identifiers, like 'Apache 2' or just 'GPL'. That way they can make an unambiguous choices in the moment.

For some idea of how "dirty" `.license` data is at the moment, see this dat project: https://github.com/kemitchell/licenses-on-npm

If you'd rather not wait for dat to slurp in all the metadata, here's a gist of fairly recent results: https://gist.github.com/kemitchell/ad3bdc6ca8888599502e

othiym23:
@kemitchell I lack permissions to view either https://github.com/kemitchell/spdx or https://github.com/kemitchell/spdx-correct – are those the correct names / URLS?

kemitchell:
My mistake, @othiym23. The correct URLs are:
- https://github.com/kemitchell/spdx.js
[...]

Issue #6241, opened 2014-09-19
inform users who publish with empty license of legal implications


seldo:
A polite message saying "you have no license field. Without one, it's unclear who can legally use your code and how. Are you sure you want to publish without a license?"

Alternative suggestions welcome.

kemitchell:
Does npm have publicly posted terms of service for repository hosting, in addition to the Artistic License for the command-line client itself? I can't seem to find any. The closest is a ["code of conduct"](https://www.npmjs.com/policies/conduct).

[GitHub's terms of service](https://help.github.com/articles/github-terms-of-service/) have:

> F.1. [...] However, by setting your pages to be viewed publicly, you agree to allow others to view your Content. By setting your repositories to be viewed publicly, you agree to allow others to view and fork your repositories.

seldo:
We have a Privacy Policy https://www.npmjs.com/policies/privacy but it is fairly generic and focused on the website. We are in the process of clarifying + expanding our policy documents to better reflect private module hosting, and I'll add this to the list of things to make clearer.

othiym23:
This is partially addressed by the resolution of #8179 – multiple licenses can be included by using an SPDX license expression, `npm init` now prompts for an SPDX-compliant license expression, and `npm install` warns when `"license"` is missing or isn't parseable as an SPDX license expression. See [the release notes for `npm@2.10.0`](https://github.com/npm/npm/releases/tag/v2.10.0) for details.

There are some aspects of this discussion – like enforcement of the presence of a license on the registry side, including clear language warning about the legal implications on validation, and improving the terms of service for hosting – that aren't yet addressed by npm, so I'm leaving this issue open for now. Once we've got something in place that @seldo thinks fulfills the spirit of his request in the original post, we can split the remaining issues out into separate threads and close this issue.

kemitchell:
@npm should consult its legal counsel about all three remaining aspects---enforcing licensing, warnings, and terms of service. Unfortunately, I can't weigh in on those topics via GitHub, for various "totally lame" professional reasons.

seldo:
The warnings as implemented in 2.10.0 are sufficient for what I had in mind. You can close this ticket as far as I'm concerned, @othiym23.

@kemitchell: we are definitely not going to stride into the minefield of attempting to enforce licensing, nor @sindresorhus are we going to require a license if you ignore our nudge that it's a good idea to have one; that seems like added friction and we are against that :-)

[...]

PR #8179, opened 2015-05-05
Clarify license metadata guidelines

kemitchell:
This PR changes the documentation visible with `npm help 7 package.json` to make a few clarifications about license metadata in `package.json`:
1. `"license"` should be a string SPDX license expression. Thus `"GNU Public License"` is not good metadata,  while `"GPL-3.0"` is fine. So is `"(MIT OR GPL-2.0)"` for a "multi-licensed" project.
2. Old-style `{type:..., url:...}` objects and `"licenses"` arrays are deprecated.
3. If you're non-standard enough that SPDX expressions don't do it, leave `"license"` out, and make sure you include a `LICENSE` file in the package.

There are a couple of open issues on point, both of which would benefit by an official word: #6241 and #4473.

I have also opened PRs to make SPDX expression validation part of `npm init` and metadata normalization: npm/init-package-json#42 and npm/normalize-package-data#61.

There are several ways to to specify machine-readable license metadata that would seem "right". Past affirmatively requiring a valid SPDX identifier, it's a bikeshed, since so few projects are multi-licensed. Some important ones are, but they are few.

I only bring this up early because so many of the most-used npm packages are older, and haven't had their metadata updated since the `licenses: [{...}]` days. Normalizing those packages would help spread the word, and give us a shot at making machine-readable licensing the norm on npm.

### DIFF
--- a/doc/files/package.json.md
+++ b/doc/files/package.json.md
@@ -120,23 +120,30 @@ idea to include a LICENSE file at the top level of the package.
 Some old packages used license objects or a "licenses" property containing an
 array of license objects:
 
+    // Not valid metadata
     { "license" :
-      { "type" : "MIT"
-      , "url" : "http://opensource.org/licenses/MIT"
+      { "type" : "ISC"
+      , "url" : "http://opensource.org/licenses/ISC"
       }
-	}
+    }
 
+    // Not valid metadata
     { "licenses" :
       [
-        { "type" : "MIT"
-        , "url" : "http://opensource.org/licenses/MIT"
+        { "type": "MIT"
+        , "url": "http://www.opensource.org/licenses/mit-license.php"
+        }
+      , { "type": "Apache-2.0"
+        , "url": "http://opensource.org/licenses/apache2.0.php"
         }
       ]
     }
 
-Those styles are now deprecated. Instead, use an SPDX expression, like this:
+Those styles are now deprecated. Instead, use SPDX expressions, like this:
 
-    { "license": "MIT" }
+    { "license": "ISC" }
+
+    { "license": "(MIT OR Apache-2.0)" }
 
 ## people fields: author, contributors
 
--- a/doc/files/package.json.md
+++ b/doc/files/package.json.md
@@ -134,7 +134,9 @@ array of license objects:
       ]
     }
 
-Those styles are now deprecated.
+Those styles are now deprecated. Instead, use an SPDX expression, like this:
+
+    { "license": "MIT" }
 
 ## people fields: author, contributors
 
--- a/doc/files/package.json.md
+++ b/doc/files/package.json.md
@@ -98,9 +98,8 @@ If a url is provided, it will be used by the `npm bugs` command.
 You should specify a license for your package so that people know how they are
 permitted to use it, and any restrictions you're placing on it.
 
-The simplest way, assuming you're using a common license such as BSD-3-Clause
-or MIT, is to just specify the standard SPDX ID of the license you're using,
-like this:
+If you're using a common license such as BSD-2-Clause or MIT, add a
+current SPDX license identifier for the license you're using, like this:
 
     { "license" : "BSD-3-Clause" }
 
@@ -108,8 +107,34 @@ You can check [the full list of SPDX license IDs](https://spdx.org/licenses/).
 Ideally you should pick one that is
 [OSI](http://opensource.org/licenses/alphabetical) approved.
 
-It's also a good idea to include a LICENSE file at the top level in
-your package.
+If your package is licensed under multiple common licenses, use an SPDX license
+expression syntax version 2.0 string, like this:
+
+    { "license" : "(ISC OR GPL-3.0)" }
+
+If you are using a license that hasn't been assigned an SPDX identifier, or if
+you are using an uncommon or custom license, do not include a "license" string
+in package.json. In those cases especially, but also more generally, it's a good
+idea to include a LICENSE file at the top level of the package.
+
+Some old packages used license objects or a "licenses" property containing an
+array of license objects:
+
+    { "license" :
+      { "type" : "MIT"
+      , "url" : "http://opensource.org/licenses/MIT"
+      }
+	}
+
+    { "licenses" :
+      [
+        { "type" : "MIT"
+        , "url" : "http://opensource.org/licenses/MIT"
+        }
+      ]
+    }
+
+Those styles are now deprecated.
 
 ## people fields: author, contributors
 
--- a/doc/files/package.json.md
+++ b/doc/files/package.json.md
@@ -107,8 +107,8 @@ You can check [the full list of SPDX license IDs](https://spdx.org/licenses/).
 Ideally you should pick one that is
 [OSI](http://opensource.org/licenses/alphabetical) approved.
 
-If your package is licensed under multiple common licenses, use an SPDX license
-expression syntax version 2.0 string, like this:
+If your package is licensed under multiple common licenses, use an [SPDX license
+expression syntax version 2.0 string](http://npmjs.com/package/spdx), like this:
 
     { "license" : "(ISC OR GPL-3.0)" }
 
--- a/doc/files/package.json.md
+++ b/doc/files/package.json.md
@@ -113,9 +113,11 @@ expression syntax version 2.0 string](http://npmjs.com/package/spdx), like this:
     { "license" : "(ISC OR GPL-3.0)" }
 
 If you are using a license that hasn't been assigned an SPDX identifier, or if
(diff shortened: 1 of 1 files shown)


kemitchell:
I've just put together [a gist showing nearly a quarter of the 1,000 most-depended-upon npm packages use the old-style "licenses" array of objects](https://gist.github.com/kemitchell/303b5911b6799846a752).

kemitchell:
Only one of the packages listed in the gist, `thrift`, lists a license URL that isn't on github.com.

kemitchell:
The vast majority of `url` properties are to LICENSE files in-repo on GitHub. A great many others are to opensource.org or apache.org. I count about 10 oddballs pointing to Wikipedia, gnu.org, or project- or person-specific domains.

kemitchell:
Of the 1,000 most-depended-upon npm modules, only the following have `licenses` arrays with more than one element:

- `amdefine` ("BSD", MIT)
- `fs.extra` (MIT, Apache-2.0)
- `json-schema` (AFL-2.0, BSD)
- `node.extend` (MIT, "GPL")
- `promised-io` (AFL-2.1, "BSD")
- `rc` (BSD, MIT, Apache-2.0)
- `requirejs` (BSD, MIT)
- `walk` (MIT, Apache-2.0)

sindresorhus:
:+1: This is so much better. @kemitchell thanks for taking the lead on cleaning up this mess :)

kemitchell:
@sindresorhus: Thanks for your work on [spdx-license-list](https://github.com/sindresorhus/spdx-license-list)! Thanks to @shinnn, it's is now an indirect dependency of [spdx.js](https://github.com/kemitchell/spdx.js).

I'd love to make this metadata guidance part of npm@3.x if at all possible. Given the growth rate of npm packages, I get the feeling it's early or never for compliance as the norm.

kemitchell:
I force-pushed a small change that makes clear that license identifiers should come from the most recent SPDX license identifier list (http://spdx.org/licenses), while the syntax for license expressions is version 2.0 of the spec.

doc/files/package.json.md
```diff
@@ -108,8 +107,34 @@ You can check [the full list of SPDX license IDs](https://spdx.org/licenses/).
...
+
+    { "licenses" :
+      [
+        { "type" : "MIT"
+        , "url" : "http://opensource.org/licenses/MIT"
+        }
+      ]
+    }
+
+Those styles are now deprecated.
```
othiym23: Can you add a little to this deprecation, including a terse explanation of how to rewrite the above into SPDX syntax?

doc/files/package.json.md
```diff
@@ -98,9 +98,8 @@ If a url is provided, it will be used by the `npm bugs` command.
 You should specify a license for your package so that people know how they are
 permitted to use it, and any restrictions you're placing on it.
 
-The simplest way, assuming you're using a common license such as BSD-3-Clause
-or MIT, is to just specify the standard SPDX ID of the license you're using,
-like this:
+If you're using a common license such as BSD-2-Clause or MIT, add a
+current SPDX license identifier for the license you're using, like this:
```
othiym23: Can you make this a link to whichever piece of documentation of SPDX you think is most approachable for developers who are unfamiliar with it? I had never heard of it before you got in touch with me, and I have an enterprise background, so I imagine many other people will be confused as well.

doc/files/package.json.md
```diff
@@ -108,8 +107,34 @@ You can check [the full list of SPDX license IDs](https://spdx.org/licenses/).
...
-your package.
+If your package is licensed under multiple common licenses, use an SPDX license
+expression syntax version 2.0 string, like this:
+
+    { "license" : "(ISC OR GPL-3.0)" }
+
+If you are using a license that hasn't been assigned an SPDX identifier, or if
+you are using an uncommon or custom license, do not include a "license" string
+in package.json. In those cases especially, but also more generally, it's a good
+idea to include a LICENSE file at the top level of the package.
```
othiym23: Will they still get a warning in these cases, or are your changes to `normalize-package-data` aware of the existence of a LICENSE file? i.e. how can we make sure that users don't get a warning when they're doing what these docs recommend?

doc/files/package.json.md
othiym23: It might be useful to make the second example an actual array with two licenses so you can show another example of using a conjunction in an SPDX expression.

doc/files/package.json.md
kemitchell: Great point. It didn't think about making sure warning scenarios are coextensive with guidelines noncompliance.

I take it `normalize-package-data` should remain a pure function of the `package.json` object, so hitting the file system or the API to `stat` a `LICENSE` is out.

That leaves us at least three ways to go at the problem:

1. SPDX does provide for a way to reference licenses that don't have assigned identifiers, `LicenseRef-X`, but they are designed as references to other RDF objects within a larger package definition, of which license IDs and expressions are just a part. npm could define a special value, like `LicenseRef-LICENSE` that means "check the LICENSE file" in the ecosystem.

2. Acknowledge that failure to indicate license terms is a warning-worthy offense. But ... private modules.

3. Don't issue warnings when `license` is missing. Accept a far greater number of false negatives (should have `license`, doesn't see a warning) for fewer false negatives (shouldn't have `license` for lack of a way to specify it as an SPDX expression, but warn anyway).

othiym23: I think it may make sense to do some special filtering of the warnings reported by `init-package-json` in npm itself to do something commonsense, like looking for "see LICENSE file" (or some other, bikesheddable phrase) in that field and then verifying that `./LICENSE` does in fact exist. I've been meaning to do some work on how those warnings are handled for a while now, because I'm not sure it's useful to print the warnings for dependencies on install.

I don't think those changes need to be part of landing this change, though. We can get to them later. It just means that npm's install warnings are going to be noisier than usual for a while.

kemitchell: I am down for getting this done right the first time. `LicenseRef-LICENSE` is Magic Words, but it works.

Alas, even detecting whether a `LICENSE` file exists is messy. Sometimes `LICENSE`, other times `LICENSE.md`, `LICENSE.markdown`, `LICENSE.BSD`, &c. That's ignoring case and licenses pasted at the end of README, which has the same problem.

kemitchell: Even if `npm` checked for `LICENSE`, the guidelines recommend that all packages have a `LICENSE` file. It won't be clear when a `LICENSE` exists whether `package.json` should have `license` or not, since `LICENSE` could contain a standard or non-standard license.

Though [GitHub](https://developer.github.com/v3/licenses/#get-a-repositorys-license) and others are trying it, detection of standard licenses in plaintext is non-trivial. I may go there someday, but I doubt `npm` really wants to, or should.

othiym23: I also don't think that getting fancy and trying to parse things out of LICENSE is a good idea. I think we should follow the lead of `fstream-npm` and use a [similar regexp or glob pattern](https://github.com/npm/fstream-npm/blob/master/fstream-npm.js#L98-L99) to just verify that if `"license"` in `package.json` says `LicenseRef-LICENSE`, that glob pattern matches a file. It's heuristic, but all of this behavior must be documented anyway, so as long as those docs exist, I see only upside to making the rules very simple and deterministic.

kemitchell:
@othiym23, I have pushed a commit explaining the use of `"LicenseRef-LICENSE"`.

kemitchell:
@othiym23, would it help if I sent a first-stab PR to implement the `if (LicenseRef-LICENSE) { glob(...) }` logic?

othiym23:
I went ahead and rebased, squashed, and landed what you had so far as 8669f7d88c472ccdd60e140106ac43cca636a648 and b01ba1affc00ea8d6d5f31ca16e65ebd1fecc21c. I also landed eb18245f55fb4cd62a36867744bcd1b7be0a33e2, so npm no longer warns on missing READMEs or invalid license stanzas on transitive dependencies (it logs them at `info` level instead, so you can see the full list by using either `-d` or `--loglevel=info` on the command line). This should make this feature both useful and non-annoying while everybody converts to using SPDX license stanzas. 

If you want to take a shot at implementing the logic behind `LicenseRef-LICENSE`, I wouldn't object, but this is definitely enough functionality to get things started. Thank you for putting this together, being so responsive, and making `spdx.js` in the first place!

kemitchell:
@othiym23 thanks for landing the PR! This probably closes #4473 and #6241, too.

othiym23:
#4473 is definitely closed, but there's a bunch of extra discussion of other features and documents in #6241 that I don't think we've quite finished. Regardless, thanks for the reminder!

PR #8201, opened 2015-05-07
fixes broken http auth install

derEremit:
http auth install for dependencies like:
"package": "http://user:password@server/npmrepo/package-0.1.2.tgz" 
 is broken since 2.1.17 or 6b7c5eca6b65e1247d0e51f6400cf

I'm not sure if maptoregistry should be used at all for packages of type remote, but if so, the auth data should be pulled from the uri and not from the registry

### DIFF
--- a/lib/utils/map-to-registry.js
+++ b/lib/utils/map-to-registry.js
@@ -46,7 +46,10 @@ function mapToRegistry(name, config, cb) {
   log.silly("mapToRegistry", "registry", registry)
 
   var auth = config.getCredentialsByURI(registry)
-
+  //get auth data from uri for remote type
+  if (data.type === "remote") {
+    auth = config.getCredentialsByURI(data.spec)
+  }
   // normalize registry URL so resolution doesn't drop a piece of registry URL
   var normalized = registry.slice(-1) !== "/" ? registry+"/" : registry
   var uri = url.resolve(normalized, name)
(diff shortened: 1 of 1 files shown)


othiym23:
Since I broke support for `http://user:password@server/npmrepo/package-0.1.2.tgz`-style  tarball URLs by accident (sorry!), I'm not averse to fixing it again. However, this patch currently breaks the fix for #7011, which is the bug that 6b7c5eca6b65e1247d0e51f6400cf2637ac880ce was intended to address. In addition, it has no tests of its own (and causes a couple other tests to break), and tests are a requirement for any code changes altering how npm interacts with registries or remote resources.

It's probably going to be a little tricky to support both private registries using bearer tokens and basic auth embedded in the URL itself, but #7011 needs to be be satisfied by whatever solution you eventually come up with.

Thanks for getting this started, and let me know if you need any further help!

derEremit:
Hi, i'm not really an experienced node dev, unfortunately. I'm just an ops guy, who realized i could not install our private libs when switching to io.js 2.0.
And then fiddled around to get it working for me, so sorry, i don't think i'm the right person to do it right.

I just think there could be other people out there who use http auth and will not know what happens when they switch to a higher npm version.

othiym23:
OK, I'm going to ensure that this problem gets solved within the next couple releases, but will probably be using a somewhat different approach to making sure that URLs are normalized properly and consistently. Thanks for bringing this problem to my attention!

PR #8207, opened 2015-05-07
handle error from readPackage

smikes:


othiym23:
This causes some really weird breakage in the tests by itself, and probably needs a bunch more hacking to be mergeable.

smikes:
I'll look at it early next week. (You'd think it would be a totally innocuous change, no?)

othiym23:
Or not! I ran the tests and everything worked, so I was probably thinking about something else. Rebased and landed as 0e2c091a539b61fdc60423b6bbaaf30c24e4b1b8. THANX XAM

PR #8212, opened 2015-05-08
Add documentation for the --registry option

mischkl:
This is a really useful feature and there is zero documentation for it at the moment!

### DIFF
--- a/doc/cli/npm-publish.md
+++ b/doc/cli/npm-publish.md
@@ -4,8 +4,8 @@ npm-publish(1) -- Publish a package
 
 ## SYNOPSIS
 
-    npm publish <tarball> [--tag <tag>] [--access <public|restricted>]
-    npm publish <folder> [--tag <tag>] [--access <public|restricted>]
+    npm publish <tarball> [--tag <tag>] [--access <public|restricted>] [--registry <registry-url>]
+    npm publish <folder> [--tag <tag>] [--access <public|restricted>] [--registry <registry-url>]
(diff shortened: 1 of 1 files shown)


othiym23:
It's true that this isn't documented in `npm-publish.md`, but that's by design. There's plenty of documentation on using `registry` elsewhere in `config`, and the documentation is getting pretty unwieldy. It's actually comparatively rare that people want / need to publish to a registry other than the one they've configured to be their default, and with the advent of scoped packages, it's increasingly unlikely that you'd need to one-off publish to a different registry. We try to keep the documentation for command-line flags for the various npm subcommands as focused on the core user experience as possible, and put lesser-used options into `doc/misc/npm-config.md` and similar pages.

Thanks for taking the time to put this together (really!), but we're not going to take this patch.

mischkl:
Hi, I know this isn't relevant for NPM Enterprise customers, but when using NPM with other NPM registries based on the concept of combining multiple registries into one "virtual registry", for instance with the Artifactory, it is necessary to use the publish command like this *always*. That means it is a very real and often-needed use case.

TBH I don't understand what the point of CLI documentation is if not  to document *every* option that is available. If I go to the trouble of looking up NPM CLI command documentation on your website, I expect it to be complete. And I hardly think one additional sentence here is making anything *unwieldy".

Request for a second opinion, please.

doublerebel:
:+1: for documenting publishing to different registries

PR #8213, opened 2015-05-08
Clarified how to use the --registry option

mischkl:
The ability to use the --registry option for individual publish commands somehow seemed buried, I've put it in its own paragraph and added "for that publish command" in order to make things clearer.


othiym23:
A useful tweak! Rebased and landed as f70dea9b4766f6eaa55012c3e8087e9cb04fd4ce. Thanks, @mischkl!

PR #8214, opened 2015-05-08
deps: remove unused foo.js & bar.js from lru-cache

Fishrock123:
These files are not found in https://github.com/isaacs/node-lru-cache and appear to have been erroneously committed in https://github.com/npm/npm/commit/530a2e369128270f3e098f0e9be061533003b0eb.

(see also: https://github.com/iojs/io.js/pull/1658#discussion_r29950685)

### DIFF
--- a/node_modules/lru-cache/bar.js
+++ b/node_modules/lru-cache/bar.js
@@ -1,1 +1,0 @@
-console.log("bar");module.exports = "bar"; require("./foo.js")
--- a/node_modules/lru-cache/foo.js
+++ b/node_modules/lru-cache/foo.js
@@ -1,1 +1,0 @@
-console.log(require("."))


othiym23:
Since it doesn't hurt anything, I rebased this and landed it as 79f8a9664199c638d7fd576a4830a15ee9813af2. I'll reiterate, though, that the way the npm team deals with bundled dependencies is just to run `npm i -S 'package@~version'` and then checking in the results (as long as the tests continue to pass). Trimming or curating those installs is a pointless use of the team's time, and if you have issues with the stuff that gets checked in as a result, the issues are better taken up with the upstream packagers.

Fishrock123:
Right, but these aren't included in that repo, and the tests/etc don't seem to produce them?

othiym23:
They're in the package as it was published to the registry, hence my suggestion to take it up with the packager (which, in this case, is @isaacs). As much as possible, to keep things moving along, we need to be able to treat `node_modules` as a black box, in much the same way that I expect the downstreams (Node.js and io.js) to treat `deps/npm` as a black box. It really isn't a productive use of my time to clean up insignificant cruft like this.

Fishrock123:
> They're in the package as it was published to the registry

Ohhhh, right, I didn't check that. Yes, I agree it does not matter in that case, messy or not.

Pssst, ya'll should consider using the `files:` package.json thing for all internal modules. ;)

PR #8219, opened 2015-05-09
Improve shasum check sanity

nadimkobeissi:
Previously, npm's behavior in case it did not receive a shasum was to:
1. Itself generate a shasum for whatever package is downloaded over HTTP,
2. Cache this arbitrary shasum as a valid, genuine shasum, exactly as if it was sent via TLS from `registry.npmjs.com`.

I would argue that this behavior is not reasonable. Instead, npm should insist on a shasum being delivered and do diligent comparison for any requested packages, without exception.

I am a new contributor to this project, so please excuse me in case I am missing something. I have done some testing, however, and feel this solution is legitimate. I also look forward to investigating in the future the possibility of adding feasible, easy to use code-signing features into npm.

### DIFF
--- a/lib/cache/add-remote-tarball.js
+++ b/lib/cache/add-remote-tarball.js
@@ -86,12 +86,10 @@ function fetchAndShaCheck (u, tmp, shasum, auth, cb) {
 
     tarball.on("finish", function () {
       if (!shasum) {
-        // Well, we weren't given a shasum, so at least sha what we have
-        // in case we want to compare it to something else later
-        return sha.get(tmp, function (er, shasum) {
-          log.silly("fetchAndShaCheck", "shasum", shasum)
-          cb(er, response, shasum)
-        })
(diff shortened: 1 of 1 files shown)


isaacs:
This removes the ability to install a local folder/file or a URL.

I think a better solution is to ensure that all registry requests include a shasum. Generation one locally is also a valid way to protect against mangling the file in subsequent fs operations.

othiym23:
As @isaacs said, this breaks several important existing use cases in npm, and causes several tests to fail besides. The correct solution to this problem would involve splitting apart the code paths that deal with caching packages from the registry from those that deal with locally-sourced packages and self-built tarballs, and would probably come along with a better-defined notion of the chain of custody between publishing and installing. As is, this breaks too much to land it. Thanks for taking the time to put it together, though!

kaepora:
(Will give it another shot when I have time, thanks!)

PR #8257, opened 2015-05-13
cache: fix error messages for private registries

hayes:
When trying to install a package from a registry that is not hosted at the domain root (eg: `http://localhost:1337/registry`) 404s will result in an error that says module registry was not found rather than the actual module name, this is due to registry-client not knowing how to determine the package name from a request, and falling back to grabbing the first segment.  see https://github.com/npm/npm-registry-client/issues/80

### DIFF
--- a/lib/cache/add-named.js
+++ b/lib/cache/add-named.js
@@ -17,6 +17,17 @@ var path = require("path")
 module.exports = addNamed
 
 function getOnceFromRegistry (name, from, next, done) {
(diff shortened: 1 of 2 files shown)


othiym23:
I futzed with it a little, rebased, and landed it as dc7752013ffce13a3d3f13e518a0052c22fc1158, with a test added for scoped packages in 5f247838c7a29040465c64b21a857c775280eb94. Thanks for putting this together, and thanks for including a test!

Issue #8185, opened 2015-05-05
When $NAME not found in registry is invalid, should not suggest "use the name yourself"


smikes:
I was trying to remember the name of `JSONStream` and ran an `npm view` command --

```
Svarog:github smikes$ npm view Jsonparse
npm ERR! Darwin 14.3.0
npm ERR! argv "node" "/Users/smikes/.nvm/v0.10.38/bin/npm" "view" "Jsonparse"
npm ERR! node v0.10.38
npm ERR! npm  v2.9.0
npm ERR! code E404

npm ERR! 404 Not Found: Jsonparse
npm ERR! 404 
npm ERR! 404 'Jsonparse' is not in the npm registry.
npm ERR! 404 You should bug the author to publish it (or use the name yourself!)
npm ERR! 404 
npm ERR! 404 Note that you can also install from a
npm ERR! 404 tarball, folder, http url, or git url.

npm ERR! Please include the following file with any support request:
npm ERR!     /Users/smikes/src/github/npm-debug.log
```

Since "Jsonparse" would not be accepted as a new module name, we should not be saying "You should bug the author to publish it (or use the name yourself!)"

othiym23:
This is pretty edge-casey, and as such is something that the CLI team are not going to have the time to get to anytime soon, but if somebody wanted to patch npm so that it checked that a name matches the naming rules before suggesting that a user publish it, we would accept that patch.

smikes:
This might be a good "low-hanging-fruit" feature for onboarding new contributors.

thefourtheye:
> a name matches the naming rules 

What are the actual naming rules?

thefourtheye:
@smikes @othiym23 Does this look okay?

    diff --git a/lib/utils/error-handler.js b/lib/utils/error-handler.js
    index 1c80ab5..bf10833 100644
    --- a/lib/utils/error-handler.js
    +++ b/lib/utils/error-handler.js
    @@ -216,8 +216,12 @@ function errorHandler (er) {
       case "E404":
         var msg = [er.message]
         if (er.pkgid && er.pkgid !== "-") {
    -      msg.push("", "'"+er.pkgid+"' is not in the npm registry."
    -              ,"You should bug the author to publish it (or use the name yourself!)")
    +      msg.push("", "'" + er.pkgid + "' is not in the npm registry.");
    +
    +      if (er.pkgid.toLowerCase() === er.pkgid) {
    +        msg.push("You should bug the author to publish it (or use the name yourself!)")
[...]

PR #8286, opened 2015-05-15
Fix for 8185

thefourtheye:
This is actually an attempt to fix https://github.com/npm/npm/issues/8185
1. Included `validate-npm-package-name` to have the name validated
2. If the name is valid, we will include the `...bug the author` suggestion, otherwise we list out the problems reported by the `validate-npm-package-name` module.


othiym23:
I did a little tweaking and rebasing and landed this as 8a46136f42e416cbadb533bcf89d73d681ed421d. Thanks, @thefourtheye! This is a nice little improvement!

Issue #7906, opened 2015-04-10
How about to put a hook for version to lifecycle?


watilde:
Hi.

I tried to search similar issue but it was not found. So I  decided to propose this.
`npm version` is really convenient and powerful. In order to update the version correctly, I'd like to do that before through to pass the all test and build code. Actually it has already can do that by using `npm-run-script` currently:
```
$ cat package.json
...
scripts: {
  preversion: "npm test & npm run build",
  version: "npm version"
}
...

$ npm run version
```

If I could do that with lifecycle, I think it become more user friendly. Thanks!

smikes:
I like the idea, but one issue is that the `pre` scripts are run unconditionally; they don't really function as guards.  I think a better approach is to write a single lifecycle script , and use that instead of `npm version`:

```
"scripts": {
   "my-version": "npm test && npm run build && npm version"
}
```

Then `npm run my-version` will do what you desire.

Please let me know if I'm misunderstanding your feature request, though!

martinheidegger:
@smikes That script of your example will need to pass all arguments for `my-version` to `npm version` shouldn't it look something like;

```
"scripts": {
   "my-version": "npm test && npm run build && npm version \"$@\""
}
```

My use-case would be to generate screenshots and other things that should be in the versioning system (shown on github) but not generated after every push.

_Note: I was not aware that the scripts don't function as guards, awkward._

smikes:
Hmm, well I don't have a module to test it with, but you should be able to pass parameters after "--" so if "npm version" is the last thing on the line you can use

npm run-script my-version -- args

martinheidegger:
_(Just referenced a duplicate)_

I still think it would be good to have hooks for version. Just for consistency and pragmatic sake. I wonder if they could be made "guarding" in case of `version` only but I am not fixed on it.

watilde:
@smikes Thanks for your comment! Yeah I've known that it can hook using `npm run-script`, but if `npm version` is in the lifecycle it just become more user friendly. And one more thing, I actually think this feature likely helps many people who would like to bump version after passed the all test.

bebraw:
Any update on this?

@watilde Did you figure out something that works? I would love to have proper hooks for `npm version` to keep things simple but I can manage with something simpler.

watilde:
@bebraw Yeah I think so too. I just tried to create a PR last night, will try waiting for a response.

zanona:
That was a great addition —Thanks guys.

[...]

PR #8287, opened 2015-05-15
version: put a hook for version to lifecycle

watilde:
I thought including `npm version` in lifecycle sctips is good solution to prevent bump version for the package that has not passed test. Any thoughts on this please? See also: https://github.com/npm/npm/issues/7906

### DIFF
--- a/lib/version.js
+++ b/lib/version.js
@@ -11,6 +11,7 @@ var semver = require("semver")
   , npm = require("./npm.js")
   , git = require("./utils/git.js")
   , assert = require("assert")
+  , lifecycle = require("./utils/lifecycle.js")
 
 version.usage = "npm version [<newversion> | major | minor | patch | prerelease | preminor | premajor ]\n"
               + "\n(run in package dir)\n"
@@ -26,11 +27,6 @@ function version (args, silent, cb_) {
 
   var packagePath = path.join(npm.localPrefix, "package.json")
   fs.readFile(packagePath, function (er, data) {
-    function cb (er) {
-      if (!er && !silent) console.log("v" + data.version)
-      cb_(er)
-    }
-
     if (data) data = data.toString()
     try {
       data = JSON.parse(data)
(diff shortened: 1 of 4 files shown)


lib/version.js
```diff
@@ -25,20 +27,7 @@ function version (args, silent, cb_) {
...
-
-    if (data) data = data.toString()
-    try {
-      data = JSON.parse(data)
-    }
-    catch (er) {
-      data = null
-    }
-
+  readJson(packagePath, function (er, data) {
```
iarna: We don't want to use readJson here for a host of reasons*.  This should remain `fs.readFile + JSON.parse`.

\* Those reasons being:
* readJson can read package metadata from non-`package.json` sources (eg, `index.js` header comments)
* readJson passes the result through `normalize-package-data` which will mutate the json a variety of ways.
* readJson does a bunch of its own normalization
* readJson reads a bunch of standalone files and injects them into the resulting structure– AUTHORS, README*…
* And more… it does an awful lot.

othiym23:
This is a useful addition, @watilde, and something that I'll use myself. In addition to @iarna's feedback, I'd like to see some tweaks to `doc/cli/npm-version.md` with an example or two of how the new hooks work, because I suspect almost nobody will see that this feature exists if it's not in the help for `npm version`. If you take a first pass at it, I can fine-tune the wording.

I'd also like to see `test/tap/ignore-scripts.js` updated to incorporate `preversion`, `version`, and `postversion`, and I'd like to see an integration test that verifies that the the pre and post hooks are working. Thanks (as always) for putting this together!

watilde:
Thanks for your comments @iarna and @othiym23! I'll try to make improvement by your comments and do rebase.

watilde:
Ok I've done to repace `fs.readFile` + `JSON.parse` with `readJson`, to write additional info to docs and to update test case in `test/tap/ignore-scripts.js`. Thanks!
// Sorry for that I missed to close this PR.

iarna:
LGTM!

iarna:
Merged w/ b07f7c7

watilde:
Thanks for your review and fine-update(▰∀◕)ﾉ

PR #8290, opened 2015-05-16
partial workaround for cache collisions on case-folding fs

smikes:
cache names with lowercase-only unchanged;
names with uppercase characters are stored as
name_xxxxxxxx where the `x`'s are the first 8 chars 
of the sha1 of the mixed-case name

add new dependency uniquify-case
use in lib/cache/add-named
use in dependency npm-cache-filename

stalling on integration test (need to build mock with correct sha)

### DIFF
--- a/lib/cache/cached-package-root.js
+++ b/lib/cache/cached-package-root.js
@@ -1,5 +1,6 @@
 var assert = require("assert")
 var resolve = require("path").resolve
+var caseUniquify = require("case-uniquify")
 
 var npm = require("../npm.js")
 
@@ -10,5 +11,7 @@ function getCacheRoot (data) {
   assert(data.name, "package metadata must include name")
   assert(data.version, "package metadata must include version")
 
-  return resolve(npm.cache, data.name, data.version)
+  var name = caseUniquify(data.name)
+
+  return resolve(npm.cache, name, data.version)
 }
--- a/test/tap/install-cache-mixcase.js
+++ b/test/tap/install-cache-mixcase.js
@@ -1,0 +1,197 @@
+var common = require('../common-tap.js')
+var test = require('tap').test
+var npm = require('../../lib/npm.js')
+var mkdirp = require('mkdirp')
+var rimraf = require('rimraf')
+var path = require('path')
+var mr = require('npm-registry-mock')
+var sha = require('sha')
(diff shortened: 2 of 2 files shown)


smikes:
See related PR https://github.com/npm/npm-cache-filename/pull/3

othiym23:
As I said on npm/npm-cache-filename#3, I think this is something better handled as part of the semi-mythical cache rewrite (which will probably be split across `npm@4` and `npm@5` – I realllllly need to update the roadmap with my current thinking, after I get a chance to go over it with @iarna), which is going to end up nuking everyone's caches anyway and pretty comprehensively restructuring how everything's stored inside the cache. The fix for the mixed-case problems we have right now is more in allowing legacy package owners to update their packages than try to deal with these cache collisions, because the inability to do the former is what led to the latter being an issue in `JSONStream`'s case.

Thanks for putting this together!

smikes:
No worries.  I am happy to leave it lying.

However I am confused - I didn't think this would totally trash the cache, since only modules with uppercase chars in the name get the _sha disambiguator.  

My goal was to address the potential (but uncommon) bug on a case-folding filesystem where for two consecutive `npm install` commands, `foo` is cached on install 1, `Foo` is needed for install 2, but `foo` is returned from cache as still fresh.

othiym23:
> However I am confused - I didn't think this would totally trash the cache, since only modules with uppercase chars in the name get the _sha disambiguator.

It won't totally trash it, but it can leave it inconsistent, in that it's hard to say which of `foo`, `Foo`, `fOo` will get created, and what's in it – the `JSONStream` case would have been find if there hadn't been overlapping versions of the two / they were depended on in similar ways.

The right way to do this is to use some kind of unambiguous strategy like uniquify (my personal preferred strategy is to lowercase everything and use a character that's illegal in names (like `,`) to indicate capitalization), but to do that right should happen at a semver-major boundary just so we don't end up with any leftover detritus from old caches. The next version of the cache is going to look _very_ different from what's there now.

smikes:
Ah, I see. If the user loads `foo`, waits, then `Foo`, then updates npm, the cache directory for `foo` contains `Foo` but this is not detectable by client code.  Thanks for pointing that out, I hadn't thought of that failure mode. :+1:

Issue #8314, opened 2015-05-20
npm README.md should suggest more caution wrt programmatic API


smikes:
<https://github.com/npm/npm#using-npm-programmatically> says:

>If you would like to use npm programmatically, you can do that. It's not very well documented, but it is rather simple.

>Most of the time, unless you actually want to do all the things that npm does, you should try using one of npm's dependencies rather than using npm itself, if possible.

>Eventually, npm will be just a thin cli wrapper around the modules that it depends on, but for now, there are some things that you must use npm itself to do.

Contrast with <https://github.com/npm/npm/issues/8216#issuecomment-100397681>

> All of this code is going to change very soon in npm@3, but even then, the CLI's API is meant for use by the CLI only, and we make no guarantees for its fitness for use for any other purpose. We're very slowly working towards a better long-term solution, which is breaking the CLI up into a set of components that fulfill its various roles, and which will have their own semver-versioned APIs. Until then, the right thing to do is to drive npm from the command line if you want to do a set of installs and then uninstalls.

I propose adding an edited version of this paragraph to the README, something like 'the programmatic API does not necessarily follow semver; semver applies to the command-line interface. If you want to use `npm` to reliably perform some task, the safest thing to do is to invoke the `npm` command with appropriate arguments.'

smikes:
Cc @ashleygwilliams

Unfortunately I don't have as much time as I used to but I'm happy to contribute if you want to spread some work around the community

smikes:
Actually this was done by someone else, so closing!

PR #8322, opened 2015-05-21
Addresses #8314 with CLI API stability warnings

claycarpenter:
Addresses #8314 by updating the README's "Using npm programmatically" section with more warnings about relying upon the CLI API.

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,8 +129,11 @@ you have chosen.
 
 ## Using npm Programmatically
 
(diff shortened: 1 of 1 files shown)


othiym23:
Reformatted, rebased, and landed as f62ee05333b141539a8e851c620dd2e82ff06860, with some further rearrangement, editing, and clarification in 68405a33ba8c7aaa46832cb17f43f93389106b97. Thanks! We've needed something like this for a while!

PR #8324, opened 2015-05-21
fix windows open node-gyp.js with editor instead of run it

bangbang93:


othiym23:
Reworded the commit message, rebased, and landed as 9bcf5730bd0316f210dafea898afe9103849cea9. I think this closes a few existing issues, but I hadn't figured out what the cause was before now. Thanks, @bangbang93!

PR #8336, opened 2015-05-22
Bugfix for --no-registry

Ellezavr:


lib/cache/caching-client.js
```diff
@@ -67,6 +67,22 @@ function get (uri, params, cb) {
   var cacheBase = cacheFile(npm.config.get("cache"))(uri)
   var cachePath = path.join(cacheBase, ".cache.json")
 
+  if (parsed.host === "noregistry") (function() {
```
KenanY: Why does this have to be wrapped in an anonymous function?

othiym23:
Thank you for taking the time to put this together, but I'm not going to take this patch. We've been working on having a cleaner separation of concerns between the cache and the rest of the code base, and coupling the thing that maps from packages to registry URLs to the caching logic goes the other way. Also, using sync APIs in the caching layer isn't a good idea for a host of reasons. We do intend to re-add support for offline operation (as a first-class option, not a hackaround like it's been up to now), but that will be when the cache rework happens, sometime after the release of `npm@3`.

PR #8365, opened 2015-05-26
git: Add support for GIT_SSH_COMMAND environment variable

nmalaguti:
Git 2.3 supports GIT_SSH_COMMAND as a way to specify command line
parameters to pass to git. It is usually used as a way to specify a
particular identity.

`GIT_SSH_COMMAND='ssh -i /path/to/identity' npm install`


othiym23:
Thanks! If you wanted to take a stab at updating the docs to demonstrate how that might be useful to people with dependencies under private git hosting, that would be gratefully accepted!

nmalaguti:
@othiym23 I added some docs in #8565

PR #8381, opened 2015-05-27
Point out that .npmrc needs chmod 600

colakong:
This could save people some troubleshooting, if they are generating the content of an .npmrc file without using _npm adduser_ interactively.


othiym23:
Landed in cf109a682f38a059a994da953d5c1b4aaece5e2f with some additional wordsmithing (and a move) in 14aceddaf15f8310b6ec29c2c7dafac7695ca22b (and a typo fix in 2dc8c610df6df11b0ca0dc2a13eb0138efe63e9a thanks to @KenanY). Thanks for the clarification!

Issue #8401, opened 2015-05-30
Have npm --outdated list alphabetically


binarykitchen:
Here an example, the list is not sorted alphabetically:

```
$ npm outdated --depth=0
Package            Current  Wanted  Latest  Location
angular             1.3.15  1.3.15   1.4.0  angular
angular-route       1.3.15  1.3.15   1.4.0  angular-route
browserify          10.2.1  10.2.1  10.2.3  browserify
require-directory    2.1.0   2.1.0   2.1.1  require-directory
cors                 2.6.0   2.6.0   2.7.1  cors
semver               4.3.4   4.3.4   4.3.5  semver
```

I think it should be because whenever you do an `npm install --save`, npm automatically makes an entry in the package.json and sorts the dependencies alphabetically which means, the developer parses that list resting assured it's sorted. Now imagine I scan that list of outdated packages, which is NOT sorted and compare it with the package.json which is sorted: you scroll up and down a couple of times. Not very "developer-user-friendly" (DX).

I know, it's a very small, unimportant, feature request. But it's always good to make the life of a hard-working developer a bit easier heh.

Dashed:
Going to +1 to show some demand for this tiny detail.

watilde:
Seems nice! +1

KenanY:
This is available as of [`npm@2.11.2`](https://github.com/npm/npm/releases/tag/v2.11.2) :confetti_ball:

zkat:
looks like another issue I forgot to close!

watilde:
Thank you!

PR #8463, opened 2015-06-04
outdated: Sorted the package list alphabetically

watilde:
Fixes: https://github.com/npm/npm/issues/8401

### DIFF
--- a/lib/outdated.js
+++ b/lib/outdated.js
@@ -48,6 +48,12 @@ function outdated (args, silent, cb) {
   outdated_(args, dir, {}, 0, function (er, list) {
     if (!list) list = []
     if (er || silent || list.length === 0) return cb(er, list)
+    list.sort(function(a, b) {
+      var aa = a[1].toLowerCase()
+        , bb = b[1].toLowerCase()
+      return aa === bb ? 0
+           : aa < bb ? -1 : 1
+    })
     if (npm.config.get("json")) {
       console.log(makeJSON(list))
     } else if (npm.config.get("parseable")) {


othiym23:
Rebased and landed by @zkat as 31aada4ccc369c0903ff7f233f464955d12c6fe2. @zkat also wrote some tests for this change as f42ba69c82890a9ed872f2800d865f8f205496f9. Thanks once more / as always watildeさん!

image

watilde:
@zkat Thank you so much for some tests you wrote!

@othiym23 Thanks for your kind comment. What a amazing kawaii cats picture :heart_eyes_cat:

lib/outdated.js
```diff
@@ -48,6 +48,12 @@ function outdated (args, silent, cb) {
   outdated_(args, dir, {}, 0, function (er, list) {
     if (!list) list = []
     if (er || silent || list.length === 0) return cb(er, list)
+    list.sort(function(a, b) {
+      var aa = a[1].toLowerCase()
+        , bb = b[1].toLowerCase()
+      return aa === bb ? 0
```
thefourtheye: Why only first characters are compared? This will not sort properly.

othiym23: Good catch – if you wanted to patch this to be

```javascript
var aa = a.toLowerCase()
var bb = b.toLowerCase()
return aa == bb ? 0 : aa < bb ? -1 : 1
```
as well as a test that would catch this, we'd be happy to merge it.

thefourtheye: @othiym23 nvm. I was wrong. `a` and `b` are arrays in the form,

    [ '/home/thefourtheye/git/npm/test/tap/outdated',
      'underscore',
      '1.3.1',
      '1.3.1',
      '1.5.1',
      '1.3.1' ]

So, the code works fine only I guess.

othiym23: I should have waited until I was less sleepy to take a look at this, so sorry for the erroneous feedback. I'm glad all is well!

PR #8469, opened 2015-06-04
version: set working directory for preversion and postversion scripts

alexkwolfe:
Explicitly sets the working directory for preversion and postversion scripts. Before this change, the preversion and postversion scripts cannot be found relative to the root directory.

The previous behavior is due to the way that the lifecycle module [resolves a default working directory](https://github.com/npm/npm/blob/f42ba69c82890a9ed872f2800d865f8f205496f9/lib/utils/lifecycle.js#L37). Specifying the working directory solves the problem.

### DIFF
--- a/lib/version.js
+++ b/lib/version.js
@@ -48,11 +48,13 @@ function version (args, silent, cb_) {
     if (data.version === newVersion) return cb_(new Error("Version not changed"))
     data.version = newVersion
 
+    var where = path.resolve(npm.dir, "..")
+
     chain([
-          [lifecycle, data, "preversion"]
+          [lifecycle, data, "preversion", where]
         , [version_, data, silent]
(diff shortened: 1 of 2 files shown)


othiym23:
I made a few small tweaks to this before landing it:

* `version` should be getting passed the working directory as well as `preversion` and `postversion` (in reality, `version` and `postversion` are basically synonyms)
* although it's rare that it would matter, the correct working directory is `npm.prefix`, thanks to the command-line options `--prefix` / `-C`
* I applied `standard --format` to the test (thank you very much for including a test!)
* I shrunk the subject of the commit message to 50 characters

Rebased, squashed, and landed as 458a91925d8b20c5e672ba71a86745aad654abaf. Thanks so much for the patch, @alexkwolfe!

Issue #5827, opened 2014-07-30
dedupe git deps


jonathanong:
right now, i have 16 copies of esprima harmony, and each of those are 6mb large. dedupe doesn't work, maybe because i have like 5 other versions of esprima. 

so... if deduping git deps actually works, feel free to close this.

iarna:
Closing this as fixed in [npm/npm#multi-stage](https://github.com/npm/npm/tree/multi-stage). At least, in so far as we're able. For git deps you need to have requested exactly the same thing for them to be considered a duplicate.

PR #8472, opened 2015-06-05
Dedupe modules from git/http/file

golbin:
At this time, if there are modules from GitHub repositories or local directories, they won't be de-duplicated. It's required for testing and some of situations.

So, I made this patch and it use the first one if there are many duplicated modules from URIs.

### DIFF
--- a/lib/dedupe.js
+++ b/lib/dedupe.js
@@ -240,6 +240,13 @@ function findVersions (npm, summary, cb) {
     var versions = data.versions
 
     var ranges = data.ranges
+
+    // If there is a module from git/http/file, use the first one.
+    var uriMatch = hasFromUri(ranges);
+    if (uriMatch !== undefined) {
+      return cb(null, [[name, has, loc, data.versions[0], uriMatch, locs]])
+    }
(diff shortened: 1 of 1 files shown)


iarna:
Tentatively I like this, but I'd be a lot more comfortable with some tests for it.

(This class of change _is_ extra painful to write tests for– they'll be much bigger than the patch. Still, all the same…)

golbin:
@iarna I found some comments for this feature which is to reduce git/http modules will be included in multi-stage install milestone. (maybe this? https://github.com/npm/npm/issues/5827#issuecomment-87085081) If it's right and coming soon, I'll close this. OR if not, I'll make some test for this.

iarna:
goblin: If you're happy waiting then yes, the deduper in `multi-stage` does do this. (The deduper generally is less needed in `multi-stage` as it won't create dupes under normal circumstances during install.)

golbin:
@iarna Great! I hope it will be come very soon. And my team is using this very widely. So, if you need some kind of test, plz let me know.

Issue #8398, opened 2015-05-29
Report version information in install failure


mfowlewebs:
As a sample, right now I'm getting

```
gyp WARN install got an error, rolling back install
gyp ERR! configure error 
gyp ERR! stack Error: connect ECONNREFUSED
gyp ERR! stack     at errnoException (net.js:905:11)
gyp ERR! stack     at Object.afterConnect [as oncomplete] (net.js:896:19)
gyp ERR! System Linux 3.19.0-14-generic
gyp ERR! command "node" "/usr/local/lib/node_modules/npm/node_modules/node-gyp/bin/node-gyp.js" "rebuild"
gyp ERR! cwd /opt/code/node_modules/winston-syslog/node_modules/unix-dgram
gyp ERR! node -v v0.10.38
gyp ERR! node-gyp -v v1.0.3
gyp ERR! not ok
```

To begin to debug this, I'll have to
- lookup my winston-syslog's version in my package.json, 
- check it against an npm info winston-syslog to see what actual version I'm getting, 
- manually check out winston-syslog, 
- check out it's package.json for unix-dgram, then 
- compare that against npm info for unix-dgram to again figure out what version is being attempted. 

All this before I can begin to replicate the problem in isolation. If this was a third or fourth level dependency, I'd have to chase down the chain even further!

It would be very handy if failures always reported the version of what was failing.

othiym23:
I agree that more context on native build failures would be more helpful. At the very least, npm should print out the package specifier for a package when `node-gyp` fails (although because of the semi-generic way that `node-gyp` is invoked by the npm package lifecycle, this may require a bit of care to get correct).

I think you should file a similar feature request against `node-gyp` to print out the package and version that it's trying to build. That will ensure that the information is always available, regardless of what npm's doing, and `node-gyp` is the best-placed component to understand what it was trying to do when a build failed.

We have a bunch of minor platform integration tasks to do as part of the next major version of npm (which is mostly intended to roll up a number of deprecations and small breaking changes, not to be a major feature release like `npm@3` was), and this would be a good thing to include as part of that. If you want to take a stab at making the necessary changes, go ahead and start a PR and we can help you figure out what's necessary. Otherwise, the CLI team will get to this as part of that major release. Thanks for your time and patience!

PR #8485, opened 2015-06-07
install: Report version information in install failure

alexanderGugel:
Addresses #8398


iarna:
@alexanderGugel `node-gyp` is not an npm maintained module, it's bundled. Changes to it should be made via its repo: https://github.com/TooTallNate/node-gyp

alexanderGugel:
Thanks!

PR #8501, opened 2015-06-09
link: don't rebuild or run lifecycle scripts for `npm install <pkgname>`

### LINKED ISSUE #7198: npm link X requires write access
If you install (say protobuf) globally as user root (or it is delivered via packaging)

```
npm install -g protobuf
```

Then, in your app you utter:

```
npm link protobuf
```

it fails because it attempts to do a rebuild in the globally installed package.

murgatroid99:
This should fix #7198.

### DIFF
--- a/lib/link.js
+++ b/lib/link.js
@@ -97,14 +97,13 @@ function linkInstall (pkgs, cb) {
 
     function next () {
       chain
-        ( [ [npm.commands, "unbuild", [target]]
-          , [function (cb) {
+        ( [ [function (cb) {
               log.verbose("link", "symlinking %s to %s",  pp, target)
               cb()
             }]
           , [symlink, pp, target]
-          // do run lifecycle scripts - full build here.
-          , rp && [build, [target]]
+          // do not run any scripts
+          , rp && [build, [target], npm.config.get("global"), build._noLC, true]
           , [ resultPrinter, pkg, pp, target, rp ] ]
         , cb )
     }


othiym23:
That "should" is worrisome. ;) This needs a failing test that reproduces the problem, and the code to make it run. Thanks for the time you've put into it so far!

murgatroid99:
OK, I'm trying to write a test to check that lifecycle scripts are called when a `npm link` is initially called, but not when `npm link foo` is called. To do this, I modified a copy of `test/tap/link.js` so that `readJSON` has the extra script lines `preinstall: 'echo once >>preinstall.txt'`, and similar for `install` and `postinstall`. However, when I try to do this, it fails, and the `npm-debug.log` file in the package directory has these log lines:

```
16 verbose stack Error: foo@1.0.0 preinstall: `echo once >>preinstall.txt`
16 verbose stack spawn sh ENOENT
16 verbose stack     at exports._errnoException (util.js:746:11)
16 verbose stack     at Process.ChildProcess._handle.onexit (child_process.js:1053:32)
16 verbose stack     at child_process.js:1144:20
16 verbose stack     at process._tickCallback (node.js:355:11)
```

This does not happen if I instead manually run `npm install` in that directory. Do you know what might be happening here, or how else I might test that the lifecycle scripts are run exactly once?

othiym23:
You can't do that for two reasons:

1. `echo` is a shell built-in, and as such that won't work as part of the child process call.
2. `>>` isn't going to work on Windows.

You might want to write a little sample script that gets written out to a test directory before the `npm install` is run that creates a file or prints a message to stderr as a byproduct of the script being invoked. If you look through the tests, there are a few examples of that.

zkat:
Hi! We're still hoping for a test for this, but will check this again next week so we can get it in a release :)

othiym23:
I added a test as e987496dda36aef3efcbdc428873b430d95daa0f (which turned out to be a little more involved than I'd hoped, but I feel good about the test), and with the original commit rebased, reworded, and landed as 932381dda385ffb60247261c35fa30f5467ef1c8. Thanks for putting this together, @murgatroid99!

PR #8548, opened 2015-06-12
view: Remove extra newline

eush77:
`npm view` always prints extra newline at the end which in certain cases prevents its output from being easily recognized by other CLI tools.

None of these work:

```
curl $(npm v express dist.tarball) |tar xz
open $(npm v express bugs.url)
git log $(npm v express gitHead)
```

I have to manually remove the newline to make it work which is annoying:

```
curl $(npm v express dist.tarball |head -1) |tar xz
open $(npm v express bugs.url |head -1)
git log $(npm v express gitHead |head -1)
```

This patch makes `npm view` output more suitable for automatic consumption by removing extra newline at the end if there's no obvious reason for it.

As a special case it preserves extra newline if output starts with an empty line (for aesthetic reasons):

```
$ npm v express keywords

[ 'express',
  'framework',
  'sinatra',
  'web',
  'rest',
  'restful',
  'router',
  'app',
  'api' ]

```

### DIFF
--- a/lib/view.js
+++ b/lib/view.js
@@ -254,7 +254,11 @@ function printData (data, name, cb) {
     })
   })
 
-  console.log(msg)
+  if (/^\s*\n/.test(msg)) {
+    msg += "\n"
+  }
+
+  process.stdout.write(msg)
   cb(null, data)
 }
 function cleanup (data) {


othiym23:
This is great! I made a few tweaks to the style (mostly so I could add some comments explaining why the output is written this particular way), rebased this against current master, and landed it as 9d8cac94a258db648a2b1069b1c8c6529c79d013. Thanks so much for putting this together!

PR #8565, opened 2015-06-14
Add details about git enviroment variables to install docs.

nmalaguti:


othiym23:
Nice! Rebased, commit message tweaked, and landed as 5018335ce1754a9f771954ecbc1a93acde9b8c0a. Thanks for putting this together! This was definitely a hole in the existing docs.

PR #8566, opened 2015-06-14
Outdated should return error if registry cannot be reached

ryantemple:


othiym23:
Thanks for putting this together! It would have been rad to have a test, because this is something I can see easily regressing in the future, but maybe next time. Rebased, commit message reworded, and landed as 765fd4bfca8ea3e2a4a399765b17eec40a3d893d.

PR #8570, opened 2015-06-15
changelog: fix links

kenany:


othiym23:
Whoops! Good catch! Thanks for catching this, Kenan! We should have caught these in review, but the flaky connection from NodeConf made that a little tougher than usual. Rebased, tweaked the commit message, and landed as bc2ffce55b7b7521bad6dab801bc26820985880b.

PR #8571, opened 2015-06-15
Added npm-dist-tag.md

zachfedor:
Using `npm dist-tag` with a non-unique package name fails without providing adequate [feedback](https://gist.github.com/zachfedor/38b66191647c49d5420b). Added a note to provide info for those failures by encouraging use of `@<user>/<pkg>` naming convention.

### DIFF
--- a/doc/cli/npm-dist-tag.md
+++ b/doc/cli/npm-dist-tag.md
@@ -22,6 +22,8 @@ Add, remove, and enumerate distribution tags on a package:
   Show all of the dist-tags for a package, defaulting to the package in
   the current prefix.
 
+Note: If `<pkg>` is not unique, use `@<user>/<pkg>@<version>`.
+
 A tag can be used when installing packages as a reference to a version instead
 of using a specific version number:
 


zachfedor:
Alright, I'm confused. I added one line to the doc file. How is the build failing?

othiym23:
Thanks for taking the time to put this together!

What is a "non-unique package name", in your understanding? With scoped packages, the scope is part of the name, so the note you've added feels more confusing than clarifying. As such, I'm not going to land this patch.

If you think the documentation about working with scoped versus unscoped packaging should be more explicit about what constitutes the name of the package, then I'm open to reviewing tweaks to the [scope documentation](https://github.com/npm/npm/blob/master/doc/misc/npm-scope.md), but every piece of npm treats scoped package names the same way, including `npm dist-tag`.

othiym23:
> Alright, I'm confused. I added one line to the doc file. How is the build failing?

npm's test suite is flaky when run under Travis. This is a known issue, and I apologize for the confusion, because you're right, a simple tweak to the docs shouldn't cause a build failure.

zachfedor:
Thanks for the clarification @othiym23. I'm new to npm and had never been pointed to the documentation on scope. That clears everything up for me. I just didn't know where to look to clear up my confusion. My failures with dist-tag were just the first time my ignorance caused an issue, so I figured that would be a good place to start. Also, this is the first time I have done a pull request, and even though my change seems unnecessary now, I fully appreciate your thoughtfulness and willingness to help me out.

PR #8598, opened 2015-06-18
utils: use a safer mkdirp with chownr

### LINKED ISSUE #8595: permissions not set to correct user for _locks when npm run as root
`EACCES` errors on `.npm/_locks` are the consequence of someone running npm for the first time with sudo. All of the cache subdirectories should be created using the same logic as `getCacheStat`, which sets the permissions upon creation correctly.

othiym23:
Makes paths more consistently. Fixes #8595.

**r**: @zkat 
**r**: @iarna


othiym23:
In particular, I'm curious as to whether there are adequate comments in `correct-mkdir.js` that all of its various ins and outs can be picked up from context.

zkat:
:+1: nothing bad jumps out at me here. Read through it a few times.

othiym23:
Woot! :sheep: :ship:

PR #8554, opened 2015-06-12
fixed typo on version manual page

ghost:

PR #8607, opened 2015-06-18
fix typo

AlexKVal:


KenanY:
https://github.com/npm/npm/pull/8554?

othiym23:
Good catch! Thanks for putting this together. As @KenanY noticed, this was also fixed in #8554, which is the version we landed as bed9edddfdcc6d22a80feab33b53e4ef9172ec72 (with some further wordsmithing on those docs in 7614c3bb48598b28a55ac14ed3264897e3e5889c). I still appreciate you bringing this to our attention, though!

AlexKVal:
Oh.. I didn't know it were addressed already. Sorry. :cherries:

Issue #8031, opened 2015-04-22
Git regression from npm@2.1.17 onward: npm can't install git+ssh dependencies anymore


madarche:
Hello,

From npm@2.1.17 onward (I've tested all the versions from 2.1.17 to 2.8.4) npm can't install git+ssh dependencies.

Here are examples of used dependencies:

    "dependencies": {
      "secret-factory": "git+ssh://git@git.example.test:secret-factory.git#v1.0.0",
      "sso-connector": "git+ssh://git@git.example.test:sso-connector-js.git#v1.0.0"
    },
    "devDependencies": {
      "sso-mockup": "git+ssh://git@git.example.test:sso-mockup.git#v3.0.0"
    },

The presence or not of an `npm-shrinkwrap.json` has no effect.

Cheers

madarche:
Here is the log for npm@2.8.4:

    npm ERR! git clone --template=/home/sweet/.npm/_git-remotes/_templates --mirror ssh://git@git.example.test/sso-connector-js.git /home/sweet/.npm/_git-remotes/ssh-git-git-example-test-sso-connector-js-git-93cc66e7: Cloning into bare repository '/home/sweet/.npm/_git-remotes/ssh-git-git-example-test-sso-connector-js-git-93cc66e7'...
    npm ERR! git clone --template=/home/sweet/.npm/_git-remotes/_templates --mirror ssh://git@git.example.test/sso-connector-js.git /home/sweet/.npm/_git-remotes/ssh-git-git-example-test-sso-connector-js-git-93cc66e7: fatal: '/sso-connector-js.git' does not appear to be a git repository
    npm ERR! git clone --template=/home/sweet/.npm/_git-remotes/_templates --mirror ssh://git@git.example.test/sso-connector-js.git /home/sweet/.npm/_git-remotes/ssh-git-git-example-test-sso-connector-js-git-93cc66e7: fatal: Could not read from remote repository.
    npm ERR! git clone --template=/home/sweet/.npm/_git-remotes/_templates --mirror ssh://git@git.example.test/sso-connector-js.git /home/sweet/.npm/_git-remotes/ssh-git-git-example-test-sso-connector-js-git-93cc66e7:
    npm ERR! git clone --template=/home/sweet/.npm/_git-remotes/_templates --mirror ssh://git@git.example.test/sso-connector-js.git /home/sweet/.npm/_git-remotes/ssh-git-git-example-test-sso-connector-js-git-93cc66e7: Please make sure you have the correct access rights
[...]

PR #8618, opened 2015-06-19
Document git+ssh true URLs with absolute path

madarche:
that work with all versions of NPM

this is the minimal fix for #8031

### DIFF
--- a/doc/files/package.json.md
+++ b/doc/files/package.json.md
@@ -419,8 +419,9 @@ install time.
 Git urls can be of the form:
 
     git://github.com/user/project.git#commit-ish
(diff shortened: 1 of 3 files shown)


madarche:
In this PR I haven't removed the other forms of git+ssh URLs, I have just put the git+ssh URL form that works *first*.

othiym23:
I think this is rendered redundant by 94ca4a711619ba8e40ce3d20bc42b13cdb7611b7, which should make the old-style relative remote refs (that aren't exactly URLs) work again. Thanks for putting this together, though, and thanks for your patience! Sorry it took us a little while to get around to addressing this!

madarche:
No problem! Thanks again!

PR #8673, opened 2015-06-25
Fix git url parsing for non-hosted repos

### LINKED ISSUE #8031: Git regression from npm@2.1.17 onward: npm can't install git+ssh dependencies anymore

zkat:
Fixes #8031 

This patch fixes normalization issues for `git+ssh` non-URLs (which are just scp-style identifiers with a special `git+ssh://` prefix), and makes sure to leave them as intact as possible. The issue was present for non-hosted git servers, which we can't normalize very aggressively because we have no guarantee that our normalizations will still point to the same resource.

### DIFF
--- a/lib/cache/add-remote-git.js
+++ b/lib/cache/add-remote-git.js
@@ -425,25 +425,19 @@ function getResolved (uri, treeish) {
 
   var parsed = url.parse(uri)
 
-  // non-hosted SSH strings that are not URLs (git@whatever.com:foo.git) are
-  // no bueno
-  // https://github.com/npm/npm/issues/7961
-  if (!parsed.protocol) return
-
-  parsed.hash = treeish
-  if (!/^git[+:]/.test(parsed.protocol)) {
-    parsed.protocol = 'git+' + parsed.protocol
+  // Checks for known protocols:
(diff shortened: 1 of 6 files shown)


othiym23:
Rebased, tweaked, and landed as e7f56e5a97fcf1c52d5c5bee71303b0126129815 and 94ca4a711619ba8e40ce3d20bc42b13cdb7611b7. Thanks for making the FINAL ULTIMATE FIX to Git ever, @zkat!

zkat:
\o/

PR #8692, opened 2015-06-26
version: allow scripts to add files to the commit

### LINKED ISSUE #8620: allow `preversion` and `version` scripts to add files to the commit
The `git version` command is useful if all you need to do is bump the value in package.json. If you try to generate/compile any files as part of your release process, then it complains:

```
npm ERR! Error: Git working directory not clean.
```

I think it should work (something) like this

  1. Check to make sure the git working directory is clean before running any scripts (scripts in future steps may add files to the commit without throwing an error).
  2. Run the `preversion` script. These scripts have access to the old `version` in package.json. Any files you want added to the commit should be explicitly added using `git add`.
  3. Bump `version` in `package.json`. 
  4. Run the `version` script. Scripts here have access to the new `version` in package.json (so they can incorporate it into file headers in generated files for example). Again, scripts should explicitly add files using `git add`.
  5. Commit and tag
  6. Run the `postversion` script.

The most important part is being able to easily incorporate generated files as part of the commit that will be tagged, and easy access to what that version number was / will be.

jamestalmage:
Closes #8620

### DIFF
--- a/doc/cli/npm-version.md
+++ b/doc/cli/npm-version.md
@@ -15,10 +15,11 @@ valid second argument to semver.inc (one of `patch`, `minor`, `major`,
 `prepatch`, `preminor`, `premajor`, `prerelease`). In the second case,
 the existing version will be incremented by 1 in the specified field.
 
-If run in a git repo, it will also create a version commit and tag, and fail if
-the repo is not clean.  This behavior is controlled by `git-tag-version` (see
-below), and can be disabled on the command line by running `npm
---no-git-tag-version version`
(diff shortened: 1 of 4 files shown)


jamestalmage:
@othiym23 
I've cherry picked my changes back to the 2.X branch.

jamestalmage:
It all passed locally.

iarna:
Landed in [`2.x`](https://github.com/npm/npm/tree/2.x) as https://github.com/npm/npm/commit/582f1704988303f00c5bc5960733c48472c12bf3

PR #8718, opened 2015-06-27
Better pipe cleanup for early shrinkwrap reading

### LINKED ISSUE #8695: npm 3: got "abort trap 6" :-(
I installed npm v3 on io.js v2.3.1, fresh-faced and excited, like a puppy greeting somebody that's been gone for 30 seconds, which is like FOREVER. I `cd`'d into https://github.com/es-shims/es7-shim, `rm -rf node_modules`, and ran `npm install`, waiting for the glorious flat tree to appear.

This is the output I received:
```shell
 npm install
loadDevDep:replace → get  ▀ ╢███████████████████████████████████████████████████████████░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░╟
Assertion failed: (ctx->mode_ != NONE && "already finalized"), function Write, file ../src/node_zlib.cc, line 122.
Abort trap: 6
```

Installing 3 or 4 of the deps one at a time allowed `npm install` to work, altho I still had to manually install a few deps after that.

Puppy sad.

iarna:
Fixes #8695


lib/fetch-package-metadata.js
```diff
@@ -304,16 +300,19 @@ function untarStream (tarball, cb) {
...
+    tounpipe.push([stream, untar])
     stream = untar
     addClose()
   }
 
   function addClose () {
     stream.close = function () {
-      tounpipe.forEach(dispose)
+      tounpipe.forEach(function (streams) {
+        unpipe(streams[0],streams[1])
```
zkat: afaict, `unpipe` only takes one argument, and it unpipes _all_ piped streams. See https://www.npmjs.com/package/unpipe (and source: https://github.com/stream-utils/unpipe/blob/master/index.js#L40)

iarna: You are correct! Good catch.

iarna:
Landed as https://github.com/npm/npm/commit/57c3cea9ed98d9d4185afb4fcd45bb5ed30a62e3

PR #8720, opened 2015-06-27
Fix package.json handling to always allow BOM at start

iarna:
(tbd: I need to look up which bug this fixes)
(also some tests would be nice)


iarna:
Closing this as I made this in my personal clone of npm by msitake.

PR #8724, opened 2015-06-27
Fix package.json handling to always allow BOM at start

### LINKED ISSUE #3358: Update NPM to support package.json with UTF BOM
I had a package.json file that was saved in UTF-8 format with a Byte Order Marker (BOM)

using `npm install PACKAGENAME --save' was failing to update said package.json

Suggest updating npm to support files that start with a BOM.

iarna:
Fixes #3358

### DIFF
--- a/lib/utils/parse-json.js
+++ b/lib/utils/parse-json.js
@@ -1,0 +1,24 @@
+'use strict'
+var parseJSON = module.exports = function (content) {
+  return JSON.parse(stripBOM(content))
+}
+
+parseJSON.noExceptions = function (content) {
+  try {
+    return parseJSON(content)
+  } catch (ex) {
+    return
+  }
+}
+
+// from read-package-json
+function stripBOM (content) {
+  content = content.toString()
+  // Remove byte order marker. This catches EF BB BF (the UTF-8 BOM)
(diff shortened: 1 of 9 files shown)


lib/utils/parse-json.js
```diff
@@ -0,0 +1,24 @@
...
+    return
+  }
+}
+
+// from read-package-json
+function stripBOM(content) {
+  // Remove byte order marker. This catches EF BB BF (the UTF-8 BOM)
+  // because the buffer-to-string conversion in `fs.readFileSync()`
+  // translates it to FEFF, the UTF-16 BOM.
+  if (content.charCodeAt(0) === 0xFEFF) {
```
zkat: This may have a bit of an awkward effect where states involving non-strong `content` will now throw the "wrong" error ("undefined is not a function"). Are we okay with that? I can see a couple of places in this PR where this seems to be possible.

iarna:
@zkat: So it turns out that `JSON.parse` will handle all sorts of data types (it actually just calls `toString` anything its passed). I've updated this to do the same, so the error messages should now match.

And more importantly, it won't explode w/ buffers.

iarna:
This landed in master as https://github.com/npm/npm/commit/28064e5070dbe38df5281a4a0e14364f4100aadc

Issue #7040, opened 2015-01-01
cannot install local dependencies with prepublish scripts


timoxley:
### tldr;

prepublish scripts always run for local dependencies on `npm install`, even when using `npm install --production`.

Consider this `package.json`

``` json
{
  "name": "npm-issue-local-dependency-prepublish",
  "version": "0.0.0",
  "dependencies": {
    "local-pkg": "file:local-pkg"
  }
}

```

And this `./local-pkg/package.json`

``` json
{
  "name": "local-pkg",
  "version": "1.0.0",
  "scripts": {
    "prepublish": "tap --help"
  },
  "dependencies": {
    "tap": "^0.4.13"
  }
}
```

Ensuring there is no `tap` installed on the PATH, and all `node_modules` are clean:

```
npm uninstall -g tap
rm -rf node_modules
rm -rf local-pkg/node_modules
```

When running `npm install` from the first package, it will produce the following error:

```
> local-pkg@1.0.0 prepublish /Users/timoxley/Projects/tests/npm-issue-local-dependency-prepublish/local-pkg
> tap --help

sh: tap: command not found

npm ERR! addLocalDirectory Could not pack "/Users/timoxley/Projects/tests/npm-issue-local-dependency-prepublish/local-pkg" to "/Users/timoxley/.npm/local-pkg/1.0.0/package.tgz"
npm ERR! addLocal Could not install /Users/timoxley/Projects/tests/npm-issue-local-dependency-prepublish/local-pkg
npm ERR! Darwin 14.1.0
npm ERR! argv "node" "/usr/local/bin/npm" "install"
npm ERR! node v0.10.35
npm ERR! npm  v2.1.16
npm ERR! code ELIFECYCLE
npm ERR! local-pkg@1.0.0 prepublish: `tap --help`
npm ERR! Exit status 127
npm ERR!
npm ERR! Failed at the local-pkg@1.0.0 prepublish script 'tap --help'.
npm ERR! This is most likely a problem with the local-pkg package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     tap --help
npm ERR! You can get their info via:
npm ERR!     npm owner ls local-pkg
npm ERR! There is likely additional logging output above.

npm ERR! Please include the following file with any support request:
npm ERR!     /Users/timoxley/Projects/tests/npm-issue-local-dependency-prepublish/npm-debug.log
```

**I believe this is incorrect behaviour.** Please correct me if I'm wrong.

[...]

PR #8725, opened 2015-06-27
cache: add-local: do full "npm install" w/ prepublish prior to packing

iarna:
Fixes: #7040 #3055

### DIFF
--- a/lib/cache/add-local.js
+++ b/lib/cache/add-local.js
@@ -13,8 +13,9 @@ var cachedPackageRoot = require('./cached-package-root.js')
 var addLocalTarball = require('./add-local-tarball.js')
 var sha = require('sha')
 var inflight = require('inflight')
-var lifecycle = require('../utils/lifecycle.js')
 var iferr = require('iferr')
+var install = require('../install.js')
+var prune = require('../prune.js')
 
 module.exports = addLocal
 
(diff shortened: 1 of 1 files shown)


iarna:
See also #7627

boneskull:
:+1:

lib/cache/add-local.js
```diff
@@ -92,13 +93,21 @@ function addLocalDirectory (p, pkgData, shasum, cb) {
...
-        var doPrePublish = !pathIsInside(p, npm.tmp)
-        if (doPrePublish) {
-          lifecycle(data, 'prepublish', p, iferr(cb, thenPack))
-        } else {
-          thenPack()
+        var isProduction = npm.config.get('production')
+        var dir = npm.dir
+        var prefix = npm.prefix
+
+        npm.config.set('production', false)
```
iarna: This stuff, in particular, is extremely scary and probably shouldn't be allowed in production code.

iarna: By "this stuff", I mean, trying to get/set things from the global config object, when who knows what else is running and could be effected by this.  To land this the `production` flag needs to be isolated to the installer object, so that we can instantiate an object here with the config we need and run that, rather than calling `install`.

iarna:
Adopting this is blocked by [§ 3.7 of the roadmap](https://github.com/npm/npm/wiki/Roadmap-area-of-focus:-api)

iarna:
Closing this, we'll repull when  [§ 3.7 of the roadmap](https://github.com/npm/npm/wiki/Roadmap-area-of-focus:-api) is resolved.

PR #8735, opened 2015-06-27
Graceful fs 4

isaacs:
This change makes it so that `npm i ember-cli yeoman generator-angular` with a `ulimit -n 64` eventually completes mostly successfully.  (There are a few errors unrelated to EMFILE or npm, because of dependencies on a tarball url that is no longer present apparently.)

The next step is to go through each of the dependencies and either upgrade their graceful-fs dep, or have them just use `require('fs')` directly.


iarna:
Landed in [`2.x`](https://github.com/npm/npm/tree/2.x) as https://github.com/npm/npm/commit/c9fb0fd9083fa5c3fa27bc530dfdfa5efe896050

iarna:
Landed in master as 3fcf35d

PR #8738, opened 2015-06-27
install: Run the "install" lifecycle in the toplevel module

### LINKED ISSUE #8736: node-gyp not executed on npm@3
I've downloaded node-canvas using git clone (I'm modifying it), but after upgrading to npm@3 and deleting the build, node_modules and deps folders, when doing ```npm install``` the dependencies get installed but node-gyp don't get executed and due to this, node-canvas module don't gets compiled. npm don't show any error nor get any error status code.

iarna:
Fixes #8736


zkat:
I don't know if there's any other place where we might need to trigger these, but this looks benign enough! :sheep:

iarna:
You're thinking of here:

https://github.com/npm/npm/blob/master/lib/install.js#L433-L447

But this is the only place it was missing.

iarna:
This has landed as https://github.com/npm/npm/commit/3cb6ad2a86a8d228fb10c389b3219855efd564b3

PR #8760, opened 2015-06-29
Emphasize the correct way to write the script

josh-egan:


othiym23:
Rebased and landed on master as c689d55 by @iarna and landed on `2.x` as 22fdc1d. Thanks for the tweak!

PR #8786, opened 2015-07-01
install: check-permissions: disable access checks on win32

iarna:
The way this was written was sometimes (always in large installs) resulting
in errors.

This is a temporary work around to make win32 work better until we have time to fix the problem more permanently.


iarna:
This landed in master as https://github.com/npm/npm/commit/b259bcc8ba4ad990a7574a1f6fcb506e01966f7e

PR #8800, opened 2015-07-02
Fixed ping

zkat:
I didn't notice that: 
- there were no tests for the npm/npm side of the `ping` command
- the command wasn't working, from the command line

Still pending: Need to look around at issues and make sure I get what the expected output of the command is in various log and unreachability situations (but we might be able to merge this anyway because it's useful already).

### DIFF
--- a/doc/api/npm-ping.md
+++ b/doc/api/npm-ping.md
@@ -1,0 +1,14 @@
+npm-ping(3) -- Ping npm registry
+================================
+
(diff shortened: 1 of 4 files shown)


zkat:
...Also I totally made this PR against the wrong branch. ~sigh~

iarna:
LGTM

iarna:
This was merged into both branches variously.

Issue #8476, opened 2015-06-05
Strange "npm view" result


pauldijou:
Today, around 5PM UDT, I played with `npm view` both with CLI and programmatic API and I got some strange results... Running `npm view npm` inside my shell returned (lot of stuff removed):

``` javascript
{ name: 'npm',
  description: 'a package manager for JavaScript',
  'dist-tags': { latest: '2.11.1', next: '2.11.2' },
  versions: 
   [ ...,
     '2.11.0',
     '2.11.1',
     '2.11.2' ],
  maintainers: [...],
  time: 
   { modified: '2015-06-05T00:54:40.257Z',
     created: '2013-07-12T18:32:48.902Z',
     ...,
     '2.11.0': '2015-05-22T06:09:54.035Z',
     '2.11.1': '2015-05-28T21:36:47.566Z',
     '2.11.2': '2015-06-05T00:54:40.257Z' },
  dist: 
   { shasum: '6197c1520e8b9110c34b0b6841175fea551e2ecf',
     tarball: 'http://registry.npmjs.org/npm/-/npm-2.11.1.tgz' } }
```

And running the following code to display the last NPM version:

``` javascript
var npm = require('npm');
npm.load({}, function () {
  npm.commands.view(['npm'], false, function (err, infos) {
    if (err) { console.log(err); }
    else { console.log(infos); }
  });
});
```

Output the same stuff as before assigned to a `2.11.1` key.

``` javascript
{ "2.11.1": {... same stuff as before...}}
```

So, here is the question. Maybe I got it wrong, but why is `2.11.1` considered the last version (both in `npm.commands.view` and in the `dist-tags` object) even if a `2.11.2` has been created (both in `versions` array and `time` object)?

Checking [NPM website](https://www.npmjs.com/package/npm) confirm the `2.11.1` as the last version. But I can actually install `2.11.2` by running `npm install npm@2.11.2`.

Is that linked to the fact that it's a [pre-release on GitHub](https://github.com/npm/npm/releases)? Or because it was published with a tag different than `latest`?

Why is it bothering me? Because I'm trying to figure the best version of a package to install but I got a conflict: if I have `npm: ^2.11.0` inside my `package.json`, checking the best version according to SEMVER inside the `versions` array will return me `2.11.2` but this version is supposed to be greater than the last version `2.11.1` so it's kind of not really valid. And so my head just exploded...

Thanks a lot for any help!

KenanY:
[...]

PR #8818, opened 2015-07-03
outdated: cap wanted with latest if possible

pauldijou:
Currently, if you have a dependency with `dist-tags: {latest: 1.1.0, next: 1.2.0}` (just like NPM) and you require `^1.0.0`, running `npm install` will install version `1.1.0` while `npm upgrade` will install version `1.2.0`. I might be wrong, but I consider that non-consistent and a bug. The reason is that `npm update` use `npm outdated` which only check the best among all possible versions without really using the `latest` version as a maximum.

`wanted` is actually capped only if `latest` is inside the required range of course.

Fixes: #8476

### DIFF
--- a/lib/outdated.js
+++ b/lib/outdated.js
@@ -385,6 +385,10 @@ function shouldUpdate (args, tree, dep, has, req, depth, pkgpath, cb, type) {
       var vers = Object.keys(d.versions)
       var v = semver.maxSatisfying(vers, r, true)
       if (v) {
+        if (semver.gt(v, l.version) && semver.satisfies(l.version, r)) {
+          v = l.version
+        }
+
         return onCacheAdd(null, d.versions[v])
       }
     }
--- a/test/tap/outdated-dist-tags.js
+++ b/test/tap/outdated-dist-tags.js
@@ -1,0 +1,134 @@
+var fs = require('graceful-fs')
+var path = require('path')
+
+var mkdirp = require('mkdirp')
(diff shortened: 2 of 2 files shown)


othiym23:
Thanks for putting this together, but the current behavior is by design. `npm install` and `npm update` are meant to do different things: `npm install` will install the the version that you'd get if you ran `npm install dependency` (respecting semver in your `package.json`), and `npm update` gives you the _newest version period_ (respecting semver in your `package.json`). The idea of doing them differently is to make it possible to quickly update all of an application's dependencies, even if they include things that are too new to be `latest`.

pauldijou:
@othiym23 Gotcha, I can understand that. Quick question then: is there a way to quickly update all dependencies but staying stable? (respecting semver in `package.json` and under `latest` dist-tags) Because nearly all of the time, I don't want to install `next` or `beta` releases into a production application. The only one I can see so far is using something like `1.1.0 - latest` for example.

thealjey:
WTF? By design? Are you kidding?
IMHO, it's a really bad "design".
I agree with Paul on this, this IS a bug.
A purposefully designed bug, but a bug nonetheless.
If I have the latest version of a package installed it should not, under any circumstances, be marked as outdated.

thealjey:
@othiym23 
At least give the user an option to turn this new behavior off somehow (maybe a package.json config).
I'm pretty sure if you did 99.9% percent of people would use it :smiley:

PR #8848, opened 2015-07-07
Fix project config not being read with npm i -g

### LINKED ISSUE #8036: Cannot specific prefix in current directory .npmrc
Create .npmrc in the current (non-home) directory

```
prefix = foo
```

But `npm config get prefix` shows `/usr/local`.

This doesn't follow the documentation on [npmrc](https://docs.npmjs.com/files/npmrc):

> Each of these files is loaded, and config options are resolved in priority order. For example, a setting in the userconfig file would override the setting in the globalconfig file.

espadrine:
Fixes #8036.

[This commit](https://github.com/npm/npm/commit/1707fcedb7503251eb77719aba7ac52931721205#diff-0274417318d45b56e7adc8820b6a76b3R110) ensured that only non-global operations read the project configuration, possibly for performance reasons. However, the project configuration `.npmrc` file can change `prefix`, which modifies where a global operation such as `npm i -g` stores files.

No tests were broken. No tests were added either; it seems inadequate to make a test to check that a piece of code isn't there when it shouldn't have been there.

### DIFF
--- a/lib/config/core.js
+++ b/lib/config/core.js
@@ -127,8 +127,7 @@ function load_ (builtin, rc, cli, cb) {
     var projectConf = path.resolve(conf.localPrefix, '.npmrc')
     var defaultUserConfig = rc.get('userconfig')
     var resolvedUserConfig = conf.get('userconfig')
-    if (!conf.get('global') &&
-        projectConf !== defaultUserConfig &&
+    if (projectConf !== defaultUserConfig &&
         projectConf !== resolvedUserConfig) {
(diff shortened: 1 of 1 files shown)


othiym23:
Following the discussion in #8036, this is actually the desired behavior. However! I do think a tweak to the documentation to indicate that project-specific `.npmrc` configuration won't be applied to global installs would be helpful, so if you want to take a stab at that in a separate PR, I'd be happy to land it. Thanks for putting this together and for engaging in the conversation on #8036!

PR #8857, opened 2015-07-08
Make "*" match "latest" if all versions are prerelease

### LINKED ISSUE #8855: '*' should match latest when there are only prerelease versions
npm v2.11.2 introduced [a change](https://github.com/npm/node-semver/issues/123) from semver@4.3.5 which caused '*' to stop matching prerelease versions. This is well and good, until you have a package that's never had a non-prerelease version, at which point it would be reasonable to pick up latest instead.

iarna:
Fix #8855 to reduce the pain from the change made in 2.11.2 that resulted in depending in `*` no longer working for modules w/o any non-prerelease versions.


othiym23:
Woop! Thanks for jumping on this, Rebecca!

Issue #8796, opened 2015-07-01
.gitignore and .npmignore conflict on publish


calebmer:
After upgrading to npm I discovered I couldn't publish my modules with a babel compile step. In my `.npmignore` file I include a `src` folder, in my `.gitignore` file I include a `lib` folder. For a pre-publish script I compile the ES6 code in `src` to ES5 code in `lib` with babel.

When publishing the compilation occurs, but then I get the following error: `npm ERR! Cannot read property 'main' of undefined`. This error only occurs in npm version 2.12.x and up, including npm version 3.x.x. I am not having any problems publishing in npm version 2.11.0

Here is the npm log:
```
0 info it worked if it ends with ok
1 verbose cli [ 'node',
1 verbose cli   '/Users/webdev/.nvm/versions/node/v0.12.5/bin/npm',
1 verbose cli   'publish' ]
2 info using npm@2.12.1
3 info using node@v0.12.5
4 verbose publish [ '.' ]
5 silly cache add args [ '.', null ]
6 verbose cache add spec .
7 silly cache add parsed spec { raw: '.',
7 silly cache add   scope: null,
7 silly cache add   name: null,
7 silly cache add   rawSpec: '.',
7 silly cache add   spec: '/Users/webdev/projects/calebmer/mongold',
7 silly cache add   type: 'directory' }
8 verbose addLocalDirectory /Users/webdev/.npm/mongold/0.5.8/package.tgz not in flight; packing
9 verbose tar pack [ '/Users/webdev/.npm/mongold/0.5.8/package.tgz',
9 verbose tar pack   '/Users/webdev/projects/calebmer/mongold' ]
10 verbose tarball /Users/webdev/.npm/mongold/0.5.8/package.tgz
11 verbose folder /Users/webdev/projects/calebmer/mongold
12 info prepublish mongold@0.5.8
13 verbose unsafe-perm in lifecycle true
14 verbose stack TypeError: Cannot read property 'main' of undefined
14 verbose stack     at Packer.applyIgnores (/Users/webdev/.nvm/versions/node/v0.12.5/lib/node_modules/npm/node_modules/fstream-npm/fstream-npm.js:108:30)
14 verbose stack     at Packer.<anonymous> (/Users/webdev/.nvm/versions/node/v0.12.5/lib/node_modules/npm/node_modules/fstream-npm/node_modules/fstream-ignore/ignore.js:61:15)
14 verbose stack     at Packer.emit (events.js:110:17)
[...]

PR #8858, opened 2015-07-08
Update fstream-npm to fix some-times crasher when publishing

iarna:
Fixes: #8796


othiym23:
Whee! Thanks for this also, Rebecca! :leftshark:

PR #8859, opened 2015-07-08
Fix uninstall lifecycle order

iarna:


othiym23:
ALSO YES PLEASE

:shaved_ice: with tests!

Dang. Now I want a shave ice. I miss :surfer: :palm_tree: :(

Issue #8862, opened 2015-07-08
global install of scoped packages cannot run scripts


tcort:
Any time a scoped package tries to run something in the `install` script (even the implicit `node-gyp rebuild`), `npm` warns that it cannot run in the working directory. At first I thought it was a permissions issue (issue #3497); however, it doesn't affect the same package with a non-scoped package name (those work fine).

Steps to reproduce:

```
git clone https://github.com/schamane/node-syslog.git
cd node-syslog
sudo npm install -g # Works great!
$EDITOR package.json # change `name` from `node-syslog` to `@foo/node-syslog`
sudo npm install -g # npm WARN cannot run in wd @foo/node-syslog@1.2.0 node-gyp rebuild (wd=...
```

Tried with npm 2.1.9, 2.11.2, 2.12.1 and node 0.10.28, 0.10.32, and 0.12.6.

I don't think it's just me. I found another person with the same issue nodejs/docker-iojs#72

**Edit** still happens with iojs v2.3.3 and npm v3.1.0

tcort:
Just looking at the code, on line 51 of `lib/utils/lifecycle.js` it's comparing the base name of the working directory to the package name:

```
    if ((wd.indexOf(npm.dir) !== 0 || path.basename(wd) !== pkg.name) &&
        !unsafe &&
        pkg.scripts[stage]) {
      log.warn(
        'cannot run in wd',
        '%s %s (wd=%s)', pkg._id, pkg.scripts[stage], wd
      )
      return cb()
    }
```

In my example above, the condition `path.basename(wd) !== pkg.name` is always always true for scoped packages because `path.basename('@foo/node-syslog')` returns the last path component `node-syslog` (i.e. without the scope) and `pkg.name` is `@foo/node-syslog` (i.e. with the scope). That's why the warning (and skipped script execution) happens with scoped packages and not with non-scoped packages. I'll put together a patch and pull request later today.

othiym23:
Thanks for investigating, and a patch would be most welcome, especially if it includes a test!

codvio:
it gives warning for gifsicle when its dependency of imagemin-gifsicle. Because wd.indexOf(pkg.name) gives first occurance value.


```
    if ((wd.indexOf(npm.dir) !== 0 ||
          wd.indexOf(pkg.name) !== wd.length - pkg.name.length) &&
        !unsafe && pkg.scripts[stage]) {
      log.warn( "cannot run in wd", "%s %s (wd=%s)"
[...]

PR #8868, opened 2015-07-08
lifecycle: fix safety check to work with scoped packages.

tcort:
A safety check compares `path.basename(wd)` to `pkg.name`.
In the case of scoped packages (e.g. `@foo/bar`), the two are
never equal. Change it to check if `wd` ends with `pkg.name`.

Fixes: https://github.com/npm/npm/issues/8862


othiym23:
Rebased and landed in `npm@2.x`  as 40981f2e0c9c12bb003ccf188169afd1d201f5af. Thanks for putting this together!

PR #8873, opened 2015-07-09
Fix problem where failed optional's deps didn't rollback with it

iarna:


othiym23:
The changes look clean, but this one is definitely subtle enough it needs a test for me to feel comfortable with the change. :sheep: once test.

PR #8876, opened 2015-07-09
Limit global tree manipulation to named pkgs

iarna:
Install and uninstall operations on globals shouldn't look at all installed things, they should only look at the things you want to install or uninstall.


othiym23:
Hoisting the global vs local paths like this is a good first step towards eventually having a more explicit delineation in the architecture of the difference between the two installer modes. I feel good about this. :sheep: once test is there.

Issue #5820, opened 2014-07-29
Bash completion changes the COMP_WORDBREAKS, breaking global completion behaviour


rparree:
Hi,

I found out the the npm completion script i had in my bash_completion.d directory broke the way bash completion works with equal sign globally. 

It turns out the completion script i have installed modifies the `COMP_WORDBREAKS` variable, which influences the behaviour of other scripts.

See my stackoverflow answer with a listing of the completion script See http://stackoverflow.com/a/25010080/103605. 

See also http://stackoverflow.com/a/12495480/103605

jasonkarns:
I ran into this bug recently. Without the equal sign, git completion breaks for any option containing an equal sign.

`git log --date=<TAB>`,  `git commit --reuse-message=<TAB>`, `git branch --set-upstream-to=<TAB>` are all broken because of npm's modification of `COMP_WORDBREAKS`.

jasonkarns:
Looks like this is a duplicate of #4530. moving my comment. this one should probably be closed.

othiym23:
Closing this as a duplicate of #4530, but at the same time, this is resolved by d7271b8226712479cdd339bf85faf7e394923e0d, which will be in the next `npm@2` release. Now all we have to do is get the completion script replaced in everyone's `bash_completion.d`...

Issue #4530, opened 2014-01-22
Shell completion script breaks bash's environment variable assignment filesystem path completion


aadamowski:
The shell completion script  `lib/utils/completion.sh` has an unfortunate effect on Bash that it breaks standard TAB completion for setting environment variable values.

As reported on [ServerFault](http://serverfault.com/questions/537460/preventing-bash-tab-completion-of-an-environment-variable-with-a-fs-path-value-f/569081#569081), without npm shell completion the correct expected behaviour is as follows:

1) the user types:

    export SOME_VARIABLE_NAME=/partial/filesystem/path/to/someth

2) the user presses TAB

3) the command line gets completed to:

    export SOME_VARIABLE_NAME=/partial/filesystem/path/to/something_existing

With npm shell completion active, the result is broken - the environment variable name gets removed:

3) the command line gets completed to:

    export /partial/filesystem/path/to/something_existing

jasonkarns:
I ran into this bug recently. Without the equal sign, git completion breaks for any option containing an equal sign.

`git log --date=<TAB>`,  `git commit --reuse-message=<TAB>`, `git branch --set-upstream-to=<TAB>` are all broken because of npm's modification of `COMP_WORDBREAKS`.

louisremi:
This might be solved using the solution proposed in [this stackoverflow thread](http://stackoverflow.com/questions/10528695/how-to-reset-comp-wordbreaks-without-effecting-other-completion-script)
I'm a complete bash noob, I can't make sense of the syntax in the accepted answer, so I won't be able to propose a patch : (

whitty:
I had a go at rewriting the completion script and it seems to work for me - see 7e4d15f

I find firstly that completions with = and @ are no longer broken outside of npm, and npm completion seems to work fairly well.

That being said I'm neither a bash_completion export, nor an npm user so I don't know if I've broken anything.

jasonkarns:
Clearly the original author of this completion script had a use case in mind for ignoring `@` and `=` in the word completions. Do we have an example command that exercises this use case so that we can test @whitty's implementation?

jasonkarns:
Also, this change seems to be working just fine for me (ie, not breaking git's tab completion)

> GNU bash, version 4.3.33(1)-release (x86_64-apple-darwin13.4.0)

anthonyfok:
[...]

PR #8892, opened 2015-07-10
Don't break global COMP_WORDBREAKS in bash_completion

whitty:
This is a candidate fix for #4530 - works for me, and confirmed by another user.  The problem causes global breakage of all bash completion on a system and thus is a big issue for those affected.

As discussed on #4530 - there is no regression test to update, and no-one on the defect knows what the use-case is that required special handling of "=" and "@".  If anyone can suggest a test-case, please do.

---

As described in #4530, #5820 modification of global variable
COMP_WORDBREAKS causes global breakage of completion:

```
dd if=/dev/ze<tab>
->
dd /dev/zero
```

### DIFF
--- a/lib/utils/completion.sh
+++ b/lib/utils/completion.sh
@@ -7,17 +7,21 @@
 # Or, maybe: npm completion > /usr/local/etc/bash_completion.d/npm
 #
 
-COMP_WORDBREAKS=${COMP_WORDBREAKS/=/}
-COMP_WORDBREAKS=${COMP_WORDBREAKS/@/}
-export COMP_WORDBREAKS
-
 if type complete &>/dev/null; then
   _npm_completion () {
+    local words cword
+    if type _get_comp_words_by_ref &>/dev/null; then
+      _get_comp_words_by_ref -n = -n @ -w words -i cword
+    else
+      cword="$COMP_CWORD"
+      words=("${COMP_WORDS[@]}")
+    fi
+
     local si="$IFS"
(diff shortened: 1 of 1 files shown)


zkat:
Hmm. This seems to have created a dependency on `bash-completion` (external package) that we didn't have before. We'll have to get rid of `_get_comp_words_by_ref` to fix that.

whitty:
Hmm,... looks like I didn't understand the consequences here.  Can you give me an example completion that should use either `@` or `=` to so I can test a few alternative approaches appropriately?

zkat:
I don't really know how right now, unfortunately -- those completions were written way before my time, as much as I'd like to help (and I've never really written shell completions) :grin:

whitty:
I've updated it to degrade to breaking on `@` and `=` for bash if the `bash_completion` method `_get_copy_words_by_ref`is unavailable

othiym23:
@zkat is going to lean on @isaacs's superior bash knowledge to make sure that this change makes sense and to maybe learn some fun new things about adding completion support in bash. WHEE FUN. :tada:

othiym23:
I had to do a little poking around, and eventually we really need to have some better testing around npm's shell completion capabilities, but I believe this patch works as advertised – the existing completion behavior is preserved, without affecting global completion behavior. Rebased and landed as d7271b8226712479cdd339bf85faf7e394923e0d. Thanks so much for helping us get this longstanding issue resolved!

PR #8917, opened 2015-07-12
Document that the project npmrc is not read in global mode

### LINKED ISSUE #8036: Cannot specific prefix in current directory .npmrc

espadrine:
For instance, when performing a global install.

Fixes #8036.

Motivated by https://github.com/npm/npm/pull/8848#issuecomment-120103424.

### DIFF
--- a/doc/files/npmrc.md
+++ b/doc/files/npmrc.md
@@ -52,6 +52,9 @@ running npm in.  It has no effect when your module is published.  For
 example, you can't publish a module that forces itself to install
 globally, or in a different location.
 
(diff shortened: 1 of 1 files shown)


zkat:
Merged as of e7a11fd

PR #8922, opened 2015-07-13
Fix complete brokenness in `npm install blah` in npm@3.1.1

iarna:
Argh. No more 4am releases. Most (but not ALL) `npm install` use cases were instant crashers. Of course, the use cases I directly tested against were the one's that didn't crash. I THEN apparently went on to run the test suite against the wrong tree. =( =(

Existing test cases already exercise this code– without this patch much explodes. Tests for the patch that this patch fixes were previously merged.

This fixes the bug introduced by #8876

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -574,19 +574,20 @@ Installer.prototype.readLocalPackageData = function (cb) {
         return cb(er)
       }
       if (!currentTree.package) currentTree.package = {}
-      if (currentTree.package._shrinkwrap) return cb()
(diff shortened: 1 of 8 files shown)


iarna:
Elsewhere, @othiym23 said:

> Rebecca: as long as you removed the bb tests because they're tested elsewhere, LGTM :sheep:

And indeed: https://github.com/npm/npm/commit/1819e4b996156c1c89addeac8c0cd0a0b58d19f8
> **test: remove the now redundent combined github+bitbucket test**
> This dates back to before npm@2 integrated the new npa and added new entirely separate tests for these.

PR #8924, opened 2015-07-13
Fixes default values on comment in cache.js correctly

watilde:
It was different in the default value [written in js comment](https://github.com/npm/npm/blob/064d62c2f251cb5e9328f63228f711844b7a1787/lib/cache.js#L13-L15) and the actual [real default value](https://github.com/npm/npm/blob/064d62c2f251cb5e9328f63228f711844b7a1787/lib/config/defaults.js#L119-L121).


zkat:
:sheep:

Thanks for

zkat:
Merged into 2.x as bad4e01

watilde:
Thank you :dango:

Issue #8962, opened 2015-07-16
npm version doesn't find git if run in a subdir


gabro:
I have several npm projects in my git repo

```
root
  |---- .git
  |---- projectA
        |---- package.json
  |---- projectB
        |---- package.json
```

```
cd projectA
npm version patch --verbose
```

The version in package.json is correctly bumped, but no git tag nor commit is created.
Inspecting the logs I spot

```
npm verb version Error: ENOENT, stat '/Users/gabro/Work/buildo/infra/hophop/.git'
npm verb version  error checking for .git { [Error: ENOENT, stat '/Users/gabro/Work/buildo/infra/hophop/.git']
npm verb version   errno: 34,
npm verb version   code: 'ENOENT',
npm verb version   path: '/Users/gabro/Work/buildo/infra/hophop/.git' }
npm verb version not tagging in git
```

The issue is in the `checkGit` function in [`npm/version.js`](https://github.com/npm/npm/blob/v2.3.0/lib/version.js#L110-L117)

```js
function checkGit (cb) {
  fs.stat(path.join(npm.localPrefix, ".git"), function (er, s) {
    var doGit = !er && s.isDirectory() && npm.config.get("git-tag-version")
    if (!doGit) {
      if (er) log.verbose("version", "error checking for .git", er)
      log.verbose("version", "not tagging in git")
      return cb(null, false)
    }
  ...
}
```

The case is which the npm package lives in a subdirectory simply doesn't seem to be taken into account.

gabro:
This is by design (see #8963)

PR #8963, opened 2015-07-16
Make checkGit more reliable

gabro:
Check the result of `git rev-parse --show-toplevel` instead of checking whether `${npm.localPrefix}/.git` exists and it's a directory.

This has the effect of making `npm version` to correctly identify a git repository when running inside
a subdirectory (issue #8962)

### DIFF
--- a/lib/utils/git.js
+++ b/lib/utils/git.js
@@ -3,6 +3,7 @@
 exports.spawn = spawnGit
 exports.chainableExec = chainableExec
 exports.whichAndExec = whichAndExec
+exports.getGitRoot = getGitRoot
 
 var exec = require('child_process').execFile
 var spawn = require('./spawn')
@@ -47,4 +48,12 @@ function whichAndExec (args, options, cb) {
 
     execGit(args, options, cb)
   })
+}
+
+function getGitRoot (cb) {
+  whichAndExec(
+    [ 'rev-parse', '--show-toplevel' ],
+    { env: process.env  },
+    cb
+  )
 }
--- a/lib/utils/git.js
+++ b/lib/utils/git.js
@@ -53,7 +53,7 @@ function whichAndExec (args, options, cb) {
 function getGitRoot (cb) {
   whichAndExec(
     [ 'rev-parse', '--show-toplevel' ],
-    { env: process.env  },
+    { env: process.env },
     cb
   )
 }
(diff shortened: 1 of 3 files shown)


gabro:
There's one test failing because it's trying to create the tag `v0.0.1` which already exists, so the error is different than the expected one.

I tried passing the `'git-tag-version': false` to the `npm.load` configuration, but the version commands just seems to ignore it (i.e. `git-tag-version` is true nonetheless)

gabro:
I just realized that all the lifecycle tests were probably working "by chance", since the git part of the `npm version` command was silently failing, hence skipping the duplicate tag error I'm hitting now.

First two options that come to my mind:
 
- disable the git tagging for those tests
- run those tests within a fresh git repo, to avoid tag name conflicts

iarna:
It seems to me that this subtly changes the semantics–

Checking `${npm.localPrefix}/.git` says _"use if this module itself is a git repo"_ versus `git rev-parse --show-toplevel` which says _"use if we're anywhere inside a git repo"_

And that, of course, is why this test is failing.

gabro:
That makes sense, although I don't understand why passing `git-tag-version': false` to `npm.load` does not affect the result. Logging the flag from inside `version` returns `true` in any case.

I agree the semantic is changed with my PR, but I think this one is a valuable use case too (having a collection of modules in a single repo is something I've seen frequently).

Perhaps is it worth to add this behavior under a flag?

iarna:
@gabro: Yeah, I agree that `git-tag-version` not working is ... surprising

othiym23:
What happens when you have two packages in a single git repo and the second gets versioned to the same tag as the first? The current behavior is the way it is by design, and for the aforementioned reason (among others) I'm not super thrilled about the idea of changing it.

gabro:
@othiym23 yep, you raise a good point, which I didn't consider.
Although the current design doesn't work in my case, I agree this is probably not a sensible change.

Thank you both for the feedback!

gelito:
Sorry for reopen an old question, but, now, we have more complexes structures (in my case, with a docker above all the npm structure).

```
/.git
/docker-compose.yml
/docker-image/
/docs/
/src/
/src/package.json
```


I understood your point of view, but, why doesn't accept, with flags, perhaps, the possibilty to add the other option?? Of course, could be that the developer has 2 projects, then will fail, but also, could be that the developer has a more complex structure and you aren't helping by default.

I think that just add a parameter on package.json or a flag on the command line will be enough and will fix the problem.

PR #8971, opened 2015-07-17
update npm to use `semver@5`

othiym23:
Includes a reinstall of `node-gyp` to bundle its own version of `semver` until Nate gets around to accepting TooTallNate/node-gyp#664.


othiym23:
WROOOONG BASE!

PR #8974, opened 2015-07-17
memoize access checks to eliminate duplicate checks

iarna:
This was causing all sorts of problems for windows users. While we're here we refactor this code and add some tests.

### DIFF
--- a/lib/install/exists.js
+++ b/lib/install/exists.js
@@ -1,0 +1,35 @@
+'use strict'
+var fs = require('fs')
+var inflight = require('inflight')
+var accessError = require('./access-error.js')
+
+// The Windows implementation of `fs.access` has a bug where it will
+// sometimes return access errors all the time for directories, even
+// when access is available. As all we actually test ARE directories, this
+// is a bit of a problem.
+// FIXME: When this is corrected in Node/iojs, update this check to be
+// a bit more specific
+var isWindows = process.platform === 'win32'
+
+// fs.access first introduced in node 0.12 / io.js
+if (fs.access && !isWindows) {
+  module.exports = fsAccessImplementation
+} else {
+  module.exports = fsStatImplementation
+}
+
+// exposed only for testing purposes
+module.exports.fsAccessImplementation = fsAccessImplementation
+module.exports.fsStatImplementation = fsStatImplementation
+
+function fsAccessImplementation (dir, done) {
+  done = inflight('exists:'+dir, done)
+  if (!done) return
+  fs.access(dir, fs.F_OK, done)
+}
+
+function fsStatImplementation (dir, done) {
+  done = inflight('exists:'+dir, done)
+  if (!done) return
+  fs.stat(dir, function (er) { done(accessError(dir, er)) })
+}
--- a/lib/install/writable.js
+++ b/lib/install/writable.js
@@ -1,0 +1,42 @@
+'use strict'
+var path = require('path')
+var fs = require('fs')
+var inflight = require('inflight')
+var accessError = require('./access-error.js')
+
+// The Windows implementation of `fs.access` has a bug where it will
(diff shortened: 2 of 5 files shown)


lib/install/memoize.js
```diff
@@ -0,0 +1,19 @@
+'use strict'
+var inflight = require('inflight')
+
+module.exports = function (name, func) {
+  var cache = {}
+  return function (dir, done) {
+    done = inflight(func + '|' + dir, done)
```
othiym23: ITYM `name` instead of `func` here.

lib/install/writable-fs-access.js
```diff
@@ -0,0 +1,7 @@
+'use strict'
+var memoize = require('./memoize.js')
+var fs = require('fs')
+
+module.exports = memoize('writable', function (dir, done) {
+  fs.access(dir, fs.W_OK, done)
```
othiym23: This is gonna splode on Node 0.8. Port over the polyfill from the old code?

othiym23: Also will kaboom on Node 0.10, so _definitely_ needs a polyfill.

lib/install/writable.js
```diff
@@ -0,0 +1,11 @@
...
+var fs = require('fs')
+// The Windows implementation of `fs.access` has a bug where it will
+// sometimes return access errors all the time for directories, even
+// when access is available. As all we actually test ARE directories, this
+// is a bit of a problem.
+// FIXME: When this is corrected in Node/iojs, update this check to be
+// a bit more specific
+var isWindows = process.platform === 'win32'
+
+module.exports = (fs.access && !isWindows) ? require('./writable-fs-access.js') : require('./writable-fs-open.js')
```
othiym23: I think it would probably be confusing to defactor this all into a single file for the purposes of npm. When would you use `writable-fs-access.js` and `writable-fs-open.js` independently?

lib/install/exists.js
```diff
@@ -0,0 +1,11 @@
...
+var fs = require('fs')
+// The Windows implementation of `fs.access` has a bug where it will
+// sometimes return access errors all the time for directories, even
+// when access is available. As all we actually test ARE directories, this
+// is a bit of a problem.
+// FIXME: When this is corrected in Node/iojs, update this check to be
+// a bit more specific
+var isWindows = process.platform === 'win32'
+
+module.exports = (fs.access && !isWindows) ? require('./exists-fs-access.js') : require('./exists-fs-stat.js')
```
othiym23: Same comment as below re: having this functionality split into three files.

lib/install/exists-fs-access.js
```diff
@@ -0,0 +1,7 @@
+'use strict'
+var fs = require('fs')
+var memoize = require('./memoize.js')
+
+module.exports = memoize('exists', function (dir, done) {
+  fs.access(dir, fs.F_OK, done)
```
othiym23: See below re: `fs.access`.

othiym23:
What kind of problems was the lack of memoization causing? My comments about polyfills aside, the refactoring looks pretty clean, but memoization of stuff at the filesystem level is sort of a smell for me at this point due to all the weird issues we've run into in the past around the program's understanding of the state of the filesystem being out of whack with the actual state of the filesystem.

lib/install/memoize.js
iarna: thank you

othiym23:
We had a long in-air discussion of how the old behavior was creating a ton of filesystem activity, 99+% of which was unnecessary, and decided that both memoizing the calls _and_ using `inflight` was excessive, and opted instead to use just `inflight`. This now looks good! :sheep:

iarna:
Merged as 9fafb18

PR #8975, opened 2015-07-17
Fix linking of scoped modules

iarna:
Linking scoped modules was failing due to a check intended to see if the module was installed in the global directory. This did the classic `path.resolve(blah, '..')` which always, ALWAYS fails for scoped modules. Alas. Adds a test for this scenario too.

### DIFF
--- a/lib/link.js
+++ b/lib/link.js
@@ -50,6 +50,14 @@ function link (args, cb) {
   linkPkg(npm.prefix, cb)
 }
 
+function parentFolder (id, folder) {
+  if (id[0] === '@') {
+    return path.resolve(folder, '..', '..')
+  } else {
+    return path.resolve(folder, '..')
+  }
+}
+
 function linkInstall (pkgs, cb) {
   asyncMap(pkgs, function (pkg, cb) {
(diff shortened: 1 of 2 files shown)


lib/link.js
```diff
@@ -50,6 +50,14 @@ function link (args, cb) {
   linkPkg(npm.prefix, cb)
 }
 
+function parentFolder (id, folder) {
```
othiym23: Maybe I'm overly paranoid, but I'd also like a check that guarantees there's a `/` in the name before adding a second parent directory into the mix. Is that too paranoid?

othiym23:
Question / paranoia aside, LGTM. :sheep:

lib/link.js
iarna: The worst case scenario here if a `/` is missing is that it refuses to link the module. And given that it's invalid to start a module name with '@' and not have it be a scoped module, I'm not concerned about that happening.

othiym23: You have more faith in software to not do the most perverse possible thing than I do, but I defer to your judgment.

PR #8977, opened 2015-07-17
fix `npm rm [-g] <linkedthing>`

iarna:
npm@3 was breaking this in a variety of exciting ways.

First, w/ the global install patch from the last release, it was no longer recognizing global symlinks as symlinks.

Second, even when it did, it would try to remove all of their deps, before removing the link to them.

And finally, even if it _didn't_ remove all the subdeps, it would complain about not being able to touch them.

This addresses all three issues.
- [x] This requires a test.

### DIFF
--- a/lib/install/filter-invalid-actions.js
+++ b/lib/install/filter-invalid-actions.js
@@ -8,13 +8,25 @@ module.exports = function (top, differences, next) {
   validate('SAF', arguments)
   var action
   var keep = []
+
+  differences.forEach(function (action) {
+    var cmd = action[0]
+    var pkg = action[1]
+    if (cmd === 'remove') {
+      pkg.removing = true
+    }
+  })
+
   /*eslint no-cond-assign:0*/
   while (action = differences.shift()) {
     var cmd = action[0]
     var pkg = action[1]
-    if (pkg.isInLink || pkg.parent.target) {
-      log.warn('skippingAction', 'Module is inside a symlinked module: not running ' +
-        cmd + ' ' + getPackageId(pkg) + ' ' + path.relative(top, pkg.path))
+    if (pkg.isInLink || pkg.parent.target || pkg.parent.isLink) {
+      // we want to skip warning if this is a child of another module that we're removing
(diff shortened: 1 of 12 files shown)


lib/install/filter-invalid-actions.js
```diff
@@ -8,13 +8,25 @@ module.exports = function (top, differences, next) {
...
+
   /*eslint no-cond-assign:0*/
   while (action = differences.shift()) {
     var cmd = action[0]
     var pkg = action[1]
-    if (pkg.isInLink || pkg.parent.target) {
-      log.warn('skippingAction', 'Module is inside a symlinked module: not running ' +
-        cmd + ' ' + getPackageId(pkg) + ' ' + path.relative(top, pkg.path))
+    if (pkg.isInLink || pkg.parent.target || pkg.parent.isLink) {
+      // we want to skip warning if this is a child of another module that we're removing
```
othiym23: A brief note of _why_ we're skipping the warning would be helpful here (e.g. "because it's very noisy and gratuitous")

test/tap/rm-linked.js
```diff
@@ -0,0 +1,136 @@
...
+})
+
+test('cleanup', function (t) {
+  cleanup()
+  t.end()
+})
+
+function cleanup () {
+  process.chdir(osenv.tmpdir())
+  try { rimraf.sync(linkRoot) } catch (e) { }
```
othiym23: Why all the trys? rimraf won't complain if they're not there, and if it blows up otherwise, I think that's something we want to know about at test time.

othiym23:
A small request for clarification in a code comment and a question about the test cleanup, but substantively this LGTM. :sheep:

test/tap/rm-linked.js
iarna: Well, it blows up if the top level path doesn't exist at all, and a failing test will result in that scenario. So not catching that results in an unclear test environment when we startup the second time.

iarna:
merged as 6d59fcf et al

PR #9002, opened 2015-07-20
docs: make a note on files always included/excluded in the package

SimenB:
Not sure how to document it, but I think it should be somehow.
I found [this](https://github.com/npm/npm/blob/master/lib/utils/tar.js#L61) when searching, is that the place these are applied?

Should I list everything? In a table? Make a note about file type not mattering much?


iarna:
This landed in `npm@2.13.3` and `npm@3.2.0`!

SimenB:
You fixed it up, awesome!

PR #9012, opened 2015-07-21
build: use npm being released during release

zkat:
There was some weirdness going on where, depending on your
local setup during release, you could get unexpected results
during build (specifically, the version on the tarball and zip
corresponded to your global npm's version)

I went ahead and also made it so the npm pack is also done
by the version getting released, to avoid any possible
issues with 2.x vs 3.x


iarna:
This landed in `npm@2.13.3`!

PR #9022, opened 2015-07-21
document the use of "access" in "publishConfig"

boennemann:
I ended up using this a few times so I thought "access" might be worthy of being added to the list of things that are useful to be overridden during publish-time.


iarna:
Thanks for this! This landed in `npm@2.13.3` and `npm@3.2.0`!

boennemann:
Awesome, thank you!

Issue #8377, opened 2015-05-27
`npm version` doesn't create a commit when in a git submodule.


trusktr:
I'm in a git submodule of another git project, and the submodule is an npm package with a package.json and everything, but running `npm version patch -m 'v%s'` doesn't create a new commit in the submodule.

Is this by design? Is there a way to make it happen?

ktarplee:
GIT uses a different structure for submodules that NPM likely needs to know about to determine if it is in a git repository.  The .git in a submodule is not a directory but rather a file that contains
gitdir: ../.git/modules/uploader

I just got burned by this so it would be nice to fix this.

ktarplee:
npm --loglevel verbose version patch

when in a submodule you will se in the log
not tagging in git

The issue is in lib/version.js:145.  You need to drop the s.isDirectory() from the expression as it is not required to be a directory.  I think it is sufficient that .git exists to claim you are in a GIT repo.

zkat:
This is fixed as of #9033 -- thanks for the patch! Took a while for us to notice :P

PR #9033, opened 2015-07-22
.git is not required to be a directory.

ktarplee:
In a submodule it is actually a file with a line in it that tells GIT where to find the actual .git directory.

### DIFF
--- a/lib/version.js
+++ b/lib/version.js
@@ -142,7 +142,7 @@ function dump (data, cb) {
 
 function checkGit (localData, cb) {
   fs.stat(path.join(npm.localPrefix, '.git'), function (er, s) {
-    var doGit = !er && s.isDirectory() && npm.config.get('git-tag-version')
+    var doGit = !er && npm.config.get('git-tag-version')
     if (!doGit) {
       if (er) log.verbose('version', 'error checking for .git', er)
       log.verbose('version', 'not tagging in git')


ktarplee:
This pull request addresses #8377

ktarplee:
Looks like this fails in an unrelated way on node 0.8.  It passes on everything else.

othiym23:
It would be great if this had a test, because I'm presuming we added that directory check for a good reason in the first place. (The implication being that none of me, @iarna, or @zkat know what that reason was.) We can probably land it without a test, but it would very helpful to have one.

To prevent this kind of confusion in the future, we're going to figure out why that test was there (lil help, @isaacs?) and then hoist the check into a function that is named to indicate to future-ourselves why it was there.

ktarplee:
As far as the history of git goes... Submodules used to have a full .git directory in them (in other words it was required to be a directory for GIT to work) until GIT reorganized things in the filesystem to keep all the versioning information in the top level git project and only use files pointing to the top level .git directory module.

othiym23:
Sorry, @ktarplee, what I meant is the history of why npm handles git the way it does. There are many edge cases that have been handled in a variety of pretty specific ways over the years, which can make it tough to figure out whether something is there to deal with a specific regression or edge case, and when it just seemed like a good idea at the time. It doesn't help that adequately testing the interactions between npm and git is pretty hard. Thanks for the additional detail either way!

ktarplee:
I understood what you meant.  I just figured having the history of GIT might help you guys figure out why NPM does what it does.  I have no idea why it is requiring a directory except for the fact that GIT used to require a directory.

zkat:
_puts on Sherlock hat and puffs bubble on her pipe_

Okay so!

like @ktarplee said, `.git` in submodules did, in fact, use to be a directory (until around 2012. See [this patch that mentions it](http://git.661346.n2.nabble.com/PATCH-v4-Teach-rm-to-remove-submodules-unless-they-contain-a-git-directory-td7567950.html)).

At the same time, that particular line of code that does `s.isDirectory()` has been completely unchanged since as far back as 2010, when it was _originally implemented_. See #394 / b1e014e.

Taking that into account, *I* think it's pretty safe to say this was almost certainly just an extra safety check added on a whim during initial implementation. I'll try and poke @isaacs again and see if he has any recollection that might say otherwise, but this seems like fairly safe code. :)

zkat:
Talked to @isaacs, this is fine. Moving on with it for this release :+1:

zkat:
This was merged into v2.13.4 as of 2e58c48. Thanks for the patch! :+1:

othiym23:
Closing as merged.

PR #9038, opened 2015-07-23
install: Update fs.access to allow io.js versions w/ fixed windows ver

iarna:
This also pushes them off to their own file, so we can stop doing
the copy-pasta dance. It also makes explaining the rational of the
check cleaner I think.

### DIFF
--- a/lib/install/exists.js
+++ b/lib/install/exists.js
@@ -2,17 +2,9 @@
 var fs = require('fs')
 var inflight = require('inflight')
 var accessError = require('./access-error.js')
+var isFsAccessAvailable = require('./is-fs-access-available.js')
 
-// The Windows implementation of `fs.access` has a bug where it will
-// sometimes return access errors all the time for directories, even
-// when access is available. As all we actually test ARE directories, this
-// is a bit of a problem.
-// FIXME: When this is corrected in Node/iojs, update this check to be
-// a bit more specific
-var isWindows = process.platform === 'win32'
-
-// fs.access first introduced in node 0.12 / io.js
-if (fs.access && !isWindows) {
+if (isFsAccessAvailable) {
   module.exports = fsAccessImplementation
 } else {
   module.exports = fsStatImplementation
--- a/lib/install/is-fs-access-available.js
+++ b/lib/install/is-fs-access-available.js
@@ -1,0 +1,24 @@
+'use strict'
+var fs = require('fs')
+var process = require('process')
+var semver = require('semver')
+
+var isWindows = process.platform === 'win32'
+
+// fs.access first introduced in node 0.12 / io.js
+if (!fs.access) {
+  module.exports = false
+} else if (!isWindows) {
+  // fs.access always works on non-Windows OSes
+  module.exports = true
+} else {
+  // The Windows implementation of `fs.access` has a bug where it will
+  // sometimes return access errors all the time for directories, even
+  // when access is available. As all we actually test ARE directories, this
+  // is a bit of a problem.
+  // This was fixed in io.js version 1.5.0
+  // As of 2015-07-20, it is still unfixed in node:
+  // https://github.com/joyent/node/issues/25657
+
+  module.exports = semver.gte(process.version, '1.5.0')
+}
--- a/test/tap/is-fs-access-available.js
+++ b/test/tap/is-fs-access-available.js
@@ -1,0 +1,47 @@
+'use strict'
+var test = require('tap').test
+var requireInject = require('require-inject')
+
+test('mac + !fs.access', function (t) {
+  var isFsAccessAvailable = requireInject('../../lib/install/is-fs-access-available.js', {
+    fs: {},
+    process: {platform: 'darwin'}
+  })
+  t.is(isFsAccessAvailable, false, 'not available')
+  t.end()
+})
+test('mac + fs.access', function (t) {
+  var isFsAccessAvailable = requireInject('../../lib/install/is-fs-access-available.js', {
+    process: {platform: 'darwin'}
+  })
+  t.is(isFsAccessAvailable, true, 'available')
+  t.end()
+})
+test('windows + !fs.access', function (t) {
+  var isFsAccessAvailable = requireInject('../../lib/install/is-fs-access-available.js', {
+    fs: {},
+    process: {platform: 'win32'}
+  })
+  t.is(isFsAccessAvailable, false, 'not available')
+  t.end()
+})
+test('windows + fs.access + node 0.10.40', function (t) {
+  var isFsAccessAvailable = requireInject('../../lib/install/is-fs-access-available.js', {
+    process: {
+      platform: 'win32',
+      version: '0.10.40'
+    }
+  })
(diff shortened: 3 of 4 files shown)


lib/install/exists.js
```diff
@@ -2,17 +2,9 @@
 var fs = require('fs')
 var inflight = require('inflight')
 var accessError = require('./access-error.js')
+var isFsAccessAvailable = require('./is-fs-access-avalable.js')
```
thefourtheye: `avalable` should have been `available`?

zkat:
:sheep:

lib/install/is-fs-access-available.js
```diff
@@ -0,0 +1,24 @@
...
+
+// fs.access first introduced in node 0.12 / io.js
+if (!fs.access) {
+  module.exports = false
+} else if (!isWindows) {
+  // fs.access always works on non-Windows OSes
+  module.exports = true
+} else {
+  // The Windows implementation of `fs.access` has a bug where it will
+  // sometimes return access errors all the time for directories, even
```
thefourtheye: sometimes returns ... all the time?

thefourtheye:
LGTM with a nit

lib/install/is-fs-access-available.js
iarna: Under some circumstance that I don't understand, it always claims directories to be unwritable. These circumstances don't show up all the time, but when they do, they always result in failures.

At least, that was how I understood someone else's explanation of the bug. TBH, I haven't read the source material enough to summarize it directly.

lib/install/is-fs-access-available.js
```diff
@@ -0,0 +1,24 @@
+'use strict'
+var fs = require('fs')
+var process = require('process')
```
thefourtheye: This is not necessary.

lib/install/is-fs-access-available.js
thefourtheye: No problem. We can leave it as it is then.

test/tap/is-fs-access-available.js
```diff
@@ -0,0 +1,47 @@
...
+  var isFsAccessAvailable = requireInject('../../lib/install/is-fs-access-available.js', {
+    fs: {},
+    process: {platform: 'darwin'}
+  })
+  t.is(isFsAccessAvailable, false, 'not available')
+  t.end()
+})
+test('mac + fs.access', function (t) {
+  var isFsAccessAvailable = requireInject('../../lib/install/is-fs-access-available.js', {
+    process: {platform: 'darwin'}
```
thefourtheye: Shouldn't `fs` be passed in all the cases?

test/tap/is-fs-access-available.js
```diff
@@ -0,0 +1,47 @@
...
+var requireInject = require('require-inject')
+
+test('mac + !fs.access', function (t) {
+  var isFsAccessAvailable = requireInject('../../lib/install/is-fs-access-available.js', {
+    fs: {},
+    process: {platform: 'darwin'}
+  })
+  t.is(isFsAccessAvailable, false, 'not available')
+  t.end()
+})
```
thefourtheye: New line between tests would improve readability

test/tap/is-fs-access-available.js
iarna: yes =)

lib/install/is-fs-access-available.js
iarna: Yup. I didn't even know I could require process until I went to write these tests =D

test/tap/is-fs-access-available.js
iarna: Fair

test/tap/is-fs-access-available.js
```diff
@@ -0,0 +1,47 @@
...
+  var isFsAccessAvailable = requireInject('../../lib/install/is-fs-access-available.js', {
+    process: {
+      platform: 'win32',
+      version: '0.10.40'
+    }
+  })
+  t.is(isFsAccessAvailable, false, 'not available')
+  t.end()
+})
+test('windows + fs.access + node 2.4.0', function (t) {
```
thefourtheye: We might never have node 2.4.0 (but io.js) ;)

thefourtheye:
Thanks. Learned a lot from this PR. LGTM, if that matters :)

test/tap/is-fs-access-available.js
iarna: :trollface:

sindresorhus:
@iarna fyi => https://github.com/sindresorhus/fs-access

Many Node.js polyfills here: https://github.com/sindresorhus/awesome-nodejs#polyfills

iarna:
@sindresorhus You're just stating and checking the results there? Is that actually all the underlying mechanism does? I guess I should go look at `libuv`... =D If that's all its doing then it probably isn't doing what I want anyway.

My main concern is that there are a lot of other, OS specific, ways of making a directory unwritable that can't be sussed from unix-style permissions. My approach is to actually try to make a file and see if that works, which I guess is the duck-typing of folder permission checking. If `fs.access` really is just a user friendly way of looking at the results of `stat` then I'll have to rethink my use of it.

iarna:
Thanks for the critical eye here @thefourtheye – This landed as 28ebc6c227c0347ac429ac01bbf01ce33f2eda5c with some changes, as it turns out you can only `require('process')` on io.js. =D 

I was able to instrument things in a different way for node 0.12 and io.js, but 0.10 eluded me and ultimately I ended up making the test skip on that target. =/

lib/install/is-fs-access-available.js
lukeapage: > Under some circumstance that I don't understand, it always claims directories to be unwritable. These circumstances don't show up all the time, but when they do, they always result in failures.

from trial and error: if directories would require admin access they are reported as unwritable even if running as admin

lukeapage: does it matter that the fix in io. js was to always return true in windows for directories? 
Are you just relying on someone adding functionality to fs.access? what happens if it doesnt detect access..  will npm not roll back correctly?

iarna: @lukeapage
If `fs.access` is just blindly returning `true` on Windows, then yes, that is a problem. I'm going to have to go digging through the `libuv` source, clearly.

The only problem with the permissions checks being too permissive, is that you get failures later rather than earlier. It shouldn't otherwise cause issues, but I'd still like to fix it. My fallback code just tries making a file in the directory to see if it has permission, and perhaps we'll have to permanently use that.

lukeapage: ok..  i linked to the libuv code and fix in that issue i raised,  should
give you a head start.

iarna: That would be much appreciated, but I don't see any links?

iarna: Ok, I dug up the source, and yeah, I see it. Blerg. Ok, I guess I'll just ditch using access entirely.

Issue #8637, opened 2015-06-22
multi-stage: npm install allows to install a project into itself


mantoni:
This might be intentional, but `npm#master` responds with `Refusing to install foo as a dependency of itself`.

I also looked into the behavior of `npm link`, and `#multi-stage` as well as `#master` allows me to `--link` a project into itself.

iarna:
Happily this is now live in `npm@3.2.0`!

Issue #8921, opened 2015-07-13
npm@3 should not be trying to install modules that arch/os say are incompatible


cchamberlain:
Using Windows 8.1 and NPM 3.1 I seem to always get node-gyp errors for packages that don't affect the Windows platform (fsevents).  If I use --no-optional on all npm installs, I don't see the errors.  Is there any way to set a default to always use --no-optional?  Simple export of an environment variable like npm_config_install_nooptional would solve a lot of short-term Windows issues.

Maybe something like this already exists and I just can't find it?

iarna:
It shouldn't actually be trying to install an optional fsevents on Windows. (fsevents marks itself as OS specific.) I'll be correcting this…

cchamberlain:
I saw that fsevents is set to only run for OSX in the gyp file but I couldn't get it to respect that.

Here are dumps of my environment.  I'd be happy to test.  Still think that a global no-optional setting would be helpful (unless I am missing something).

from ZSH:
`uname` => `MSYS_NT-6.3`
`env` => http://pastebin.com/Pstc7MZu
`node -p process` => http://pastebin.com/ad2PuTMa
`npm config list -g` =>

    ; cli configs
    global = true
    user-agent = "npm/3.1.0 node/v0.12.6 win32 ia32"

    ; environment configs
    prefix = "C:\\Users\\ColeChamberlain\\local\\npm"

    ; userconfig C:\Users\ColeChamberlain\.npmrc
    shell = "bash"

    ; globalconfig C:\Users\ColeChamberlain\local\npm\etc\npmrc
    loglevel = "warn"

    ; node bin location = C:\Users\ColeChamberlain\local\npm\node.exe
    ; cwd = C:\Users\ColeChamberlain\cchamberlain\npm
    ; HOME = C:\Users\ColeChamberlain
    ; "npm config ls -l" to show all defaults.

iarna:
I updated the title to reflect the underlying cause

iarna:
Welp, I've found the code that does this. It exists. I didn't forget it. =D It just, apparently, doesn't work for some reason. And apparently we didn't have any tests for it. So we'll be fixing both of those things.

cchamberlain:
Awesome! All the recent fixes with npm 3 have been making my life much easier.

iarna:
\o/ Awesome, glad to hear it! =)

iarna:
Hey! Good news, this should be fixed in the just released `npm@3.2.0`!

[...]

PR #9039, opened 2015-07-23
refactor per module checks

iarna:
This in turn fixes https://github.com/npm/npm/issues/8637 and https://github.com/npm/npm/issues/8921

It also fixes a here-to-unknown engine checking bug (that is to say, it wasn't checking them– if it had been, it would have been inappropriately exploding <.<).


zkat:
Lots of stuff, so you maaaybe want another set of eyeballs, too? LGTM, though. I say :sheep: if your tests pass :)

iarna:
This was released w/ `npm@3.2.0`.

Issue #5779, opened 2014-07-23
`npm shrinkwrap` should be idempotent


othiym23:
Reason: noisy diffs, potential nondeterminism.

Key order within the dependencies and other properties should have a stable ordering (i.e. be sorted). A useful implementation strategy would also involve centralizing all the calls to `JSON.stringify` so that we have a standard way of marshalling objects for external use.

iarna:
Reopening because this was not addressed by #6933

wagenet:
From a project without an existing node_modules or shrinkwrap:

```
npm install
npm shrinkwrap
rm -rf node_modules
npm install
npm shrinkwrap
```

The shrinkwrap file changes.

iarna:
Do check out the somewhat confusingly named https://www.npmjs.com/package/npm-shrinkwrap module  in the interim.

mixonic:
Also related to https://github.com/npm/npm/issues/7292.

We're looking into dropping the `from` field or not rewriting it in 2.x. Very happy to work with y'all to get something upstream. If the `from` field is only used for logging context and not logic, we should be able to figure something out.

othiym23:
The original version of the shrinkwrap algorithm [saved only versions](https://github.com/npm/npm/commit/d54ce3154dfe5283fcfeffc13d4e003bbade6370):

```patch
commit d54ce3154dfe5283fcfeffc13d4e003bbade6370
Author: Dave Pacheco <dap@joyent.com>
Date:   Tue Feb 21 15:32:16 2012 -0800

    add "npm shrinkwrap"

diff --git a/lib/install.js b/lib/install.js
--- a/lib/install.js
+++ b/lib/install.js
@@ -138,1 +173,4 @@
-      fn(args, where, family, ancestors, true, data, cb)
+      rv["dependencies"] = {}
+      for (key in wrap)
+        rv["dependencies"][key] = wrap[key]["version"]
+      log.verbose([rv["dependencies"]], "readDependencies: returned deps")
```

A later version added support for `from` to [handle dependencies installed from URLs](https://github.com/npm/npm/commit/4bb884d2f8a30701c05ed81f7ace62be53147cae) better (which also eventually added support for installing from GitHub repo branches):

```patch
commit 4bb884d2f8a30701c05ed81f7ace62be53147cae
Author: isaacs <i@izs.me>
Date:   Tue Feb 28 17:25:18 2012 -0800

    shrinkwrap: Behave properly with url-installed deps

[...]

PR #9068, opened 2015-07-26
remove sorting all of package.json

Qix-:
As per a discussion at chalk/ansi-styles#13, NPM shouldn't automatically sort `package.json` contents. It makes `--save` and `--save-dev` incredibly useless for collaborators on open source projects, making diffs annoyingly complex and invalidating thought-out and precise ordering of `package.json` elements (yes, some people are just _that_ anal-retentive about things).

Further, it doesn't seem to provide any real benefit to begin with.

This PR removes the two places in code where `package.json` is deep-sorted, and removes the deep-sort code itself since these two places are the only locations its used. ~~Further, it removes the test that checks for a sorted package.~~ It introduces a new sort function using the originally included `sorted-object` package (still net 0 added dependencies) that simply sorts dependency fields. Tests only appeared to check dependencies anyway - it looks like this PR fixes the _originally intended_ functionality.

I can confirm the failed test `test/tap/bin.js` is not a regression; it was present as of b3c0e86662a1f64132d42df81019e1fbec9d6846.

Docs don't seem to mention `package.json` sorting, so nothing to be changed there (let me know if this isn't true).

// @sindresorhus


sindresorhus:
It should only sort the dependencies objects (dependencies, devDependencies, etc), not the other fields.

Qix-:
Fixing now.

KenanY:
@Qix- Are you using `npm@3`? Looks to me like https://github.com/npm/npm/commit/2f3c83a59f7387cff41ee0894707d1cffed04825 introduced this.

Qix-:
@KenanY I am.

othiym23:
@Qix-: can you help me understand how npm/npm#5779 is related? As far as I can tell, it only deals with shrinkwrap files, and not `package.json`. As much as possible, we try to treat shrinkwrap files as generated artifacts, which is at least a little different from how we think about / manage `package.json`.

@iarna what was your original intention with 2f3c83a59f7387cff41ee0894707d1cffed04825. Was it indeed just to make sure that the various dependency sections were alphabetized, or was there an intention to make the whole `package.json` file more consistent?

Qix-:
@othiym23 that is the issue referenced by the commit that introduced the sort-everything functionality. :)

iarna:
huh, I thought I had taken that back out, but apparently not…

iarna:
1b71c55 is the actual commit responsible for current package.json sorting behavior

iarna:
Ok, I'm gonna reject this PR, but thanks for raising this issue.

`install/update-package-json.js` is used to update `package.json` files IN your node_modules folders. Those are not user editable artifacts. It's behavior is correct.

By contrast `install/save.js` is for saving YOUR `package.json` and reasonably shouldn't be reshuffling unrelated keys on you.

Qix-:
Great thanks @iarna

PR #9079, opened 2015-07-27
Stop sorting anything other than the package.json fields being updated

iarna:
This provides an alternate fix to #9068


iarna:
This shipped in 3.2.1

PR #9089, opened 2015-07-28
windows: search for a user-installed npm

joaocgreis:
This changes npm.cmd and the npm script when running in Cygwin to look for an npm installation in prefix, running it if found. The default npm is used to get the prefix. This implements the changes described in https://github.com/joyent/node/issues/8528 .

Fixes npm/npm#6412
Fixes joyent/node#8528
Refer joyent/node#8554

### DIFF
--- a/bin/npm
+++ b/bin/npm
@@ -7,8 +7,21 @@ case `uname` in
     *CYGWIN*) basedir=`cygpath -w "$basedir"`;;
 esac
 
-if [ -x "$basedir/node.exe" ]; then
-  "$basedir/node.exe" "$basedir/node_modules/npm/bin/npm-cli.js" "$@"
-else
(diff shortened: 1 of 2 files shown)


othiym23:
We need to review and test this further, but don't have a problem with the change in principle. We specifically don't support Cygwin, though, so if we need to nerf stuff to make it work better for plain Windows or Git bash mingw, then that will happen.

othiym23:
Rebased and landed as e40e71f2f838a8a42392f44e3eeec04e323ab743. Thanks a bunch for putting this together, and for your patience @joaocgreis! This is included in `npm@2.14.1`, which will become `npm@latest` in a week.

othiym23:
(And at that point we can downstream it into Node and hopefully a bunch of our upgrading issues with npm will become a thing of the past! :tada:)

Issue #9050, opened 2015-07-23
npm 3: No warning when peer dependency also listed as regular dependency cannot be met


papandreou:
Hi!

The [unexpected.js](https://unexpected.js.org/) team is currently looking into npm 3 in order to come up with a plan for how to ensure that only compatible plugins are installed into the library. We have some complex scenarios where some plugins depend on certain versions of other plugins being installed into the libary instance, and until now we have used `peerDependencies` to express those requirements. Also, we have been relying on the somewhat controversial feature that the mere presence of a peer dependency will cause the module to be installed.

All in all it seems like npm 3 can lead to a much better experience for our users. The only unsolved problem for us is that scenario where a plugin requires another plugin to be installed in a specific version (range). Reading through the discussion on #6565 we were led to believe that we could achieve that by listing the second level plugin under both `dependencies` and `peerDependencies` of the plugin that needs it.

As an experiment we tried publishing four packages to our private npm repo:

* `bogus_a`@1.0.0, lists `bogus_b`@1.0.0 and `bogus_c`@1.0.0 under `dependencies`
* `bogus_b`@1.0.0, lists `bogus_d`@1.0.0 under both `dependencies` and `peerDependencies`
* `bogus_c`@1.0.0, lists `bogus_d`@2.0.0 under both `dependencies` and `peerDependencies`
* `bogus_d`@1.0.0 and @2.0.0

We expected that installing `bogus_a`@1.0.0 would cause a warning because both of the `bogus_d` peer dependencies cannot be fulfilled at the same time. However, it seems like the peer dependency is ignored in this case:

```
 $ npm install bogus_a
/home/andreas/work/test
└─┬ bogus_a@1.0.0 
  ├─┬ bogus_b@1.0.0 
  │ └── bogus_d@1.0.0 
  └─┬ bogus_c@1.0.0 
    └── bogus_d@2.0.0 
```

The multistage/dedupe process does exactly the right thing, as it cannot install a `bogus_d` at the top level that would be satisfied by both `1.0.0` and `2.0.0` -- but there's no warning about the peer dependencies not being met.

Would it be possible to get that warning?

sunesimonsen:
Would it be possible to only validate peer dependencies against top-level dependencies and issue a warning if the peer dependency is not satisfied? Then you could supply both a dependency and a peer dependency if dedupe hoist the right dependency to the top-level everything is good; otherwise you get a warning.

iarna:
[...]

PR #9169, opened 2015-08-05
Resolve peer deps from the perspective of the parent module

iarna:
This stops it from considering a direct dependency to be acceptable. The exception to this is if you're operating with the module with the peer deps as your top level. This means you're developing that module and having its peer deps installed inside it is appropriate.

This also updates `ls` to display this new perspective correctly.

Fixes: #9050


iarna:
Released in 3.2.2

Issue #9077, opened 2015-07-27
npm install --save git+ssh://host/foo.git results in crash


iarna:
```
53 verbose stack Error: Unsupported URL Type: ssh://host/foo.git
53 verbose stack     at parseUrl ($HOME/.nvm/versions/v0.12.2/lib/node_modules/npm/node_modules/npm-package-arg/npa.js:168:13)
53 verbose stack     at npa ($HOME/.nvm/versions/v0.12.2/lib/node_modules/npm/node_modules/npm-package-arg/npa.js:75:12)
53 verbose stack     at isProdDep ($HOME/.nvm/versions/v0.12.2/lib/node_modules/npm/lib/install/deps.js:55:10)
53 verbose stack     at isDep ($HOME/.nvm/versions/v0.12.2/lib/node_modules/npm/lib/install/deps.js:35:19)
53 verbose stack     at addRequiredDep ($HOME/.nvm/versions/v0.12.2/lib/node_modules/npm/lib/install/deps.js:112:8)
53 verbose stack     at $HOME/.nvm/versions/v0.12.2/lib/node_modules/npm/lib/install/deps.js:185:12
53 verbose stack     at $HOME/.nvm/versions/v0.12.2/lib/node_modules/npm/node_modules/iferr/index.js:13:50
53 verbose stack     at $HOME/.nvm/versions/v0.12.2/lib/node_modules/npm/lib/install/deps.js:426:7
53 verbose stack     at $HOME/.nvm/versions/v0.12.2/lib/node_modules/npm/node_modules/iferr/index.js:13:50
53 verbose stack     at addBundled ($HOME/.nvm/versions/v0.12.2/lib/node_modules/npm/lib/fetch-package-metadata.js:209:39)
```

iarna:
Released in 3.2.2

PR #9174, opened 2015-08-05
Prefer rawSpec over spec when saving

iarna:
This is necessary because npa mangles spec in ways that are not compatible npa itself. For reasons. Specifically, `git+ssh` urls have their `git+` stripped from the front, which, for reasons, is then fed BACK into npa, which then explodes.

Fixes: #9077

### DIFF
--- a/lib/install/deps.js
+++ b/lib/install/deps.js
@@ -171,10 +171,12 @@ exports.loadRequestedDeps = function (args, tree, saveToDependencies, log, next)
         child.isGlobal = true
       }
       if (saveToDependencies) {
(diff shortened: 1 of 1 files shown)


othiym23:
It's probably time to review `npm-package-arg` and decide whether the interface it exports is the interface we need, and adjust things accordingly. Editorializing aside, this LGTM. :sheep:

iarna:
othiym23: Agreed– I'd like to do an `npm-package-arg` 5.0 that folds in (optional) `realize-package-specifier` functionality and cleans up its return value.

iarna:
Released in 3.2.2

Issue #8996, opened 2015-07-19
npm@3 modules that link the same bin can remove a bin it doesn't own


phated:
As per @iarna at https://github.com/gulpjs/gulp-cli/issues/7#issuecomment-121037259

I haven't tested this, but a gulp core contributor tested and linked results at https://github.com/gulpjs/gulp-cli/issues/7#issuecomment-120907981

iarna:
Argh, yes, thank you for making these. =D They've been sitting on my todo list for the past few weeks. <.<

contra:
Any ETA on fixing this? We aren't able to release gulp 4.0 until this is fixed

iamstarkov:
im waitin gulp4 for almost one year, pls fix it 🐱

iarna:
You'll note that this is scheduled for today's release.

phated:
Awesome! Thanks @iarna!

iarna:
Hey! Good news! This is out and live in 3.3.0!

phated:
Hooray! :fireworks:

PR #9198, opened 2015-08-07
Fix bug where removing one package could result in another's bin being removed

iarna:
(In the case of conflicting bins)

This very much needs tests before it's cooked, as gently-rm is touchy code that's not directly tested much.

Fixes: https://github.com/npm/npm/issues/8996


iarna:
This just landed in 3.3.0!

PR #9201, opened 2015-08-07
Revert "doc: build partial html content as well as full-baked pages"

isaacs:
This reverts commit 8b58ad09b719295461167a34f666c5aa4d6e26f2.

As it turned out, those partial docs aren't actually used by the
http://docs.npmjs.com site, so there's no sense slowing down the build.


othiym23:
Rebased, reworded, and landed as 286f3d97103812f0fd84b70352addbe899e258f9. THANKS BOSS

PR #9227, opened 2015-08-11
Add silly logging for lifecycle.js

saper:
This is to reduce the amount of frustraction while troubleshooting various custom lifecycle script environment.

Cons: user's privacy, if any

### DIFF
--- a/lib/utils/lifecycle.js
+++ b/lib/utils/lifecycle.js
@@ -47,6 +47,7 @@ function lifecycle (pkg, stage, wd, unsafe, failOk, cb) {
     if (er) return cb(er)
 
     unsafe = unsafe || npm.config.get('unsafe-perm')
+    log.silly('lifecycle[' + stage + ']', 'valid working directory:', wd)
 
     if ((wd.indexOf(npm.dir) !== 0 ||
           wd.indexOf(pkg.name) !== wd.length - pkg.name.length) &&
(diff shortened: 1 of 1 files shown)


iarna:
FYI: We update the authors file automatically as part of our release process, so we don't look for that in pull requests.

As for this pull request, I'm happy to take this, but I think I'm going to remove the env logging as I fear that would make our already large logs much too big.  If you'd like to pick out some key env vars, I'd probably be up for that though.

saper:
I had this doubt too, but first I have no idea whether `PATH` or `LD_LIBRARY_PATH` or `PATHEXT` or `COMSPEC` or who-knows-what breaks pretty fragile lifecycle scripts. On a second though, having thousands of `addNameRange` logged already doesn't make it so bad, given that mostly up to three lifecycle scripts are invoked per npm run.

othiym23:
Landed, with a bit of tweaking by @iarna, as d088b7df7a6d620d0adac53a339c0108fe5f9685, included in `npm@3.3.1`. Thanks for putting this together and your time, @saper!

saper:
Posted one little nit. Can we have this in 2.x as well?

othiym23:
> Can we have this in 2.x as well?

The team decided that it wasn't worth the time to backport this to `2.x` – `npm@3` is very close to coming out of beta, and going forward, we want to switch to mostly fixing only critical or security issues in `npm@2`, not adding new functionality.

saper:
Can we discuss the 2.x issue again? This was a missing feature from the start. I got another issue with running lifecycle script in some kind of restricted environment (https://github.com/sass/node-sass/issues/1138) and [the log](https://gist.githubusercontent.com/cibulka/b83fb2543990c0212e50/raw/f5ab835695091675cddbf87057fe34f06c212826/npm-debug.log) says only "17721 info node-sass@3.3.2 Failed to exec install script". npm 2.x will be around for some time and it is really difficult to troubleshoot anything related to custom lifecycle scripts this way....

Issue #8812, opened 2015-07-03
Unintelligible console output


williamboman:
When running `$ npm install` on `npm@3.1.0` `node@0.12.5` the console output is pretty unintelligible. Attached gif for demo purposes.

![npm-install](https://cloud.githubusercontent.com/assets/6705160/8500571/bb40df24-219d-11e5-85df-0bc622c0e7c5.gif)

Somewhat related to #8761. This issue doesn't only refer to the carriage return/new line issue.

iarna:
Yes, if you change your window size while it's running it will mess it up. Uh... don't do that?

Yeah, we can try to catch SIGWINCH and look at the terminal size again although that's not universally available, so.

williamboman:
@iarna This behaviour is also present when working with split panes, which I do constantly. It's so malfunctioning I have to change the way I work in the terminal. No other package manager makes me have to do that. I really think it should be addressed.

iarna:
Hmm, I use split panes all the time and I've never encountered an issue… something somewhere must be misreporting the terminal size on you. Hmm…

williamboman:
@iarna Oh, no issues with terminal sizing. It's just that I like to work in other panes (often creating new ones) while an installation is ongoing. Listening to SIGWINCH and resizing the progress bar accordingly would fix this.

What's the status on the progress bar? Is it a WIP or is it considered pretty much done?

iarna:
Pretty much done. The module providing the progress bar visuals is here:

https://github.com/iarna/gauge

At some point, I intend to make a `meta-gauge` library that exposes that API but allows plugging w/ a host of different progress indicators. That will probably happen when we start looking at making npm more embedable. (Specifically, one of those views would be to just print out progress updates in a parsable format to allow things calling npm (like Visual Studio) to display their own progress bars based on those updates.)

iarna:
Completion tracking is a separate concern and is isolated to https://github.com/iarna/are-we-there-yet, which has its top level change event trigger gauge `show` calls.

`npmlog` triggers `pulse` calls with every log line.

iarna:
The "listen to column changes" fix is in, though some cruft still shows up, so I'm going to defer calling this fixed until we get that cleaned up.

williamboman:
The "cruft" is pretty much unavoidable afaik. Perfectly acceptable imo.

iarna:
Some of it I think I can fix 😁(Probably not when you shrink of course, but it shouldn't cruft up when you grow.)

[...]

PR #9240, opened 2015-08-11
[npm@3] Use the progress bar only with quiet logging levels

saper:
Adding -d or -ddd to npm invocation
causes a strange mix of the progress bar
along with the verbose output.

### DIFF
--- a/lib/npm.js
+++ b/lib/npm.js
@@ -358,7 +358,10 @@
 
         log.resume()
 
-        if (config.get('progress')) {
+        if (config.get('progress') &&
+             (log.level == 'warn'  ||
+              log.level == 'error' ||
+              log.level == 'silent')) {
           log.enableProgress()
         } else {
           log.disableProgress()


saper:
Might improve #8812 a bit.

iarna:
You link to 8812, so maybe I'm not understanding the problem you're seeing? Can you describe that a bit more?

For me I get the log scrolling in the window and a progress bar at the bottom. It sometimes isn't visible when there's a lot of stuff being printed, but when it slows down the progress bar is visible, and when the install finishes it goes away. This all seems fine to me…

Is that not what you're seeing?

saper:
I think this is really related to #9313. It looks like this: silly messages scroll off the screen - and to the right of every message there is a piece of progress bar attached.

It shows my age but I dislike vt control sequences in the output because I am very often using `script(1)` or excellent GNU screen logging facility to capture my work (batch or interactive) to be able to troubleshoot problems later.

Processing/analyzing a log full of vt sequences is cumbersome (yes, I know how to strip the sequences or use `less -R`...), so I set `color=false` by default.

iarna:
I'm going to close this, because between `--no-progress` and `--no-color` I think the problem here is adequately addressed. Plus we WILL be fixing #9313.

saper:
Thanks, this was opened before I have identified #9313 as a probable cause.

Issue #8751, opened 2015-06-29
npm should use /d command line argument to prevent Autorun commands when spawning cmd.exe


jkwuc89:
Please see https://github.com/npm/npm/issues/7688 for background information regarding this issue.
Specifically, these comments:

* https://github.com/sass/node-sass/issues/1012#issuecomment-115715502
* https://github.com/sass/node-sass/issues/1012#issuecomment-116608862

When npm spawns cmd.exe /c to start any process under Windows, it should also use the /d command line argument to prevent Autorun script commands from impacting the process being started.

saper:
I think Autorun scripts can have its legitmate uses, for example if somebody wants to log all invocations of cmd.exe or something like that.

saper:
This is one of possible solutions to #7333 problem, not a feature request per se.

othiym23:
Fixed by 204ebbb3e0cab696a429a878ceeb4a7e78ec2b94 / #9245, included in `npm@2.14.1`.

Issue #7333, opened 2015-02-10
Trying to install modules locally to a project and there's 3 modules that keep giving me issues "Error: Cannot find module 'C:\Users\USER\install.js'


kvntrn:
![image](https://cloud.githubusercontent.com/assets/4700929/6136084/7f1d872a-b133-11e4-9fa1-28ab3022d499.png)

Is there anything I can do to fix this? Been looking everywhere and so far no solution.

smikes:
It looks like something is causing `npm` to try to run post-install scripts relative to your home directory instead of the install directory.  I haven't seen a bug like that before, but since you are running a pretty old version of `npm` (not your fault, `node` hasn't been updated for a while and it used to bundle an old version; new `node` versions will come with 2.5.1)

There have been a lot of improvements to npm -- especially around conflicts and race conditions during install -- since 1.4.28. Can you try updating your npm installation and see if the problem persists?

To update npm on Windows, follow the instructions here: https://github.com/npm/npm/wiki/Troubleshooting#upgrading-on-windows 

Also, next time, could you please upload your npm-debug.log as a  gist <http://gist.github.com>? If possible, run npm with at least one `-d` flag to increase the logging level.

Thanks!

kvntrn:
I updated my npm to 2.5.1 and I am still getting the same errors.

Here is the npm-debug.log:

https://gist.github.com/2e76f520eca24034ec4a.git

Thanks for helping smikes!

smikes:
OK, it's still failing to run the install scripts but at least it's looking in the right place now (at least, it looks that way).

Both `phantomjs` and `node-sass` will require either that you have a certain set of tools installed (Python 2.x and a C++ compiler – Visual Studio Express for Windows Desktop), or else that you download pre-compiled versions of the libraries.  It may be that the problem is that the `postinstall` scripts are trying to build/download the appropriate libraries but can't.

Can you try the following steps:

1. Download and install phantom js <http://phantomjs.org/download.html>
2. npm -d install -g phantomjs

and post the npm-debug.log from step 2 (if any).  If that works, then retry your `npm install`.  

If `npm install` doesn't work even though the `npm install -g phantomjs` did work, then can you please post your `package.json` to a gist – remove any proprietary information, of course – to help me in analyzing the problem.

kvntrn:
Okay, so I installed phantom js, ran npm -d install -g phantomjs and the install failed.

https://gist.github.com/a03b571b77bd3b6fccca.git

[...]

PR #9245, opened 2015-08-11
lifecycle: add /d and /s to cmd.exe

saper:
`/d` disables `cmd.exe` AutoRuns feature, where registry entry points to the script that will be executed when cmd.exe starts. Surprisingly a lot of users have `CD \` or similar command there which causes lifecycle scripts to fail, since relative path to the script no longer works (and `package.json` has no way to deduce absolute path)

/s enables handling of quotes, so that you can have "C:\Program Files\Node\node.exe" quoted in your script. Node's `child_process.exec()` is doing the same. We prevent additional quoting by libuv already by setting `uv_spawn()` flag `UV_PROCESS_WINDOWS_VERBATIM_ARGUMENTS`.

If those flags are not set there is no way package.json can reliably tell us to run "node" executable with a relative script path:

``` json
    "scripts": {
      "install": "node scripts/install.js"
    }
```

Without /d "node" invocation may end up in a random directory as a result of the AutoRuns command.

Fixes:
    https://github.com/npm/npm/issues/8751
    https://github.com/npm/npm/issues/7333
    https://github.com/sass/node-sass/issues/1027
    https://github.com/sass/node-sass/issues/1012
    https://github.com/sass/node-sass/issues/659
    https://github.com/sass/node-sass/issues/603

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -306,3 +306,4 @@ Kyle M. Tarplee <kyle.tarplee@numerica.us>
 Derek Peterson <derekpetey@gmail.com>
 Greg Whiteley <greg.whiteley@atomos.com>
 murgatroid99 <mlumish@google.com>
+Marcin Cieślak <saper@marcincieslak.com>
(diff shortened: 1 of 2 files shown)


othiym23:
Rebased, tested by hand on Windows (because I have yet to figure out how to test these kinds of things automatically), and landed as 204ebbb3e0cab696a429a878ceeb4a7e78ec2b94. Thanks, @saper!

saper:
Thanks. I think traditional unit testing is not well suited for integration issues like this.

Issue #9223, opened 2015-08-10
Symlink cycles between deps makes npm explode


iarna:
Our tree walking gets burned by this in multiple places.

iarna:
This just landed in 3.3.0!

PR #9261, opened 2015-08-13
Ensure walking the tree doesn't result in infinite loops

iarna:
So much tree walking. SO MUCH. So many infinite loops.

Fixes: #9223


iarna:
This just landed in 3.3.0!

PR #9269, opened 2015-08-13
Bring in the fix for `npm init` including subdeps

### LINKED ISSUE #8716: npm init brings in flattened dependencies in v3
I usually do `npm init` right before I publish, so I've already got a ton of files in `node_modules/` with the new flatter behavior in v3. When it comes time to `npm init`, npm thinks all of those flattened dependencies belong in my package.json as top-level dependencies.

iarna:
Fixes #8716


iarna:
This landed in 3.3.0!

PR #9274, opened 2015-08-14
Make install order more consistent for directly dependencies

### LINKED ISSUE #8995: npm@3 given the same module list, not linking in same order
As per @iarna at https://github.com/gulpjs/gulp-cli/issues/7#issuecomment-121037259

I haven't tested this, but a gulp core contributor tested and linked results at https://github.com/gulpjs/gulp-cli/issues/7#issuecomment-120907981

iarna:
The install order is determined by:

1) The location of the thing to be installed relative to the root module.
2) Dependency order, such that a -> b -> c will install modules as: c, b, a

1 is deterministic, regardless of what's being installed.

But 2 can change the order of things higher in the dep tree.  Eg, b, or a
might get sorted earlier if c requires them.  This mostly doesn't matter,
but it does mean that if you have two modules with conflicting bins, they
_can_ get installed in different orders.  This changes sorts all of the top
level modules to be LAST, in location order (1), and then sorts all the rest
per (2).  This ensures that top level modules have a deterministic install
order. (Non top level modules can't have bin conflicts as that's treated
the same as a version conflict and the conflicting module would be hoisted.)

Fixes #8995 
- [x] Needs a test

### DIFF
--- a/lib/install/diff-trees.js
+++ b/lib/install/diff-trees.js
@@ -59,6 +59,12 @@ function requiredByAllLinked (node) {
   return node.requiredBy.filter(isLink).length === node.requiredBy.length
 }
 
+function isNotReqByTop (req) {
+  return req !== '/' &&     // '/' is the top level itself
+         req !== '#USER' && // #USER
+         req !== '#EXTRANEOUS'
+}
+
 var sortActions = module.exports.sortActions = function (differences) {
   var actions = {}
   differences.forEach(function (action) {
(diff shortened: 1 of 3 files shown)


lib/install/diff-trees.js
```diff
@@ -68,7 +68,21 @@ var sortActions = module.exports.sortActions = function (differences) {
...
+
+  var sortedlocs = Object.keys(actions).sort(sortByLocation)
+
+  // Do top level deps first, this stops the sorting by required order from
+  // unsorting these deps.
+  var toplocs = sortedlocs.filter(function (location) {
+    var mod = actions[location][1]
+    if (!mod.package._requiredBy) return true
+    var deepDeps = mod.package._requiredBy.filter(function (req) {
+      return req !== '/' && req[0] !== '#'
```
iarna: put in function

othiym23:
Reviewed by me over ScreenHero, with one small suggestion to improve comprehensibility. :sheep:

lib/install/diff-trees.js
```diff
@@ -68,22 +74,36 @@ var sortActions = module.exports.sortActions = function (differences) {
...
   }
   function sortByDeps (action) {
-    var cmd = action[1]
-    if (added[cmd.package._location]) return
-    added[cmd.package._location] = action
-    cmd.package._requiredBy.forEach(function (location) {
+    var mod = action[1]
+    if (added[mod.package._location]) return
+    added[mod.package._location] = action
+    mod.package._requiredBy.forEach(function (location) {
```
iarna: sort `_requiredBy`

othiym23:
Landed as d395a6b8767853eede6ad4502266ac99504f51a8. Consistency is one of our top-level goals, so this is a great thing to get in. Thanks once more, Rebecca!

PR #9279, opened 2015-08-14
Clarify which ware comes with which.

janaspage:


othiym23:
The "now" can probably go altogether at this point, as that chunk of the text was added in 2011 (and was probably intended to be tongue in cheek at the time). Either way, rearranging the words to be clearer in this case also makes the sentence less accurate. I took a stab at fixing that in #9315, but if you want the credit for the change (which I encourage you to take!) a new push of this PR with that wording also works.

othiym23:
We went ahead and merged #9315, landing it as a138c61415dae3f4fb490cbcfd2cb73f4db041d3, but thanks for putting this together and prompting us to make this wording clearer!

janaspage:
Cheers.

PR #9284, opened 2015-08-15
doc: remove misleading API documentation

othiym23:
npm's internal API is awkward, brittle, and frequently counterintuitive. Having it documented sets an unrealistic expectation of it being supported for use by developers outside the team. Until we have an API we can stand behind, having it be undocumented is actually a net win because people won't be misled into thinking they're gonna have a good time by using it.

**r**: @iarna 
**r**: @zkat


othiym23:
This is going to need a review from a stakeholder on @npm/docs, because it will remove a section of docs.npmjs.com when `npm@3` goes live as `latest`. @rockbot, @jefflembeck, @aredridel – who wants to go over this with me?

othiym23:
@rockbot as @ceejbot said elsewhere, you can remove that section from the docs site before we remove the section from the published docs, and in fact that would probably be the best way to handle this without risking any footguns. We were targeting making that section disappear along with the release of `npm@3` to the wider community. Just FYI, the target for that is currently a week from Thursday, on the 17th.

othiym23:
This is landed in the next release branch (`release-3.3.3`) as 2fb60bfef323d8dcf8f8c972bcb8e4d167e5e25d. Even after this change lands in a release, it won't affect what gets pushed to `docs` until we release `npm@3` as `latest`, which as I said will be next week, most likely.

PR #9304, opened 2015-08-17
Fix the progress-config test when run from travis or other CI

iarna:
We have special code to disable the progress bar on travis (and other CI) to
save ourselves and our users a lot of unhelpful output. Unfortunately as
this test didn't take that into account, it immediately exploded in this
envs.

So we fix that up, PLUS we add tests for the various env options for
suppressing progress bars.

### DIFF
--- a/lib/config/defaults.js
+++ b/lib/config/defaults.js
@@ -172,7 +172,7 @@ Object.defineProperty(exports, 'defaults', {get: function () {
     parseable: false,
     prefix: globalPrefix,
     production: process.env.NODE_ENV === 'production',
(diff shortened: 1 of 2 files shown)


iarna:
Not fixed all the way yet ;.; https://travis-ci.org/npm/npm/jobs/75915883#L3826-L3859

iarna:
Fixed!

othiym23:
Reviewed by me via ScreenHero. :sheep:

othiym23:
Landed as ad3714c2e52778c99c69c4be8e3beb350396708b. Woop! Thanks, Rebecca!

PR #9338, opened 2015-08-19
Added links as requested from npm/docs issue 106

ChicagoDev:
See: https://github.com/npm/docs/issues/106

### DIFF
--- a/doc/misc/npm-config.md
+++ b/doc/misc/npm-config.md
@@ -85,7 +85,7 @@ example:
 
 ## Per-Package Config Settings
 
-When running scripts (see `npm-scripts(7)`) the package.json "config"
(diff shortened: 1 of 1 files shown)


othiym23:
Thanks for putting this together! What npm/docs#106 is describing is a change to the process of _building_ the docs, not to the docs files themselves. It would be great to change the Markdown-to-HTML build process to automatically generate the crosslinks. Just replacing them by hand is going to be difficult to maintain and error-prone.

As such, I'm going to close this PR. I think it _does_ make sense to make the changes in npm/npm and not npm/docs, so if you're interested in helping move npm/docs#106 forward, take a look at https://github.com/npm/npm/blob/master/scripts/doc-build.sh for some ideas of where these changes should be made. And thanks again!

PR #9343, opened 2015-08-19
Elminate warnings about missing package.json at top level

### LINKED ISSUE #9113: 3.2: npm install completes before complaining about missing package.json
in npm 3.2, installing a package to a project without a package.json fails (ENOENT package.json).. but it performs all network/downloading first.. 

i work remotely on mobile data plans so pulling a lot of data, only to be told it failed, seems inefficient.. 

is it possible to test for package.json presence *before* tanning the network?

iarna:
npm@2 didn't emit these and to be honest, it's not that interesting. The test change actually just reverts a previous test change when these warnings were introduced.

This will half fix #9113

### DIFF
--- a/lib/ls.js
+++ b/lib/ls.js
@@ -146,7 +146,8 @@ function getLite (data, noname) {
     lite.problems.push('extraneous: ' + getPackageId(data) + ' ' + (data.path || ''))
   }
 
-  if (data.error && data.path !== path.resolve(npm.globalDir, '..')) {
+  if (data.error && data.path !== path.resolve(npm.globalDir, '..') &&
+      (data.error.code !== 'ENOENT' || noname)) {
     lite.invalid = true
     lite.problems = lite.problems || []
     var message = data.error.message
--- a/test/tap/ls-top-errors.js
+++ b/test/tap/ls-top-errors.js
@@ -1,0 +1,71 @@
+'use strict'
+var fs = require('fs')
+var path = require('path')
+
+var test = require('tap').test
+var mkdirp = require('mkdirp')
+var rimraf = require('rimraf')
+
+var common = require('../common-tap')
+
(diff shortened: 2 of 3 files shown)


lib/ls.js
```diff
@@ -146,7 +146,8 @@ function getLite (data, noname) {
     lite.problems.push('extraneous: ' + getPackageId(data) + ' ' + (data.path || ''))
   }
 
-  if (data.error && data.path !== path.resolve(npm.globalDir, '..')) {
+  if (data.error && data.path !== path.resolve(npm.globalDir, '..') &&
+      (data.error.code !== 'ENOENT' || noname)) {
```
iarna: add test that DOES trigger an error at the bottom from a top level package. Invalid package.json.

othiym23:
Reviewed by me via ScreenHero, with the suggestion of adding a sad path test. :sheep: once that's added.

test/tap/ls-top-errors.js
```diff
@@ -0,0 +1,77 @@
...
+var expected =
+  'npm ERR! error in ' + pkg + ': Failed to parse json\n' +
+  "npm ERR! Unexpected token 'b' at 1:2\n" +
+  'npm ERR! {broken json\n' +
+  'npm ERR!  ^\n'
+
+test('ls-top-errors', function (t) {
+  common.npm(['ls'], {cwd: pkg}, function (er, code, stdout, stderr) {
+    t.ifErr(er, 'install finished successfully')
+    t.is(stderr, expected)
```
iarna: these would be be less fragile w/ a regex match

othiym23:
Landed as 02420dc73ae185144267912e461da2effe90341d. It's fitting that these trees are getting "elminated"! Thanks, Rebecca!

PR #9344, opened 2015-08-19
Make sure user installed modules are attached to root of logical tree

### LINKED ISSUE #9113: 3.2: npm install completes before complaining about missing package.json

iarna:
If you had a set of modules with a dep cycle that never touched the root of the tree (that is, the top level was extraneous) then it would be entirely invisible to `npm ls`. It got detached from the root and never attached anywhere else.

Ordinarily extraneous modules were explicitly left in place, but it turns out that the way I was determining... er, extraneousness, was to see if anything required the module.

This will half fix #9113

### DIFF
--- a/lib/install/logical-tree.js
+++ b/lib/install/logical-tree.js
@@ -46,6 +46,9 @@ var logicalTree = module.exports = function (tree) {
     requiredBy.forEach(function (parentNode) {
       parentNode.children = union(parentNode.children, [node])
     })
+    if (node.package._requiredBy.some(function (nodename) { return nodename[0] === '#' })) {
+      newTree.children = union(newTree.children, [node])
+    }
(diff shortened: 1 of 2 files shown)


othiym23:
Reviewed by me via ScreenHero. :sheep:

othiym23:
Landed as d119ea67804e4bd2ea00025bd1d779cf7ec8ca18. Awesome! Radical! Other words co-opted to express generic affirmation during the 1980s!

PR #9358, opened 2015-08-20
Document aliases for npm cache

markstos:
'npm cache clear' does not fail like `npm cache boom` does, but it doesn't
provide any output either. What does it do?

A StackOverflow answer suggests it's not the same, but the source code shows
it's a silent alias for `npm cache clean`, along with `npm cache rf`.

Documenting these aliases clears up of the confusion.

The aliases for 'ls' are now documented as well, and the formatting of the
page is improved slightly as well.

### DIFF
--- a/doc/cli/npm-cache.md
+++ b/doc/cli/npm-cache.md
@@ -16,20 +16,20 @@ npm-cache(1) -- Manipulates packages cache
 
 Used to add, list, or clear the npm cache folder.
 
-* add:
+* **add:**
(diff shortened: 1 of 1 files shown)


othiym23:
npm provides a variety of useful, undocumented aliases for many of its commands. What's in the documentation is the official, recommended way to use those commands. Including all of the aliases would clutter up the docs (and probably also people's heads), so the omission of all the variant spellings is on purpose. As such, we're not going to take this, but thanks for taking the time to put it together!

PR #9360, opened 2015-08-21
Make it easier to run just _some_ tests in the same way as the test suite

iarna:
This restructures the test related scripts such that if you want to run ALL the tap tests you now run:

`npm run test-tap`

But if you want to run just a few of the tests:

`npm run tap /path/to/test`

This ensures that environmental differences between running the tests directly, via tap and via `npm run` don't result in tests that succeed when run directly but that fail when run via `npm test-all`.


othiym23:
The 2.x version of this was backported and landed as 26774579c739c5951351e58263cf4d6ea3d66ec8.

othiym23:
This was also landed on master as 55502060c9689670a0a47cddf6dfd6e69c1d7178, and is included in `npm@3.3.1`. Thanks for putting this together, Rebecca! This is useful!

Issue #9205, opened 2015-08-09
Local dependencies broken in npm v3


timoxley:
npm 3 crashes with an `ENOENT` error during an `npm install` if your app contains relative `file:` path local dependencies that have the dependency structure outlined below.

I believe this is caused by npm 3 looking for the child dependency relative to location of the parent inside node_modules, i.e. '../child' == 'node_modules/parent/../child' rather than relative to the location of the local file's source.

Tested on npm v3.2.2.
The structure installs just fine on npm v2.13.5.
I have cleared the cache and *not* cleared the cache between tests and witnessed no difference in result.

```dot
digraph {
  root -> A
  root -> B
  A -> B
  A -> C
  C -> B
}
```
![image](https://cloud.githubusercontent.com/assets/43438/9153410/77e3caaa-3e19-11e5-84ae-4fd204e0e201.png)

### Note that almost any other configuration will work. 

For example:

It works fine if you remove the root package's dependency on `B` 

```dot
digraph {
  root -> A
  // root -> B
  A -> B
  A -> C
  C -> B
}
```

It works if you flip the dependency between `B` and `C` i.e. `B -> C` instead of `C -> B`:

```dot
digraph {
  root -> A
  root -> B
  A -> B
  A -> C
  // C -> B
  B -> C
}
```

It works if you completely remove the dependency from `C -> B`:

```dot
digraph {
  root -> A
  root -> B
  A -> B
  A -> C
  // C -> B
}
```

----

### Concrete Example

See [timoxley/npm3-localdeps-issue](https://github.com/timoxley/npm3-localdeps-issue) for a concrete example of the failing structure.

```
git clone git@github.com:timoxley/npm3-localdeps-issue.git
cd npm3-localdeps-issue
npm install
```

Should produce the following error:
```
npm install
npm ERR! addLocal Could not install /Users/timoxley/Projects/tests/npm3-localdeps/node_modules/banana
npm ERR! Darwin 14.4.0
npm ERR! argv "/usr/local/bin/node" "/usr/local/bin/npm" "install"
npm ERR! node v2.0.1
npm ERR! npm  v3.2.2
npm ERR! path /Users/timoxley/Projects/tests/npm3-localdeps/node_modules/banana
npm ERR! code ENOENT
npm ERR! errno -2
npm ERR! syscall open
...
```

npm 3 also doesn't want to install this structure from the registry, with a similar yet different error:

`npm install npm3-localdeps-issue`

```
> npm install  npm3-localdeps-issue
[...]

PR #9367, opened 2015-08-21
Use package relative paths for specifier realization

iarna:
We were using this when calling `fetchPackageMetadata` but not when calling `realizePackageSpecifier`. 

Not having this meant that we were resolving relative paths for local modules incorrectly, be they directories or tarballs.

Fixes: #9205


othiym23:
Reviewed 1:1 with @iarna via ScreenHero. Looks good! :sheep:

othiym23:
Ease up on the trigger finger there, buddy!

othiym23:
Landed as 4468c92eccd0814b567606897e0792e49640e056. Thanks, Rebecca!

Issue #9308, opened 2015-08-17
`npm repo` runs prepublish


aredridel:
This surprised me very much!

```
:; npm repo

> newww@2.1.0 prepublish /Users/aredridel/Projects/npm/newww
> gulp build

module.js:338
    throw err;
    ^
Error: Cannot find module 'gulp-rev-all'
    at Function.Module._resolveFilename (module.js:336:15)
    at Function.Module._load (module.js:278:25)
    at Module.require (module.js:365:17)
    at require (module.js:384:17)
    at Object.<anonymous> (/Users/aredridel/Projects/npm/newww/gulpfile.js:14:14)
    at Module._compile (module.js:460:26)
    at Object.Module._extensions..js (module.js:478:10)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Module.require (module.js:365:17)
    at require (module.js:384:17)

npm ERR! addLocal Could not install /Users/aredridel/Projects/npm/newww
npm ERR! Darwin 15.0.0
npm ERR! argv "node" "/usr/local/bin/npm" "repo"
npm ERR! node v0.12.7
npm ERR! npm  v3.3.0
npm ERR! code ELIFECYCLE
npm ERR! newww@2.1.0 prepublish: `gulp build`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the newww@2.1.0 prepublish script 'gulp build'.
npm ERR! This is most likely a problem with the newww package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     gulp build
npm ERR! You can get their info via:
npm ERR!     npm owner ls newww
npm ERR! There is likely additional logging output above.

npm ERR! Please include the following file with any support request:
npm ERR!     /Users/aredridel/Projects/npm/newww/npm-debug.log
```

othiym23:
That _is_ very surprising! Did you hold onto the `npm-debug.log`? I'm sure this should be rep(r)oducible, but it might save us a little time in figuring out why the lifecycle was getting involved at all.

aredridel:
Here you are!

```
0 info it worked if it ends with ok
1 verbose cli [ 'node', '/usr/local/bin/npm', 'repo', '.' ]
2 info using npm@3.3.0
3 info using node@v0.12.7
4 verbose node symlink /usr/local/bin/node
5 silly fetchPackageMetaData .
6 silly fetchOtherPackageData .
[...]

PR #9369, opened 2015-08-21
Add direct support for local modules

iarna:
So, `fetch-package-metadata` has direct support for fetching metadata from the registry. For anything else it adds the package to the cache and then gets the metadata from there.

This adds support for loading metadata directly from local directories. This turns out to be important for commands like `npm repo` that want metadata about the module in the current directory.

Fixes: #9308


othiym23:
Reviewed 1:1 with @iarna via ScreenHero. Looks good! :sheep:

othiym23:
Landed as f4a578424b1f9d12f58210c229e7af633d952587. Whoop! Thanks, Rebecca!

PR #9376, opened 2015-08-23
pack: always ignore .npmrc

mantoni:
npm should ignore it's own dotfiles when running `npm pack`. This patch adds `.npmrc` to the internal ignore list.

Background story: I must have published packages with my credentials. I use `.npmrc` at work for my open source projects. I'm ignoring them with a global git ignore, so I tend to forget that they're there. Your bot suggests to add .npmrc to .npmignore, which would work of cause, but I'd rather have it in npm itself instead of adding new .npmignore files to other peoples projects.

@zkat @iarna I'm on npm@3, but maybe this should go to both branches. I tried finding a test case to build on. Could find one, sorry.

### DIFF
--- a/lib/utils/tar.js
+++ b/lib/utils/tar.js
@@ -84,7 +84,8 @@ BundledPacker.prototype.applyIgnores = function (entry, partial, entryObj) {
       entry.match(/^\..*\.swp$/) ||
       entry === '.DS_Store' ||
       entry.match(/^\._/) ||
-      entry === 'npm-debug.log'
+      entry === 'npm-debug.log' ||
+      entry === '.npmrc'
     ) {
     return false
   }


iarna:
So we actually _just_ added this over here:

https://github.com/npm/fstream-npm/commit/b6181728ca68d79d96cca27469c2e45d3e5e7eaf

But! Your pull request is really helpful because it points to a second point that (npm@3 only) needs to update to match that. npm@3 has very awkward integration w/ fstream in this code path, because it needs ignore logic that's currently unavailable to external modules.

My inclination is to port the commit I linked, and then add a new issue to the roadmap to decouple these (which will mean refactoring how fstream-npm does ignores).

mantoni:
@iarna Sounds good. However, wouldn't it make sense to merge this to fix the issue right now in both npm streams? If I understand correctly, the whole block would be removed in favor of the new fstream implementation in the next step, right?

iarna:
@mantoni I'm not sure I get what you mean...

This issue and then some is already fixed in `npm@next`. npm@3's master branch ALSO has the new `fstream-npm` but is going to require an additional change for it to take effect.

iarna:
(The comments about the roadmap and refactor are really more for the benefit of other npm folks, who may be wondering why this is so weird.)

mantoni:
@iarna Then it's fine. I was wondering if the code block I was patching was still in use. Couldn't that be removed altogether then?

iarna:
That code block is only in npm@3, and is why refactoring the integration between fstream-npm and npm would be awful nice. Once that's done then yeah, for sure.

mantoni:
Just to confirm: when I revert my change locally and `npm i fstream-npm@1.0.5` in npm, the `.npmrc` is not in the tar anymore. Thanks for explaining.

iarna:
Final followup: Yeah, turns out the fstream-npm _does_ fix npm@3 too, which is contrary to what I said before. So yay! \o/

PR #9385, opened 2015-08-24
Exclude .npmrc by default (fixes #9384)

### LINKED ISSUE #9384: Exclude .npmrc files by default
The documentation states that .npmrc files may contain sensitive information and that they have no effect on a published module.

Is there any good reason they should not be included in the list of files ignored regardless of .npmignore settings?

I recently got an email from the npm security bot informing me that I had accidentally published my password. The problem turned out to be a CI server writing a .npmrc file. Since there is not normally a local .npmrc, there was no reason for it to be in .gitignore, and it never showed up on github, which would have been more noticeable. 

Yes, I screwed up the configuration, but it seems like an easy mistake to make impossible.

tqc:


iarna:
Hey there! Thanks for putting this together. As it happens, this was fixed in npm@2 last week and will be fixed with npm@3 this Thursday.

Issue #6590, opened 2014-10-29
Improved message when package.json is missing


toomasr:
Currently when you run `npm install` and you are missing package.json or have it misspelt like I had (packages.json vs package.json) the error you get is very cryptic.

    npm WARN package.json package.json@0.0.0 No repository field.
    npm ERR! addLocal Could not install /Users/username/projects/project
    npm ERR! Darwin 14.0.0
    npm ERR! argv "node" "/usr/local/bin/npm" "install" "."
    npm ERR! node v0.10.32
    npm ERR! npm  v2.1.5
    npm ERR! code EISDIR
    npm ERR! errno 28
    
    npm ERR! EISDIR, read
    npm ERR!
    npm ERR! If you need help, you may report this error at:
    npm ERR!     <http://github.com/npm/npm/issues>
    
    npm ERR! Please include the following file with any support request:
    npm ERR!     /Users/username/projects/project/npm-debug.log

This error is not very informative and difficult to google (EISDIR, errno 28). I've created a StackOverflow issue to get a better ranking for the message. See http://stackoverflow.com/questions/26586830/unable-to-npm-install-my-package/ I was advised to create an issue here also to maybe improve the message.

iarna:
This was fixed at um, some point? I'm not sure when, but 2.7.3 says:
```
npm ERR! package.json ENOENT, open '/Users/rebecca/code/npmtest/package.json'
npm ERR! package.json This is most likely not a problem with npm itself.
npm ERR! package.json npm can't find a package.json file in your current directory.
```
Anyway, I'm gonna close it now that we're actually telling you that your package.json is missing.

alexey-shaposhnikov:
@iarna 
```
npm install git+https://github.com/cyrfer/x2js.git
npm ERR! addLocal Could not install /var/folders/c1/4310dxxx07d5mv7_mzcsvlfr0000gp/T/npm-86137-20be5949/git-cache-2ab3176e22b5/06d51bf460c3442a0f30a552883613adc038acdb
npm ERR! Darwin 14.0.0
npm ERR! argv "node" "/usr/local/bin/npm" "install" "git+https://github.com/cyrfer/x2js.git"
npm ERR! node v0.12.0
npm ERR! npm  v2.11.1
npm ERR! code EISDIR
npm ERR! errno -21

npm ERR! EISDIR, read
npm ERR! 
npm ERR! If you need help, you may report this error at:
npm ERR!     <https://github.com/npm/npm/issues>

[...]

Issue #9297, opened 2015-08-16
helpful errors when package has no package.json


trusktr:
I'm getting an error like in #8881. I have

```
 ❯ git --version                                                      [09:54:35]
git version 2.3.2 (Apple Git-55)

 ❯ npm -v                                                             [09:54:38]
3.3.0
```

and package.json looks like this

``` js
{
...
  "dependencies": {
    "army-knife": "^0.2.1",
    "async": "^0.9.0",
    "big.js": "^2.5.1",
    "bignumber.js": "^1.4.1",
    "bootstrap": "^3.2.0",
    "bootstrap-datepicker": "^1.3.0",
    "famous03": "git+https://github.com/trusktr/famous.git#jspm",
    "infamous": "^0.0.37",
    "james-bond": "^0.0.0",
    "jquery": "^2.1.1",
    "jquery-tooltipster": "git+https://github.com/iamceege/tooltipster.git#3.2.6",
    "jsfromhell": "git+https://github.com/jonasraoni/JSFromHell.git#master",
    "jss": "^0.10.1",
    "jss-nested": "^0.1.4",
    "leaflet": "git+https://github.com/RickMohr/Leaflet.git#better-inertial-scrolling",
    "operative": "^0.3.2",
    "string": "^3.0.0"
  }
}
```

The error looks like the following after npm install.

```
npm ERR! Command failed: git clone --template=/Users/josephpea/.npm/_git-remotes/_templates --mirror git+https://github.com/iamceege/tooltipster.git /Users/josephpea/.npm/_git-remotes/git-https-github-com-iamceege-tooltipster-git-3-2-6-07b3db04e00cbb8bcf32cecf4dd6d490
npm ERR! Cloning into bare repository '/Users/josephpea/.npm/_git-remotes/git-https-github-com-iamceege-tooltipster-git-3-2-6-07b3db04e00cbb8bcf32cecf4dd6d490'...
npm ERR! fatal: Unable to find remote helper for 'git+https'
```

Is NPM 3 forgetting to remove the `git+` before passing the URL to git or something?

cc: @othiym23 @iarna

trusktr:
I'm unable to make a reproduction because now I get a different error when trying `npm install` in the reproduction. I'm trying to reproduce it here: https://g.trusktr.io/trusktr/npm-issue-9297.git

But when I `npm install` I get

```
 ❯ npm install                                                                                                                                                           [10:18:44]
[...]

PR #9396, opened 2015-08-24
error-handler: improve EISDIR error message for the most common case

kenany:
Reports of `EISDIR` on the issue tracker are almost exclusively caused by users trying to install something that does not have a `package.json`. However, it is clearly difficult for many users to discern this simple problem from the error code alone. So let's expand upon it.

Fixes https://github.com/npm/npm/issues/6590, https://github.com/npm/npm/issues/9297

### DIFF
--- a/lib/utils/error-handler.js
+++ b/lib/utils/error-handler.js
@@ -366,6 +366,14 @@ function errorHandler (er) {
               ].join("\n"))
     break
 
+  case "EISDIR":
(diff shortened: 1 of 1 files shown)


othiym23:
Landed as 0313e40ee0f757fce8861be590ad668c23d7be53, included in `npm@2.14.2` and `npm@3.3.1`. Thanks a bunch for putting this together, Kenan!

PR #9404, opened 2015-08-25
Fix npm.commands.publish when in interactive shell

alanclarke:
When running node's interactive shell, require.main is undefined because no specific file is being executed by node.

This pr fixes the issue by falling back on process.cwd() if require.main is undefined, enabling people to test and use npm.commands.publish in node's interactive console.


othiym23:
We really don't encourage people to drive npm programatically (the API is terrible for use outside the product, which is why #9284 exists), and it's _really_ not meant to be driven from the REPL, even / especially for interactive debugging. We're very slowly working on decomposing npm into a set of components that can be used independently of each other, which is the API surface we want to support. As such, I'm not going to take this patch. Thanks for taking the time to put this together, though!

PR #9443, opened 2015-08-30
add 'npm it' command

isaacs:
I keep wanting a command for `npm i && npm t` so I could do `npm it`.

Needs tests and docs, but leaving this here for my fellow CLI peops to weigh in on whether this is a good thing or an excessive thing.

### DIFF
--- a/doc/cli/npm-install-test.md
+++ b/doc/cli/npm-install-test.md
@@ -1,0 +1,25 @@
+# npm install-test(1) -- Install package(s) and run tests
+
+## SYNOPSIS
+
+    npm install-test (with no args, in package dir)
+    npm install-test [<@scope>/]<name>
+    npm install-test [<@scope>/]<name>@<tag>
+    npm install-test [<@scope>/]<name>@<version>
+    npm install-test [<@scope>/]<name>@<version range>
+    npm install-test <tarball file>
+    npm install-test <tarball url>
+    npm install-test <folder>
+
+    alias: npm it
+    common options: [--save|--save-dev|--save-optional] [--save-exact] [--dry-run]
+
+## DESCRIPTION
+
+This command runs an `npm install` followed immediately by an `npm test`. It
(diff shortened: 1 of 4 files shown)


othiym23:
Aaglaglgalgaglaglahaghaga I want this ALL THE TIME. It might confuse some people's muscle memory, but YES THIS IS A GOOD IDEA:+1::japanese_goblin::ski::sunflower:

othiym23:
Sounds good, we'll try to get this out next week if you want to put together some tests and docs!

othiym23:
(and let's go ahead and do this against both 2 and 3)

zkat:
Missing tests so I'm bumping until next week. Should be quick for me to write some basic ones for this.

zkat:
Woe is me, if only SOMEONE would be willing to write tests + docs for this ;)

(I can take care of it, but it'll be a bit before I have time)

iarna:
I'm removing the milestone from this because, at this moment, writing tests for this (which is what's needed to land it) is not scheduled for this week or next.

mzgol:
Sometimes pruning is also needed, would it be worthwhile to have a shortcut for:
```
npm prune && npm install && npm test
```
?

othiym23:
I don't think that's the dominant use case, @mzgol, and I'd like to keep this a simple, focused addition if we can.

mzgol:
OK, fair enough.

ashleygwilliams:
documentation added.

iarna:
Tests look good: :sheep:

othiym23:
Rebased against `release-3.4.0` and landed as 1ac3e088f495bdcba45949ede36e2ad5578f60aa, bcb04f67b3cd1cd454cbfa1f78bf80257fb1cde2, and b6c17ddc65ebcd68091c39fae1f04501cde792a6. This was a real team effort. Thanks, everybody!

PR #9476, opened 2015-09-03
test: create .npmrcs if they're gone

othiym23:
This is what happens as a result of the changes to `fstream-npm` – when you try to integrate npm downstream, those files won't be there more because `make release` calls `npm pack`, which uses `fstream-npm`.


zkat:
Merged as 01a3428

Issue #9433, opened 2015-08-28
Invalid dependency format can result in EMISSINGARG


jamietre:
Not a high priority but caused some confusion for a while, e.g.

    "devDependencies": {
      "babel-core": "npm:babel-core@^5.1.13"
    }.

will result in

    > npm install
    npm ERR! Windows_NT 6.1.7601
    npm ERR! argv "node" "...\npm\\node_modules\\npm\\bin\\npm-cli.js" "install"
    npm ERR! node v0.12.7
    npm ERR! npm  v3.3.0
    npm ERR! code EMISSINGARG

    npm ERR! typeerror Error: Missing required argument #3
    npm ERR! typeerror     at exports.findRequirement (...\npm\node_modules\npm\lib\install\deps.js:475:3)
    npm ERR! typeerror   .. (etc)

This was a JSPM format dependency that I didn't fix properly. I would expect it to give an error more like how

    "devDependencies": {
      "babel-core": "foobar"
    }.

results in...

    > npm install
    npm ERR! No compatible version found: babel-core@foobar
    npm ERR! Valid install targets:
    npm ERR! ["5.8.23","5.8.22","5.8.21" ...]

othiym23:
Yeah, this error message isn't ideal, but it's also not a bug, per se, so a patch to return a better error message in this case would be appreciated.

iarna:
This was fixed a week or two ago! You should now get this instead:
```
npm ERR! Unsupported URL Type: npm:babel-core@^5.1.13
```

PR #9480, opened 2015-09-03
Report error gracefully with invalid URL format dependencies

jamietre:
Resolves issue #9433


iarna:
I've reviewed this, it looks great! Thank you! I've made two small changes, first, we've been prefixing the commit message w/ the thing it modifies. And second, we use a small module called `iferr` to explicitly route error conditions in scenarios like this. You can see the version I'm planning to land below:

https://github.com/npm/npm/commit/7c39f4f8f35fe503626b64c97e5ada3212ceab12

jamietre:
Makes sense, glad to know about the conventions. Thanks!

iarna:
Thanks again for this, it's live with npm@3.3.4

PR #9483, opened 2015-09-03
Add `silent` param to `npm pack` docs

xizhao:


othiym23:
Thanks for making this tweak! Because [the API docs are going away](https://github.com/npm/npm/pull/9284) as part of the general release of `npm@3`, though, it doesn't really make sense to take it. See #9284 for the details on why we're taking that seemingly drastic step. But thanks again for your time!

PR #9492, opened 2015-09-04
Clarify documentation wrt `.{npm,git}ignore` behaviour

addaleax:
I was a bit confused when I found out that `.gitignore` files are treated as replacements for `.npmignore` files even in subdirectories, given that the documentation basically only talks about `.npmignore` in the root directory.
I hope this clears things up a little for other users.


zkat:
Thanks! Merged into `npm@2.x` as 475daf5 and into `npm@3` as 2153297.

addaleax:
Great, thanks! :)

Issue #9505, opened 2015-09-06
npm link broken on windows with node 4


jon-hall:
Using npm 3.3.2, after updating to node v4.0.0-rc.1, 'npm link' throws an error on Windows complaining that the node version is pre-0.7.9.

jnfingerle:
Since this is a real world problem that happens to be a hassle for me: Is there any way I could help to work on this bug?

othiym23:
As soon as Node 4.0.0 is released (which should be imminent), this problem will go away. The underlying issue is that semver range matching doesn't include prerelease versions like `4.0.0-rc.1`. We're still discussing how to deal with this inside npm, because it definitely makes Node RCs and nightlies less useful than they should be.

jnfingerle:
You're right, the problem has disappeared for me with 4.0.0. That said, being able to test against future prerelease versions would be of much help for my team.

othiym23:
This is fixed in 6665e5489bae2418696695c5e8f15642ea2339b1 and c4dd5213b2f3283ea0392845e5f78cac4573529e for `npm@3` and `npm@2` respectively. Thanks for filing this!

PR #9506, opened 2015-09-06
allow npm link to run on windows with prerelease versions of node 

jon-hall:
Addresses #9505


othiym23:
@iarna Double-check with @isaacs that the behavior of `.gte()` here is as expected vs `">=0.7.9"`.

iarna:
Thank you for this! It shipped with 3.3.3 and 2.14.4.

Issue #9394, opened 2015-08-24
Uninstall refusing to delete .bin files - npm 3.3.0


bdukes:
Using npm 3.3.0, I get an error trying to uninstall the module _nativefier_:
 > Refusing to delete: C:\Code\npm-uninstall-test\node_modules\.bin\nativefier.cmd not in C:\Code\npm-uninstall-test\node_modules\nativefier

I was able to uninstall a different package.  I do not get the error in npm 2.14.0.  I'm running Windows 10.

[See silly log output](http://pastebin.com/raw.php?i=pg07sg8w)

othiym23:
```
npm ERR! gentlyRm C:\Code\npm-uninstall-test\node_modules\.bin\nativefier.cmd is outside C:\Code\npm-uninstall-test\node_modules\nativefier and not a link
```

Whoops! Looks like our deletion code needs to be made aware that `cmd` shims are a thing, @iarna. :sob: 

Thanks for the report, @bdukes, and we'll try to get this fixed soon!

iarna:
That was surely caused by https://github.com/npm/npm/pull/9198

robertbaker:
@iarna please fix this and release https://github.com/npm/npm/issues/9095 asap. It's really annoying

amyspark:
Using npm 3.3.2, this bug is still present.
`node install gulp-sass@"~2.0.4" --save`
```
npm ERR! Refusing to delete: d:\Documents\Amyspark\TheSilvermoonSentinelFrontend\node_modules\node-sass\node_modules\.bin\supports-color.cmd not in d:\Documents\Amyspark\TheSilvermoonSentinelFrontend\node_modules\node-sass\node_modules\supports-color
File exists: d:\Documents\Amyspark\TheSilvermoonSentinelFrontend\node_modules\node-sass\node_modules\.bin\supports-color.cmd
Move it away, and try again.
```

iarna:
Good news, this should be fixed as of 3.3.3!

vjpr:
I'm having this same issue with npm 2.14.8.

I use `npm link ~/dev/gulp` then I get:

```
npm ERR! Refusing to delete: ~/dev/gulp4/node_modules/.bin/gulp not in ~/dev-live/foo
File exists: ~/dev/gulp4/node_modules/.bin/gulp
Move it away, and try again. 
[...]

PR #9537, opened 2015-09-10
Fix removing of cmd-shims when removing modules

iarna:
When removing modules on windows, there was a bug that would result it in refusing remove any of the associated command shims.

This fixes: https://github.com/npm/npm/issues/9394

The root cause was the change made in #9198, which understood how to read symlinked bins but not how to read command shims.


iarna:
🐑from @othiym23 during screenhero code review yesterday

othiym23:
:sheep: :sheep: :sheep:

iarna:
This was shipped with 3.3.3!

Issue #9544, opened 2015-09-10
process.binding is being deprecated.


ChALkeR:
You use `process.binding` in [lib/adduser.js#L10](https://github.com/npm/npm/blob/v3.3.2/lib/adduser.js#L10).

There is an ongoing discissuon on deprecating it: https://github.com/nodejs/node/pull/2768, that code should be rewritten without `process.binding`.

It should be fairly simple, I guess.

ChALkeR:
This is fixed and should have been closed with #9549.

PR #9549, opened 2015-09-10
adduser: remove obsolescent process.binding use

othiym23:
There's a use of `process.binding('crypto')` that dates back to 2010, when the `crypto` module wasn't an automatically compiled-in dependency of Node. As far as I know, `crypto` has been part of Node by default since at least 0.6, and either way, the `require('crypto')` in a try / catch should be sufficient on its own.

Fixes: #9544


zkat:
:sheep:

iarna:
Hey, this landed back in 3.3.3 with 2b97a57!

PR #9553, opened 2015-09-11
refactor(lifecycle): Swap out custom logic with add-to-path module

kentcdodds:
I ripped off most of your code to create [add-to-path](npm.im/add-to-path) to accomplish this for [something else](https://github.com/gtramontina/ghooks/issues/26) and thought I'd give it back. It's 100% tested and written with all the :heart: I have.

### DIFF
--- a/lib/utils/lifecycle.js
+++ b/lib/utils/lifecycle.js
@@ -9,19 +9,10 @@ var path = require('path')
 var fs = require('graceful-fs')
 var chain = require('slide').chain
 var Stream = require('stream').Stream
-var PATH = 'PATH'
+var addToPath = require('add-to-path')
+var getPathVar = require('add-to-path/dist/get-path-var')
 var uidNumber = require('uid-number')
 var umask = require('./umask')
-
-// windows calls it's path 'Path' usually, but this is not guaranteed.
-if (process.platform === 'win32') {
-  PATH = 'Path'
-  Object.keys(process.env).forEach(function (e) {
-    if (e.match(/^PATH$/i)) {
-      PATH = e
-    }
-  })
-}
 
 function logid (pkg, stage) {
   return pkg._id + '~' + stage + ':'
@@ -89,9 +80,7 @@ function lifecycle_ (pkg, stage, wd, env, unsafe, failOk, cb) {
   // we also unshift the bundled node-gyp-bin folder so that
   // the bundled one will be used for installing things.
   pathArr.unshift(path.join(__dirname, '..', '..', 'bin', 'node-gyp-bin'))
-
-  if (env[PATH]) pathArr.push(env[PATH])
-  env[PATH] = pathArr.join(process.platform === 'win32' ? ';' : ':')
+  addToPath(pathArr)
 
   var packageLifecycle = pkg.scripts && pkg.scripts.hasOwnProperty(stage)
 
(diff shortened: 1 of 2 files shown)


lib/utils/lifecycle.js
```diff
@@ -89,9 +80,7 @@ function lifecycle_ (pkg, stage, wd, env, unsafe, failOk, cb) {
   // we also unshift the bundled node-gyp-bin folder so that
   // the bundled one will be used for installing things.
   pathArr.unshift(path.join(__dirname, '..', '..', 'bin', 'node-gyp-bin'))
-
-  if (env[PATH]) pathArr.push(env[PATH])
-  env[PATH] = pathArr.join(process.platform === 'win32' ? ';' : ':')
+  addToPath(pathArr);
```
kentcdodds: whoops, just realized this has a semi-colon. I'll update the PR to match the conventions.

kentcdodds:
Fixed

iarna:
@kentcdodds I tried to bring this in this week, but in the patch, you do this:

```
var PATH = require('add-to-path/get-path-var')()
```

But that doesn't work with your module as published– I get:

```
node
> var PATH = require('add-to-path/get-path-var')()
Error: Cannot find module 'add-to-path/get-path-var'
    at Function.Module._resolveFilename (module.js:336:15)
    at Function.Module._load (module.js:286:25)
    at Module.require (module.js:365:17)
    at require (module.js:384:17)
    at repl:1:12
    at REPLServer.defaultEval (repl.js:164:27)
    at bound (domain.js:250:14)
    at REPLServer.runBound [as eval] (domain.js:263:12)
    at REPLServer.<anonymous> (repl.js:392:12)
    at emitOne (events.js:82:20)
```

Also, in `src/index.js`, the entry point for your module, I see a `=>` function, which obviously we can't use in npm, which still nominally supports all the way back to 0.8.

It looks like you were originally intending to run babel over your source tree?

kentcdodds:
Oh! Yeah, sorry about that. Let me look into that. Thanks!

iarna:
So, the only use of `PATH` in `lib/utils/lifecycle` is to fetch the value of the path attribute from the environment.

I think that API would look cleaner (for npm anyway) if you made it so that the exported function returns the value of the path, eg

```
var getPathValue = require('add-to-path/get-path-value')
```
And then later, instead of:
```
  log.verbose('lifecycle', logid(pkg, stage), 'PATH:', env[PATH])
```
it becomes
```
  log.verbose('lifecycle', logid(pkg, stage), 'PATH:', getPathValue())
```

What do you think?

kentcdodds:
Great feedback. Will do. Thanks!

kentcdodds:
That should do it! Hopefully CI passes. Thanks for iterating with me on this @iarna :-) Any additional feedback?

iarna:
:+1:

kentcdodds:
ping? :-)

iarna:
We didn't do any PRs last week, but this week... THIS WEEK, we are, so it's coming in for tomorrow. =)

iarna:
Thanks for the work on this! It shipped with 3.3.7!

kentcdodds:
Hooray! :-) modularity for the win!

iarna:
We had to revert this as it was mutating (and growing forever) the global environment's path env, which is not what the old code did. I totally missed that when I was reviewing!

kentcdodds:
Ah!!! I'm so sorry! I'll give it another look :-/ How embarrassing :-(

kentcdodds:
If you have any ideas as to where/how that's happening, I'd be all ears.

iarna:
You're modifying `process.env`, where as the original code works with a lexically scoped clone of it. Having your module hold the environment doesn't work, it needs to take that as an argument.

iarna:
I would love to see something like this, but I'm gonna be a little more specific about the kind of interface I'd like to see. Don't worry to much about the actual module/function names, I'm just sketching here:

```
var managePath = require('manage-path')
…
managePath.push(env, '/single/path')
managePath.push(env, ['/path/to/one/thing', '/path/to/other/thing'])
managePath.shift(env, string)
managePath.shift(env, array)
managePath.get(env) // returns value of the path attribute in env
```

Alternatively...
```
var ManagePath = require('manage-path')
var runPath = new ManagePath(env)
runPath.push(…)
runPath.shift(…)
runPath.get()
```

Issue #9583, opened 2015-09-14
[docs:scripts] Reference to node-waf


ronkorving:
This *has* to be some overlooked legacy, right? See: https://docs.npmjs.com/misc/scripts

ref: https://github.com/npm/newww/issues/1323
cc @aredridel

aredridel:
Sure is. node-waf went out with the arrival of node 0.8; the current default involves some shenanigans around node-gyp instead. #9584 looks right to me, anyway!

PR #9584, opened 2015-09-14
doc: replace `node-waf` with `node-gyp` in default `install` script

kenany:
The behavior described here has been inaccurate for a long time. Not only was the `wscript` bit removed from `read-package-json` over 2 years ago (https://github.com/npm/read-package-json/commit/bb5f88a9809e0382d36fb47db184a0fc6e84fb96), the `install` script is what is set to the default (now `node-gyp rebuild`), not `preinstall`.

Fixes https://github.com/npm/npm/issues/9583


zkat:
Hooray! Merged as cf1b39f

Thanksw @KenanY!

PR #9667, opened 2015-09-18
Fix stack overflow on pack

iarna:
So... under some circumstances we're seeing stack overflows when people initiate actions that result in a package being created. This can be from installing a directory, running `npm publish` or directly running `npm pack`.

While walking back through the tarball creation code, I realized that the npm@3 bundled dependency support, which needs to check to see if an installed package is a subdep of a bundled dep, was hanging on to its copy of the install tree by sticking it in the `props` attribute. The problem with this is that the `props` attribute gets written to the tarball's headers in a recursive fashion. If there any dependency cycles then this recursion would go on forever and we'd see a stack overflow.

This patch stops storing the tree in the object and instead closes over it with the class we'll be using to feed the tarball. This results in simpler and (I believe) easier to understand code.

While I was working on this, I realized that we were needlessly building out the tree in the first place in the case of modules w/o any bundled dependencies. So this pull request also includes a patch to stop doing that.

### DIFF
--- a/lib/utils/tar.js
+++ b/lib/utils/tar.js
@@ -20,6 +20,7 @@ var myUid = process.getuid && process.getuid()
 var myGid = process.getgid && process.getgid()
 var readPackageTree = require('read-package-tree')
 var union = require('lodash.union')
+var flattenTree = require('../install/flatten-tree.js')
 
 if (process.env.SUDO_UID && myUid === 0) {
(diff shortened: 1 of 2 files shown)


iarna:
Review in person by @othiym23: 🐑

iarna:
Landed in 3.3.5!

Issue #8402, opened 2015-05-30
Lifecycle: script that runs just before (and only before) `npm publish`


olalonde:
It would be nice to have a script that runs only when running `npm publish`. For example, this would allow developers to compile files (coffeescript, babel, etc), automatically tag a git repository with the package's version, generate documentation, run tests, etc. 

There is currently a trigger called `prepublish` but it also runs on `npm install` which is not ideal for the use cases mentioned (see https://github.com/npm/npm/issues/3059).

AlastairTaft:
+1

mtscout6:
I would like to re-emphasize @dfellis's point (https://github.com/npm/npm/issues/3059#issuecomment-12257387) that the current behavior is not semantically the same as all the other lifecycle hooks.

I also understand the caveats of whether or not this should be classified as a "breaking change" even though it was initially released as a Minor change. If that is still a concern then can this please make it into the upcoming 3.0.0 release?

othiym23:
> If that is still a concern then can this please make it into the upcoming 3.0.0 release?

It most definitely is still a concern, and for that reason `npm@3` (which is really very close to completion – see #7774 for work remaining) is too soon to release this change. A message could be added to warn for top-level packages using prepublish saying that the behavior will change (although I need to decide what, exactly, it's going to change to – the current behavior is _not_ the same as running the same hooks for `install` and `prepublish`, as @iarna helped me understand), but this is  a change that will need to be trailed for a longer period than the few weeks to ~month remaining before `npm@3`'s release.

mtscout6:
If npm3 hasn't been released then I don't see why it's too late for this. Are you planning a v4 release soon after v3?

othiym23:
If npm is to deprecate or change behavior that packages rely upon, the change must be trailed in documentation and / or warning messages for at least a full major release cycle or 6 months prior. This may seem like a straightforward fix to the people on this thread, but I already know it's going to cause pain.

david0178418:
@othiym23 Great to hear this is still on the radar.  I'm looking forward to seeing the timeline for this change set.

dperetti:
+1...

flockonus:
afaik, npm is the best package manager for any platform, in a big part because it's so intuitive to use and powerful.
This particular behaviour is completely counter intuitive, and despite being mildly infuriating every time i `npm i` and slowing down my test pipeline by a lot, i don't see myself moving away from npm just because of it.
[...]

Issue #9721, opened 2015-09-23
--literally flag.


dominictarr:
when a command is invoked with the --literally flag enabled (may also be configured in your local npm configuration) then all commands will be taken literally and performed exactly as you asked them, without npm trying to guess what you meant or do you any special favors.

for example, `npm install --literally` would install dependencies, but not run the prepublish script.

dominictarr:
this would resolve the long standing issues
 
https://github.com/npm/npm/issues/3059
https://github.com/npm/npm/issues/9590
etc

danmactough:
I `npm install --ignore-scripts` does what you want, here.

dominictarr:
I don't want to ignore all scripts, just ignore the scripts that i didn't literally ask for. so literally should still run preinstall postinstall etc

danmactough:
Fair point

sindresorhus:
:+1: Alternative name: `--fix-all-the-things`.

dominictarr:
--literally would also make npm use only `.npmignore` and not look in `.gitignore`

PR #9722, opened 2015-09-23
Add lifecycle script that actually runs before (and only before) prepublish

dominictarr:
This PR adds a lifecycle script that runs before (and only before) publish.
This PR closes https://github.com/npm/npm/issues/8402
This PR does not break legacy behavior of `prepublish` script.
If merged, users who need a lifecycle script that actually runs before prepublish may opt into this change
by installing the latest npm and adding an `"actuallyprepublish"` script to their `package.json`.

If there is a better way to implement this change, please advise.
If you would merge this please let me know and I'll add documentation.

This PR supercedes this feature request https://github.com/npm/npm/issues/9721

There is a bit of a bikeshed about what this script should be called, I settled on `actuallyprepublish` but I considered `prepublishliterally` and `pregoddamnpublish`. Any other suggestions are welcome.

### DIFF
--- a/lib/publish.js
+++ b/lib/publish.js
@@ -71,6 +71,7 @@ function cacheAddPublish (dir, didPre, isRetry, cb) {
     var cachedir = path.resolve(cachedPackageRoot(data), 'package')
     chain(
       [
+        [lifecycle, data, 'actuallyprepublish', cachedir],
         !didPre && [lifecycle, data, 'prepublish', cachedir],
         [publish_, dir, data, isRetry, cachedir],
         [lifecycle, data, 'publish', didPre ? dir : cachedir],


ForbesLindesay:
:+1:

samccone:
:+1: would be highly useful on many of my projects :)

Fishrock123:
This would be great. I've wanted to script actions that run prior to publish only on npm modules before but was unable to do so.

iarna:
@othiym23 has been promising to drop by this issue…

But till then, https://www.npmjs.com/package/in-publish let's you do exactly this today without new lifecycles or backwards compatibility issues.

dominictarr:
@iarna thanks I am aware of in-publish, but reading it's code i see it's more complicated than my PR, and relies on a obscure npm api (i didn't know that the config is passed to the as JSON in an env var!)

I guess I'd be okay with in-publish (although that solution is uglier) if that was officially supported _by npm_.

Actually, I discussed this with @othiym23 at andyet conf the other day, he said he wouldn't merge this, because he did not want to add (more) cruft to npm, and that he was committed to (eventually) fixing this by changing the prepublish behavior. That is what i recall - I'll let @othiym23 correct me though.

I see his point of view, but i do disagree. Yes it is cruft, but npm already has _lots_ of cruft. does it really hurt to have a little more, carefully considered cruft? If we graph cruft over time and measure the area under the graph, how do we minimize total cumulative cruft? I argue that a new script command is that minimal cruft solution. 
offically supporting in-publish is a bit more crufty, in my opinion, but i'll reluctantly accept that otherwise...

Fishrock123:
RE: cruft, which node code also very carefully considers:

I always state that things that people need and are painful for users to make in userland probably belong in core. (Usually c++ things in our case.)

Considering this has been painful to people for literally years, and has only recently been solved by an employee of npm by using obscure workarounds, I personally think that fits the bill for taking it in.

othiym23:
Closing in favor of #10074. I hope the discussion in #10074 accurately captures the in-person discussion that you and I had about prepublish, @dominictarr. Thank you for that conversation and for putting this together, Let me know over there if you think I captured it incorrectly.

PR #9733, opened 2015-09-23
run `build` not `prepublish` script before `npm install <noargs>`

ahdinosaur:
hey. o/

[everyone knows](https://terrytao.wordpress.com/2011/04/07/the-blue-eyed-islanders-puzzle-repost/) that [`prepublish` before `npm install <noargs>` is a misfeature](https://github.com/npm/npm/issues/3059#issuecomment-58283393), we honestly should [fix it in some upcoming release](https://github.com/npm/npm/issues/8402) (4.0.0?).

here's a strawman for a particular implementation based on:

[[0](https://github.com/npm/npm/issues/3059#issuecomment-12273443)]

> @dfellis: I use `prepublish` to actually take care of chores that need to be done
> prior to publishing, such as generating a minified version for
> browsers, building documentation from the source code, making the
> commit for these things and tagging that commit with the new version
> number.

[[1](https://github.com/npm/npm/issues/3059#issuecomment-60379536)]

> @jussi-kalliokoski: When you think of prepublish as build, things start making sense. For
> example:
> - Travis CI default setup is `npm install && npm test`, so basically the
>   idea is that after install, your package is ready, then you test it.
> - You download the sources of a service, then to get it up and running,
>   you run `npm install && npm start`.
> - You scp the sources of a node module from somewhere, then you go into
>   the directory, you run `npm install` and you expect it to be ready to used
>   from outside.
> 
> So basically, the idea is that `npm install` without arguments gives you
> the same result as installing the package from the npm repositories,
> where it has already been built.

### DIFF
--- a/lib/cache/add-local.js
+++ b/lib/cache/add-local.js
@@ -5,6 +5,7 @@ var chownr = require('chownr')
 var pathIsInside = require('path-is-inside')
 var readJson = require('read-package-json')
 var log = require('npmlog')
+var chain = require('slide').chain
 var npm = require('../npm.js')
 var tar = require('../utils/tar.js')
 var deprCheck = require('../utils/depr-check.js')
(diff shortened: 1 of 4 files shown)


othiym23:
Closing in favor of #10074, which I think has a solution that will work for everyone.

BangLy:
[LICENSE.txt](https://github.com/npm/npm/files/1147519/LICENSE.txt)

PR #9741, opened 2015-09-23
When normalize-package-data throws, capture that as a warning

iarna:


iarna:
Reviewed in person by @othiym23: 🐑

iarna:
Landed in 3.3.5!

PR #9743, opened 2015-09-24
Release 3.3.5 updates

iarna:
Module updates for 3.3.5

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -127,6 +127,7 @@ npm-debug.log
 /node_modules/underscore.string/
 /node_modules/underscore/
 /node_modules/unicode-length/
+/node_modules/win-spawn/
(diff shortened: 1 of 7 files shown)


iarna:
3.3.5 is released

jrit:
I believe in the release notes where it says Windows users should `npm install -f -g npm` it should say `npm install -f -g npm@v3.3.5` or you may just end up with the same version that is already installed. This was my experience anyway.

iarna:
Well, it would get you `npm@3.3.4` which is the current latest release. If you actually want the current testing release then yes, `npm@3.3.5` or `npm@next`.

jrit:
Yeah, I should probably look at the tags on the releases :smile: - thanks

Issue #9695, opened 2015-09-21
 npm v3.3.3: npm can no longer self-update


CKarper:
npm won't update to v3.3.4. This is the output I get:
```
d:\>npm i npm -g
npm ERR! Windows_NT 6.1.7601
npm ERR! argv "C:\\Program Files\\nodejs\\node.exe" "C:\\Users\\ckarper.DEV\\AppData\\Roaming\\npm\\node_modules\\npm\\bin\\npm-cli.js" "i" "npm" "-g"
npm ERR! node v4.0.0
npm ERR! npm  v3.3.3
npm ERR! code ENOSELF

npm ERR! Refusing to install npm as a dependency of itself
npm ERR!
npm ERR! If you need help, you may report this error at:
npm ERR!     <https://github.com/npm/npm/issues>

npm ERR! Please include the following file with any support request:
npm ERR!     d:\npm-debug.log
```

I've linked the [npm-debug.log here](https://gist.github.com/CKarper/dd27c1fb40ba068849c6).

othiym23:
Marking as a big bug because we don't want to leave npm in a state where it can't update itself for very long. This is one for @iarna's attention.

iarna:
Hey you all, this landed in 3.3.5! You can work around it in the time being by passing `-f` to install, to tell npm to ignore the "error" that it found. (eg `npm i npm -g -f`)

gdibble:
@iarna good tip, thanks :+1:

JessicaR:
Thanks @iarna - really saved me some time!

Sogl:
@iarna Many thanks! 
It works on Windows too.

jorrit:
I guess this issue can be closed.

colorgap:
@iarna -f worked. Thanks

charlescarrari:
@iarna Thanks! :)

ShepelE:
@iarna Thanks from me too! :)
this works also for  
sudo npm install express --save - npm WARN install Refusing to install express as a dependency of itself

annliz:
@iarna thanks a lot ! it did the trick !

kayGweb:
@larna This was a awesome tip thanks!!!!!!

xgqfrms-GitHub:
npm i npm -g -f

# --force

xgqfrms-GitHub:
**https://github.com/npm/npm/issues/14433**

# Error  
[...]

PR #9744, opened 2015-09-24
Stop inferring package names in directories w/o package.json files

iarna:
This fixes: https://github.com/npm/npm/issues/9695

This fixes that issue because, on Windows, your globals directory is a `$APPDATA/npm`. And of course, your globals directory doesn't have a package.json because's not a package. So previously we were intuiting its name, from directory name, which was `npm`, and then we tried to install `npm` in and and "eek, those look identical!!"


iarna:
Reviewed in person with @othiym23: :sheep:

iarna:
Landed in 3.3.5!

PR #9819, opened 2015-10-02
(1) seemed unnecessary

didyouknowthaticantwritecode:


othiym23:
That's a convention used in UNIX documentation that indicates what man section that doc is in. As such, we'll be keeping it, but thanks for putting this together!

PR #9820, opened 2015-10-02
doc: correct `bindings.gyp` to `binding.gyp`

kenany:
`node-gyp` looks for `binding.gyp`.


KenanY:
My last doc commit was amended before landing to use the subsystem `docs:` but since `doc:` is more popular I will continue to submit PRs with it :)

``` sh
$ git log --oneline doc/ | awk '{print $2}' | sort | uniq -c | sort -nr
     83 doc:
     64 Add
     36 Update
     29 Document
     27 docs:
```

iarna:

PR #9835, opened 2015-10-03
install: Update --only option to install only the argument env regardless of the NODE_ENV

### LINKED ISSUE #9463: CLI flag `--only` is overridden by `NODE_ENV`
# Background
Here's my understanding of how this new feature should work: npm@3.3.0 includes a new CLI flag `--only={prod[uction]|dev[elopment]}`. When this flag is used, only the `dependencies` or `devDependencies` listed in _package.json_ should be installed (and `devDependencies` will not be installed recursively :+1: ). 

This behavior was arrived at in issue #9024.

Here is what the docs say about the `--only` flag:
> The --only={prod[uction]|dev[elopment]} argument will cause either only devDependencies or only non-devDependencies to be installed.

# Problem
Using the `--only=dev` flag when `NODE_ENV` is set to _production_, only the _non-devDependencies_ are installed.

# Solution(s):
Just some ideas for more intuitive behavior (in order of preference):

1. Make it so that the `--only` flag trumps the environment variable (haven't looked at how this could be implemented).
2. If the `--only` flag is used when `NODE_ENV === 'production'` display a warning, or exit with error.
3. Update the docs to explain that `NODE_ENV` trumps the `--only` flag.

# Test-case

## Use this _package.json_:
```
  {
    "dependencies": {
      "lodash": "^3.10.1"
    },
    "devDependencies": {
      "jshint": "^2.8.0"
    }
  }
```
## Test `--only` behavior with  `NODE_ENV` as _development_ (works as expected): 
```
  export NODE_ENV=development
  npm i --only=dev
  npm i --only=prod
```
## Delete _node_modules_, and re-test with `NODE_ENV` as _production_:
```
rm -rf node_modules
export NODE_ENV=production
npm i --only=dev
```
### Expected output
```
npm WARN prefer global jshint@2.8.0 should be installed with -g
npmtest@1.0.0 /coins/npmtest
└─┬ jshint@2.8.0
  ...
```
### Actual output
```
npmtest@1.0.0 /coins/npmtest
└── lodash@3.10.1
```

watilde:
Hello! I just try to fix #9463. Here is the condition of setting the value of the development / production option.
## npm install
- dev: true
- prod: true
## --only=dev
- dev: true
- prod: false
## --only=prod || NODE_ENV=production
- dev: false
- prod: true

Thanks.

### DIFF
--- a/doc/cli/npm-install.md
+++ b/doc/cli/npm-install.md
@@ -273,7 +273,7 @@ The `--nodedir=/path/to/node/source` argument will allow npm to find the
 node source code so that npm can compile native modules.
 
 The `--only={prod[uction]|dev[elopment]}` argument will cause either only
-`devDependencies` or only non-`devDependencies` to be installed.
+`devDependencies` or only non-`devDependencies` to be installed regardless of the `NODE_ENV`.
 
 See `npm-config(7)`.  Many of the configuration params have some
 effect on installation, since that's most of what npm does.
--- a/lib/install.js
+++ b/lib/install.js
@@ -201,7 +201,7 @@ function Installer (where, dryrun, args) {
   this.noPackageJsonOk = !!args.length
   this.topLevelLifecycles = !args.length
   this.npat = npm.config.get('npat')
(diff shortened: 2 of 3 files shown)


iarna:
Looks good to me!

iarna:

PR #9870, opened 2015-10-07
tmatch@1.0.2

watilde:
I've noticed a diff has come when I run `make link` in the local npm directory.

```
$ pwd
/Users/watilde/Development/npm

$ make link
...


$ git status
On branch master
Untracked files:
  (use "git add <file>..." to include in what will be committed)

    node_modules/tmatch/

nothing added to commit but untracked files present (use "git add" to track)
```

Additional information:
- All log of running `make link` in local npm directory is here: https://gist.github.com/watilde/2fb8f1804a03c0a5a555
- tmatch was added in tap@2.0.0: https://github.com/isaacs/node-tap/commit/f8da831750b5010e67d3b627e3d02aec247b3fc3

Thanks.


iarna:
Thanks for this– what we actually want to do here is add `tmatch` to the `.gitignore`. (We don't bundle development deps, and as you noted this is a dep of tap's.) I'll be doing that with this week's release along with a bunch of others that have crept in over the past week.

watilde:
I was not sure about the manegement of deps and now I got it! Thanks.
Also :+1: https://github.com/npm/npm/wiki/npm-maintained-dependencies

PR #9879, opened 2015-10-07
shrinkwrap: do not fail on optional dependencies

mantoni:
If an optional dependency was not installed, shrinkwrap fails with a message saying that a required dependency is missing. This patch checks whether a missing dependency was listed in the optionalDependencies.

I was seeing this with `fsevents` not being installed as a dependency of `chokidar`. I made sure the test case produces the exact same error and that the patch fixes it.


mantoni:
Fixed the stylistic issues which the build complained about. It's all shiny green now, except for the Node 0.8 build which has some issues with the install process. I can't do anything about this one, I guess.

@iarna @zkat Is this an acceptable solution? Is there anything else that should be improved?

iarna:
Yeah, this seems like the right thing to do. I'm adding this to tomorrow's release.

iarna:

mantoni:
Oh cool, that went into a release really quickly! Thanks.

PR #9880, opened 2015-10-07
doc: remove unused --tag for install

Swaagie:
Discussed this with Aria over IRC a few days ago. Encountered a snippet in the CLI install documentation that mentioned `--tag` to overrule any semver targets. However the functionality does not seem to be available atm. 

What works: `npm install package@tag`
What doesn't seem to work, but is implied from the docs: `npm install package@semver --tag=something`. Also, if no semver is specified it will still install latest, even if the tag specifies some other available release.

From what I got from checking source code , it feels changing around [the if logic here](https://github.com/npm/npm/blob/master/lib/cache/add-named.js#L57-L66) could make it possible, but also might result in more issues than solving stuff, so for now it might be best to just remove the docs snippet.

### DIFF
--- a/doc/cli/npm-install.md
+++ b/doc/cli/npm-install.md
@@ -242,10 +242,6 @@ For example:
 
     npm install sax@">=0.1.0 <0.2.0" bench supervisor
 
-The `--tag` argument will apply to all of the specified install targets. If a
-tag with the given name exists, the tagged version is preferred over newer
-versions.
-
 The `--dry-run` argument will report in the usual way what the install would
 have done without actually installing anything.
 
--- a/doc/cli/npm-install.md
+++ b/doc/cli/npm-install.md
@@ -242,6 +242,10 @@ For example:
 
     npm install sax@">=0.1.0 <0.2.0" bench supervisor
 
+The `--tag` argument will apply to all of the specified install targets. If a
+tag with the given name exists, the tagged version is preferred over newer
+versions.
+
 The `--dry-run` argument will report in the usual way what the install would
 have done without actually installing anything.
 
--- a/lib/cache/add-named.js
+++ b/lib/cache/add-named.js
@@ -255,8 +255,7 @@ function addNameRange (name, range, data, cb) {
     // if the tagged version satisfies, then use that.
     var tagged = data['dist-tags'][npm.config.get('tag')]
     if (tagged &&
-        data.versions[tagged] &&
-        semver.satisfies(tagged, range, true)) {
+        data.versions[tagged]) {
       return addNamed(name, tagged, data.versions[tagged], cb)
     }
 


iarna:
As weird as the `--tag` option _is_, it's supposed to be supported (there's lots of code to support it) and so I'd kind of prefer to see this as either a patch to remove that code or a patch to fix the bug.

It's not clear immediately where it's going wrong, but... the following should probably be changed to fetch that config var:

* https://github.com/npm/npm/blob/v3.3.6/lib/cache/add-named.js#L268
* https://github.com/npm/npm/blob/v3.3.6/lib/outdated.js/#L376

In fact, for me it works with my test case:

```
npm install -g --tag=latest-2 npm
```

Does indeed install the latest npm@2 release.

My suspicion is that you tried this with a module that didn't have any non-prerelease versions? If that were the case you'd be burned by the `add-named` code above.

iarna:
Oh, but I see...

```
npm install -g --tag=latest-2 'npm@>2'
```
Picks the wrong version... veeery interesting.

I don't know why, but that's definitely a bug.

Swaagie:
Thank you for doing the deep dive into this. Afaik the prerelease/pre-latest version was there and tagged to `prod` (in my case). I just tried the `latest-2` and that indeed works as expected. However `npm install minimize@prod` (prod@1.6.1 vs latest@1.7.2) is installing 1.7.2. Also tried to tag minimize@0.9.0 as prod, still will install 1.7.2, what is the difference we are looking at here?

Swaagie:
Found the root cause, but not sure if it can simply be removed. https://github.com/npm/npm/blob/master/lib/cache/add-named.js#L259 will only install tags if within range, adding a additional restriction. The same restriction doesn't seem to apply to `addNameTag` https://github.com/npm/npm/blob/master/lib/cache/add-named.js#L86-L87

Swaagie:
Reverted the doc changes and removed the line, most tests (some unrelated? are not) seem to pass on Travis, but can't seem to get them to work locally

iarna:
`addNameRange` is what's used when you specified a version-range – it will have an upper bound of whatever was specified by `--tag`. By contrast, `addNameTag` is only used when you requested a tag by name, eg `npm install foo@beta`, and therefore it quite reasonably ignores `--tag`.

iarna:
As best as I can tell, https://github.com/Swaagie/npm/commit/b9f16256a28c8cf9d21685bd7065feb53283a8a0 would make it ignore ranges entirely if the tag exists, so you'd _always_ get what `--tag` was or `latest` even if you specified an range that should match something less than that.

iarna:
Further, as this is happening in npm@3 but not in npm@2, it's unlikely to be in the cache code, which mostly didn't change. I'd be most suspicious of https://github.com/npm/npm/blob/master/lib/fetch-package-metadata.js#L117-L139 but I don't see how that code could be failing this…

Swaagie:
How to run the tests though? Not having the ability to check chances or confirm changes with tests is making this a bit painful to contribute.

Swaagie:
Btw I don't think the conclusion: that this works for npm@2 and only is broken in npm@3 is correct. It fails for both versions for me.

othiym23:
@Swaagie You should be able to run `npm run test-all` and get a successful test run on your local instance – the tests all pass for the CLI team developers on Node 0.12 and 4.2.1.

Swaagie:
So this is on latest master, with everything freshly installed, node@4.2.1 and osx@10.10.5

```
ok 22 npm test -ddd

+npm test -ddd
$$$$$$ cd /var/folders/wv/s_txn2293r9f3f807cb_b9yr0000gn/T/npm-test-18711/root/lib/node_modules/npm-test-bundled-git; PATH=/Users/mswaagman/.nvm/versions/node/v4.2.1/bin:/var/folders/wv/s_txn2293r9f3f807cb_b9yr0000gn/T/npm-test-18711/root/bin:/var/folders/wv/s_txn2293r9f3f807cb_b9yr0000gn/T/npm-test-18711/root/node_modules/.bin:/Users/mswaagman/.nvm/versions/node/v4.2.1/lib/node_modules/npm/bin/node-gyp-bin:/Users/mswaagman/projects/npm/node_modules/.bin:/Users/mswaagman/.nvm/versions/node/v4.2.1/lib/node_modules/npm/bin/node-gyp-bin:/Users/mswaagman/projects/npm/node_modules/.bin:/Users/mswaagman/.nvm/versions/node/v4.2.1/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/opt/X11/bin:/usr/local/MacGPG2/bin:/Users/mswaagman/bin:/usr/local/mysql/bin:/usr/local/sbin /var/folders/wv/s_txn2293r9f3f807cb_b9yr0000gn/T/npm-test-18711/root/bin/npm test -ddd
$$$$$$ after command /var/folders/wv/s_txn2293r9f3f807cb_b9yr0000gn/T/npm-test-18711/root/bin/npm test -ddd /var/folders/wv/s_txn2293r9f3f807cb_b9yr0000gn/T/npm-test-18711/root/lib/node_modules/npm-test-bundled-git
not ok 23 npm test -ddd
1..23
/Users/mswaagman/projects/npm/test/run.js:192
  if (er) throw er
          ^

Error: failed npm test -ddd
    at /Users/mswaagman/projects/npm/test/run.js:103:10
    at exithandler (child_process.js:210:5)
    at ChildProcess.errorhandler (child_process.js:217:5)
    at emitOne (events.js:77:13)
    at ChildProcess.emit (events.js:169:7)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:198:12)
    at onErrorNT (internal/child_process.js:344:16)
    at doNTCallback2 (node.js:439:9)
    at process._tickCallback (node.js:353:17)
```

Swaagie:
Confirmed they work after removing my code change, so that has rather big effects. Not sure if that was the culprit before, but now I can investigate with confidence :)

iarna:
I'm going to make an issue for the work related to this and close this PR with a reference when I do that.

iarna:
Closing in favor of #11038

Swaagie:
Ok :) thanks for keeping track of this

PR #9890, opened 2015-10-07
Refactor all the code dealing with module/package names

iarna:
The recent change to `read-package-tree` to set `package.name` if it wasn't already set had broader fallout than I'd originally realized. This PR includes a refactoring of all of the places unguarded uses of `package.name` existed to use a set of standardized and tested functions.


zkat:
:sheep:

PR #9891, opened 2015-10-07
rebuild bundles via npm3 mechanisms

iarna:
Previously this was being done implicitly via `build.linkStuff`– we're now avoiding that.

Not doing this meant that failures of optional deps during those rebuilds wasn't catchable as an optional failure and would kill the entire install.  My test case for this was `npm install -g karma`. For reasons I'm not super clear on, it didn't do this for non-global installs.  Regardless the new way of rebuilding is the right way.

Similarly, I'm pretty sure that this was responsible for needless rebuilds when running `npm install` on already installed trees that have bundled deps.


zkat:
:sheep:

PR #9892, opened 2015-10-07
Remove dead code around npm.ROLLBACK global

iarna:
`npm.ROLLBACK` was used in npm@2, but is no longer used at all in npm@3.


zkat:
:sheep:

PR #9903, opened 2015-10-08
[docs] Add documentation for npm install --prefix

nikolas:
This is a useful option that I couldn't find documentation on.

### DIFF
--- a/doc/cli/npm-install.md
+++ b/doc/cli/npm-install.md
@@ -13,7 +13,10 @@ npm-install(1) -- Install a package
     npm install <folder>
 
     alias: npm i
-    common options: [--save|--save-dev|--save-optional] [--save-exact] [--dry-run]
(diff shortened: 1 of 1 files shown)


othiym23:
Thanks for putting this together! We're not going to take this patch, but I appreciate you spending the time to write it. `--prefix` is documented [here](https://docs.npmjs.com/misc/config#prefix), which is maybe not so easy to find (a problem that @ashleygwilliams is now working to resolve). If you read the second sentence there, it says

> If set on the command line, then it forces non-global commands to run in the specified folder.

So it's not that it's telling npm where to install the package, but where npm should be performing its actions relative to. The description in this PR is misleading about what's going on, and is likely to confuse people as a result.

That said, we're actively looking at finding ways to better get the more generic config documentation closer to the places where it will be useful to users, and this PR is in the spirit of that effort, so I really do appreciate you taking the time to send this patch along.

PR #9920, opened 2015-10-09
add size when logging installation results

AntouanK:
- add a small recursive function to get the size of a directory ( in Bytes )
- add the size of a package when printing the installation results
- test manually on node 0.10.x 0.12.x 4.x
- add "pretty-ms" dependency
#### How it looks like

screen shot 2015-10-09 at 12 54 02
#### Why

I don't know whether it was discussed before, but I believe that logging the size of the packages you install would be very useful, and it would also make users/authors more aware of the size of their libs and their dependencies. 
Because most projects out there get their npm packages size in very high numbers and they are not aware until it becomes a problem at some stage of the pipeline. And then it's too late to start removing stuff.

Also linux package managers ( like apt-get ) do something similar, by printing the size of the file they will download, and the size that it will occupy on the disk.
#### Anything else?

Well, mainly I want to know whether it's a feature you want to add and whether I've done something wrong so far.

I guess I could extend my PR to also capture the size of the tar files it will download, and maybe print a total at the bottom to quickly check the total size in a big list.
I could use some help on where to look/tap into, since I just did that PR in a couple of hours today, and I'm not aware of the codebase.
And I also have to add tests and probably check how it works on other OS.

Thanks.


othiym23:
Right! I was going to give feedback on this two weeks ago! I have a piece of paper with my notes on it that I've been carrying around since then. My apologies for not transcribing that piece of paper onto this PR before now!

The idea is appealing, but this needs refinement:

1. First off, the fact that it's missing tests, and that the existing tests don't pass, is a pretty big red flag. `npm ls` is actually pretty close to a core piece of machinery in the npm installer, so anything that changes its behavior needs pretty careful testing.
2. The current implementation has some pretty major performance considerations because it's calculating the sizes independently of the install process, which means another pass over the filesystem. Since performance of `npm install` is already something of a sensitive issue, coming up with a way to piggyback the size-gathering on the back of the unpacking or moving from staging into place processes would be extremely helpful.
3. The mapping of sizes to colors feels arbitrary and not very helpful. Packages have sizes, and it's really going to depend on the application whether a set of dependencies is "big" or "small". I'd ditch that part.
4. The size information should also go into the `--json` and `--parseable` output, which means that it should probably be gathered in `read-package-tree` so it can just be passed as part of the result object that comes out of reading the package tree (or something like that; see #2 for some complicating factors).
5. This should be opt-in, mostly because of #2 and #4, and because it's otherwise a semver-major breaking change. The format of the `npm ls` output is an important part of npm's contract to users, and as such has semver implications. For starters, maybe adding it to the `--long` output works, or behind a `--with-sizes` option. If it's the latter, people who want this can run `npm config set with-sizes true` and get the behavior every time (although that might also cause packages that depend on the `npm ls` output looking a certain way to break).

Like I said up top, this is an interesting idea, and a good start! Thanks for putting it together! Because it's changing one of npm's key features, it's going to need a little extra love. Let me know if any of the preceding is unclear.

AntouanK:
@othiym23 Thanks a lot for the feedback. I know you guys are busy with more important stuff.

To your points:
1. 
I left the test out, because I didn't know whether that feature is useful or not, and more importantly I didn't actually know where to start from...
So, how would I go making a test for this?
As in, where to put it in the `/test` directory, and what exactly am I testing for? ( can we specify what we expect from this feature? Is the `pretty-bytes` output/dependency acceptable? )

2.
I'm sure there's a more performant way to do this. But I had a hard time figuring out the flow of things happening when someone does `npm install`. Maybe my code style is different so I cannot read the codebase easily.
I did the whole thing in a couple of hours anyway. I'll try to give it more time now.

Can you point me to where you actually get the tar file ( so we can print to the user the size that will need to be downloaded ) and where you unpack it? ( so I can get from there the uncompressed final size )

3.
That was helpful for me to quickly spot "heavy" packages. But like you said, it's subjective. I'm fine with skipping it for now.

4. 
I'll have to read the codebase more to understand what you mean exactly. Never used those flags myself, so...

5.
Makes sense. I'm fine with a flag. And if I can configure it for me to be always on, even better.
Any small hint on where to add the flag? ( that would be a good point to start my next commit )


Thanks for taking the time to have a look at this.

othiym23:
I've had a chance to review this again, and it's getting there! As you've probably noticed by now, this is actually a pretty good-sized new feature, and there's a lot of things that need to be chased down, but this is looking a lot closer to something that's ready to land. Hang in there! And thanks for your patience thus far!

I have some more questions and comments for you:

1. This will need to have `pretty-bytes` checked in under `node_modules/` (preferably installed with `npm@2`, because of issues that some old versions of npm have upgrading directly to `npm@3` if `npm@3` is used to package things for `npm@2`).
2. The tests mostly pass, but when I use this version of npm to install itself, I end up with a bunch of modules missing `package.json` files, the installed copy of `minimatch` marked as extraneous, and `npm ls` no longer works because `ls.fromTree()` is now making some incorrect assumptions about what types the arguments are. It seems likely that your changes got out of sync with some recent work @iarna's been doing in there, so rebase and update?
3. The more I look at the proposed change, the more it seems like the size-calculation functionality belongs in either `npm ls`, or perhaps in the npm dependency `read-package-tree`. `--with-sizes` is a very useful option for `npm ls`. See also my point #4 above -- if the sizes are coming out of `read-package-tree`, then it's easy to include them in the `--json` and `--parseable` output. I'm happy to work with you to get the relevant changes and tests made to `read-package-tree`.
4. A good test would be to use either `npm-registry-mock` or static fixtures created from within the tests (of which there are many examples in `test/tap`) to create a prepopulated tree that you can then install as a local package, both with and without `--with-sizes` to ensure that it gives the expected output in both cases. Bonus points for including tests for the output of `--json`.
5. It's not a big deal, but there was a change in your patch that didn't pass the part of npm's tests that run `standard` over the tree -- an errant blank line in `lib/ls.js`. `npm run test-all` is the gold standard for figuring out if the tests are in good shape.
6. `--with-sizes` and all of the stuff you've set up for documenting it and setting its defaults is perfect (except the docs should say "its size" rather than "it's size", but that's quick to fix).

> But I had a hard time figuring out the flow of things happening when someone does npm install. Maybe my code style is different so I cannot read the codebase easily.

Don't blame yourself for this -- the npm code base is gradually getting more conventional, but it's still an idiosyncratic body of JavaScript, and it takes everyone who deals with it a while to understand how all the pieces are put together.

> Can you point me to where you actually get the tar file (so we can print to the user the size that will need to be downloaded) and where you unpack it? (so I can get from there the uncompressed final size)

* [fetching the tarball](https://github.com/npm/npm/blob/master/lib/cache/add-remote-tarball.js#L56-L122)
* [unpacking the package once it's in the cache](https://github.com/npm/npm/blob/master/lib/cache.js#L331-L337)
* you should also check out [`lib/fetch-package-metadata.js`](https://github.com/npm/npm/blob/master/lib/fetch-package-metadata.md), which the `npm@3` installer uses to coördinate dealing with the registry. (Arguably it should be its own package, but @iarna and I are still trying to figure out what the best factoring for all those packages is.)

Unfortunately, right now we only know the tarball size after it's been downloaded. The best way to figure out the compressed and expanded sizes is probably to add passthrough streams to both the download and install code that tallies the total size of all the chunks passed through them, although again that may be something better handled by pushing the code into one of the subsidiary modules like `read-package-tree` that's already reading the package tree, and could be fairly easily extended to calculate the installed tree's size.

Again, it's looking pretty good, and thanks for your patience and all the hard work you've put into this so far! Let me know if there are any further questions I can answer.

AntouanK:
@othiym23 Thanks for the feedback!
I'll go through all those points, and I'll try to do separate commits fixing/working on each one of them.

I'll ping you once I'm done so you can have a look.

davej:
Looks like a great feature, a couple of comments:

1. `npm ls --with-sizes` would be essential for sure.

2. Should module sizes not include their dependencies sizes? Or at least should there be a way of getting that info? My most common use-case for this feature would be `npm ls --depth=0 --with-sizes`. I would expect that this would list the modules size including dependencies size. I have no control of sub-dependencies so it's more important for me to check the sizes of level 0 deps. This is especially important with flat deps in npm3 because I can't go into the node_modules directory and inspect a module's folder to see it's total size.

3. For easier comparison of module size it might be worth considering normalizing the output (to `x.xxxMB` for example) and right-aligning. Depending on the length of the line this may negatively effect tracking though.

othiym23:
> Should module sizes not include their dependencies sizes?

@davej, I guess my question for you is, what would you do with this information if you had it? Decide not to use the dependency? File an issue on it urging the maintainer of the other issue to slim down their dependencies? Knowing how you want to use it helps me figure out what the default behavior should be here. (I generally use `du -sh <dirname>` a ton, mostly when I'm trying to find directories to nuke).

> For easier comparison of module size it might be worth considering normalizing the output (to `x.xxxMB` for example) and right-aligning.

I personally prefer the current automatic scaling of sizes to units provided by `pretty-bytes`. If this were to be made configurable, I think I'd prefer a kilobytes option (again, like `du -sk`) to megabytes / [mebibytes](https://en.wikipedia.org/wiki/Mebibyte).

davej:
@othiym23 

> what would you do with this information if you had it? Decide not to use the dependency? File an issue on it urging the maintainer of the other issue to slim down their dependencies?

Yeah, either of those is a possibility and I have swapped packages that were too heavy in the past. I'm mainly using Electron so I'm probably more sensitive to this issue than the average node user.

My use case:

In Electron it's important to have a handle on the size of modules so that you can keep the final binary as small as possible. I'm trying to keep my app <100MB zipped (Electron by itself is ~40MB). I switched to npm3 a few weeks ago and this has made it more difficult to keep track of the first-level dependency sizes. With npm2 I could just do `du -sh` on the module, that's not as useful with flat deps because it doesn't include the sub-dependencies. So what I've ended up doing is keeping a mirror of the project that uses npm2, I'm doing this simply so I can keep a handle on the module sizes.

> I personally prefer the current automatic scaling of sizes to units provided by pretty-bytes. If this were to be made configurable, I think I'd prefer a kilobytes option (again, like du -sk) to megabytes / mebibytes.

I'd be happy with kilobytes as opposed to (mega/mebi)bytes. My main point was to format the output in a way that makes it easier to compare and quickly grok the output.

For example, I find this is much easier to scan and compare:
```
└─┬ standard@4.5.4           2382 kB
  ├─┬ deglo@1.0.1            1139 kB
  │ ├─┬ glob@5.0.15           930 kB
  │ │ ├─┬ inflight@1.0.4       90 kB
  │ │ │ └── wrappy@1.0.11      70 kB
  │ │ ├── inherits@2.0.1      243 kB
```

than this:
```
  └─┬ standard@4.5.4  (2.38 MB)
    ├─┬ deglo@1.0.1  (1.13 MB)
    │ ├─┬ glob@5.0.15  (930 kB)
    │ │ ├─┬ inflight@1.0.4  (90 kB)
    │ │ │ └── wrappy@1.0.11  (70 kB)
    │ │ ├── inherits@2.0.1  (243 kB)
```

AntouanK:
Glad to see there is discussion on this.
Unfortunately I have personal matters to attend to for a few days. 
I should be able to jump back to this PR afterwards that and join the discussion.

Dignifiedquire:
@othiym23 I'm seeing the same use case as @davej, not only specific for `electron` but for example there is an effort to reduce the total install size of karma (https://github.com/karma-runner/karma/issues/1694) and it's very hard now with npm@3 to get reliable details on how large a dependency is (including its dependencies).

othiym23:
@AntouanK Have you had any more time to look at this? It's a neat feature, but the CLI team don't have the bandwidth to pick it up and work on it. It'd be great to get this into good enough shape to merge.

AntouanK:
@othiym23 Got into other things and I forgot that PR. 
Will try to review it and see what else I can commit within the coming week. I'll update you next week.
( seen the comments for alignment which is a good idea, but I doubt I can get this in easily since I can think of many edge cases on that. might go in on a second iteration )

nicbarker:
Hi guys, any updates on this? Would be really cool to have.

AntouanK:
@nicbarker I completely forgot about this, sorry. 
I've seen that npm has already changed a bit, so the PR probably has to start from scratch. I'll try to start it in the weekend if possible.

wayofthefuture:
https://github.com/mikermcneil/kit#kit-deps

iarna:
We no longer display a tree on install. A summary with sizes would be pretty neat, but either way, this PR is pretty obsolete now. I'm going to close this now, but please feel free to open a new PR if you come back to this. Thanks!

PR #9927, opened 2015-10-10
Show available versions as lists instead of JSON

iarna:


iarna:
todo: add ux issue for flow on wide output

iarna:
This landed in 3.3.9

PR #9928, opened 2015-10-10
mark our installable flag as internal

iarna:
This is just a dumb little flag I had in the package.json that should have been polluting the main namespace of package.json files.


iarna:
This landed in 3.3.9

Issue #9520, opened 2015-09-08
Make installing specific modules, eg, `npm install foo`, faster, better


iarna:
### If the module isn't already installed there's no need to:

1) Scan deps for existing modules (I think we already avoid this). But we can even avoid recalculating metadata, because without pruning, we don't need it
2) Prune (removing things on a fresh install is weird anyway)

### If a module _does_ already exist

We would get a much cleaner result if we first removed it from the tree, pruned, and then added it back.

This solves a problem where, given:

```
A@1
B@1 (required by A)
```

`npm install A@2` (which requires B@2), would result in:
```
A@2
    B@2
````

This was happening because A@2 would replace A@1, then it would need B@2, but B@1 would already be in the top level so it will be installed inside A@2.  Then pruning occurs and B@1 is removed, leaving B@2 mysteriously not at the top level.

iarna:
Completed by #9929

Issue #9703, opened 2015-09-22
npm@v3: npm shrinkwrap is broken


fundon:
My package.json is below:

``` json
  "dependencies": {
    "bluebird": "^2.10.0",
    "connect-redis": "1.4.5",
    "decimal.js": "^4.0.2",
    "dotenv": "^1.2.0",
    "ejs": "~0.8.4",
    "file-type": "^2.10.2",
    "fs-extra": "^0.24.0",
    "grunt": "^0.4.2",
    "grunt-contrib-clean": "~0.5.0",
    "grunt-contrib-coffee": "~0.10.1",
    "grunt-contrib-concat": "~0.3.0",
    "grunt-contrib-copy": "~0.5.0",
    "grunt-contrib-cssmin": "~0.9.0",
    "grunt-contrib-jst": "~0.6.0",
    "grunt-contrib-less": "0.11.1",
    "grunt-contrib-uglify": "~0.4.0",
    "grunt-contrib-watch": "^0.6.1",
    "grunt-sails-linker": "^0.9.6",
    "grunt-sync": "~0.0.4",
    "hashids": "^1.0.2",
    "html-pdf": "^1.4.0",
    "include-all": "~0.1.3",
    "ipaddr.js": "^1.0.3",
    "ipware": "0.0.7",
    "is-svg": "~1.1.1",
    "joi": "^6.7.0",
    "jsonwebtoken": "^5.0.5",
    "kue": "^0.9.4",
    "lodash": "^3.10.1",
    "method-override": "^2.3.5",
    "mime": "^1.3.4",
    "moment": "^2.10.6",
    "morgan": "^1.6.1",
    "mz": "^2.0.0",
    "native-or-bluebird": "^1.2.0",
    "node-uuid": "~1.4.3",
    "node-xlsx": "^0.6.0",
    "otp.js": "^1.0.4",
    "passport": "^0.3.0",
    "passport-local": "^1.0.0",
    "pg": "^4.4.1",
    "pg-native": "^1.9.0",
    "pg-promise": "^1.10.5",
    "pingpp": "^2.0.6",
    "pinyin": "^2.6.2",
    "qr-image": "^3.1.0",
    "raw-body": "^2.1.3",
    "rc": "^0.5.5",
    "request-promise": "^0.4.3",
    "sails": "^0.11.0",
    "sails-disk": "^0.10.8",
    "sails-hook-email": "^0.12.1",
    "sails-migrations": "^2.0.7",
[...]

PR #9929, opened 2015-10-10
Refactor upgrades

iarna:
By which I mean, when npm decides it needs to upgrade an existing module, this changes how it does that. This applies to both `npm install foo@newversion` and updating a package.json and running `npm install`. In either case, it does a better job of cleaning out the older deps, which means that the resulting tree is flatter!

Fixes: #9520

### DIFF
--- a/lib/install/deps.js
+++ b/lib/install/deps.js
@@ -133,6 +133,16 @@ function addRequiredDep (tree, child) {
   return true
 }
 
+function removeObsoleteDep (child) {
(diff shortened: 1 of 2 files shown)


othiym23:
See also #9703.

iarna:
I don't feel like this got enough of a review, so I'm pushing it out a week. Plus CI was being weird with it and that concerns me.

iarna:
This landed in 3.3.10

PR #9950, opened 2015-10-12
Encourage users to update node & npm before contacting package maintainers

ForbesLindesay:
This could vastly reduce the e-mail volume to package maintainers.


iarna:
:+1:

iarna:
Good news, this landed in 3.3.10!

PR #9954, opened 2015-10-13
Standard

sebdeckers:
Fixes some minor syntax warnings. :see_no_evil:

### DIFF
--- a/bin/npm-cli.js
+++ b/bin/npm-cli.js
@@ -73,5 +73,4 @@
     if (er) return errorHandler(er)
     npm.commands[npm.command](npm.argv, errorHandler)
   })
-
 })()
--- a/lib/build.js
+++ b/lib/build.js
@@ -133,12 +133,10 @@ function shouldWarn (pkg, folder, global, cb) {
 
     // current searched package is the linked package on first call
     if (linkedPkg !== currentPkg) {
-
       // don't generate a warning if it's listed in dependencies
       if (Object.keys(topPkg.dependencies || {})
           .concat(Object.keys(topPkg.devDependencies || {}))
           .indexOf(currentPkg) === -1) {
-
         if (top && pkg.preferGlobal && !global) {
           log.warn('prefer global', packageId(pkg) + ' should be installed with -g')
         }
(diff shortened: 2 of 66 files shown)


cbas:
Build only fails in Node 0.8 ... but is this related at all to the style change?
https://travis-ci.org/npm/npm/jobs/85043844

(Still trying to get `npm test` to pass on my local machine. Somehow getting a few errors with node 4.1.2 on OSX 10.11 using the latest `master` branch. :sweat_smile: )

iarna:
our build has failed on 0.8 for a long while now– ever since some of our testing related deps dropped 0.8 support =/

cbas:
@iarna Thanks for clarifying. :smile: Hope this PR is good to merge then. I'm wondering though why `npm test` didn't pick up all these warnings when it also runs `standard`. I only found them by opening up the files with Atom's `linter-js-standard`.

othiym23:
@cbas The version of `standard` that npm installs itself as a `devDependency` is an older one that doesn't complain about the issues you cleaned up (this leads to a confusing result if you have the latest `standard` globally installed but haven't run `npm install` against npm itself, because then when you run `npm test` it will produce errors, but won't after `npm install` is run). I was going to take on getting this cleaned up this week myself, so thank you so much for picking this up!

cbas:
Ohh that explains it! Thanks :relieved: 

I've gone and bumped to the latest `standard@~5.3.1` and ran `standard --format` to fix the remaining files (mostly the `test/` folder). HTH.

cbas:
@iarna @othiym23 Should I continue to keep this branch up to date with the latest pushes to master?

iarna:
@cbas We'll take this during this week! If you can update it to match master that would be fantastic, but if not we'll take care of it.

cbas:
@iarna :ok_hand: Updated.

iarna:
Thank you! =)

cbas:
Rebased the commits for your convenience. Thanks for the tip @timoxley

iarna:
Good news and thank you for all this work, this landed in 3.3.10!

Issue #9935, opened 2015-10-10
Nested npm run scripts


cbas:
What are the thoughts on adding support to nest scripts?

Rough idea:
```json
{ "scripts": {
  "test": {
    "tdd": { "js": "mocha --watch" },
    "lint": { "js": "standard", "css": "postcss", "html": "htmllint frontend" },
    "audit": "node-license-validator --deep"
  },
  "dev": "npm run test.tdd & http-server",
  "start": "npm run audit && forever"
} }
```

This would avoid the repetition in property names that I see in many [substack-esque, pseudo-nested packages](http://substack.net/task_automation_with_npm_run).

I am not yet familiar with the `npm run` implementation but happy to take a look and do a PR if this idea is not deemed too terrible. :stuck_out_tongue_winking_eye:

faiq:
I'm a heavy user of npm scripts, and I don't really see myself using this feature. I usually just combine things together the pseudo nested way and its worked just fine.

Also from an implementation standpoint, adding this kind of complexity seems pretty hard.

ehsalazar:
@cbas Thanks for submitting the idea. It looks you received a response via #9970. I'll close this issue out as you'd closed the other as well.

cbas:
@ehsalazar :+1: Thanks!

PR #9970, opened 2015-10-14
Nested scripts

sebdeckers:
Request for feedback. This is a proof of concept for https://github.com/npm/npm/issues/9935

The goal is to eliminate pseudo-nesting of scripts by repeating a prefix. JSON supports nesting, might as well use it.
##### Example

``` json
{ "scripts": {
  "test": "npm run test-lint && npm run test-js",
  "test-lint": "standard",
  "test-js": "mocha"
} }
```

becomes

``` json
{ "scripts": {
  "test": {
    "lint": "standard",
    "js": "mocha"
  }
} }
```

Where `npm run test` recurses into `test` and iterates over `lint` and `js`.

Dot notation to run sub-tasks: `npm run test.lint`. Currently uses [getobject](https://www.npmjs.com/package/getobject) but alternatively [JSONPath](https://www.npmjs.com/package/JSONPath) or [lodash.get](https://lodash.com/docs#get).
##### Concerns

:question: How to handle arrays vs maps? Does order matter?
:question: What about `pre`/`post` hooks? Do they apply only to the deepest property? Or also to all higher level properties?
:exclamation: Did not yet implement recursion or any data sanitisation for this simple proof of concept. Wanted to get some feedback. Afraid it will get shut down hard by core contributors. :cold_sweat:

:poop: Hacks were made to two dependencies: `read-package-json` and `normalize-package-data`. This is to support non-string properties inside `scripts`. Separate pull requests would, of course, need to be made for those packages.

### DIFF
--- a/lib/run-script.js
+++ b/lib/run-script.js
@@ -6,6 +6,7 @@ var path = require('path')
 var readJson = require('read-package-json')
 var log = require('npmlog')
 var chain = require('slide').chain
+var getobject = require('getobject')
 
 runScript.usage = 'npm run-script <command> [-- <args>...]' +
                   '\n\nalias: npm run'
@@ -131,7 +132,7 @@ function run (pkg, wd, cmd, args, cb) {
       'prestart', 'start', 'poststart'
     ]
   } else {
-    if (!pkg.scripts[cmd]) {
+    if (!getobject.exists(pkg.scripts, cmd)) {
       if (cmd === 'test') {
         pkg.scripts.test = 'echo \'Error: no test specified\''
       } else if (cmd === 'env') {
@@ -151,6 +152,10 @@ function run (pkg, wd, cmd, args, cb) {
     cmds = [cmd]
   }
 
+  // todo:
+  // - how does this apply to nested scripts?
+  // - pre/post at every level or only the deepest?
+  // - recurse/iterate over all nested scripts?
   if (!cmd.match(/^(pre|post)/)) {
     cmds = ['pre' + cmd].concat(cmds).concat('post' + cmd)
   }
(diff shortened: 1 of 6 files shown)


othiym23:
Thanks for putting this together! Feedback was requested, and thus here I am. The short version is that the time is probably past for making these kinds of sweeping changes to how commonly-used sections of `package.json` can be formatted.

There are two primary problems here: backwards compatibility and added complexity. The backwards compatibility issue is pretty easy to explain, and probably enough of a deal-killer all on its own to rule this change out: I'm 95% sure the proposed syntax will straight-up cause every version of npm currently extant to explode, probably without a super useful error message. The _best case_ here is that existing versions of npm will try to run "[Object object]" and complain about that not working. Given the number of sites still running `npm@1.3` in production, this introduces a support headache that I don't think we're prepared to take on.

Beyond that, I'm now of the opinion that having polymorphic behavior for fields in `package.json` based on the types of their values is a bad idea, and one we should at least avoid introducing more of in the future. It complicates the package-handling code (which is spread all over the place, as you've now learned – ironing that out is on the CLI team's list of things to do, but it's a major project), it makes it harder to understand, and it's more difficult to concisely and accurately document. So, put another way, it's bad user experience.

I get the urge to make it easier to scale to large number of package scripts, especially given the push within the community to use them to do so many things, but this approach is too difficult to transition to for it to be the way forward. Again, thanks for putting this together, and thanks for framing this as a proposal. I really appreciate the time and effort.

cbas:
@othiym23 Thanks for the feedback. :bow: That makes perfect sense.

Really curious in what the above referenced `npm-run-all` discussion will result. They probably have more freedom, and are a more appropriate place, for sweeping changes to the task runner system.

PR #9982, opened 2015-10-15
make tests always use right node + npm

iarna:
Our test lifecycles regularly run npm commands as a part of them– this patch ensures that they're run with the same npm and node that the original was run.


othiym23:
Reviewed in person: :sheep:

iarna:
This landed in 3.3.10.

Issue #9647, opened 2015-09-18
npm@3: `npm install -S` reverts shrinkwrap with devDependencies


wbinnssmith:
Hi all! Thanks for npm and congratulations on the stable release of npm@3 :) Apologies if this issue is known; I couldn't find anything similar.

I absolutely love the fact that `npm install --save` will update an existing shrinkwrap file if it's there. Unfortunately, it looks like if that shrinkwrap file was generated with `npm shrinkwrap --dev`, npm will re-write the shrinkwrap file with just the production dependencies (and often output a ton of WARN messages that it's excluding devDependencies).

What's interesting is that the shrinkwrap I generated *initially* with `npm shrinkwrap --dev` includes the devDependencies, but they're actually listed under the `dependencies` key. Maybe this has something to do with it — I imagine npm would/could use the presence of a `devDependencies` key as a hint to continue generating a shrinkwrap with devDependencies on events like `npm install --save`.

**tl;dr reproduction steps on npm@3.3.3 and 3.3.4 on a project with both dependencies and devDependencies:**

1. `npm shrinkwrap --dev`
2. the generated shrinkwrap will contain devDependencies inside a "dependencies" key
2. `npm i -S xtend` (or any other module)
3. npm will install the module, but will begin shrinkwrapping without dev dependencies and issue WARNs
4. the resulting shrinkwrap won't include devDependencies that were present after step 1

Thanks and congratulations again!

othiym23:
Yeah, this looks like an edge case we weren't catching. We should probably have some way of checking whether `devDependencies` were included in `npm-shrinkwrap.json`, and preserve them when upgrading the shrinkwrap file. It would also be nice to do this without adding any additional metadata to `npm-shrinkwrap.json` (which might break backwards compatibility for the file), but we'll see how it goes. @iarna, what do you think?

wbinnssmith:
I don't mind tackling this myself, actually, maybe with guidance from @iarna? I can't seem to find any test cases around saving the shrinkwrap post-save, so I'll probably PR that first and we'll go from there. What do you think?

micahr:
This bug introduced some very unexpected behavior when installing new dependencies (namely deleting lots of lines from our shrinkwrap file). Is there any work around while the bug is being worked on, short of going back to NPM 2?

iarna:
This PR didn't make it under last week's release cut off, so it'll be included in this week's release instead.

PR #9992, opened 2015-10-15
If a shrinkwrap already has dev deps, make sure they're kept

iarna:
When running `install --save` in a folder with a shrinkwrap that has
dev-only dependencies in it, it's important that we don't then throw
away the dev dependencies just because they didn't say
`install --save --also=dev`.

Fixes: #9647

### DIFF
--- a/lib/install/save.js
+++ b/lib/install/save.js
@@ -46,9 +46,28 @@ function saveShrinkwrap (tree, next) {
     var save = npm.config.get('save')
     var saveDev = npm.config.get('save-dev')
     var saveOptional = npm.config.get('save-optional')
-    if (!saveOptional && saveDev) return next()
+
+    var shrinkwrap = tree.package._shrinkwrap || {dependencies: {}}
+    var hasDevOnlyDeps = tree.requires.filter(function (dep) {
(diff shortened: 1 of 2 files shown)


othiym23:
Reviewed in person: :sheep:

markstos:
I triggered this issue and tried this branch as a fix, but it didn't work for me with either `npm install --save` or `npm install --save --save-exact`. However, I did notice  regression which appears to be triggered by some other new code. This no longer displays the `install` docs anymore:

```
 npm help install
```

Instead, I get some sort of "search results" report, that starts with:

`Top hits for "install"`

iarna:
This landed in 3.3.10!

iarna:
@markstos `npm help install` works fine for me. That makes me suspicious that there's something broken with your installation of this branch.  Please try again with the released version! (If it still doesn't work for you, please open an issue (and if at all possible include a copy of your `package.json` and `npm-shrinkwrap.json`).

So, we're looking for `#DEV` in the realized tree, which is an internal npm@3 data structure. Specifically it's looking in `module.package._requiredBy`– everything in `module.package` will be saved to the `package.json` for that module if npm touches it, but that's just a snapshot of its brain for debugging purposes and not used when it's loaded back from disk.

It's computed in `lib/install/deps.js` in the `addRequiredDep` function, which is called when we find the module in the tree.

Issue #7600, opened 2015-03-11
Request: cli shortcut for shorthands


RichardLitt:
Typing `npm help 7 config` in order to see a list of shorthands is difficult to remember and oddly longwinded. Can there be a shortcut to show these?

othiym23:
From my comment on #7601:

> It makes more sense to me to follow Git's model here and start pushing more of the shortcuts out to the documentation for the subcommands that use the shortcuts, which is something that people have been doing in fits and starts over the last couple months. I also think it would be nice to have Git-style command-line help tailored per subcommand.

Do either of those sound like things you'd be interested in hacking on?

RichardLitt:
Yeah, I'm down. Sounds good to me. I'll see what I can do.

RichardLitt:
@othiym23 Is this still an issue, or has npm 2 fixed it? Happy to hack away, back burned it until now.

othiym23:
It would still be great to have a slightly less terse output for `npm --help`! If you're just starting it now, I'd probably do it against `master` / `npm@3` (which is veeeerrry close to general release now). Thanks for circling back to this!

RichardLitt:
That's off of this commit, right? https://github.com/npm/npm/commit/1569304243ea91e01b5d8548950ace75a6e9fbfc

I haven't forked off of a tag before, just checking.

othiym23:
No, just pull down master and target that. (`npm@2` is currently being developed off the `2.x` branch – apologies for not being clear).

PR #10015, opened 2015-10-18
Added shortcuts to all instances in doc/

RichardLitt:
In response to #7600. I went through and added the shortcuts to all instances in the `doc` folder; however, I still we think we should also farm these out to the console.logs in `lib`, and I'm not sure that this was what @othiym23 was suggesting in #7600.

### DIFF
--- a/doc/cli/npm-bin.md
+++ b/doc/cli/npm-bin.md
@@ -3,7 +3,7 @@ npm-bin(1) -- Display npm bin folder
 
 ## SYNOPSIS
 
-    npm bin [--global]
+    npm bin [-g|--global]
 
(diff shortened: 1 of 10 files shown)


othiym23:
`-reg` is not idiomatic and only works because of some quirks of `nopt`, but the rest of these look goods, so if you pull the references to it out, this will be good to land. Thanks for putting this together!

RichardLitt:
@othiym23 Done!

iarna:
Thank you for this! It landed in 3.3.10!

RichardLitt:
No problem! Glad to help.

PR #10020, opened 2015-10-18
Update npm-config.md

lady3bean:
Adds the command to see the list of config defaults to the section on Default Configs

### DIFF
--- a/doc/misc/npm-config.md
+++ b/doc/misc/npm-config.md
@@ -35,8 +35,8 @@ See npmrc(5) for more details.
 
 ### Default Configs
 
-A set of configuration parameters that are internal to npm, and are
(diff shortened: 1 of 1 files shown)


othiym23:
It's `npm config ls -l`, instead of `npm config -ls -l`, but once you make that tweak, this looks good to merge! Thank you!

lady3bean:
Wow. What's funny is that two nights ago, in my dream, I was like like, oh no, I made a typo in my pr! but when I woke and thought to check it, I dismissed the worry because, why would I introduce a typo when adding to documentation... In any case, it's fixup'ed now.

iarna:
Thank you for this! It landed in 3.3.10!

Issue #10032, opened 2015-10-19
"Did you mean to 'npm install' first?"


aredridel:
I forget to run `npm install` on fresh clones _all the time_.

It'd be great that if there's no `node_modules` and there are `dependencies` or `devDependencies` in the package.json that if a lifecycle script fails, the CLI's error message made it super obvious that I forgot to do this.

othiym23:
This would indeed be handy to have. Attached to the road map!

Toltar:
Could we implement it where when it knows its missing `dependencies`, it would just do the installation for you when prompted the error? If the CLI makes an obvious statement, couldn't that be caught and then install all of the `dependencies` and`devDependencies` for you. I'm pretty open to discussing this.

othiym23:
This is a nice-to-have, buuuttt the CLI team isn't going to have time to work on it themselves this year. If @aredidel gets inspired and turns this into a patch, we'll land it, which also extends to the rest of the community. Because the team isn't going to be able to get to it, I'm closing this feature request. Thanks for the suggestion and your time, @aredridel and @Toltar!

KenanY:
Pending patch from @timkrins: https://github.com/npm/npm/pull/10043

PR #10043, opened 2015-10-20
Implement reminder if using run/run-script without 'install'

timkrins:
For #10032

Implements warning if using run/run-script with locally defined dependencies but without a node modules directory.

Need help writing tests - couldn't figure out how to achieve a pass/fail through using npm in tap.


lib/run-script.js
```diff
@@ -3,6 +3,7 @@ module.exports = runScript
 var lifecycle = require('./utils/lifecycle.js')
 var npm = require('./npm.js')
 var path = require('path')
+var fs = require('fs')
```
KenanY: Might as well use `graceful-fs` I think.

KenanY:
I wonder if this is best done [here](https://github.com/npm/npm/blob/8946cb3f321995792a3009f6372a6f87e81873ed/lib/utils/lifecycle.js#L240-L251) so that this warning only prints if a script actually failed to run.

Also would:

``` js
fs.stat(path.resolve(pkgdir, 'node_modules/'), ...)
```

be "better" than having to do `fs.readdir` then the `indexOf` check?

othiym23:
@KenanY Agreed. @timkrins if you can make that change, this will get merged shortly thereafter. Thanks!

timkrins:
Will get onto it tomorrow, cheers!

iarna:
This landed in 3.6.0! =)

thomblake:
@timkrins I just encountered this, great work, very helpful!

PR #10046, opened 2015-10-20
Close stream if possible, otherwise destroy stream. Should fix some TravisCI

timkrins:
Towards #7842 with Node 0.8.28

Noticed all recent pull requests were failing cryptically on `packages/npm-test-shrinkwrap` (`npm ERR! Object [object Object] has no method 'close'`), because `fetch-package-metadata` was being passed a non-gzip/tarball and Node 0.8.28's file stream does not have a close method, only a destroy method.

### DIFF
--- a/lib/fetch-package-metadata.js
+++ b/lib/fetch-package-metadata.js
@@ -283,7 +283,8 @@ function untarStream (tarball, cb) {
     } else if (hasTarHeader(c)) {
       doUntar()
     } else {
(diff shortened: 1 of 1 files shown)


timkrins:
TravisCI on 0.8 now fails much later due to Nock

iarna:
Thank you! We landed essentially the same fix in 5ee1491c

PR #10066, opened 2015-10-21
refactor(lifecycle): Swap out custom logic with manage-path module

kentcdodds:
Here's we go trying #9553 again! cc @iarna

I implemented the API similar to what [you suggested](https://github.com/npm/npm/pull/9553#issuecomment-147483156) and even renamed the module to what you suggested because I thought it made more sense. Let me know if you have questions or concerns!

### DIFF
--- a/lib/utils/lifecycle.js
+++ b/lib/utils/lifecycle.js
@@ -9,19 +9,10 @@ var path = require('path')
 var fs = require('graceful-fs')
 var chain = require('slide').chain
 var Stream = require('stream').Stream
-var PATH = 'PATH'
+var managePath = require('manage-path')
+var getPathVar = require('manage-path/dist/get-path-var')
 var uidNumber = require('uid-number')
 var umask = require('./umask')
-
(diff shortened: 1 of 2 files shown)


kentcdodds:
Just a friendly ping @iarna :-) Let me know if you don't like those or think they're unnecessary. Otherwise I'll keep it up every few days. I personally find it helpful for my projects.

iarna:
Hey, I'm really sorry to do this to you, after going through a round of revisions on this, but I'm no longer comfortable with the premise of this PR.

It takes some very simple code with a single purpose and makes it something much bigger with options and features that we don't use. It doesn't feel like a win for npm. I'm going to, with regret, close this PR. Thank you for your time working on this. We really appreciate it, even if in this case it didn't result in something we're going to bring into the project.

kentcdodds:
All's well. I'm disappointed, but that's OK. I'll just have to find some other way to contribute :-) Have a good evening/afternoon/morning :-)

Issue #5403, opened 2014-06-04
selectively install only dependencies and / or devDependencies from shrinkwrap


andersjanmyr:
We use `npm shrinkwrap` to lock down our `devDependencies` so that our CI machine does not randomly fail.

We also use `npm shrinkwrap` to lock down our dependencies for production deployment. 

Currently we cannot do this (without unwanted file manipulation) since `npm shrinkwrap` only saves one tree of dependencies into `npm-shrinkwrap.json`.

If we use `npm shrinkwrap --dev`, the `dependencies` section contains all dependencies and if we use `npm shrinkwrap`, it only contains production dependencies.

I propose that `npm-shrinkwrap.json` mirrors the `devDependencies` and `dependencies` sections of `package.json` in order to support locking down both development and production dependencies at the same time.

SirZach:
is this going to be fixed in NPM 3? would love to have this feature

marcello3d:
Just hit a build failure because devDependencies is not covered by shrinkwrap and a devDependency's dependency introduced a breaking change.

SirZach:
That was the cause for finding this issue

iarna:
So this is not currently in npm@3 (which is already out). We're certainly not going to delay its exit from beta for any new features, but if there were a pull request we would absolutely consider it– new stuff goes in every week.

jackyzhai:
+1 for this feature. =)

othiym23:
This will be resolved when #10073 is (finally) merged – see that pull request for further details, as well as what remains to be done to get it over the line and landed.

busches:
#10073 was merged in back in August. This should now be resolved. :)

cc\ @KenanY

KenanY:
woot!

Issue #2679, opened 2012-08-06
shrinkwrap and optionalDependencies


seanmonstar:
We have to shrinkwrap, since some of our dependencies are loose, and that's no good for production. Still, some of the dependencies are optional, depending on the OS.

If we include a shrinkwrap file, then npm doesn't even try to install optional dependencies. And if we shrinkwrap after the optional has been installed, it's no longer optional.

jumoel:
+1

This is still a problem.

sarith:
+1

We just ran into this as well.
Any workarounds?

othiym23:
This is now tagged onto the [npm road map](https://github.com/npm/npm/wiki/Roadmap#refinements-to-shrinkwrap-and-bundleddependencies). I can't promise that this is something we'll get to immediately, but the npm CLI team does plan to address this within the medium term (6-12 months). It will probably require making some hard choices about how to handle this case in a robust way, and may end up being a semver-major change (in that shrinkwraps that fix this may end up not being interoperable with older versions of npm).

Also, this is at least partially a duplicate of #6801.

jackwanders:
Hi @othiym23, my organization has also been suffering from this issue. I've read through the v3 source involved in installing and shrink-wrapping dependencies, and from a technical angle, it seems feasible to create a shrinkwrap process that can account for the installation failure of optional dependencies.

The two scenarios that seem to require resolution are:

1. Properly allowing a shrinkwrapped optional dependency to fail on unsupported platforms
2. Properly installing an optional dependency of a shrinkwrapped dependency that itself was not included on the shrinkwrap (i.e. the shrinkwrap occurred on the unsupported platform)

For the first scenario, would it be possible to simply check if the package's parent has it listed in its `optionalDependencies`, and if so, adding an optional failure rollback to its `installSteps`?

The second scenario seems a bit more complicated. It would seem that, during the shrinkwrap process, you would want to iterate through all optional dependencies, generating an `optionalDependencies` block within `npm-shrinkwrap.json`, including data for all such packages, regardless of whether they are currently installed. Essentially, we want to provide as much data as possible about what WOULD have been installed on the system that's doing the shrinkwrap, assuming the installation did not fail. This would both generate exact specs of non-installed optional dependencies and allow future installations to attempt their own installation of these dependencies, rolling back on failure.

I'm curious to know if there is anything glaringly infeasible about this approach. Given the impact this is having on my team, I'm invested in at least discovering if a resolution is possible, and contributing to the cause if it is.

[...]

PR #10073, opened 2015-10-22
shrinkwrap: handle devDeps separately

bengl:
We've been running into issues caused by #5403, so this a PR to implement it. 

As a result of this PR, the following things happen:
- `devDependency`-only items in the shrinkwrap are marked with `dev: true`.
- `npm i --only=prod` will not install anything in the shrinkwrap marked `dev: true`.
- `npm i --only=dev` will only install things in the shrinkwrap marked with `dev: true`.

### DIFF
--- a/lib/install/inflate-shrinkwrap.js
+++ b/lib/install/inflate-shrinkwrap.js
@@ -24,6 +24,8 @@ function inflateShrinkwrap (topPath, tree, swdeps, finishInflating) {
   var onDisk = {}
   tree.children.forEach(function (child) { onDisk[moduleName(child)] = child })
   tree.children = []
+  var dev = npm.config.get('dev') || (!/^prod(uction)?$/.test(npm.config.get('only')) && !npm.config.get('production')) || /^dev(elopment)?$/.test(npm.config.get('only'))
+  var prod = !/^dev(elopment)?$/.test(npm.config.get('only'))
   return asyncMap(Object.keys(swdeps), doRealizeAndInflate, finishInflating)
 
   function doRealizeAndInflate (name, next) {
@@ -34,6 +36,7 @@ function inflateShrinkwrap (topPath, tree, swdeps, finishInflating) {
     return function (requested) {
       var sw = swdeps[name]
       var dependencies = sw.dependencies || {}
+      if ((!prod && !sw.dev) || (!dev && sw.dev)) return next()
       var child = onDisk[name]
       if (child && (child.fromShrinkwrap ||
                     (sw.resolved && child.package._resolved === sw.resolved) ||
--- a/lib/install/is-dev.js
+++ b/lib/install/is-dev.js
@@ -21,7 +21,7 @@ function andIsOnlyDev (name) {
   }
 }
 
-exports.isOnlyDev = function (node) {
+var isOnlyDev = exports.isOnlyDev = function (node) {
   return node.requiredBy.every(andIsOnlyDev(moduleName(node)))
 }
 var isTreeDev = exports.isTreeDev = function (node, tree) {
@@ -30,7 +30,7 @@ var isTreeDev = exports.isTreeDev = function (node, tree) {
   //
   // If a node is not required by anything, then we've reached
   // the top level package.
-  if (!node || (node && node.package && node.package._requiredBy.length === 0)) {
+  if (!node || (node && node.requiredBy && node.requiredBy.length === 0)) {
     return false
   }
 
@@ -40,7 +40,7 @@ var isTreeDev = exports.isTreeDev = function (node, tree) {
   }
 
   // Otherwise, walk up one step.
-  return node.package._requiredBy.every(function (pkg) {
+  return node.requiredBy.every(function (pkg) {
     return isTreeDev(tree[pkg], tree)
   })
 }
@@ -48,11 +48,11 @@ var isTreeDev = exports.isTreeDev = function (node, tree) {
 var isTreeOptional = exports.isTreeOptional = function (node, tree) {
   // If a node is not required by anything, then we've reached
   // the top level package.
-  if (!node || (node && node.package && node.package._requiredBy.length === 0)) {
+  if (!node || (node && node.requiredBy && node.requiredBy.length === 0)) {
     return false
   }
 
-  return node.package._requiredBy.every(function (pkg) {
-    return tree[pkg] && tree[pkg].package.optionalDependencies[node.package.name] || isTreeOptional(tree[pkg], tree)
+  return node.requiredBy.every(function (pkg) {
+    return pkg && pkg.package.optionalDependencies[node.package.name] || isTreeOptional(pkg, tree)
   })
 }
--- a/lib/install/is-dev.js
+++ b/lib/install/is-dev.js
@@ -24,3 +24,23 @@ function andIsOnlyDev (name) {
 exports.isOnlyDev = function (node) {
   return node.requiredBy.every(andIsOnlyDev(moduleName(node)))
 }
+var isTreeDev = exports.isTreeDev = function (node, tree) {
+  // If we don't have the node in the tree, assume the node
+  // is not a dev dependency (as we're in a subtree)
+  //
+  // If a node is not required by anything, then we've reached
+  // the top level package.
+  if (!node || (node && node.package && node.package._requiredBy.length === 0)) {
+    return false
+  }
+
+  // A package is only a devDependency if nothing else requires it
+  if (isOnlyDev(node)) {
+    return true
+  }
+
+  // Otherwise, walk up one step.
+  return node.package._requiredBy.every(function (pkg) {
+    return isTreeDev(tree[pkg], tree)
+  })
+}
--- a/lib/install/is-dev.js
+++ b/lib/install/is-dev.js
@@ -44,3 +44,15 @@ var isTreeDev = exports.isTreeDev = function (node, tree) {
     return isTreeDev(tree[pkg], tree)
   })
 }
+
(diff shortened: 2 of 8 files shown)


othiym23:
This is promising, and I understand the use case for your team and think that this is the least disruptive way to handle it, so where is this at?

bengl:
I should have time to work on this some more in the next few days. Sorry for the delay!

bengl:
#10265 seems super-helpful for this.

bengl:
OK, by including @terinjokes's [PR](https://github.com/npm/npm/pull/10265) (thanks!), this became a whole lot easier, so here it is, including tests. I edited the title and description of this PR to show completion and include the net effects. :+1: :boom:

othiym23:
Awesome! I'll talk to @iarna about including this in next week's release (it's a little late in the week to be including in tomorrow's, especially because we've been doing a lot of cheese-moving around shrinkwrap over the last few weeks, so things are maybe not quite as solid as we'd like, and we need to review these kinds of patches carefully). Thanks to both of you for putting this together! (We'll also probably land @terinjokes's change first and then rebase this patch on top of that, so plz don't squash any commits until we get a chance to do that.)

avindra:
Thanks for putting this together. This will allow us to do better CI in our project.

@andersjanmyr  : Take a look. P.S., have you found any workarounds in the meantime (other than merging your "dependencies" and "devDependencies" lists?

andersjanmyr:
@avindra We ended up just packaging everything up in a tar-file for production after a clean build
without the shrinkwrap. This has worked fine, since we are using the same architecture for development and for production. We mainly use shrinkwrap to pin down our development dependencies, especially for grunt-plugins.

terinjokes:
What is the expected behavior of shrinkwrapping optionalDependencies?

Right now optionalDependencies are included in the shrinkwrap file, and can cause an installation failure when they fail to hydrate. For example, when a shrinkwrap is created on an OS X machine, thus including "fsevents", and an install is then attempted on Linux.

Should we tag optionalDependencies, as this pull request does for devDependencies, and allow them to fail on hydration?

bengl:
Provided something like `isTreeOptional` can be made, sure. Sounds like it would solve #2679.

terinjokes:
@bengl I've got some free time tomorrow. I'll look into if it can be made.

evocateur:
@terinjokes Any progress on the `isTreeOptional` stuff? I'm feeling the pain with shrinkwrapped `fsevents` on a linux host, so I'm willing to help out however I can.

terinjokes:
@evocateur not yet, i'll pop it off the queue shortly.

Whoaa512:
Also, what happens in these cases:

Prerequisites:
- Existing `npm-shrinkwrap.json` which includes both dev and prod deps

Cases:
- `npm install --save-dev new_package` - Will the `new_package` be added to the shrinkwrap?
- `npm install --save-dev existing_package@NEW.VERSION` - Will the `existing_package` be updated in the shrinkwrap?
- `npm install --save existing_package@NEW.VERSION` - Will the existing dev deps remain untouched in the shrinkwrap?

evocateur:
@Whoaa512 In all cases, you need to run `npm shrinkwrap` after the install commands, otherwise there will be no changes to `npm-shrinkwrap.json`.

Whoaa512:
What about when `--also=dev` is added to the aforementioned cases? Looking at the install code and from local testing `saveShrinkwrap` inside install does not respect this setting, even though the docs/changelog would tell you otherwise.

iarna:
@terinjokes I've verified that failing optional shrinkwrapped dependencies don't break the install, they just warn. I see this behavior going as far back `3.3.12`.

Now there have been some more complicated wrinkles around these, so I could believe there's a bug lurking, but a reproducer would be really helpful here!

bengl:
@iarna here's a contrived example:


package.json:
```js
{
  "optionalDependencies": {
    "test-c": "7.0.0"
  }
}
```

npm-shrinkwrap.json:
```js
{
  "dependencies": {
    "test-c": {
      "version": "7.0.0",
      "from": "test-c@7.0.0",
      "resolved": "https://registry.npmjs.org/test-c/-/test-c-7.0.0.tgz"
    }
  }
}
```

The optionalDependency is a version that's not on the registry, and so the shrinkwrap file could never really be generated, but I think this properly simulates a version having been unpublished.

* `npm install` with these two files fails, with fetch timeouts on the non-existent package.
* `npm install` without the shrinkwrap succeeds, with warning.

iarna:
Ok, I'm sorry I'm just really digging into the proposed solution now but…

> devDependency-only items in the shrinkwrap are marked with dev: true.
> npm i --only=prod will not install anything in the shrinkwrap marked dev: true.
> npm i --only=dev will only install things in the shrinkwrap marked with dev: true.

Why are we adding anything to the shrinkwrap? Why not do these checks in inflate-shrinkwrap?

bengl:
@iarna Hmm, it's been a while since I did this, but as I recall, it was my understanding that knowledge of dev-or-not was not available in inflate-shrinkwrap. Is it there in the tree object?

iarna:
It is! So somewhere in the asyncMap I'd…

https://github.com/npm/npm/blob/master/lib/install/inflate-shrinkwrap.js#L22

Add a check like:
`if (tree.package.devDependencies[name]) {`

If you're skipping it just `return next()`

bengl:
Oh awesome! That simplifies things. I'd still need `isTreeDev` (and the soon to arrive `isTreeOptional`) though, right? Otherwise we're only looking at one level.

iarna:
I was thinking that at first too, but no, because you never add that child to the tree, you never recurse to any of its children, and so there's never any need to check on them.

The only other detail, which is probably an outstanding bug, is:

https://github.com/npm/npm/blob/master/lib/install/deps.js#L110

which should account for `only=dev(elopment)` and `only=prod(uction)` too.

terinjokes:
@iarna when installing from shrinkwrap everything is (mostly) flat including dependencies of that devDependency.

iarna:
@terinjokes Ah yeah, fair, I was thinking the work flow based around resolving deps initially, but you're right, that doesn't apply here. Instead determining it at run time means walking up `requiredBy` and seeing if all of the ultimate top levels are dev deps.

Rush:
Is there a chance of merging this PR any time soon?

terinjokes:
@Rush I think they're waiting on me to update to the new style. I plan on getting to it before early next week.

graingert:
@terinjokes what's the status of this PR?

terinjokes:
@graingert I need to make changes to make npm happy, and I haven't had the time or energy to do finish.

jgrund:
What changes need to be made to get this to land? I'd really like to have this functionality.

iarna:
Ok, I'll restate what's remaining on this PR:

1. This will need to be rebased on to master.

2. Previously I'd asked that the use of `child.package._requiredBy` be rewritten to use a newer data structure `child.requiredBy`. Since then (as of 3.10.0) `_requiredBy` has gone away entirely and as such is no longer available to use at all. So this is no longer a "would make a better PR" and is now "the PR won't work without this change".

graingert:
ping @bengl

bengl:
@iarna I'm giving this another go this afternoon/evening to see if I can rebase and use `child.requiredBy`.

bengl:
OK, seems to be working as intended, after rebasing on master and making the required tweaks. I think this is good now, modulo squashing?

bengl:
Looks like an OOM failure on travis and appveyor seemed to have git perm issues? Either way, unrelated.

graingert:
@bengl I think every PR is doing this.

@bengl you need to rebase again.

graingert:
@othiym23 @iarna any chance of getting this into a release?

evocateur:
I rebased it locally to latest `master`, the only tricksy part was re-establishing the `dev` and `prod` checks in [`inflate-shrinkwrap.js`](https://github.com/npm/npm/compare/master...evocateur:devdepshrink-rebased#diff-1429e2e6780fcf5f0902e6ed7d4c3afbR27).

I would really, really like this to get into the next release.

graingert:
So say we all!

bengl:
Cool I rebased again. This is gonna need some changes when https://github.com/npm/npm/compare/iarna/fix-dev-prod-env is merged.

terinjokes:
@bengl thanks for getting this to move forward. let me know if there's anything I can do to help.

iarna:
`fix-dev-prod-env` is not a blocker on this (it's a minor structural change and will need tests to land). 

@bengl Where is this at here? Do you think this is ready to land?

bengl:
@iarna It passes the tests I had for it. I think it's good to go.

graingert:
@bengl @iarna any updates on this?

iarna:
I'm putting this on my list to finish and get landed.

graingert:
@iarna casual bump.

iarna:
Ok, the current status of this is:

This adds new attributes to `npm-shrinkwrap.json` for `optional` and `dev` to record what kind of dependency they were saved as.

It makes it so that `npm install --only=prod` _won't_ install deps that were saved to the shrinkwrap as dev deps.

Currently at adds no new behavior for the optional flagging.

iarna:
I'm ok with landing this as is and we can add behavior around optional deps later as desired.

othiym23:
SGTM.

iarna:
I'm just running the tests for the ported and squashed version and if all looks good I'll land it.

iarna:
Looks good!

iarna:
Merged to our `release-next` branch. This will be in the next release of npm (scheduled for September 8th).

PR #10122, opened 2015-10-27
lifecycle: Fix "close" handler of proc

### LINKED ISSUE #10114: Npm test swallows stream to process.stdout
This is by far the weirdest bug I have ever encountered.

When you pipe to `process.stdout` during execution of `npm test` the output will only be displayed in the shell when the streamed string has a certain length. (except the string contains a `\n`. Then it will always get printed).

I couldn't figure out, however, an exact value for the length. It seems to depend on the width of the terminal window, but no idea what the exact relation is.

I use Mac OS X 10.11.1 and it's replicatable both in the iTerm and Terminal app.

Minimal example:

Make a directory `test` including two files `test.js` and `package.json`

File `test.js`:

``` es6
var stream = require('stream')
var hasRead = false

new stream.Readable({
        read: function (n) {
            if (!hasRead){
                this.push('test')
                hasRead = true
            }
        }
    })
    .pipe(new stream.Transform({
        transform: function (chunk, encoding, done) {
            done(null, 'abcde')
        }
    }))
    .pipe(process.stdout)
```

`package.json`

``` json
{
  "name": "test",
  "scripts": {
    "test": "node ./test.js "
  }
}
```

When you now run `npm test` in the `test` directory, nothing gets printed to stdout:

``` sh
~/test$ npm test

> test@ test /Users/adrian/Desktop/test
> node ./test.js

~/test$
```

If you keep increasing the length of the `abcde` String, it will start to print to stdout at some point:

``` sh
adrians-mac:~/Desktop/test$ npm test

> test@ test /Users/adrian/Desktop/test
> node ./test.js

abcde abcde abcde abcde abcde abcde abcde abcde abcde abcde abcde abcde abcde abcde abcde abcde abcde abcde abcde abcde abcde abcde abcde adrians-mac:~/Desktop/test$
```

This only happens during execution of the script with `npm test` and not when it's executed directly with `node test.js`

Any ideas?

alexanderGugel:
Fixes #10114

Steps to reproduce + further description in #10114 
Not sure what the best way to test this is.

### DIFF
--- a/lib/utils/lifecycle.js
+++ b/lib/utils/lifecycle.js
@@ -230,8 +230,8 @@ function runCmd_ (cmd, pkg, env, wd, stage, unsafe, uid, gid, cb_) {
       process.kill(process.pid, signal)
     } else if (code) {
       var er = new Error('Exit status ' + code)
+      procError(er)
     }
-    procError(er)
   })
 
   function procError (er) {


othiym23:
The way that's relying on variable hoisting makes me a little twitchy, so I'd like to see the declaration of `er` pulled up and out of the if / else block.

Also, since this is fixing a regression, it pretty much needs a test to land. It's definitely difficult to test, especially if, as I suspect, it's due to changes in behavior in Node itself. But just landing this as-is makes me pretty nervous that we're going to cause problems elsewhere.

alexanderGugel:
@othiym23 You are correct.

The issue is actually in procError:

```js
   if (progressEnabled) log.enableProgress()

**Comment by alexanderGugel** [comment]
@othiym23 You are correct.

The issue is actually in procError:

```js
  function procError (er) {
    if (progressEnabled) log.enableProgress()
    if (er) {
      log.info('lifecycle', logid(pkg, stage), 'Failed to exec ' + stage + ' script')
      er.message = pkg._id + ' ' + stage + ': `' + cmd + '`\n' +
                   er.message
      if (er.code !== 'EPERM') {
        er.code = 'ELIFECYCLE'
      }
      er.pkgid = pkg._id
      er.stage = stage
      er.script = cmd
      er.pkgname = pkg.name
    }
    return cb(er)
  }
```

`log.enableProgress()` causes it. When taken out, everything works fine. The issue is not actually in npm, but somewhere in [gauge](https://github.com/iarna/gauge/blob/master/progress-bar.js#L128).

Not sure if this should be closed then or left open. I don't have time to look into this, but this is definitely **not** a npm issue.

othiym23:
It does seem like this is a bug in `gauge`, if in fact this is a bug – this is a known consequence of setting the terminal output to be in buffered mode when using terminal control codes. The proper thing to do would be to pick it up over on https://github.com/iarna/gauge/issues. Thanks for tracking this down, @alexanderGugel!

Issue #10026, opened 2015-10-18
npm@3 fails to install serialport@2, but npm@2 works


aredridel:
```
serialport-2@1.0.0 /Users/aredridel/Projects/serialport-2
├── async@0.9.0  extraneous
├── bindings@1.2.1  extraneous
├── debug@2.2.0  extraneous
├── nan@2.0.9  extraneous
├── optimist@0.6.1  extraneous
└── sf@0.1.7  extraneous

npm WARN EPACKAGEJSON serialport-2@1.0.0 No repository field.
npm ERR! Darwin 15.0.0
npm ERR! argv "/usr/local/lib/node_modules/node-darwin-x64/bin/node" "/usr/local/bin/npm" "i" "--save" "serialport@2"
npm ERR! node v4.1.0
npm ERR! npm  v3.3.9
npm ERR! path /Users/aredridel/Projects/serialport-2/node_modules/serialport/node_modules/node-pre-gyp/node_modules/request/node_modules/har-validator/bin/har-validator
npm ERR! code ENOENT
npm ERR! errno -2
npm ERR! syscall chmod

npm ERR! enoent ENOENT: no such file or directory, chmod '/Users/aredridel/Projects/serialport-2/node_modules/serialport/node_modules/node-pre-gyp/node_modules/request/node_modules/har-validator/bin/har-validator'
npm ERR! enoent This is most likely not a problem with npm itself
npm ERR! enoent and is related to npm not being able to find a file.
npm ERR! enoent

npm ERR! Please include the following file with any support request:
npm ERR!     /Users/aredridel/Projects/serialport-2/npm-debug.log
```

debug log at https://gist.github.com/aredridel/c1284461af09f4cf47c7

iarna:
I can't reproduce this in `3.3.9` or `3.3.10`

nemrod:
@iarna this happened to me last week but I've already downgraded so I'm not sure exactly what version I had nor can I post debug logs. Sorry.

iarna:
@nemrod specifically with `serialport@2` or with some other modules?

nemrod:
@iarna: ``serialport@2`` indeed

XeCycle:
`3.3.9` is still failing to install it in my Gitlab CI.  It is a private project depending on `serialport@2` so sorry I cannot post debug logs.

XeCycle:
A test of only `npm -g i serialport@2` succeeded, but installing all dependencies failed.  Can you try a `package.json` with these dependencies?

[...]

PR #10147, opened 2015-10-29
Make finalize ok w/ package.json resulting in bundled dep overrides

iarna:
This is the cause of at least some, if not many, of the ENOENT errors we've seen, because npm ends up installing a different version of a dependency than it _thinks_ it did, and then tries to do stuff to the version that's not there.

Fixes: #10026


iarna:
Fixes: #10026

iarna:
This landed in 3.3.10

Issue #9187, opened 2015-08-05
Upgrading to v3.2.0 npm install throws "Object.keys called on non-object"


madcapnmckay:
First time using npm 3 beta. I installed globally and then deleted node_modules and called npm install.

The result was:

```
npm ERR! Object.keys called on non-object
npm ERR!
npm ERR! If you need help, you may report this error at:
npm ERR!     <https://github.com/npm/npm/issues>
npm ERR! Darwin 14.4.0
npm ERR! argv "node" "/Users/imckay/.nvm/v0.12.7/bin/npm" "install"
npm ERR! node v0.12.7
npm ERR! npm  v3.2.0
```

Seems like this was fixed in v2 but has resurfaced.

Any suggestions?

madcapnmckay:
I think I found the root cause of this. In npm/lib/install/deps.js

```javascript
  var binaryMatches = tree.children.some(function (child) {
    return Object.keys(child.package.bin || {}).some(function (bin) {
      return pkg.bin && pkg.bin[bin]
    })
  })
```

The above code does not take into account package.json with a string bin property value. This is valid according to the docs. In my case a sub dependency had such a bin entry.

IgorAufricht:
+1

iarna:
Hmm, I was under the impression that one of the normalization layers was removing the variation here, but maybe not…

iarna:
Ok, so yeah, when I tired to reproduce this I did indeed find that the `bin` field in package.json is expanded into an object even if you specified it as a string.  I've not been able to find a scenario that reproduces this.

Can you share the `package.json` of your project?  The `npm-debug.log` from when it fails would help too!

ashchander:
I ran into this error as well using 3.3.9.  I wrapped `Object.keys(child.package.bin || {})` in a try-catch and logged child.package in the catch.

Here's one of the child.package objects that it threw the error for:

```
{ description: 'Extremely fast HTTP Archive (HAR) validator using JSON Schema',
  author:
   { name: 'Ahmad Nassri',
     email: 'ahmad@ahmadnassri.com',
     url: 'https://www.ahmadnassri.com/' },
  homepage: 'https://github.com/ahmadnassri/har-validator',
  repository: 'ahmadnassri/har-validator',
  license: 'ISC',
  main: 'lib/index',
  bin: 'bin/har-validator',
  keywords: [ 'har', 'http', 'archive', 'validate', 'validator' ],
  engines: { node: '>=0.10' },
[...]

PR #10151, opened 2015-10-29
Try to ensure string-type `package.bin` don't cause crashers

iarna:
I've not yet been able to reproduce the root cause– my current pet theory is that it's caused by third party registries returning unnormalized package.json data.

So this patch does two things, first it makes sure that whatever route the module is getting to us, `fetch-package-metadata` can only ever return normalized results, and second it insures the proximate cause of the crash can't make a crash happen, even if it slips in through some door I can't see/think of.

Fixes: #9187


iarna:
This landed in 3.3.10

iarna:
This landed in 3.3.10

PR #10153, opened 2015-10-29
fix interactions between modules managed by shrinkwraps and those not

iarna:
Shrinkwraps need to let you override dependency versions specified in
package.json files. Indeed, this was already supported, but it was a bit
to keen on this.  Previously, if you had a dep in your shrinkwrap then anything
that required that dependency would count as a match, regardless of version.

We're changing this so that it only does this broad matching when the
matched module is a direct child of the module doing the requiring.

This fixes one of the several, probably unrelated issues in #9703.


iarna:
This landed in 3.3.10

PR #10185, opened 2015-10-30
preserve array order in package.json

### LINKED ISSUE #10063: Sorting arrays in package.json
I'm sure this will be a duplicate but I couldn't find the issue. I use browserify transforms in a 3rd party module. They are defined in the package.json like this:
```json
"browserify": {
   "transform": ["jadeify", "babelify", "brfs"]
}
```
The problem i'm having is when the module is installed with `npm install` the array gets sorted and the transforms are then in an invalid order. It looks like its on line 6 of npm/lib/utils/deep-sort-object.js https://github.com/npm/npm/blob/master/lib/utils/deep-sort-object.js#L6 and its called from here https://github.com/npm/npm/blob/master/lib/install/update-package-json.js#L14. I removed this sort and tested and I get the array in the order I expect (the order it was defined).

ghost:
Fixes #10063: before, arrays were getting erroneously sorted in package.json. With this patch, only non-array objects are sorted.


terinjokes:
(y)

lib/utils/deep-sort-object.js
```diff
@@ -3,7 +3,9 @@ var sortedObject = require('sorted-object')
 
 module.exports = function deepSortObject (obj, sortBy) {
   if (obj == null || typeof obj !== 'object') return obj
-  if (obj instanceof Array) return obj.sort(sortBy)
+  if (obj instanceof Array) return obj.map(function (x) {
```
iarna: How about?
```
if (obj instanceof Array) return obj
```

othiym23:
Also, this is something that needs a test to guard against regressions before it lands. Once the above discussion is resolved to everyone's satisfaction and a test is landed, we can merge and release this.

lib/utils/deep-sort-object.js
substack: Then nested objects inside arrays would be unsorted. I don't know what purpose the sorting serves in the first place, so this way seems most cautious.

terinjokes: This originally came in with https://github.com/npm/npm/commit/2f3c83a59f7387cff41ee0894707d1cffed04825 to resolve #5779 by making the npm-shrinkwrap.json output consistent.

npm-shrinkwrap.json no longer includes the entire package manifest, so I'm not sure how useful it is anymore.

substack:
Test added.

iarna:
LGTM

iarna:
This landed with 3.7.0! =)

PR #10192, opened 2015-10-31
When replacing an existing module with a bundled one, remove it first

iarna:
If a bundled module is going to be replacing a module that's currently on disk (for instance, when you upgrade a module that includes bundled dependencies) we want to select the version from the bundle in preference over the one that was there previously.

We achieve this by removing the obsoleted module before extracting the bundled dependency.

This fixes the problems npm is currently seeing in Travis, where it has trouble upgrading to 3.3.11 from a 2.x branch.


iarna:
(For deploy monday with a special 3.3.12 release)

iarna:
This landed in 3.3.12

PR #10207, opened 2015-11-01
Travis CI test stable and LTS Node.js

fscherwi:
Test not all Node.js 4.x and 5.x and so on!
Tests the 4.2 LTS and the latest stable version!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: node_js
 node_js:
-  - "4.1"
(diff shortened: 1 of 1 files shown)


othiym23:
We landed a similar tweak to this in 791ec6b, only following the Node.js project conventions for release series (i.e. `4` and `5`, instead of the more implicit `stable`). Thanks for noticing that we'd drifted out of date and putting this together!

PR #10232, opened 2015-11-02
Fix tests in 0.8

iarna:
This is all that should be required! No upgrading nock eeeever again. =D


othiym23:
I am in favor of anything that moves the test suite closer to being all-green. :sheep: at will.

iarna:
Ok, so this didn't _fix_ 0.8 compatibility, but it did improve it. It revealed that some further polyfills for buffer changes will be required.

iarna:
This landed in 3.4.1

PR #10233, opened 2015-11-02
Update tested node versions

iarna:
Just test the 4.x latest, as that's LTS, plus 5


othiym23:
rubberstamp :sheep:, even if tests fail (especially if tests fail)

PR #10244, opened 2015-11-03
Remove arguments from npm-dedupe.md

bengotow:
The example invocations of npm dedupe show arguments provided, but the copy says "Arguments are ignored. Dedupe always acts on the entire tree.", and that appears to be correct.

### DIFF
--- a/doc/cli/npm-dedupe.md
+++ b/doc/cli/npm-dedupe.md
@@ -3,8 +3,8 @@ npm-dedupe(1) -- Reduce duplication
 
 ## SYNOPSIS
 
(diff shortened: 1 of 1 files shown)


othiym23:
Right you are! Confirmed against the code and updated the usage message to match. Reworded the commit message and landed as 4659f1c5ad617c46a5e89b48abf0b1c4e6f04307 for `npm@3.5.1`.

othiym23:
It turns out that `npm@2` _does_ use the arguments, so this only applies to `npm@3`, and so this is now ready for release. Thanks for putting this together!

Issue #9693, opened 2015-09-21
npm link; npm ls invalid package?


leshy:
Hi, when linking packages, my linked packages are marked as invalid from within packages depending on them (when doing npm ls)
also, when doing npm install in the parent, I get this message:

npm WARN update-linked node_modules/helpers needs updating to 1.0.0 from 1.0.0 but we can't, as it's a symlink

why is it trying to update from 1.0.0 to 1.0.0?

also, when a (linked) dep tree is deeper, npm seems to start behaving extra strange, deleting deps from deeper deps, and installing extraneous packages in the parent node_modules.

I can't seem to google out these error messages, couldn't find anyone with problems like this.

example:

```shell
~/coding/nodelibs/helpers > npm link
helpers@1.0.0 /home/lesh/coding/nodelibs/helpers
└── underscore@1.8.3 

/usr/lib/node_modules/helpers -> /home/lesh/coding/nodelibs/helpers



~/coding/nodelibs/backbone4000 > npm link helpers
/home/lesh/coding/nodelibs/backbone4000/node_modules/helpers -> /usr/lib/node_modules/helpers -> /home/lesh/coding/nodelibs/helpers



~/coding/nodelibs/backbone4000 > npm link
npm WARN skippingAction Module is inside a symlinked module: not running remove underscore@1.8.3 node_modules/helpers/node_modules/underscore
npm WARN update-linked node_modules/helpers needs updating to 1.0.0 from 1.0.0 but we can't, as it's a symlink
/usr/lib/node_modules/backbone4000 -> /home/lesh/coding/nodelibs/backbone4000



~/coding/nodelibs/backbone4000 > npm ls
backbone4000@0.0.1 /home/lesh/coding/nodelibs/backbone4000
├── backbone@1.1.2
├─┬ helpers@1.0.0 -> /home/lesh/coding/nodelibs/helpers invalid
│ └── underscore@1.8.3
└── underscore@1.8.3

npm ERR! invalid: helpers@1.0.0 /home/lesh/coding/nodelibs/backbone4000/node_modules/helpers

```

these are the repos in question:

https://github.com/leshy/backbone4000

https://github.com/leshy/nodejs-helpers

thank you

baelter:
Also seeing this.

qimingweng:
:+1: Is this expected behaviour?

qimingweng:
@othiym23 Hi, just wondering if there was any update or if there's anything we can do to help. I see the bug label, is it safe to ignore this warning for now?

PR #10268, opened 2015-11-05
install: blast linked modules if direct request

boneskull:
- addresses issue #9693
- `npm install foo` will overwrite any existing symlinks to package `foo`

This may not be npm's intended behavior, but it makes sense to me.  Currently (`master` as of this writing):

``` sh
$ cd foo
$ npm link
$ cd ../bar
$ npm link foo
$ npm install ../foo # doesn't seem to do much of anything
```

This PR makes `npm install ../foo` just go ahead and install `../foo`.  

This PR _does not_ address the inherent problems with the `update-linked` wording, which can suggest a package needs to be replaced with itself.  Especially confusing here is that if a package has the same `version` property, but otherwise a slightly (or wildly!) different `package.json`, a user may receive the warning.

So, I wouldn't consider #9693 to be resolved by this PR, but addresses _what I consider to be_ unexpected behavior surrounding it.  I can't recall if npm v2 would simply overwrite the symlinked package?

Questions: 
1.  I didn't see anywhere where the nuances of `install`'s behavior against existing linked packages are documented.  If this is to be written, please let me know where.
2.  My local tests pass, so I imagine there is no coverage here.  If you are interested in this PR, please let me know what you want tested, specifically, and I will create a new test suite.
3.  If this is not acceptable/intended/desired behavior for npm, what is an effective solution to #9693?
4.  If the current behavior is not a "bug", then this is a breaking change.  If it _is_ a bug, then this still might be a breaking change.  :smile:

### DIFF
--- a/lib/install/diff-trees.js
+++ b/lib/install/diff-trees.js
@@ -135,7 +135,7 @@ function diffTrees (oldTree, newTree) {
       if (pkg.oldPkg) differences.push(['remove', pkg])
     } else if (pkg.oldPkg) {
       if (!pkg.directlyRequested && pkgAreEquiv(pkg.oldPkg.package, pkg.package)) return
-      if (!pkg.isInLink && (isLink(pkg.oldPkg) || isLink(pkg))) {
+      if (!pkg.directlyRequested && !pkg.isInLink && (isLink(pkg.oldPkg) || isLink(pkg))) {
         differences.push(['update-linked', pkg])
       } else {
         differences.push(['update', pkg])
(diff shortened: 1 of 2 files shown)


boneskull:
are the v0.8 errors expected?

boneskull:
Hoping someone can comment on this?

iarna:
Tests should all be passing now, if you rebase on master!

This would really need tests for us to accept this.

The code that you changed is intended to stop us from changing the _internals_ of a linked module. We definitely should let you replace the linked module itself, however.

boneskull:
@iarna Great.  Can you direct me to where these tests should live?

iarna:
@boneskull Tests go in `test/tap`– generally you'd make a new test file in the style of others there.

iarna:

boneskull:
@iarna Does npm@5 have the desired behavior here?  If not, I'll try to create a new PR (as I think rebasing at this point would be unwise)

Issue #9980, opened 2015-10-15
`npm unlink` won't remove global links because gentlyRm takes its job very seriously


othiym23:
Only affects `npm@3`, with the following output:

```
npm ERR! gentlyRm containing path /usr/local/lib/node_modules/npm isn't under npm's control
npm WARN install:npm Refusing to delete: /usr/local/bin/npm not in /usr/local/lib/node_modules/npm
npm verb install:npm Error: Refusing to delete: /usr/local/bin/npm not in /usr/local/lib/node_modules/npm
```

Repro case is to run `make link` from a checkout of the npm repository, and then run `npm unlink` (with no arguments) afterwards.

denysdovhan:
I have the same problem.

jasonkarns:
dupe of #9680 ?

denysdovhan:
Looks like it is the same

iarna:
I didn't get time to get this in for 3.3.10, so I'm pushing this to next week.

denysdovhan:
No problem. I'm just still use Unix `rm` for unlinking.

iarna:
This was pushed out One More Week, because while I have a PR available, we would like to take the opportunity to write some unit tests for it, now that it's been rewritten to be unit-testable.

denysdovhan:
Okay, not problem.

vproman:
v3.4.1 fixes this issue (per #10270) if `npm install -g npm` doesn't get you 3.4.1+ because it's not yet in the repo, do `npm install -g npm@v3.4.1`.

PR #10270, opened 2015-11-05
The great gently rm refactor

iarna:
Fixes: #9980

This was done to try (again) to make gently-rm easier to reason about. Plus fix the bug where globally linked modules couldn't be removed.

Some squashing is probably ultimately in order.

The final, refactored form of the module with full context is here:

https://github.com/npm/npm/blob/60da484/lib/utils/gently-rm.js

And the test is here:

https://github.com/npm/npm/blob/60da484/test/tap/gently-rm-linked-module.js


iarna:
This landed in 3.4.1

Issue #8581, opened 2015-06-16
cache-min & unknown version


timoxley:
Currently, if npm is trying to install a package not yet in the npm cache, it will reach out to the registry and grab the package, even with `cache-min=Infinity`. This is desirable.

This allows high values for cache-min to work well, but a problem occurs in the following scenario: 

1. package A in cache, knows all versions up to v1.0.0.
2. package A releases version 1.1.0
3. package B not in cache, gets downloaded
4. package B requests package A ^1.1.0.
5. `npm install --cache-min=Infinity`
6. npm errors with a "no matching version found" error, listing all versions up to 1.0.0.

To resolve this, one has to update the cache's version info e.g. `npm info packagename` or lower the `cache-min` setting. This makes use of high `cache-min` values particularly error prone, especially on first installs.

It would be good if unknown versions were treated the same as unknown package names, i.e. npm should fall back to checking the remote registry before reporting "no compatible version found".

timoxley:
@othiym23 what are your reservations about this?

timoxley:
bump

JamesMGreene:
Arrrrrgh, I keep running into this. :+1:

iarna:
We chatted more about this internally and reservations are withdrawn– We would welcome a patch that, if a dependency version lookup failed with a cached version of the package's metdata, tried again with a fresh version. (This may be a bit tricky due to the architecture around the config being a painfully global object, but with sufficient surgery it's plausible.) However, it's not on our timeline for the coming year to add it ourselves.

As such, @othiym23 has marked this patch-welcome, and I'm going to close this issue.

timoxley:
For the record, I disagree that this issue should be closed, and I disagree with labelling it as a "feature request". It's a race condition bug that *sometimes* causes a completely valid `npm install` to fail.

quarterto:
Also for the record: this issue was fixed by #12084 which was released in npm 3.9. `--cache-min=Infinity` to your heart's content.

gumallett:
Has there been a regression? We are currently using 3.10.8 and this logic is no longer working.  we have cache-min=Infinity and after publishing a new version of a dependency (e.g 1.0.7), running npm install dep@1.0.7 gives an error ETARGET.

legodude17:
That could be it.
Pinging @timoxley, as well as @iarna and @othiym23.

alienlike:
I'm still experiencing this at v4.1.2, with `npm update`.

```
[...]

PR #10331, opened 2015-11-10
add "-offline" shorthand for "--cache-min 999999"

lewiscowper:
I was going through @ashleygwilliams's [slides](http://ashleygwilliams.github.io/you-dont-know-npm) and came across [slide 15](http://ashleygwilliams.github.io/you-dont-know-npm/#15) where I discovered how to install a package from the cache, which is super helpful.

She suggested adding an alias of `--offline`, so I thought I'd explore the possibility that this PR would solve that problem. Although it does appear that it will actually shorthand to `-offline` which isn't quite what was suggested. Maybe there's another way of doing this to maintain the double dash syntax for full words that'd be better in this case, and if so, please let me know whether I can do this inside this PR.

Happy to close this if there's a totally different way of doing it, and I'm more than happy to add tests, but I couldn't find any other tests to test whether the shorthands expanded to the intended arguments, so I figured it's not seen as a useful unit test, which is reasonable.

### DIFF
--- a/doc/misc/npm-config.md
+++ b/doc/misc/npm-config.md
@@ -53,6 +53,7 @@ The following shorthands are parsed on the command-line:
 * `-C`: `--prefix`
 * `-l`: `--long`
 * `-m`: `--message`
+* `-offline`: `--cache-min 999999`
 * `-p`, `--porcelain`: `--parseable`
 * `-reg`: `--registry`
 * `-f`: `--force`
--- a/lib/config/defaults.js
+++ b/lib/config/defaults.js
@@ -371,6 +371,7 @@ exports.shorthands = {
   'local': ['--no-global'],
   l: ['--long'],
   m: ['--message'],
+  offline: ['--cache-min', '999999'],
   p: ['--parseable'],
   porcelain: ['--parseable'],
   g: ['--global'],


hemanth:
:+1:

othiym23:
I think you only need the second of the two changes – "shortcuts" are for genuine shortcuts. The piece that's missing is documentation, which is going to cause the tests to complain.

That saiiiiid, this is a nice thought, but @ashleygwilliams led you astray here, because as it stands this isn't actually going to guarantee offline behavior. npm will still try to connect to the registry if a package has dependencies on other packages that aren't in the cache. See also #8581 for another edge case that can arise when trying to use a high `cache-min` setting.

My concern is that landing this as-is would create confusion among users who don't understand the subtleties of how npm's cache works (which, going by this issue tracker, includes almost everybody). Instead of doing something that's going to cause a bunch more issues to be filed when people notice the "offline mode's" edge cases, I'd prefer to wait until full support for offline operation is added. This would entail making sure that npm doesn't generate any network traffic, and either errors or gives a strong warning that certain things aren't going to work when `offline` is set.

I leave this up to you, @lewiscowper – we don't like to leave PRs with open-ended tasks on them open for too long, but offline support would be very useful to a lot of people (at the very least, if you fixed #8581, you'd have the gratitude of @timoxley). Is this something you're interested in continuing to work on?

lewiscowper:
That might be a big ask given my lack of experience in this area.

But I'm more than willing to give it a go, if you had anyone cli-side that might want to mentor/be available for Qs that would also be really helpful.

I'd say give me around 2-3 weeks (I have a few talks to give in between now and then), and I'll try and explore it, and if I can't solve it by then (so the end of November), then we should shut the PR completely.

othiym23:
If you ask questions here, one of us can answer them, and some of us are generally available on #npm on Freenode (although we're generally only paying close attention to IRC during our work hours, which are roughly 8 a.m. to 6 p.m. US/Pacific time).

> I'd say give me around 2-3 weeks (I have a few talks to give in between now and then), and I'll try and explore it, and if I can't solve it by then (so the end of November), then we should shut the PR completely.

That sounds reasonable! Thanks, Lewis!

iarna:
I have a little experimental branch that does something like this– it's an experiment because it lacks docs or tests and may or may not be complete, but it might be helpful in getting an idea as to what direction you might take:

https://github.com/npm/npm/tree/iarna/experiment/disable-networking

iarna:
For instance, it returns `ECONNREFUSED` if networking is disabled, something new like maybe `EOFFLINE` would probably be more appropriate. (This would also need an entry in `lib/utils/error-handler` to give user friendly error messages.)

What this doesn't do is disable `git clones`, which a `--offline` probably should.

lewiscowper:
Hi @iarna. I'd had a look through the wiki, and found your experiments. Thanks so much for them (and for publicising their existence), they are nice little explorations into a specific thing. I'm going to have a look into them over the next few evenings.

Can I ask though, is there a preferred way to run a local version of npm itself? I'd be worried of breaking something in npm completely, and then it being linked to my global npm install and having to do some acrobatics to fix it.

lewiscowper:
19 days have passed, December is upon me, and I haven't had a change to mess around with this exploratory PR any more than I did 19 days ago. I will close this PR to keep your stats looking good, because the changes proposed here don't work for reasons @othiym23 explained above. If I ever manage to get to the point that I might be closer than @iarna to figuring it out, I'll open a PR with the changes for discussion. I don't however, expect that to be soon, due to time, knowledge, and reasons.

Thanks for your time, help, clarifications, and your continued work over at `npm`.. (I did manage to get the local version of npm cli running via `make link`, which is sort of in the docs, but because it looked different to the other guides, I skipped it. <<)

othiym23:
Thank you for the update and I look forward to finding another opportunity for you to contribute to npm, Lewis. :bow:

SimenB:
Any news on this?

Would make stuff like https://twitter.com/ReBeccaOrg/status/789569913111269376 unnecesarry, right?

PR #10337, opened 2015-11-10
Allow partial and full disabling of flat trees

iarna:
Rename `--no-flat` to `--legacy-bundling`, document it's purpose as making a tree that's compatible with the npm 1.2 that ships with node 0.8.

Rename `--partial-flat` to `--global-style`.

(The secret is that both of these produce 0.8 compatible bundles.)


iarna:
@othiym23 (in person) says :sheep:

PR #10338, opened 2015-11-11
Fix two bugs introduced by my not returning from `filter` functions

iarna:
=/

The meaty commit is:

https://github.com/npm/npm/commit/33aedafd9531b897dca6f650ed09aab483ca3163

Which fixes a regression introduced in 971fd47


othiym23:
Someday we'll be able to use arrow functions... soooomedaaaaay.

LGTM :sheep: (since we talked about this already)

iarna:
This landed in 3.4.1

Issue #9683, opened 2015-09-20
npm 3.3.3: remove bar for short-running commands?


indexzero:
From what I've gleaned the progress bar is pretty awesome for `npm install` and `npm outdated`, but not as useful for the following commands:

```
npm owner
npm view
```

Local commands appear to have it removed, e.g.:

```
npm config
npm ls
npm test
```

... so I guess it's just a network request thing? What are your thoughts on removing it for commands where the network requests are all or nothing?

iarna:
I generally agree– I think we shouldn't be showing progress bars for commands that complete quickly. That said, having used npm on really, really bad networks, having some indication that it's doing something is really nice.

I think my preferred solution would be to only show the progress bar (and thus the spinner) if the command has taken more than some amount of time.

However, this isn't on the roadmap for the coming year, and so the core team is unlikely to get to it. We're not keeping open feature requests that we don't expect to get to in the next year, so I'm going to close this. We would accept a PR to this effect though.

PR #10373, opened 2015-11-13
Optional non-atomic move for use with Docker/cross-device moves

haggholm:
See #9863


package.json
```diff
@@ -61,6 +61,7 @@
     "lodash.uniq": "~3.2.2",
     "lodash.without": "~3.2.1",
     "mkdirp": "~0.5.1",
+    "node-fs-extra": "~0.8.1",
```
vsimonian: I think you meant to require `fs-extra` here and in the code as well - `node-fs-extra` points to an old fork for which the linked source repository doesn't exist anymore.

othiym23:
As described on #9683, this is not a good solution to the problem, because it doesn't address the underlying race condition the `fs.rename()` is intended to solve, where multiple installs could end up writing to the install target at the same time, leaving the installed package tree in an inconsistent or corrupt state. We need a solution for Docker users, but at this point I think that has more to do with how and where Docker mounts the filesystem used by npm, and not in npm itself.

That aside, thanks for taking the time to put this together!

graingert:
@othiym23 do you mean #9863

othiym23:
Yes! Thanks for the heads up – corrected inline.

frank-dspeed:
DOCKERFILE THAT APPLYS RENAME TO BE MOVE !!!! FOR DOCKER USERS use it if you understand this ISSUS:

~~~
FROM node:5.6
WORKDIR /usr/local/lib/node_modules/npm
RUN npm install --save fs-extra
RUN sed -i s/graceful-fs/fs-extra/g /usr/local/lib/node_modules/npm/lib/utils/rename.js
RUN sed -i s/fs.rename/fs.move/g /usr/local/lib/node_modules/npm/lib/utils/rename.js
WORKDIR /app
#...... main part of the Dockerfile greetings.
~~~

Issue #10348, opened 2015-11-11
link=true on 3.x no longer symlinks the majority of modules


JonDum:
I have a stupid amount of projects with npm dependancies. So much so that my poor little SSD keeps running out of space for the 30-100mb+ that each node_modules folder ends up taking. It's really ridiculous. Thankfully, adding the `link=true` to my `~/.npmrc` fixed the problem very well in npm 2.x. Almost every module was symlinked from my `$prefix` and I was a happy camper since even though there was some waste in the global registry, it wasn't being compounded by the number of projects I had with deps.

I've since updated to 3.4.0 and this has come crashing to a halt. Now, with the whole 'flat' dep management thing, only the root packages are symlinked and all of their dependancies are added as copies to the root of `node_modules/`. And this has exacerbated the file size problem again. Is it possible to tell `npm` to install *every* package to `$prefix` and symlink flat style from there?

othiym23:
This is a regression from `npm@2`'s behavior, but I'm not sure we're going to have time to tackle this soon. What `npm@3` _should_ do, is not install any of the `--link`ed packages' dependencies, because when they're installed (globally), their dependencies will be installed then.

If you want to learn (a bunch) about the `npm@3` installer, you could put together a patch for this. :grin::thumbsup::tada: @iarna and I can answer your questions if you feel like giving this a shot.

JonDum:
I'll give it a shot!

PR #10388, opened 2015-11-14
Fix linking from global for installing when link=true

sarendipitee:
A la #10348 

So this may be super naive, since I couldn't exactly decipher why you would only want to link if requiredBy was a top level package or by `"#USER"`, but at the _very_ least this allows modules to get through to `isLinkable()` where before they would never get passed  `if (!isReqByTop && !isReqByUser && !isExtraneous) return next()` condition under what appeared to be any circumstances. It may be entirely possibly to remove that entirely, since aggressive symlinking is what we want anyways, right?

Anyways, this makes symlinking work pretty well again. Would like some feedback on a better way to go about this since I've now identified the cause.

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -410,7 +410,7 @@ Installer.prototype.computeLinked = function (cb) {
     var isReqByTop = pkg.package._requiredBy.filter(function (name) { return name === '/' }).length
     var isReqByUser = pkg.package._requiredBy.filter(function (name) { return name === '#USER' }).length
     var isExtraneous = pkg.package._requiredBy.length === 0
-    if (!isReqByTop && !isReqByUser && !isExtraneous) return next()
+    if (!self.link && !isReqByTop && !isReqByUser && !isExtraneous) return next()
     isLinkable(pkg, function (install, link) {
       if (install) linkTodoList.push(['global-install', pkg])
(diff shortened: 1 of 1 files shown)


iarna:
I think this actually points to the entire approach being bogus, and requiring reworking–

My preferred solution would be to ditch this code entirely and instead, in `lib/install/deps.js`'s `resolveWithNewModule` check to see if it should link (eg, is this a top level dep), add a node that's flagged as a new global link somehow.

Then `lib/install/diff-trees.js` could use the `isLinkable` logic to generate the `global-install` and `global-link` actions itself.

JonDum:
@iarna sounds like you know more about it then me. That would probably be a much cleaner solution than my hack. If you submit a PR I can help test to verify it works as well.

iarna:
@JonDum I don't currently have fixing this bug planned for the immediate future (which sees out about a month right now). As it sounds like we're agreed that we won't take this approach I'm going to close this PR.

JonDum:
I've been using my fix successfully for several months with no issues. That's really disappointing that you'd rather do nothing and keep it broken for everyone rather than accept a sub-optimal fix for now, pending a correct fix later. It's literally one additional check in an if statement.

Issue #10401, opened 2015-11-15
`install` silently prunes misnamed git dependencies


olizilla:
Adding a dependency on a git url fails silently if the name given doesn't match the name property of the remote package.json

This is an error on the users part, but the result is the remote dep is resolved correctly, but then pruned:

``` sh
npm sill install loadAllDepsIntoIdealTree
npm sill fetchOtherPackageData x@github:select2/select2#6be96cfaa175448ef70b978648de7d6cb5822

...

verb addRemoteGit data._resolved: git://github.com/select2/select2.git#6be96cfaa175448ef70b978648de7d6cb5822e89
npm sill cache afterAdd Select2@4.0.0
npm verb afterAdd /Users/oli/.npm/Select2/4.0.0/package/package.json not in flight; writing
npm verb afterAdd /Users/oli/.npm/Select2/4.0.0/package/package.json written
npm sill resolveWithNewModule Select2@4.0.0 checking installable status
npm sill loadAllDepsIntoIdealTree Finishing
npm sill idealTree:prePrune some-whizzy-app-built-with-npm3@1.0.0
npm sill idealTree:prePrune └── Select2@4.0.0
npm sill loadIdealTree Finishing
npm sill currentTree some-whizzy-app-built-with-npm3@1.0.0
npm sill idealTree some-whizzy-app-built-with-npm3@1.0.0

...

npm verb exit [ 0, true ]
npm info ok
```

**The user gets no warning of their mistake, and the dep is not installed.**

This is most noticeable when coupled with casing issues. Given a dependency declaration like:

``` json
"dependencies": {
  "select2": "select2/select2#6be96cf"
}
```

...where we've accidentally lowercased the dep name to `"select2"` (_in savant-like readiness for the new way_) but the remote is actually uppity cased as [`"Select2"`](https://github.com/select2/select2/blob/7e2ade09652f66f3a70787f6f0e2918d171a6dcb/package.json#L2);
in `npm@2` the dep is installed regardless. 

In `npm@3` the dep is silently pruned as if you never declared it, which is surprising when transitioning an old project.

Digging around, I can see that it's because the resolved dep name doesn't match any listed deps, and is correctly pruned, so I reckon this sort of error should be caught earlier, in a similar way to [how tarballs are verified](https://github.com/npm/npm/blob/65a64c9277184b1a0665d78fce0a9b00f930d9bc/lib/cache/add-local-tarball.js#L124)

olizilla:
[...]

Issue #8588, opened 2015-06-17
Local deps where the dep name and the name in the package.json differ don't result in an error


iarna:
So if in the main package.json:

```{ "dependencies": { "foo": "../path/to/foo" } }```

And in `../path/to/foo:
```{ "name": "bar" }```

This should result in an error. Instead, `npm@2` happily installs it and `npm@3` adds it to the ideal tree and then prunes it as it appears to be extraneous, so it never ends up installing it.

piranna:
For the record: on npm@2, it install it with "bar". It should throw an error because the name on the dependencies and the name of the package are not the same, or install it with the name defined on the dependencies instead (like an alias).

PR #10410, opened 2015-11-16
Error on install where dep name !== resolved name

olizilla:
Where the name given for a git url dep doesn't match the name given
in the remote package.json, fail the install and let the user know.

Existing `npm@3.4.0` behaviour is to siltenly prune the dep from the tree.
with no warning.
- Minimal fix for #10401
- Resuses the error message from [tarball verification](https://github.com/npm/npm/blob/65a64c9277184b1a0665d78fce0a9b00f930d9bc/lib/cache/add-local-tarball.js#L124)

### DIFF
--- a/lib/fetch-package-metadata.js
+++ b/lib/fetch-package-metadata.js
@@ -87,6 +87,9 @@ function fetchOtherPackageData (spec, dep, where, next) {
   validate('SOSF', arguments)
   log.silly('fetchOtherPackageData', spec)
   cache.add(spec, null, where, false, iferr(next, function (pkg) {
+    if (dep.name !== pkg.name) {
+      return next(new Error('Invalid Package: expected ' + dep.name + ' but found ' + pkg.name))
+    }
     var result = clone(pkg)
     result._inCache = true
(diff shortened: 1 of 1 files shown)


olizilla:
Please consider this a proposal, or a starting point for a discussion on pulling remote dep verification into a common lib. Happy to re-jig this PR as required.

othiym23:
I think this is probably the correct thing to do in this case, but we should soften the rough edges here a tiny bit. Also, it needs a test!

iarna:
See also: https://github.com/npm/npm/issues/8588

olizilla:
I will clean this PR for socks.

@othiym23 totally agree on the need for a test. I found the existing tests a little difficult to wrap my head around at first glance, but I'll figure it out. Any guidance welcome though. 

Also, happy to polish things... I know it was forever ago now, but can you clarify on
> we should soften the rough edges here a tiny bit.

busches:
@olizilla are you still planning to write a test to get this functionality merged?

iarna:

PR #10419, opened 2015-11-16
doc: remove refs to nonexistent `npm-rm(1)` manpage

kenany:
The correct manpage is `npm-uninstall(1)`, **or** `doc/cli/npm-uninstall.md` should be renamed to `doc/cli/npm-rm.md`.

### DIFF
--- a/doc/cli/npm-uninstall.md
+++ b/doc/cli/npm-uninstall.md
@@ -1,4 +1,4 @@
-npm-rm(1) -- Remove a package
+npm-uninstall(1) -- Remove a package
 =============================
 
 ## SYNOPSIS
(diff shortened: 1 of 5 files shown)


doc/cli/npm-uninstall.md
```diff
@@ -1,4 +1,4 @@
-npm-rm(1) -- Remove a package
+npm-uninstall(1) -- Remove a package
```
KenanY: And maybe change this to "Uninstall a package" while I am here?

othiym23:
I think this is the right call, and I think leaving it as "Remove a package" instead of "uninstall a package" is the more precise way of describing what the command is doing, so I left that alone. Reworded the commit message a bit and landed as 0a25e2956e9ddd4065d6bd929559321afc512fde for `npm@3.5.1`.

othiym23:
Cherry-picked for `npm@2.14.13` as 1ae2dbe759feb80d8634569221ec6ee2c6d1d1ff. Thanks for putting this together, Kenan! Also, fixing the docs entitles you to socks! Do you want wdu socks? If so, send support@npmjs.com a mailing address and the information that I said you need socks!

Issue #9642, opened 2015-09-17
node_modules/history included in npm pack'd tarball


lukekarrys:
When running `npm pack` in a directory that contains `node_modules/history/`, the resulting tarball contains `node_modules/history/` when I think it should be ignored. I'm using a `.gitignore` to try and ignore `node_modules`. Other directories within `node_modules/` are ignored, expect for `history`.

I created [a gist](https://gist.github.com/lukekarrys/7ba04eb61a8d88f933fc) with the steps I used to reproduce, which I also pasted below.

A quick glance at the code makes it seem like the possible culprit could be [this regex](https://github.com/npm/npm/blob/v2.14.4/node_modules/fstream-npm/fstream-npm.js#L102) from `fstream-npm` (which is also in the [`v3.3.3` release](https://github.com/npm/npm/blob/v3.3.3/lib/utils/tar.js#L72)).

**Steps to repro**
- `git clone git@gist.github.com:7ba04eb61a8d88f933fc.git`
- `cd 7ba04eb61a8d88f933fc`
- `npm install`
- `npm pack`
- `tar xzf history-huh-1.0.0.tgz`
- `tree -L 2 package`

**Output**
```sh
> tree -L 2 package
package
├── README.md
├── node_modules
│   └── history
└── package.json
```

**Expected**
```sh
> tree -L 2 package
package
├── README.md
└── package.json
```

othiym23:
Your diagnosis of what's going on looks absolutely correct. The fix would be to modify `fstream-npm` to verify that the entry is a file as well as being named something like "history" or "changelog", and off the top of my head, I don't know how tough that would to be. From the outside, this looks pretty innocuous, along the lines of an involuntary `bundledDependency`. How is this affecting you, @lukekarrys? (We'd be happy to review and land patches that fix this, but otherwise it might be a while until we have the cycles to get to this, unless it's causing users serious problems.)

lukekarrys:
@othiym23 No serious problems. I noticed it and was mostly curious why it was happening, as it just seemed odd.

It affected me when `node_modules/history` it ended up in the tarball of [`redux-react-router`](https://github.com/rackt/redux-react-router/issues/54) where it's listed as `devDependency` but no a `dependency`. I usually run `npm ls` as a `npm run validate` step on projects to make sure everything is good with my deps, but in this case it errors since `redux-react-router/node_modules/history` is extraneous.

[...]

PR #10445, opened 2015-11-18
Test if entries are directories before mandatorily including them

lukekarrys:
This matches the behavior in fstream-npm@1.0.6 from d82ff81403e906931fac701775723626dcb443b3.

I was tempted to completely remove the `applyIgnores` method (which is the only thing `BundledPacker` extends from `fstream-npm`), but it seems like there is slightly different logic in how bundled dependencies are packaged in [`npm`](https://github.com/npm/npm/blob/65a64c9277184b1a0665d78fce0a9b00f930d9bc/lib/utils/tar.js#L130-L131) vs [`fstream-npm`](https://github.com/npm/fstream-npm/blob/d57b6b24f91156067f73417dd8785c6312bfc75f/fstream-npm.js#L167-L178). The bundled dependencies logic was the only thing I could find that `npm` does in addition to `fstream-npm`, everything else was additional logic from `fstream-npm`.

Here's a full diff of the two functions: https://gist.github.com/lukekarrys/4bf53ef42ac41d395a32

Ref #9642

### DIFF
--- a/lib/utils/tar.js
+++ b/lib/utils/tar.js
@@ -62,17 +62,19 @@ function BundledPacker (props) {
 inherits(BundledPacker, Packer)
 
 BundledPacker.prototype.applyIgnores = function (entry, partial, entryObj) {
-  // package.json files can never be ignored.
-  if (entry === 'package.json') return true
+  if (!entryObj || entryObj.type !== 'Directory') {
+    // package.json files can never be ignored.
+    if (entry === 'package.json') return true
 
-  // readme files should never be ignored.
-  if (entry.match(/^readme(\.[^\.]*)$/i)) return true
+    // readme files should never be ignored.
+    if (entry.match(/^readme(\.[^\.]*)$/i)) return true
 
-  // license files should never be ignored.
-  if (entry.match(/^(license|licence)(\.[^\.]*)?$/i)) return true
+    // license files should never be ignored.
+    if (entry.match(/^(license|licence)(\.[^\.]*)?$/i)) return true
 
-  // changelogs should never be ignored.
-  if (entry.match(/^(changes|changelog|history)(\.[^\.]*)?$/i)) return true
+    // changelogs should never be ignored.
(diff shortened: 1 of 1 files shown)


othiym23:
This looks good, but I'd like a test before landing it. Is that something you could add, Luke? Thanks for putting this together!

lukekarrys:
Yeah, I could add a test for this.

Would it be best to create a new test specifically for `lib/utils/tar.js`? Or is there a test file that already exists where I could add a test for this specific functionality?

I wasn't sure since I think `pack()` is only used by `lib/cache/add-local.js` which I could only find in `test/` as stubbed out [here](https://github.com/npm/npm/blob/master/test/tap/cache-add-localdir-fallback.js#L52-L54). But that is only from a quick look at the code.

othiym23:
I think an integration test where you create the tree structure you're trying to pack / filter, and then looking at the results after you install that packed tarball would do the trick.

That code is not super friendly to unit testing right now, but we're in the process of cleaning up the test suite, and we might have bright ideas on how to simplify the integration tests some in the process. We could probably help you write the tests, but it might take us a month or so, as we've got our hands full at the moment.

taion:
Would it be possible to amend the npm v3.4.0 release notes (https://github.com/npm/npm/releases/tag/v3.4.0) to remove the note about this bug having been fixed there? This caused us some issues earlier.

glenjamin:
Just ran into this, anything I can do to help land this?

lukekarrys:
@glenjamin I never found time to add a test for this. If you wanted to add a commit with tests for this behavior, I imagine this would be ready to merge.

othiym23:
Yes, with a test case, this would be ready to land. Feel free to submit a new PR to replace this one if you do that, @glenjamin! And thanks to you and @lukekarrys!

taion:
Given https://github.com/npm/npm/pull/11995, can we close this PR?

lukekarrys:
Yep, I'll close.

Issue #9204, opened 2015-08-08
Couldn't install optional dependency: Unsupported


StreetStrider:
Hello. I've faced some issues, possibly it was caused by updating npm from v2 to v3 on my machine. I've tried to solve this by myself but had no success.

I'm using my homedir as a place for common-used npm-packages.
So, when I do `npm i fibonacci` in `~` I get `~/node_modules/fibonacci`. This works even if I have no `~/package.json` (I recieve warning, but this is ok for me). I was using this for tests: any node script in my home subtree is able to access modules from `~/node_modules`. I also have my own packages in `~/node_modules` as git-repositories, synchronized with my github repos. This works in `v2`, but in `v3` I got this error on **every** install:
```
npm WARN install Couldn't install optional dependency: Unsupported
npm WARN ENOENT ENOENT, open '/home/streetstrider/package.json'
npm ERR! Linux 3.13.0-61-generic
npm ERR! argv "node" "/home/streetstrider/opt/node/bin/npm" "i" "lodash"
npm ERR! node v0.12.7
npm ERR! npm  v3.2.1
npm ERR! path /home/streetstrider/node_modules/repl.js
npm ERR! code EISGIT

npm ERR! git Appears to be a git repo or submodule.
npm ERR! git     /home/streetstrider/node_modules/repl.js
npm ERR! git Refusing to remove it. Update manually,
npm ERR! git or move it out of the way first.
```
Here I'm trying to install `lodash`; and `repl.js` is my own package, installed as git-repo. So, when I have any git repositories in `~/node_modules` they prevent me from installing **any** other ordinary package.

How I'm trying to fix this:
1. I've searched through my packages tree and found nothing broken.
2. I've also searched for «Unsupported»-stuff in tree, nothing too.
3. Then I hide all my git repositories and install proceeded successfully.
4. Then I unhide one of my git-packages and error shows again.
5. I've searched through issues here and found nothing similar. Maybe #7790 has some similarities, though.
Please, help me to overcome this issue.

StreetStrider:
I'm switching to `npm@2` for now, since this issue bothers me too much. I also have npm's edge installed beside, so if fix arrives I can help to reproduce & check it.

trusktr:
I also have some "Unsupported" option dependency error:

```
npm WARN install Couldn't install optional dependency: Unsupported
```

What does it mean?

cchamberlain:
Same here.

[...]

PR #10465, opened 2015-11-19
Give better diagnostics for optional dependency failures

iarna:
Optional dependency errors were printing out the error object alone, which often lacked context. This updates the output to use the same code as error-handler

Fixes: #9204

### DIFF
--- a/lib/install/deps.js
+++ b/lib/install/deps.js
@@ -297,6 +297,18 @@ var failedDependency = exports.failedDependency = function (tree, name_pkg) {
   return false
 }
 
+function top (tree) {
+  if (tree.parent) return top(tree.parent)
+  return tree
+}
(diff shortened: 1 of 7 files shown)


lib/install/deps.js
```diff
@@ -306,8 +319,7 @@ function andHandleOptionalErrors (log, tree, name, done) {
     var isFatal = failedDependency(tree, name)
     if (er && !isFatal) {
       tree.children = tree.children.filter(noModuleNameMatches(name))
-      log.warn('install', "Couldn't install optional dependency:", er.message)
-      log.verbose('install', er.stack)
+      treeWarn(tree, name, er)
```
zkat: where does this arning actually get logged?

Issue #10482, opened 2015-11-19
When we update a dependency of a bundled dependency, the bundled dep is elided


iarna:
So if you have:
```
a bundles b requres c
```
And the version of bundled `c` in `b` doesn't match the `package.json` it will end up installing a new copy of `c` under `b` and ditching the contents of `b`.

This is because the `finalize` step assumes that anything deeper than us must already have been installed, due to the guarantees we make about install order. This is true with any kind of module EXCEPT bundled dependencies, where they all come together in a single giant glob.

PR #10487, opened 2015-11-20
Refactor the bundled module model (again)

iarna:
This rejiggers bundled modules, such that they're moved out of the bundling module's folder at extract time and places in staging by themselves like any other module. This allows the rest of npm to pretend they're normal modules, which lets us remove a whole host of special cases and overall simplifies our code.

Fixes: #10482

### DIFF
--- a/lib/install/action/extract.js
+++ b/lib/install/action/extract.js
@@ -1,19 +1,77 @@
 'use strict'
+var path = require('path')
+var iferr = require('iferr')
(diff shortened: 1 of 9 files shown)


othiym23:
:sheep: :sheep: :sheep: This is a _huge_ improvement to bundled dependency handling, @iarna. Thanks so much for banging this out on such short notice!

PR #10488, opened 2015-11-20
doc: break up changelogs by major version

othiym23:
Also archive the older ones into a separate folder to reduce clutter.

**r**: @zkat


othiym23:
In-air :sheep: from @zkat

Issue #9948, opened 2015-10-12
Stop telling users to e-mail me


ForbesLindesay:
Like the vast majority of package maintainers, my preferred method for receiving bug reports is on GitHub.  Apart from anything, most people who go to GitHub realise that someone else has already had their problem and solved it.  Unfortunatley npm keeps telling them to email me:

https://github.com/npm/npm/blob/2724e30b71ac79f5c6e1aa821474c70ab4f0a222/lib/utils/error-handler.js#L203-L204

Could we direct them to the relevant github issues instead?

Stephn-R:
I dont immediately see how this is even feasible. How would you go about building it?

aredridel:
Ooh, if there's package data in the package.json, that could be a lot better!

KenanY:
Something like this?:

``` diff
$ git diff
diff --git a/lib/utils/error-handler.js b/lib/utils/error-handler.js
index cd61f79..a8364b4 100644
--- a/lib/utils/error-handler.js
+++ b/lib/utils/error-handler.js
@@ -200,8 +200,8 @@ function errorHandler (er) {
         'not with npm itself.',
         'Tell the author that this fails on your system:',
         '    ' + er.script,
-        'You can get their info via:',
-        '    npm owner ls ' + er.pkgname,
+        'You can report this error here:',
+        '    npm bugs ' + er.pkgname,
         'There is likely additional logging output above.'
       ].join('\n')
     )
```

Stephn-R:
I like that idea @KenanY , and then just make sure to provide a `CONTRIBUTING.md` file

sinkingshriek:
I really like how `Atom` handles this. I believe they just check the error msg/status(?)  and also checks if it's already been posted in github and allows you to jump to the posted issue. Not sure if it's quite feasible in `npm` but that would definitely bring down number of tickets in issue tracker too.

ForbesLindesay:
It should be possible for cases where the project is hosted in GitHub (and has the repository field configured correctly).

juliangruber:
@trodrigues's change above looks good to me

othiym23:
This is fixed in `npm@3` by b25efc88067c843ffdda86ea0f50f95d136a638e, which somebody will need to backport to `npm@2` if they want it there. Thanks to @trodrigues for the fix!

trodrigues:
Also added it for `npm@2` at #10628

PR #10496, opened 2015-11-20
error-message: Tell users to open an issue

trodrigues:
Urge users to open an issue when an error occurs on a package instead of
telling them to contact the author by default.

See #9948

### DIFF
--- a/lib/utils/error-message.js
+++ b/lib/utils/error-message.js
@@ -42,7 +42,9 @@ function errorMessage (er) {
           'not with npm itself.',
           'Tell the author that this fails on your system:',
           '    ' + er.script,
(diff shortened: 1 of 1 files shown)


othiym23:
Landed as b25efc88067c843ffdda86ea0f50f95d136a638e for `npm@3.5.1`. Error handling has changed a bit since `npm@2`, so I wasn't able to cherry-pick this into `npm@2`, so if you want this change included in Node.js 0.12 LTS or 4 LTS, you should make a similar change to the `2.x` branch of npm. Thanks for putting this together!

trodrigues:
Ah thanks, I'll look into it!

PR #10501, opened 2015-11-20
Fix excessive "replacing bundled versions" bug

iarna:
These are terrible. Everything is terrible. Let's patch this out maybe?

### DIFF
--- a/lib/install/action/extract.js
+++ b/lib/install/action/extract.js
@@ -2,7 +2,7 @@
 var path = require('path')
 var iferr = require('iferr')
 var asyncMap = require('slide').asyncMap
-var rename = require('graceful-fs').rename
+var fs = require('graceful-fs')
 var gentlyRm = require('../../utils/gently-rm.js')
 var updatePackageJson = require('../update-package-json')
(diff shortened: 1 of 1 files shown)


lib/install/action/extract.js
```diff
@@ -2,7 +2,7 @@
 var path = require('path')
 var iferr = require('iferr')
 var asyncMap = require('slide').asyncMap
-var rename = require('graceful-fs').rename
+var fs = require('graceful-fs')
```
hemanth: Isn't `var rename = require('graceful-fs').rename` lighter?

othiym23:
Landed as e05e2961ed125abf39da895d9c851650a4485e2d, along with some test cleanup in d804c3c662cebcc0dc0ca43dfc0b5c64c6d7f96e to get the tests passing. Thanks, Rebecca!

fengmk2:
Can we add a test case for this bugfix?

othiym23:
It works in concert with the tests added to `npm@3.6"5.0` for the bundling bug, but we'd welcome more tests around this behavior for sure!

PR #10505, opened 2015-11-21
Fix ls depth when outputting json

MarkReeder:
This fixes the `depth` option when used in combination with the `json` option.

As an example, prior to this diff:

`npm ls -depth=0`

```
r-dom@2.1.0 /root/temp/r-dom
├── classnames@1.2.2
├── for-each@0.3.2
...
```

`npm ls -depth=0 --json`

```
{
  "name": "r-dom",
  "version": "2.1.0",
  "dependencies": {
    "classnames": {
      "version": "1.2.2",
      "from": "classnames@^1.1.4",
      "resolved": "https://registry.npmjs.org/classnames/-/classnames-1.2.2.tgz"
    },
    "for-each": {
      "version": "0.3.2",
      "from": "for-each@^0.3.2",
      "resolved": "https://registry.npmjs.org/for-each/-/for-each-0.3.2.tgz",
      "dependencies": {
        "is-function": {
          "version": "1.0.1",
          "from": "is-function@~1.0.0",
          "resolved": "https://registry.npmjs.org/is-function/-/is-function-1.0.1.tgz"
        }
      }
    },
...
```

and after this change:

`npm ls -depth=0`

```
r-dom@2.1.0 /root/temp/r-dom
├── classnames@1.2.2
├── for-each@0.3.2
...
```

`npm ls -depth=0 --json`

```
{
  "name": "r-dom",
  "version": "2.1.0",
  "dependencies": {
    "classnames": {
      "version": "1.2.2",
      "from": "classnames@^1.1.4",
      "resolved": "https://registry.npmjs.org/classnames/-/classnames-1.2.2.tgz"
    },
    "for-each": {
      "version": "0.3.2",
      "from": "for-each@^0.3.2",
      "resolved": "https://registry.npmjs.org/for-each/-/for-each-0.3.2.tgz"
    },
...
```

Note that the dependencies of `for-each` are incorrectly included prior and correctly excluded after.

### DIFF
--- a/lib/ls.js
+++ b/lib/ls.js
@@ -139,13 +139,14 @@ function isCruft (data) {
   return data.extraneous && data.error && data.error.code === 'ENOTDIR'
 }
 
-function getLite (data, noname) {
+function getLite (data, noname, depth) {
   var lite = {}
 
   if (isCruft(data)) return lite
 
   var maxDepth = npm.config.get('depth')
 
+  if (typeof depth === 'undefined') depth = 0
   if (!noname && data.name) lite.name = data.name
   if (data.version) lite.version = data.version
   if (data.extraneous) {
@@ -213,6 +214,9 @@ function getLite (data, noname) {
           lite.problems.push(pdm)
         })
         return [d, { required: dep, peerMissing: true }]
+      } else if (npm.config.get('json')) {
+        if (depth === maxDepth) delete dep.dependencies
+        return [d, getLite(dep, true, depth + 1)]
       }
       return [d, getLite(dep, true)]
     }).reduce(function (deps, d) {
--- a/test/tap/ls-depth-cli.js
+++ b/test/tap/ls-depth-cli.js
@@ -113,6 +113,52 @@ test('npm ls --depth=Infinity', function (t) {
   )
 })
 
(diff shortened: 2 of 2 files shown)


ganemone:
+1

othiym23:
This looks good! It needs a test before it can be landed, though. We're in the process of doing a bunch of work to the npm tests, and part of that will be writing a new guide on how to write a good npm test, but in the meantime, if you want to take a look at what's in `test/tap` and copypasta one of the existing tests, that would be a good start. Thanks for putting this together!

MarkReeder:
I added some tests a couple days ago. They're copypasta'd from similar tests in test/tap/ls-depth-cli.js - let me know if you'd like me to make any changes.

lxe:
Ah I am observing this issue as well. Would be nice to get this merged!

othiym23:
Merged and rebased and landed for `npm@3` as 71c9590be61b6a7b7fa8b6dc19baa588cda26a27. Will do the same for `npm@2` shortly.

othiym23:
Cherry-picked for `npm@2` as f4826645dc6b5c0f05c5f9187efb28c1a293554f. Thanks for putting this together, and thanks for adding the tests!

PR #10578, opened 2015-11-27
used stripBOM module instead of a function

noamokman:
Hi!
Just replaced a `stripBOM` function with an [existing module](https://github.com/sindresorhus/strip-bom).
I tried to run the tests, but they failed before I changed anything.

Anything I am missing?

Thanks!


package.json
```diff
@@ -171,6 +172,7 @@
     "slide",
     "sorted-object",
     "strip-ansi",
+	"strip-bom",
```
sindresorhus: Wrong indent

sindresorhus:
npm bundles all dependencies, so you have to do so with this dependency too.

othiym23:
Is there a particular reason to switch to the module? The current implementation is dirt simple and documents exactly what it's doing (and I doubt it's going to need to change without something extraordinary happening).

noamokman:
@sindresorhus What do you mean? 
> npm bundles all dependencies, so you have to do so with this dependency too.

@othiym23 This [module](https://github.com/sindresorhus/strip-bom) does the same,
The current implementation has this comment:
```js
// from read-package-json
```
This helps reduce the npm code base.

package.json
noamokman: Opps :sweat_smile: fixed in  6d69890

othiym23:
> We should strive for more modularization.
> This helps reduce the npm code base.

I think this is a good general principle, but in this case, the amount of functionality replaced is too small to be worth the added dependency. We're getting to the point where there's so much startup time caused by npm reading all its source files that I'm starting to be more careful about splitting things up into smaller pieces or adding more dependencies. Because of that, I'm not taking this patch. Thanks for putting this together and for answering my questions!

Issue #8982, opened 2015-07-17
Regression in npm 2.13.1


pkaminski:
When installing `grunt-contrib-imagemin` 0.9.4 (with deps `imagemin` 3.1.0, `imagemin-gifsicle` 4.1.0, `gifsicle` 2.0.1), the nested `gifsicle` dependency installs fine when using `npm` 2.13.0 and below:

```
> gifsicle@2.0.1 postinstall /root/workspace/src/github.com/Reviewable/reviewable-client/node_modules/grunt-contrib-imagemin/node_modules/imagemin/node_modules/imagemin-gifsicle/node_modules/gifsicle
> node lib/install.js
✔ gifsicle pre-build test passed successfully
```

But with 2.13.1 and 2.13.2, it fails:

```
npm WARN cannot run in wd gifsicle@2.0.1 node lib/install.js (wd=/root/workspace/src/github.com/Reviewable/reviewable-client/node_modules/grunt-contrib-imagemin/node_modules/imagemin/node_modules/imagemin-gifsicle/node_modules/gifsicle)
```

othiym23:
Yep, this is definitely a regression, introduced by 40981f2e0c9c12bb003ccf188169afd1d201f5af, which adds a check that's meant to be safer for use with scoped packages, but which causes problems when the package name appears multiple times in the working directory's full path. Thanks for filing this!

blairstark:
I also think it reintroduced the permissions issues noted in https://github.com/npm/npm/releases/tag/v2.12.1

'REMEMBER WHEN I SAID "REMEMBER WHEN I SAID THAT THING ABOUT PERMISSIONS?" '

used clean machine with npm@latest which installed 2.13.1

ran 
    yo angular myapp

got a bunch of "npm ERR! tar.unpack untar error" messages...

did 
  npm uninstall -g npm && npm install -g npm@2.12.1

then 
   yo angular myapp

ran without a problem

LoicMahieu:
Fixed with npm@3.3.5
```
> gifsicle@3.0.1 postinstall /builds/node_modules/gifsicle
> node lib/install.js

  ✔ gifsicle pre-build test passed successfully
```

elvanja:
I've stumbled upon this "multiple package name occurrence in working directory's full path" regression issue for some very strange edge cases, CentOS related only. I've created a PR #10582 that resolves this for my usage. Hope this helps.

pkaminski:
Any chance that this fix will be backported to the 2.x LTS branch?

PR #10582, opened 2015-11-27
Lifecycle: fix for safety check & package name appearing multiple times in in working directory full path

elvanja:
As noted in #8982, there was a regression in npm that caused install to fail when package name occurred in working directory's full path. The mentioned issue states that this is resolved in npm@3.3.5, but I still got the same behavior on CentOS.

So, the setup I used to reproduce the problem:
- node v4.2.1, v4.2.2 or v0.12.7
- npm v3.4.1 or v3.5.0

Steps:
- `mkdir my-time && cd my-time`
- `npm init` (accept all defaults)
- `npm install time`

"time" appears in path and it all fails with:

```
npm WARN lifecycle time@0.11.4~install: cannot run in wd %s %s (wd=%s) time@0.11.4 node-gyp rebuild /opt/my-time/node_modules/time
my-time@1.0.0 /opt/my-time
└── time@0.11.4  extraneous
```

I tried with various node version, as noted at the top, and I always got the same result. I tried also with npm v2.11.3 and it worked as expected.

Note that the issue emerged only on CentOS!
And the behavior was the same for CentOS version 6 and 7.
On Ubuntu and OSX it works as expected.

Not sure what might be the cause, but those CentOS machines were vanilla installed, with gcc from their default repos (required for `node-gyp rebuild`).

Also, I had the idea to use `wd.substring(npm.dir.length + 1) !== pkg.time` instead of `lastIndexOf` fix, but wasn't sure if the original idea was to detect on all folder depths (nested dependencies). In case that is not needed, the `substring` fix might be a safer option.


iarna:
This needs a test case, particularly seeing as it's apparently shown up more than once.

elvanja:
Update: Ubuntu fails as well. Previously I've been using my local dev machine, so my `nodenv` setup might have influenced the result.

But, I recreated this behavior in Docker images. At https://gist.github.com/elvanja/4a903a73f23d326c0177 you can find Dockerfiles for CentOS and Ubuntu setup. In both cases the error is exhibited.

Note that I've added some logging to `lifecycle.js` in those Dockerfiles.

elvanja:
@iarna Also, I'd be happy to help with writing tests for this.
Tried to add some in `test/tap/lifecycle.js` but without success.
Can't figure out how to make that `lifecycle(...)` pass, it fails e.g. at finding the dependency in `validWd` function. So, any push in the right direction would be appreciated.

iarna:
@elvanja To test this I would move this logic into an exported function (functions we're exporting just for tests we prefix with `_` by convention) and then test that function exhaustively, preferably with values gotten by logging what arguments it is currently getting when it fails for you.

elvanja:
@iarna thanks for the feedback... I've extracted only the problematic part of the code into a new function `_incorrectWorkingDirectory` and added tests for it, that use data like I extarcted from logs and fail with original implementation (from master).

I realize this is a bit odd, but if I try to test `lifecycle` method itself, then it calls `lifecycle_` and that's where the stuff breaks down.

Again, same behavior is exhibited on CentOS and Ubuntu; vanilla node/npm setup.

Hope this helps.

iarna:
I spent some time trying to reproduce this on Ubuntu with various node & npm combinations and failed to do so. (Just as an FYI– your code is obviously more correct than what we currently have, so we're landing it regardless.)

elvanja:
Well, I can attest that it happens in some corner cases.
That's why I created the docker images, where I can duplicate the problem every time, no exceptions, and on Ubuntu as well as CentOS.

My guess is that it may also relate to npm cache. E.g. once a certain gem is installed successfully, it will continue to work even in the "wrong" environment (e.g. root project name that matches this criteria) because npm has it cached. That's why docker containers can repeat the problem every time, while on my local dev machine I couldn't replicate this. Hope this helps ...

zkat:
This might fix #11304 too?

elvanja:
@zkat Replicated with `npm install imagemin-gifsicle`.
On ubuntu docker image, with node v4.2.6, with default npm v2.14.12 I get:

```bash
npm WARN package.json my-time@1.0.0 No description
npm WARN package.json my-time@1.0.0 No repository field.
npm WARN package.json my-time@1.0.0 No README data
npm WARN cannot run in wd gifsicle@3.0.3 node lib/install.js (wd=/opt/my-time/node_modules/imagemin-gifsicle/node_modules/gifsicle)
imagemin-gifsicle@4.2.0 node_modules/imagemin-gifsicle
├── is-gif@1.0.0
├── through2@0.6.5 (xtend@4.0.1, readable-stream@1.0.33)
└── gifsicle@3.0.3 (logalot@2.1.0, bin-build@2.2.0, bin-wrapper@3.0.2)
```

When I apply the `lastIndexOf` fix, it works OK.
On similar docker image, with same node v4.2.6, but with npm v3.5.3 (installed with `npm install -g npm`), it also works out of the box.
#11304 references https://github.com/imagemin/gifsicle-bin/issues/59 where node 4.2.4 is used, so my guess is that they are working with npm v2x.

Not sure why this would work in npm v3x vs npm v2x though.

iarna:

PR #10615, opened 2015-11-30
RFC: A new approach to search

aredridel:
This bugged me so poked at it this weekend.

Not perfect: it still reads all the packages to filter, but it does it streaming now!


othiym23:
I think this landing this, even if it's not perfect, would be a useful alternative to deprecating `npm search` altogether (#11035). It's going to need a little work to be land, though – I pulled down the branch to run the tests, and most of the style changes cause `standard` to be unhappy, the error-handling is a little rough, it needs tests, we need to land that change to `npm-registry-client` (with its own tests), and we should probably update the search documentation to say that it's now streaming. Given the way that search is perpetually near but not at the top of npm's list of priorities, it would be really helpful to land an incremental improvement like this.

Do you feel like poking at this some more? Thanks so much for the work you've done so far!

aredridel:
I do indeed feel like touching this more.

Not this weekend or next, but after that I might play with it again. It needs help for index updating, but the proof of concept is sound, I think.

aredridel:
Modifications to npm-registry-client are at https://github.com/npm/npm-registry-client/pull/131

zkat:
*rubs hands together*.

Aright, imma see if I can get this thing over the finish line. <3

iarna:
@zkat When @EmmaRamirez and I looked at this a couple months ago we were getting errors that only occured when:

1. You were using Node >= 6

OR

2. You didn't have auth info you in your `.npmrc`

This seemed to be some sort of interaction between node & request bugs that was tickled by @aredridel's use of request's streaming interface.

zkat:
Moving dev and discussion over to #13746 and closing this one. Thanks for the boost, @aredridel!

PR #10650, opened 2015-12-01
doc: Fixed example typo

eymengunay:
In the above example `scripts/install.js` is running for two different phases, not three.


othiym23:
Landed for `npm@3` as 6abd0e0626d0f642ce0dae0e128ced80433f15a1. Will do the same shortly for `npm@2`. Thanks for putting this together!

othiym23:
Cherry-picked onto `npm@2` as 96ebb902439e4f6f37f8beffb589769146fecf24 and dfa3ad3374ff54e843fed40304bb396f638903bf. Thanks again!

Issue #10687, opened 2015-12-04
npm outdated reports "wrong" wanted version


doberkofler:
I'm currently using npm 3.5.0 and over the last couple of weeks I've noticed that `npm outdated` on occasion reports some strange wanted versions. In the following example:
```
D:\MyDev\ljs_app\trunk\periscope\build>npm outdated
Package              Current  Wanted  Latest  Location
babel-preset-es2015   6.1.18   6.2.4  6.1.18  periscope
yargs                 3.30.0  3.31.0  3.30.0  periscope
```
It seems as if babel-preset-es2015@6.2.4 and yargs@3.31.0 would be wanted but actually those versions do not (yet) exist and the `package.json` file specifies them as follows:
```
{
  "name": "periscope",
  "license": "MIT",
  "version": "0.0.0",
  "repository": {},
  "dependencies": {
    "babel-preset-es2015": "^6.1.18",
    "yargs": "^3.30.0"
  }
}
```
Something seems wrong to me!

othiym23:
Babel has already released a new version since you filed this issue(!), so let's look at `yargs`, which I use on one of my projects as well:

```
bauchelain% npm outdated -l
Package              Current  Wanted  Latest  Location  Package Type
babel-cli              6.2.0  6.3.13  6.3.13  packard   devDependencies
babel-polyfill         6.2.0  6.3.13  6.3.13  packard   dependencies
babel-preset-es2015   6.1.18  6.3.13  6.3.13  packard   devDependencies
bluebird               3.0.5   3.0.6   3.0.6  packard   dependencies
yargs                 3.30.0  3.31.0  3.30.0  packard   dependencies
```
```
bauchelain% npm dist-tags ls yargs
latest: 3.30.0
next: 3.31.0
```

So "latest" is telling you a true story here, which is that the latest version of `yargs` is 3.30.0. What's probably not clear enough from context is that "latest" has a specific meaning here, which is _the version of yargs assigned to the `latest` [`dist-tag`](https://docs.npmjs.com/cli/dist-tag)_. You don't have to think about the fact that "latest" is a tag very often as an npm user, but an increasing number of developers are using release channels for their npm projects, which means that they're frequently publishing the newest versions of their packages to a dist-tag other than `latest` to let them get additional testing before becoming `latest`.

There are three ways of looking at packages that are interesting here:

1. What's marked as latest?
2. What's the most recently published version?
3. What's the maximum semver?

[...]

PR #10695, opened 2015-12-04
doc: explain what `npm outdated`' tells you

othiym23:
There are a few issues like #10687 where people think the output of `npm outdated` is wrong. In part this is because the column headings in the output are ambiguous / confusing, but in part that's because the `npm outdated` docs aren't clear about what the output means. I took a whack at cleaning that output up. As I say in #10687, it's hard to come up with pithy column headings that _aren't_ confusing (and, because this is part of the interface of the command, and included in the JSON output of the command that may be used programmatically, changing the label is a semver-major breaking change), so I figured making the docs clearer was a good step for now.

**r**: @ashleygwilliams 
**r**: @iarna

### DIFF
--- a/doc/cli/npm-outdated.md
+++ b/doc/cli/npm-outdated.md
@@ -10,9 +10,60 @@ npm-outdated(1) -- Check for outdated packages
 This command will check the registry to see if any (or, specific) installed
 packages are currently outdated.
 
-The resulting field 'wanted' shows the latest version according to the
-version specified in the package.json, the field 'latest' the very latest
-version of the package.
+In the output:
+
+* `wanted` is the maximum version of the package that satisfies the semver
+  range specified in `package.json`. If there's no available semver range (i.e.
+  you're running `npm outdated --global`, or the package isn't included in
+  `package.json`), then `wanted` shows the currently-installed version.
+* `latest` is the version of the package tagged as latest in the registry.
+  Running `npm publish` with no special configuration will publish the package
+  with a dist-tag of `latest`. This may or may not be the maximum version of
+  the package, or the most-recently published version of the package, depending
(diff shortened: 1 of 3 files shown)


ashleygwilliams:
hey! this is def an improvement. is there a long form page in the docs where you think it would be appropriate to put an example of this to go a step further? reasoning about this in the abstract probably is what is difficult for a lot of people.

othiym23:
This is probably the closest place in the docs to where people are going to be looking for answers when they're confused. Given that the output is the whole purpose of the command is this output, and given that understanding what `npm outdated` is telling you is essential to understanding what `npm update` does (because it uses `npm outdated`'s output to make its decisions about what to upgrade, and to which versions), I don't think it's out of place to go into detail here.

I'm happy to add a couple examples:

* `npm outdated --global`, so that people see that the `wanted` version won't prevent `npm update --global` from bringing global dependencies up to their latest versions
* `npm outdated` for a package with a `package.json`, probably with a package that uses dist-tags to manage release channels, like npm itself or yargs.

What do you think, @ashleygwilliams? Should this go here or somewhere else?

ashleygwilliams:
hey! so i def think it should go here, but i also mean that it should go somewhere in the guides so that a full example can be included. a full example is gonna be too long for the cli man pages (i think), and we don't currently have a way to show extended cli pages in the browser version

my thought was maybe in `semver` or `using a package.json` or in `publishing a package` or a few notes in each.

othiym23:
Landed (after further tweaks that @iarna reviewed) for `npm@3` as a5e8df53b8d6d75398cb6a55a44dcf374b0f1661.

othiym23:
Cherry-picked for `npm@2` as 5196893a7496f68a514b83641ff6b72f14d664dd.

PR #10700, opened 2015-12-04
search.npmjs.org seems outdated

gagern:
I can't reach a server called [`search.npmjs.org`](https://search.npmjs.org/), since there is no DNS record for it. I assume that these days one would use the search form on [`www.npmjs.com`](https://www.npmjs.com/).

### DIFF
--- a/html/index.html
+++ b/html/index.html
@@ -86,7 +86,7 @@ program that you run on your computer!</p>
   <li><a href="https://npmjs.org/doc/README.html">README</a>
   <li><a href="doc/">Help Documentation</a>
   <li><a href="doc/faq.html">FAQ</a>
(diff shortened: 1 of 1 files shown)


othiym23:
Rebased and landed for `npm@3` as 8f52833f5d15c4f94467234607d40e75198af1aa (will do the same for `npm@2` shortly). At some point, we'll probably have a better dedicated search endpoint again, but you're right, I can't remember the last time I used `search.npmjs.org`. 2012, maybe? Thanks for putting this together!

othiym23:
Cherry-picked for `npm@2` as 8e6712d4ee128858cab36c77723e35bddbb977ba. Thanks again!

PR #10706, opened 2015-12-05
Replace nock with npm-registry-mock

iarna:
Mostly this means just using `hock` via `npm-registry-mock`.  

This will bring us yet closer to 0.8 passing.


zkat:
Tests pass, travis green, like is good ^______^ :sheep:

watilde:
:tada: green green

PR #10717, opened 2015-12-06
Add 'npm version from-git'

### LINKED ISSUE #9402: Tag version on package.json
The attribute "version" could not automatically come from the git tag? Bower does this and this feature is pretty cool.

ekmartin:
Tries to use the last git tag as the new npm version, as long as it exists and is a valid version range. Fixes #9402.

Any feedback welcome!

### DIFF
--- a/lib/version.js
+++ b/lib/version.js
@@ -14,7 +14,7 @@ var assert = require('assert')
 var lifecycle = require('./utils/lifecycle.js')
 var parseJSON = require('./utils/parse-json.js')
 
-version.usage = 'npm version [<newversion> | major | minor | patch | premajor | preminor | prepatch | prerelease]' +
+version.usage = 'npm version [<newversion> | major | minor | patch | premajor | preminor | prepatch | prerelease | from-git]' +
                 '\n(run in package dir)\n' +
                 "'npm -v' or 'npm --version' to print npm version " +
                 '(' + npm.version + ')\n' +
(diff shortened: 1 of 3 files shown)


ekmartin:
Hm, seems like the tests that fail using Travis also fail in the [master branch](https://travis-ci.org/ekmartin/npm/builds/95125602) of my fork, which is kind of weird, since the same commit hash originally passed [here](https://travis-ci.org/npm/npm/builds/94788885). Same with @iarna's #10706: [original](https://travis-ci.org/npm/npm/builds/95002770), [fork](https://travis-ci.org/ekmartin/npm/builds/95172382) (same hash).

Or am I missing something here?

test/tap/version-from-git.js
```diff
@@ -0,0 +1,162 @@
...
+    t.ifError(er, 'git log ran without issue')
+    t.notOk(stderr, 'no error output')
+    t.ok(log.indexOf(version) !== -1, 'commited from subdirectory')
+    t.end()
+  }
+})
+
+test('npm version from-git with a valid tag updates the package.json version', function (t) {
+  var version = '1.2.3'
+  var tag = 'v' + version
```
othiym23: This should use `npm.config.get('tag-version-prefix')` instead of hardcoding the `v`, in order to preserve the symmetry of that configuration. (Also there should be a test case for using a `tag-version-prefix` other than `v`.)

othiym23: Never mind on the former, I was looking in the wrong place, but the latter is still necessary – that behavior is sufficiently weird that I want test cases around it to prevent regressions when people are working in this code.

lib/version.js
```diff
@@ -46,27 +37,70 @@ function version (args, silent, cb_) {
...
+function parseLastGitTag (cb) {
+  var options = { env: process.env }
+  git.whichAndExec(['describe', '--abbrev=0'], options, function (er, stdout) {
+    if (er) {
+      if (er.message.indexOf('No names found') !== -1) return cb(new Error('No tags found'))
+      return cb(er)
+    }
+
+    var tag = stdout.trim()
+    var version = semver.valid(tag)
```
othiym23: I think it would probably be good to check if `tag-version-prefix` can be stripped off the tag, and do so if it can. See my other comments for why.

othiym23:
This looks good! I have a couple suggested tweaks, but otherwise this is in good shape. We're in the process of cleaning up the test suite to make it run properly with all the versions of Node.js that npm supports, so I wouldn't worry about test failures if the tests are passing for you locally. Thank you for taking the time to put together such a thorough and careful patch!

ekmartin:
Thank you for the feedback, didn't think about `tag-version-prefix`. https://github.com/ekmartin/npm/commit/3dc4ed1130a5283bca3306cb4ad46c8e2d387cac strips the prefix if possible, with a couple new test cases. If that was what you were thinking of I can squash it into the first commit.

iarna:
The AUTHORs file is generated as a part of releases automatically from git history, and doesn't require a manual commit. =)

ekmartin:
@iarna I tried changing it to use my real name instead of my username, however it seems like that would create a duplicate entry when the script is run (because of the different `git config user.name`). Removed the commit now :smiley:.

iarna:
Yeah, we'll update the `.mailmap` at release time to have your real name, if it's not attached to your commits. =)

RumyantsevMichael:
:+1:

ekmartin:
Looks like this was merged, closing :)

PR #10765, opened 2015-12-09
Fix usage of semver module in docs

domharrington:
Previously the docs displayed only installing the `semver` module and not going into a node shell or requiring the module before displaying examples.

I think this update makes it a little clearer that `semver` is a standard node module and it is used internally in `npm`.

### DIFF
--- a/doc/misc/semver.md
+++ b/doc/misc/semver.md
@@ -4,6 +4,8 @@ semver(7) -- The semantic versioner for npm
 ## Usage
 
     $ npm install semver
+    $ node
+    var semver = require('semver')
 
     semver.valid('1.2.3') // '1.2.3'
     semver.valid('a.b.c') // null


KenanY:
I think this file might come from [node-semver's README](https://github.com/npm/node-semver).

domharrington:
Oh i see. Then i'm guessing it gets pulled in somehow during a build script when updating the dependencies? I'll make the change in there then. Thanks!

othiym23:
I'm going to close this PR, as we'll just bring in the new version once the change lands in `npm/node-semver`. This does bring up that that process (which has a test to ensure that the two stay in sync) could probably be more automated, but that seems sufficiently unrelated to this PR that it could be handled separately. Thanks for taking this on, @domharrington!

PR #10766, opened 2015-12-09
Ensure temporary files are cleaned up

ganemone:
Fixes: #6855

### DIFF
--- a/lib/npm.js
+++ b/lib/npm.js
@@ -22,6 +22,7 @@ var EventEmitter = require("events").EventEmitter
   , which = require("which")
   , CachingRegClient = require("./cache/caching-client.js")
   , charSpin = require("char-spinner")
+  , rimraf = require('rimraf')
 
 npm.config = {
   loaded: false,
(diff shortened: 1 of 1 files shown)


lxe:
:+1:

othiym23:
For this to get merged, it will need some changes:

0. It needs tests.
1. It should use nonblocking I/O – using `rimraf.sync` will just block the return to the command prompt, even though it's firing in `process.on('exit')`.
2. It should be configurable, so that people with performance concerns can disable it.
3. It should only clean up the temporary directory when the run is successful, or at least that behavior should be tied to the `rollback` setting so that users can see what's happening in the temporary directory when debugging install or build failures.

For more details, look at my comments on #6855. Thanks for your time, and for getting this started!

ganemone:
Hello!

In response,

1. I will gladly add tests, just not sure how to set up my local environment to have dev dependencies installed. Simply running `npm install` does not install dev deps. `npm install --dev` didn't work either. Can't find any documentation on running tests locally.
2. Any tasks done inside process.on('exit') must be synchronous. I believe we will also need to add a handler on SIGINT to handle user interrupts via ctrl+c
3. I did a bit of profiling and I think the performance effects are very small, but we can add that option.
4. I think it should definitely clean up temp files always. The addition of a flag would solve the issue of debugging.

ganemone:
With regards to testing, I didn't really spend that much time trying to get tests to run locally, so maybe I just need to do a little more debugging. It would be nice to have some documentation around running tests locally though.

othiym23:
> Any tasks done inside process.on('exit') must be synchronous

I don't think `process.on('exit')` is the correct way to do this, regardless. You can look at how rollback on error is implemented in `lib/error-handler.js` for ideas, but having a synchronous cleanup step there is the wrong way to solve this problem.

> I think it should definitely clean up temp files always.

I'm going to insist on files being left behind on error before this merges. The thing about postmortem debugging is that it's impossible if you don't have the pieces you need to look at what happened.

othiym23:
> With regards to testing, I didn't really spend that much time trying to get tests to run locally, so maybe I just need to do a little more debugging. It would be nice to have some documentation around running tests locally though.

We're working on reworking `CONTRIBUTING.md` to have better guidelines for _writing_ tests, but as far as running them is concerned, all you should need to do is run `npm install && npm run test-all` to get the complete test suite (which, be forewarned, takes about 7-10 minutes to run). There's a set of tests that will set up and tear down a registry for some API testing if you have CouchDB installed and on your `PATH`, but those tests are optional.

PR #10777, opened 2015-12-10
Change outdated output to match user preference

roc:
I got confused as to why when you use `npm.command.outdated` you end up with a slightly oddball looking array of arrays, whilst the console shows something completely different. 
If the user requests a list view with e.g. `--json` or `npm.config.set('json')` I feel like outdated should not just `console.log` that result but also return it.
- This PR returns `makeJSON` or `makeParseable` versions of the outdated package lists, rather than just logging them
- The benefit is that when using `npm.command.outdated` you don't have to grok `stdout` if you want the json/parseable formats.
- Default API return signature remains the same

@iarna I did a `git log -p -S 'list.map(function (item) { return [item[0].parent.path'` and got a commit from back in which changed the API structure to what it is now, from `null`.

https://github.com/npm/npm/commit/7e188d77efc7d7f75152a09c3325d9023fd5328b

I'm not sure why the `list.map` looks the way it does, and am happy to make more changes to this PR if needs be! It doesn't look like it's harming any of the tests though, now that I've changed the expected formatting output.

Tools like [updtr](https://github.com/peerigon/updtr/blob/master/lib/run.js#L35) are already using forked child processes and listening to the stdout, rather than calling npm directly. I think this could possibly change if the `outdated` module is updated.

### DIFF
--- a/lib/outdated.js
+++ b/lib/outdated.js
@@ -83,10 +83,15 @@ function outdated (args, silent, cb) {
       })
       if (er || silent || list.length === 0) return cb(er, list)
       log.disableProgress()
+
+      var listType = false
+
       if (npm.config.get('json')) {
         console.log(makeJSON(list))
+        listType = makeJSON(list)
       } else if (npm.config.get('parseable')) {
         console.log(makeParseable(list))
+        listType = makeParseable(list)
       } else {
         var outList = list.map(makePretty)
         var outHead = [ 'Package',
@@ -110,7 +115,7 @@ function outdated (args, silent, cb) {
         }
         console.log(table(outTable, tableOpts))
       }
-      cb(null, list.map(function (item) { return [item[0].parent.path].concat(item.slice(1, 7)) }))
+      cb(null, listType || list.map(function (item) { return [item[0].parent.path].concat(item.slice(1, 7)) }))
     })
   }))
 }
--- a/test/tap/outdated-local.js
+++ b/test/tap/outdated-local.js
@@ -117,39 +117,11 @@ test('outdated support local modules', function (t) {
           npm.outdated(function (er, d) {
             t.ifError(er, 'outdated success')
             t.ok(verify(d, [
-              [
-                path.resolve(__dirname, 'outdated-local'),
-                'local-module',
-                '1.0.0',
-                '1.1.0',
-                '1.1.0',
-                'file:local-module'
-              ],
-              [
-                path.resolve(__dirname, 'outdated-local'),
-                '@scoped/another-local-module',
-                '1.0.0',
-                '1.2.0',
-                '1.2.0',
-                'file:another-local-module'
-              ],
-              [
-                path.resolve(__dirname, 'outdated-local'),
-                'underscore',
-                '1.3.1',
-                '1.6.1',
-                '1.5.1',
-                'file:underscore'
-              ],
-              [
-                path.resolve(__dirname, 'outdated-local'),
-                'optimist',
-                '0.4.0',
-                '0.6.0',
-                '0.6.0',
-                'optimist@0.6.0'
-              ]
-            ]), 'got expected outdated output')
+              path.resolve(__dirname, '/outdated-local/node_modules/@scoped/another-local-module:@scoped/another-local-module@1.2.0:@scoped/another-local-module@1.0.0:@scoped/another-local-module@1.2.0'),
+              path.resolve(__dirname, '/outdated-local/node_modules/local-module:local-module@1.1.0:local-module@1.0.0:local-module@1.1.0'),
+              path.resolve(__dirname, '/outdated-local/node_modules/optimist:optimist@0.6.0:optimist@0.4.0:optimist@0.6.0'),
+              path.resolve(__dirname, '/outdated-local/node_modules/underscore:underscore@1.6.1:underscore@1.3.1:underscore@1.5.1')
+            ].join('\n')), 'got expected outdated output')
             s.close()
           })
         })
--- a/test/tap/outdated-long.js
+++ b/test/tap/outdated-long.js
@@ -48,16 +48,11 @@ test('it should not throw', function (t) {
   ]
 
   var expData = [
-    [
-      pkg,
-      'underscore',
-      '1.3.1',
-      '1.3.1',
-      '1.5.1',
-      '1.3.1',
-      'dependencies'
-    ]
-  ]
+    [pkg, 'node_modules', 'underscore'].join('/'),
+    '1.3.1',
+    '1.3.1',
+    '1.5.1:dependencies'
+  ].join(':underscore@')
 
   console.log = function () {
     output.push.apply(output, arguments)
--- a/test/tap/outdated-private.js
+++ b/test/tap/outdated-private.js
@@ -67,15 +67,12 @@ test('outdated ignores private modules', function (t) {
           bumpLocalPrivate()
           npm.outdated(function (er, d) {
             t.ifError(er, 'outdated success')
-            t.deepEqual(d, [[
-              path.resolve(__dirname, 'outdated-private'),
-              'underscore',
+            t.deepEqual(d, [
+              path.resolve(__dirname, 'outdated-private/node_modules/underscore'),
+              '1.5.1',
               '1.3.1',
-              '1.5.1',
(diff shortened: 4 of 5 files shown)


roc:
### CLI :smile_cat: 

```shell
npm outdated --json --long --depth=0
```
returns
```json

  "chai": {
    "current": "1.0.1",
    "wanted": "2.2.1",
    "latest": "3.4.1",
    "location": "node_modules/chai",
    "type": "devDependencies"
  }
```


### Module :person_frowning: 

```javascript
  npm.config.set('json', true);
  npm.config.set('long', true);
  npm.commands.outdated(function (err, packages) {
    console.log(packages);
  }
```

returns

```
[ [ '/Path/To/Current/Dir',
    'chai',
    '1.0.1',
    '1.4.1',
    '3.4.1',
    '1.0.1',
    'devDependencies' ],
  [ '/Path/To/Current/Dir',
    'chai',
    '1.0.1',
    '1.4.1',
    '3.4.1',
    '1.0.1',
    'devDependencies' ] ]
```

roc:
Huh! All tests pass for me on `iojs-v3.3.1` locally...

othiym23:
> Tools like [updtr](https://github.com/peerigon/updtr/blob/master/lib/run.js#L35) are already using forked child processes and listening to the stdout, rather than calling npm directly. I think this could possibly change if the `outdated` module is updated.

This is how the CLI team prefers that npm be used. We're so dissatisfied with the status of npm's internal API that I recently took the step of removing the API documentation from the tree. I did this because it was misleading people into thinking that using that API is supported or stable, when in fact one of my goals is to rip it out and replace it with something much less tightly coupled to the CLI's configuration and bootstrapping process. As someone who tried to consume the CLI's API before I joined the CLI team, I can tell you that it's not a good idea.

In addition, the coupling between `npm outdated` and `npm update` is sufficiently complicated that I'm hesitant to land anything that would make it more brittle. This change doesn't do that, exactly, but it does add some conditional logic that we'd need to deal with when we get around to finally unifying and redesigning `outdated`, `update`, and `install` to use the same underlying engine (which is high on the list of our architectural goals).

Right now, I'm disinclined to land this (although I appreciate the time you've put into it -- especially into writing the tests. Thank you!) because I do think it's superfluous to using `npm outdated` the way we encourage people to use it, i.e. via the command line.

roc:
@othiym23 Thanks so much for your response. I was definitely adding this in whilst feeling unsure of what the real design of this module was about outside of its CLI usage. I've read this [section](https://github.com/npm/npm#using-npm-programmatically) of the readme a bit more closely now and see the big caveat warnings.

It's just a shame because I feel like as a developer I should be able to call in npm with an internal API rather than e.g. require in npm then exec it from, or even exec the implied system npm—if I've read that right—and capture the resulting `stdout`. It feels unnecessarily circuitous, and maybe even currently avoidable. 

I don't think this PR will cause npm any harm really, but I can see why you just might not want to go there. You think adding it in will add additional complexity to tearing it back out upon redesigning, in essence? :person_with_pouting_face:

othiym23:
It does add another coupling point between `outdated` and `update` (potentially, and that's actually a point I hadn't considered – with this change, what happens if I run `npm update --json`? I fear that it might break), and it's more complexity that we'd need to handle when we move everything around. 

Thanks for being understanding! I know it seems like using the API should be simpler than using a child process, but there's a list of caveats as long as your arm about dealing with internal mutation, commands not being idempotent, etc.

roc:
Ah well! :balloon: 

I'll rewrite the current project I'm working on to just use npm in a child process and read the stdout. 

If it's of any interest, with this change if you run `npm update --json` you get the new package information logged out as json as you'd imagine.

Sample output from running on npm itself:
```
[
  {
    "name": "fs-write-stream-atomic",
    "version": "1.0.7",
    "from": "fs-write-stream-atomic@1.0.7",
    "dependencies": {}
  }
]
[
  {
    "name": "request",
    "version": "2.67.0",
    "from": "request@2.67.0",
    "dependencies": {
      "is-typedarray": {
        "version": "1.0.0",
        "from": "is-typedarray@>=1.0.0 <1.1.0",
        "dependencies": {}
      },
      "aws-sign2": {
        "version": "0.6.0",
        "from": "aws-sign2@>=0.6.0 <0.7.0",
        "dependencies": {}
      },
      "caseless": {
        "version": "0.11.0",
        "from": "caseless@>=0.11.0 <0.12.0",
        "dependencies": {}
      },
      "oauth-sign": {
        "version": "0.8.0",
        "from": "oauth-sign@>=0.8.0 <0.9.0",
        "dependencies": {}
      },
      "extend": {
        "version": "3.0.0",
        "from": "extend@>=3.0.0 <3.1.0",
        "dependencies": {}
      },
      "bl": {
        "version": "1.0.0",
        "from": "bl@>=1.0.0 <1.1.0",
        "dependencies": {}
      },
      "combined-stream": {
        "version": "1.0.5",
        "from": "combined-stream@>=1.0.5 <1.1.0",
        "dependencies": {
          "delayed-stream": {
            "version": "1.0.0",
            "from": "delayed-stream@>=1.0.0 <1.1.0",
            "dependencies": {}
          }
        }
      },
      "har-validator": {
        "version": "2.0.3",
        "from": "har-validator@>=2.0.2 <2.1.0",
        "dependencies": {}
      },
      "qs": {
        "version": "5.2.0",
        "from": "qs@>=5.2.0 <5.3.0",
        "dependencies": {}
      },
      "mime-types": {
        "version": "2.1.8",
        "from": "mime-types@>=2.1.7 <2.2.0",
        "dependencies": {
          "mime-db": {
            "version": "1.20.0",
            "from": "mime-db@>=1.20.0 <1.21.0",
            "dependencies": {}
          }
        }
      },
      "form-data": {
        "version": "1.0.0-rc3",
        "from": "form-data@>=1.0.0-rc3 <1.1.0",
        "dependencies": {
          "async": {
            "version": "1.5.0",
            "from": "async@>=1.4.0 <2.0.0",
            "dependencies": {}
          }
        }
      },
      "hawk": {
        "version": "3.1.2",
        "from": "hawk@>=3.1.0 <3.2.0",
        "dependencies": {}
      },
      "http-signature": {
        "version": "1.1.0",
        "from": "http-signature@>=1.1.0 <1.2.0",
        "dependencies": {
          "jsprim": {
            "version": "1.2.2",
            "from": "jsprim@>=1.2.2 <2.0.0",
            "dependencies": {
              "extsprintf": {
                "version": "1.0.2",
                "from": "extsprintf@1.0.2",
                "dependencies": {}
              },
              "verror": {
                "version": "1.3.6",
                "from": "verror@1.3.6",
                "dependencies": {}
              },
              "json-schema": {
                "version": "0.2.2",
                "from": "json-schema@0.2.2",
                "dependencies": {}
              }
            }
          },
          "sshpk": {
            "version": "1.7.1",
            "from": "sshpk@>=1.7.0 <2.0.0",
            "dependencies": {
              "assert-plus": {
                "version": "0.2.0",
                "from": "assert-plus@>=0.2.0 <0.3.0",
                "dependencies": {}
              },
              "ecc-jsbn": {
                "version": "0.1.1",
                "from": "ecc-jsbn@>=0.0.1 <1.0.0",
                "dependencies": {}
              },
              "jsbn": {
                "version": "0.1.0",
                "from": "jsbn@>=0.1.0 <0.2.0",
                "dependencies": {}
              },
              "asn1": {
                "version": "0.2.3",
                "from": "asn1@>=0.2.3 <0.3.0",
                "dependencies": {}
              },
              "jodid25519": {
                "version": "1.0.2",
                "from": "jodid25519@>=1.0.0 <2.0.0",
                "dependencies": {}
              },
              "dashdash": {
                "version": "1.10.1",
                "from": "dashdash@>=1.10.1 <2.0.0",
                "dependencies": {
                  "assert-plus": {
                    "version": "0.1.5",
                    "from": "assert-plus@>=0.1.0 <0.2.0",
                    "dependencies": {}
                  }
                }
              },
              "tweetnacl": {
                "version": "0.13.2",
                "from": "tweetnacl@>=0.13.0 <1.0.0",
                "dependencies": {}
              }
            }
          }
        }
      }
    }
  }
]
``` 

Thanks for all of your responses on this and good luck with the big rewrite!

PR #10778, opened 2015-12-11
fetch-package-metadata: add dist-tags to list of valid install targets

boennemann:
I was trying to update npm with `npm install -g npm@2.x-latest` instead of `npm install -g npm@latest-2`. dist-tags are missing from this list, but they would've been helpful in that output. Also this makes it more correct :)

### DIFF
--- a/lib/fetch-package-metadata.js
+++ b/lib/fetch-package-metadata.js
@@ -150,9 +150,12 @@ function fetchNamedPackageData (dep, next) {
       }
 
       // And failing that, we error out
-      var targets = versions.length
-                  ? 'Valid install targets:\n' + versions.join(', ') + '\n'
-                  : 'No valid targets found.'
+      var targets = Object.keys(pkg['dist-tags']).filter(function (f) {
+        return versions.hasOwnProperty(f)
+      }).concat(versions)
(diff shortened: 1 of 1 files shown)


othiym23:
Could we get a test for this, @boennemann? I agree that this would be a useful bit of diagnostic output to have.

boennemann:
@othiym23 I couldn't find a test for the current behaviour so I wasn't quite sure how/whether you would want that tested.

I saw that in cache/add-named tags are already used for the error message. I just updated this PR to use the implementation found there, which handles an edge case with inconsistent registry-data: https://github.com/npm/npm/blob/master/lib/cache/add-named.js#L281-L289

Please let me know how to approach tests for this and I'm happy to update the PR with one.

Best,
Stephan

iarna:
Testing this will involve using `npm-registry-mock` to return made up registry info for a package. If you grep for that in the tests you should find a bunch of examples.

zkat:
I'm gonna close this 'cause all this code is just plain gone. I'll go ahead and add it to [`npm-pick-manifest`](https://npm.im/npm-pick-manifest), since that's what does this work now.

PR #10798, opened 2015-12-12
src: always install in npm in legacy mode

othiym23:
Also, clean out the docs and prune the tree before making a release tarball, to keep as much cruft as possible out of the release tarball.

**r**: @iarna


iarna:
This landed in 3.5.4 =)

PR #10805, opened 2015-12-13
Replaced gracefull-fs with gracefull-fs-extra and lib/utils/rename.js…

### LINKED ISSUE #9863: Reinstalling npm v3 fails on Docker.
Our project initialization scripts make sure a specific version of npm is installed. If npm 3.x is accidentally installed _twice_ the install fails with `EXDEV: cross-device link not permitted`. Installing additional modules results in the same error.

npm 2.x does not have this problem.

Using Docker 1.8.2, this Dockerfile:

```dockerfile
FROM ubuntu:14.04

ENV NPM_CONFIG_LOGLEVEL info
ENV NODE_VERSION 4.1.1

RUN apt-get -y install curl

RUN curl -SLO "https://nodejs.org/dist/v$NODE_VERSION/node-v$NODE_VERSION-linux-x64.tar.gz" \
  && tar -xzf "node-v$NODE_VERSION-linux-x64.tar.gz" -C /usr/local --strip-components=1 \
  && rm "node-v$NODE_VERSION-linux-x64.tar.gz"

RUN npm install -g npm@3
RUN npm install -g npm@3 # again!
RUN npm install -g coffee-script

CMD [ "/bin/bash" ]
```

Results in the following:

```
$ docker build -t npmtest ~/misc/npmtest
Sending build context to Docker daemon 2.048 kB
Step 0 : FROM ubuntu:14.04
 ---> 91e54dfb1179
Step 1 : ENV NPM_CONFIG_LOGLEVEL info
 ---> Using cache
 ---> 8794a99ac0a2
Step 2 : ENV NODE_VERSION 4.1.1
 ---> Using cache
 ---> 3fb65b4c7d55
Step 3 : RUN apt-get -y install curl
 ---> Using cache
 ---> e1a9f7308089
Step 4 : RUN curl -SLO "https://nodejs.org/dist/v$NODE_VERSION/node-v$NODE_VERSION-linux-x64.tar.gz"   && tar -xzf "node-v$NODE_VERSION-linux-x64.tar.gz" -C /usr/local --strip-components=1   && rm "node-v$NODE_VERSION-linux-x64.tar.gz"
 ---> Using cache
 ---> 942c49d39079
Step 5 : RUN npm install -g npm@3
 ---> Using cache
 ---> 871d339a3944
Step 6 : RUN npm install -g npm@3 # again!
 ---> Running in 55fbedab36cf
npm info using npm@3.3.5
npm info using node@v4.1.1
npm info attempt registry request try #1 at 6:46:54 PM
npm http request GET https://registry.npmjs.org/npm
npm http 200 https://registry.npmjs.org/npm
npm info lifecycle npm@3.3.5~preinstall: npm@3.3.5
npm ERR! Linux 4.0.9-boot2docker
npm ERR! argv "/usr/local/bin/node" "/usr/local/bin/npm" "install" "-g" "npm@3"
npm ERR! node v4.1.1
npm ERR! npm  v3.3.5
npm ERR! path /usr/local/lib/node_modules/npm
npm ERR! code EXDEV
npm ERR! errno -18
npm ERR! syscall rename

npm ERR! EXDEV: cross-device link not permitted, rename '/usr/local/lib/node_modules/npm' -> '/usr/local/lib/node_modules/.npm.DELETE'
npm ERR!

npm ERR!     /npm-debug.log
The command '/bin/sh -c npm install -g npm@3 # again!' returned a non-zero code: 238
```

frank-dspeed:
… changed method to move

This patch fixes the EXDEV rename on multi device issus its needed for working with docker and other stuff i tested this fix and it works like a charm and dont brakes your existing functionality it can be save merged.

closes #9863

greetings


frank-dspeed:
The all checks failed is wrong only 0.8 failed and was a travis ci worker that was failing all other tests passed it will even pass on 0.8 plz review https://travis-ci.org/npm/npm/builds/96604192

othiym23:
See #9863 for why this solution is inadequate. `fs.rename()` is atomic, and this solution, which replaces it with a copy and unlink, is not. As such, this only makes the install process slower without addressing the underlying concurrency issues. I'm very interested in seeing this particular use case addressed for Docker, but this isn't going to work. Thanks for taking the time to put it together, though!

frank-dspeed:
It works realy well i will maintain a fork of this no problem and it is not slow i dont see in issu #9863
Any reason for this we simply do the following you merge in that patch and then we simply add a ENV check for switching between rename and move its simply it works and it works now and dont stops npm docker usage there are many companys working with nodejs and npm they will 100% switch away from npm if no hotfix like this gets putted in we even researched npm alternates if this situation stays as is. because we run all our infrastructure in containers as also twitter, ebay, google does you stop us all from using npm ok not us all as we no got this fork but think about it.

othiym23:
It absolutely _has_ to be slower than a simple rename, because it's copying the finished tree across file systems instead of changing an inode. More importantly, landing this patch would _negate the reason for having a rename in the first place_, because the entire reason this is done is to avoid a race condition where two processes may be trying to finish an install to the same `node_modules` at the same time. If there's no way to guarantee that step is being done atomically (and, as I said above, rename is the closest thing we have to a guaranteed-atomic file system operation), there's no reason to have an additional move operation. It would be better just to perform the installation directly to `node_modules` in that case, although, as with switching from a rename to a move, it's not safe.

In general, the project has no problem with people providing their own forks of npm – it's open source, and different sets of users have different needs – but the fact that this fork is less safe in a non-obvious way concerns me. We've put more effort than most people realize into chasing down often quite subtle sources of nondeterminism within the CLI. These concerns aren't abstract, either – I've spent weeks working with users to fix race conditions affecting only one or two (large and complicated) build environments.

It would indeed be great to find a way to solve this problem that unblocks Docker users, and ai'm happy to discuss strategies that might be successful, but this approach isn't it.

frank-dspeed:
and rename cant be success full too so we have simply no alternate till you implamented what you called a extra directory that does the stuff and then switch that back. but that will take weeks till implamented right?

iarna:
I've described what we're looking for over here: https://github.com/npm/npm/pull/13257#issuecomment-230603360

PR #10846, opened 2015-12-16
help people know they need to install peer deps

ryanflorence:
In our experience with React Router most people think the current messaging is an error.

> npm WARN EPEERINVALID react-router@1.0.2 requires a peer of history@1.13.x but none was installed.

They see this and then open an issue on our project.

Maybe somebody else can think of a better phrase, but here's my attempt at making it clearer that this is not an error, but a task the person reading it needs to complete.

### DIFF
--- a/lib/install/validate-tree.js
+++ b/lib/install/validate-tree.js
@@ -39,7 +39,7 @@ function thenValidateAllPeerDeps (idealTree, next) {
   validate('OF', arguments)
   validateAllPeerDeps(idealTree, function (tree, pkgname, version) {
     var warn = new Error(packageId(tree) + ' requires a peer of ' + pkgname + '@' +
-      version + ' but none was installed.')
+      version + ' but none is installed. You must install peer dependencies yourself.')
     warn.code = 'EPEERINVALID'
     idealTree.warnings.push(warn)
   })
--- a/test/tap/install-scoped-with-peer-dependency.js
+++ b/test/tap/install-scoped-with-peer-dependency.js
@@ -30,7 +30,7 @@ test('it should install peerDependencies in same tree level as the parent packag
   common.npm(['install', '--loglevel=warn', './package'], EXEC_OPTS, function (err, code, stdout, stderr) {
     t.ifError(err, 'install local package successful')
     t.equal(code, 0, 'npm install exited with code')
-    t.match(stderr, /npm WARN @scope[/]package@0[.]0[.]0 requires a peer of underscore@[*] but none was installed[.]\n/,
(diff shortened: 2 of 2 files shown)


hellerim:
The problem is simply the coloring of UNMET PEER DEPENDENCY - it's almost the same as the coloring indicating an error, so it is very easy to be mistaken.

rgbkrk:
The second sentence you introduced would have saved me quite a bit of time from searching around last week. :+1:

ashleygwilliams:
personally, i think this change is really great and helpful. @ryanflorence i'm working on docs re: "how npm works" -- do you think a longer form description of peer deps would help-- if so would you want to chat for 10-15min about it?

iarna:
I think I like this, but… well, let's talk this through: Assuming `a` has a peer dep:

```
User project -> a --peerdeps--> m
```

In which case I like your message:

```
npm WARN EPEERINVALID a@#.#.# requires a peer of m@#.#.# but none was installed. You must install peer dependencies yourself.
```

But in contrast if:

```
User project -> z -> a --peerdeps--> m
```

I'm less sure if I want that same message? It seems to me that if `z` isn't providing `m` itself then it should have a peer dep on it too, to pass the buck up to the top level. Otherwise it's an error in `z` no?

othiym23:
I like the underlying idea, and agree that the wording could stand to be less misleading, less brusque, and more actionable. @ryanflorence, do you have any interest in doing a little more work to refine this?

rgbkrk:
That label is funny, apropos, and scary all at the same time. Perhaps `npm` could take Ryan's current version and then refine it with follow on PRs?

othiym23:
> Perhaps `npm` could take Ryan's current version and then refine it with follow on PRs?

Given the potential for confusion and increased support load inherent in the current wording, I'd prefer to get this right the first time.

ryanflorence:
> Given the potential for confusion and increased support load inherent in the current wording, I'd prefer to get this right the first time.

Agreed, that's why I didn't bother really fixing the test because this needs a really good message.

othiym23:
I think this is ready to land once the wording is a little better, so I'm assigning this to myself for wordsmithing, and will afterwards land. Thanks to all for their patience!

iarna:
Hi, thanks for your patience as we lost track of this for so long. On reflection I think your revision is fine. I've changed my view of modules that depend on things that have peer dependencies without providing them. I was concerned that modules like that were conceptually broken, that is, that it was the module's responsibility to provide any peer dependencies of its direct dependencies. Really though, by not providing them it's effectively reexporting that dependency up the tree. There's nothing wrong with that and in plugin ecosystems I can see how this could end up being common practice.

Anyway, I've landed this as b019680db in `release-next` and it'll be in the next npm release.

Issue #4603, opened 2014-02-04
passing unix signals to child rather than dying?


evantahler:
Here's some interesting behavior I've seen when running an application which does some slower shutdown operations when receiving a termination signal.  

When running `npm start` in a *nix terminal (tested on ubuntu and OSX), the `SIGINT` (ctrl+c) command is passed from the parent npm process to the child (my running application).  npm instantly exits, and then my application handles the passed-down signal and eventually exits.  However, because the stdout/err of the session was managed by npm, strange things can happen if my application still writes (or `console.out`s).  

At the very least, the terminal which started the application will interpret the npm process' death as a return and draw the next line.  However, the child's stdout is still bound to that session.  This means you can start another application in that same terminal and still get the first app's messages intermingled.   Or, more strangely, if you then close that terminal session, the underlying app which is still running and possibly trying to write to stdout is likely to crash on the next `console.log`.

You can emulate this pretty simply:

**package.json**
```javascript
{
  "name": "test",
  "scripts": {
    "start": "node ./app.js"
  }
}
```
**app.js**
```javascript
process.on('SIGINT',  function(){ console.log("SIGINT");  shutDown() });
process.on('SIGTERM', function(){ console.log("SIGTERM"); shutDown() });

var string = ".";

var shutDown = function(){
  console.log("off-ing...");
  string = "x";

  setTimeout(function(){
    console.log("bye!");
    process.exit();
  }, 1000 * 60);
}

setInterval(function(){
  console.log(string);
}, 500)
```

1) start `app.js` with `npm start`
2) `ctrl+c` in the window you started npm with.  You will see npm exit, but still receive messages from `app.js`.
3) have fun opening vi or some other application in the same window.

I *think* a possible solution to this would to have npm catch (`process.on('x')`) and not actually exit until the child process exits and pass those commands along to the child... but I'm sure that will cause all sorts of other problems.  

Thoughts?

stuartpb:
Okay, this is a fairly significant problem for me, as it means that any apps I'm developing to use `npm start` don't get stopped when they're supposed to (in other words, every deploy I do keeps me waiting for 10 more seconds than it should as it waits for the app to respond to a SIGTERM it never gets, then force-kills the entire process group).

Does Node not provide any mechanism for accessing the process-replacing `exec()` call used by `bash  -c`?

kimmobrunfeldt:
[...]

PR #10868, opened 2015-12-17
lifecycle: propagate SIGTERM to child

daniel-pedersen:
When one runs for example `npm run-script`, any SIGTERM to npm will orphanize the child process. This is a simple fix for this (potential) issue. Now whenever npm launches a child through lifecycle, any SIGTERM signal is simply propagated to the child for the duration of the process.

This behavior of npm when getting a SIGTERM seems to be what makes sense to most people (see the discussion in #4603), and makes `npm run-script` more friendly in various development scenarios.

### DIFF
--- a/lib/utils/lifecycle.js
+++ b/lib/utils/lifecycle.js
@@ -233,6 +233,7 @@ function runCmd_ (cmd, pkg, env, wd, stage, unsafe, uid, gid, cb_) {
     }
     procError(er)
   })
+  process.once('SIGTERM', procKill)
 
   function procError (er) {
     if (progressEnabled) log.enableProgress()
@@ -248,7 +249,11 @@ function runCmd_ (cmd, pkg, env, wd, stage, unsafe, uid, gid, cb_) {
       er.script = cmd
       er.pkgname = pkg.name
     }
+    process.removeListener('SIGTERM', procKill)
     return cb(er)
+  }
+  function procKill () {
+    proc.kill()
   }
 }
 
--- a/test/tap/lifecycle-signal.js
+++ b/test/tap/lifecycle-signal.js
@@ -16,7 +16,8 @@ var json = {
   name: 'lifecycle-signal',
   version: '1.2.5',
   scripts: {
-    preinstall: 'node -e "process.kill(process.pid,\'SIGSEGV\')"'
+    preinstall: 'node -e "process.kill(process.pid,\'SIGSEGV\')"',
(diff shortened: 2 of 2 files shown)


iarna:
This looks pretty reasonable to me. It does still need a test.

daniel-pedersen:
Hmm. Something weird is going on. Testing with my linked npm in the shell, the child process is orphanized. For example, using `npm start` with the following `package.json`,
`{ "scripts": { "start": "node -e 'process.stdin.resume()'" } } `
and then doing
`$ kill <npm_pid>`
in another tab results in a orphanized node process without my first commit, with my commit the child is also terminated.

But I've now been trying to reproduce this in the commited unit-test, so that the test fails without my first commit and passes with it, but to no avail.

daniel-pedersen:
Done! The new test fails and hangs without my previous commit, which passes it.

iarna:
Awesome, thank you!

zkat:
I was considering whether to pull this into the 2.x LTS, and came to the conclusion that while folks may really not want to have orphaned zombies just from a `^C`, there is, in fact, a chance that someone out there is already relying on this behavior.

Because of that, and because LTS is supposed to be more careful about stuff like this, I'm not gonna merge this into LTS. If anyone has other thoughts or objections, please feel free to chime in, since it's not set in stone. :)

iarna:
This landed in 3.8.1!

gaastonsr:
Freaking awesome @daniel-pedersen

dchambers:
Hey guys, is this known/expected to work on Linux? I can see that it works on MacOS, but it doesn't seem to be working on our UAT box which is running Ubuntu Server?

kucharskimaciej:
Seconding @dchambers on this. Make for quite annoying experience when automating tasks with npm.

twang2218:
I'm using docker with `npm start`, and the npm doesn't forward the signal to child-processes, which caused graceful shutdown failed, and the app process be SIGKILLed instead, without a chance to release the resources.

Are you sure this PR is working on Linux? My NPM version is `3.10.10`

```bash
$ npm -v
3.10.10
$ node -v
v7.2.1
```

dchambers:
Hey @twang2218, it may be worth creating a new issue specifically for Linux support as this has been closed quite some time now.

addaleax:
Yes, this isn’t working, at least with bash; npm runs its lifecycle processes in a shell, and bash doesn’t forward SIGTERM to its children.

Issue #9809, opened 2015-10-01
Force npm 3 to install a node module always nested


Alphapage:
Hello,

npm 3 node modules installation is now flat.
I'm trying to build a frontend package, but this new feature breaks the import path in ES6.
With npm 2, all modules where nested except peerDependencies which could be set explicitly. It was very easy to find the path something like this:
```
import * from "mymodule/node_modules/mynestedmodule/file.js";
```

With npm 3, the path could be:
```
import * from "mymodule/node_modules/mynestedmodule/file.js";
```
or 
```
import * from "mynestedmodule/file.js";
```

This is a real problem to locate the package, so I think there is a need to be able to set "nestedDependencies" in package.json to find the frontend files.

Thanks in advance for your answers.

othiym23:
When you say "in ES6," I'm guessing that you mean you're using babel or a similar ES6 transpiler. I mention this because the ES module loader spec isn't yet finished, and there isn't a standard implementation of it for Node (or any other JavaScript environment) yet, which that the behavior you're describing isn't defined by the ES standard. However, it seems unlikely that any implemented version of the spec will behave in a way surprising to users of the current Node require syntax, so I think(?) that what you're talking about is more about how Node module loading works than anything related to ES6.

The issue here is that this pattern relies on using the Node module resolution logic in a way that has never been guaranteed to be safe with npm. Using full paths to files could be broken at any time by users using `npm dedupe`, and so this isn't an issue that's new or exclusive to `npm@3`. The correct thing to do is to call `import * from 'mynestedmodule/file.js'`, and if you can't require `mynestedmodule` from the package trying to use it, that means the package in which that code is located should be explicitly depending on `mynestedmodule` itself.

To answer your original question, no, this behavior isn't configurable, and it's unlikely that it will be made configurable. We _are_ probably going to add a way to make it [easier to work with front-end dependencies using npm](https://github.com/npm/npm/wiki/Roadmap-area-of-focus%3A-browser-dependencies), but that work is still a little ways out (and is itself still undefined). Since this isn't something that's likely to change, I'm going to close this issue, but I'm happy to continue to discuss the specifics of what you're trying to do here or elsewhere. Thanks for your time!

XadillaX:
BUT I NEED NESTED WAY! FLATTEN IS UGLY! I JUST HATE FLATTEN MODULE PATH! IT'S UGLY FOR ME!

How can I force install packages nested?

xinyu198736:
[...]

Issue #9808, opened 2015-10-01
Error installing grunt and grunt-contrib-clean


hellboy81:
npm i grunt@~0.4.1

npm ERR! Linux 3.13.0-63-generic
npm ERR! argv "node" "/usr/local/bin/npm" "i" "grunt@~0.4.1"
npm ERR! node v0.10.30
npm ERR! npm  v3.3.4

npm ERR! Invalid Version: 0.4.0rc8

``` sh
0 info it worked if it ends with ok
1 verbose cli [ 'node', '/usr/local/bin/npm', 'i', 'grunt@~0.4.1' ]
2 info using npm@3.3.4
3 info using node@v0.10.30
4 silly loadCurrentTree Starting
5 silly install loadCurrentTree
6 silly install readLocalPackageData
7 silly fetchPackageMetaData grunt@~0.4.1
8 silly fetchNamedPackageData grunt
9 silly mapToRegistry name grunt
10 silly mapToRegistry using default registry
...
23 verbose get saving grunt to /home/local/.../.npm/.../grunt/.cache.json
24 verbose stack TypeError: Invalid Version: 0.4.0rc8
24 verbose stack     at new SemVer (/usr/local/lib/node_modules/npm/node_modules/semver/semver.js:293:11)
24 verbose stack     at compare (/usr/local/lib/node_modules/npm/node_modules/semver/semver.js:570:10)
24 verbose stack     at rcompare (/usr/local/lib/node_modules/npm/node_modules/semver/semver.js:580:10)
24 verbose stack     at Array.sort (native)
24 verbose stack     at pickVersionFromRegistryDocument (/usr/local/lib/node_modules/npm/lib/fetch-package-metadata.js:111:48)
24 verbose stack     at /usr/local/lib/node_modules/npm/node_modules/iferr/index.js:13:50
24 verbose stack     at /usr/local/lib/node_modules/npm/lib/utils/pulse-till-done.js:19:8
24 verbose stack     at saved (/usr/local/lib/node_modules/npm/lib/cache/caching-client.js:168:7)
24 verbose stack     at /usr/local/lib/node_modules/npm/node_modules/graceful-fs/polyfills.js:210:7
24 verbose stack     at Object.oncomplete (fs.js:107:15)
26 error Linux 3.13.0-63-generic
27 error argv "node" "/usr/local/bin/npm" "i" "grunt@~0.4.1"
28 error node v0.10.30
29 error npm  v3.3.4
[...]

Issue #9726, opened 2015-09-23
NPM 3.3.3: Unable to `npm install pm2`


beeant:
```command
sudo npm install pm2 --save                                                                                                                                                                                                               master * ] 4:05 PM
npm ERR! Darwin 14.5.0
npm ERR! argv "/usr/local/bin/node" "/usr/local/bin/npm" "install" "pm2" "--save"
npm ERR! node v4.1.1
npm ERR! npm  v3.3.3

npm ERR! Invalid Version: 0.0.1alpha1
npm ERR!
npm ERR! If you need help, you may report this error at:
npm ERR!     <https://github.com/npm/npm/issues>

npm ERR! Please include the following file with any support request:
```

bengl:
Are you using an internal/private registry mirror (that does *not* use CouchDB)? If so you may be running into #9701. Some solutions are available in the comments there.

beeant:
yes, I'm using sinopia at the moment. Maybe that's the problem. I will try to use default registry.

snopeks:
Hi @beeant We haven’t heard back and we’re trying to clean up some issues. If you weren't able to install from the default registry (http://registry.npmjs.com) can you please reply and let us know? We’ll be happy to reopen if necessary.

Issue #9701, opened 2015-09-21
npm@3.3.4: "Invalid Version: 1.0.5beta"


indexzero:
This is a very strange bug. Lets start with the relevant parts of the `npm-debug.log`:

```
2643 verbose etag http://$REDACTED/handlebars from cache
2644 verbose get saving handlebars to /$REDACTED/.npm/$REDACTED/handlebars/.cache.json
2645 verbose stack TypeError: Invalid Version: 1.0.5beta
2645 verbose stack     at new SemVer (/$REDACTED/lib/node_modules/npm/node_modules/semver/semver.js:293:11)
2645 verbose stack     at SemVer.compare (/$REDACTED/lib/node_modules/npm/node_modules/semver/semver.js:346:13)
2645 verbose stack     at compare (/$REDACTED/lib/node_modules/npm/node_modules/semver/semver.js:570:31)
2645 verbose stack     at rcompare (/$REDACTED/lib/node_modules/npm/node_modules/semver/semver.js:580:10)
2645 verbose stack     at Array.sort (native)
2645 verbose stack     at pickVersionFromRegistryDocument (/$REDACTED/lib/node_modules/npm/lib/fetch-package-metadata.js:111:48)
2645 verbose stack     at /$REDACTED/lib/node_modules/npm/node_modules/iferr/index.js:13:50
2645 verbose stack     at /$REDACTED/lib/node_modules/npm/lib/utils/pulse-till-done.js:19:8
2645 verbose stack     at saved (/$REDACTED/lib/node_modules/npm/lib/cache/caching-client.js:168:7)
2645 verbose stack     at /$REDACTED/lib/node_modules/npm/node_modules/graceful-fs/polyfills.js:210:7
```

The really strange part of this is that version of `handlebars` is not being installed explicitly. I rolled back to `npm@2.14.5` and the install worked. At which point I ran:

```
my-redacted-thing@0.0.0 /path/to/my-redacted-thing
└─┬ istanbul@0.3.20
  └── handlebars@3.0.0 
``` 

If I had to guess my gut is that since the cache holds a document which represents all versions (not just the version requested – in this case `3.0.0`) then if there is logic somewhere that iterates over those versions and does a `semver.valid` naively ... :boom: ... this bug.

indexzero:
_UPDATE:_ I did indeed narrow this down to being unable to install `handlebars` at all:

```
$ npm i handlebars@3.0.0
npm http request GET http://172.19.112.179:4873/handlebars
[...]

PR #10869, opened 2015-12-17
Ignore invalid dependency versions when picking version to install

gregersrygg:
This is a fix to avoid errors for those using Sinopia and similar proxy/private registry. Related issues: #9808, #9726, #9701, #9657.

At some point npm seems to have cleaned up invalid semver versions in the registry and modify versions in flight when publishing from the CLI. This works for those that don't have a private registry/proxy, but those who do have a private registry that already have a package with invalid in the cache will get a very little informative error message:

> npm ERR! Invalid Version: 0.4.0rc8

It does not say which dependency that has an invalid version, only that something has an invalid version. It's not clear if it's the projects own package.json, a direct dependency or a transitive dependency. Nor if it's a version npm would actually resolve to or an old version we don't care about. This exception is thrown even when we don't rely on that version, it happens if any dependency has one invalid version in it's release history.

The npm-debug.log has some hints about which package this happens to, but the first time I got this error I got 4 of them, and it wasn't clear to me if the line just above the error was the same package or a log statement from another fetch happening in parallell. Unfortunately I didn't backup that log-file, but #9809 have the relevant parts from a npm-debug.log.

This fix ignores all packages with an invalid semver version. I don't think it will cause problems for public packages because the invalid versions have been fixed in the public registry, and the private registrys should just proxy the new versions. But private packages with invalid versions will also be ignored, so I added a warning for the invalid versions just in case. Please let me know if you want this to be solved in another way, or anything else :)

### DIFF
--- a/lib/fetch-package-metadata.js
+++ b/lib/fetch-package-metadata.js
@@ -122,7 +122,15 @@ function fetchNamedPackageData (dep, next) {
     }
     function pickVersionFromRegistryDocument (pkg) {
       if (!regCache[url]) regCache[url] = pkg
-      var versions = Object.keys(pkg.versions).sort(semver.rcompare)
+      var versions = Object.keys(pkg.versions)
+
+      var invalidVersions = versions.filter(function (v) { return !semver.valid(v) })
+      if (invalidVersions.length > 0) {
+        log.warn('', 'The package %s has invalid semver-version(s): %s. This usually only happens for unofficial private registries. ' +
+            'You should delete or re-publish the invalid versions.', pkg.name, invalidVersions.join(', '))
+      }
+
+      versions = versions.filter(semver.valid).sort(semver.rcompare)
 
       if (dep.type === 'tag') {
         var tagVersion = pkg['dist-tags'][dep.spec]
--- a/lib/fetch-package-metadata.js
+++ b/lib/fetch-package-metadata.js
@@ -126,7 +126,7 @@ function fetchNamedPackageData (dep, next) {
 
       var invalidVersions = versions.filter(function (v) { return !semver.valid(v) })
       if (invalidVersions.length > 0) {
-        log.warn('', 'The package %s has invalid semver-version(s): %s. This usually only happens for unofficial private registries. ' +
+        log.warn('pickVersion', 'The package %s has invalid semver-version(s): %s. This usually only happens for unofficial private registries. ' +
             'You should delete or re-publish the invalid versions.', pkg.name, invalidVersions.join(', '))
       }
 
--- a/lib/fetch-package-metadata.js
+++ b/lib/fetch-package-metadata.js
@@ -130,7 +130,7 @@ function fetchNamedPackageData (dep, next) {
             'You should delete or re-publish the invalid versions.', pkg.name, invalidVersions.join(', '))
       }
 
-      versions = versions.filter(semver.valid).sort(semver.rcompare)
+      versions = versions.filter(function (v) { return semver.valid(v) }).sort(semver.rcompare)
 
       if (dep.type === 'tag') {
         var tagVersion = pkg['dist-tags'][dep.spec]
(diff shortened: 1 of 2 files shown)


iarna:
This is great! Thank you!

philk:
This sort of worked for me, had to change line 129 to:

```javascript
versions = versions.filter(function (v) { return semver.valid(v) }).sort(semver.rcompare)
```

Not really sure why just passing the semver.valid function directly like you did doesn't work for me, didn't dig too much. Thanks for the fix though!

iarna:
I would like to see a test for this, however.

gregersrygg:
@philk You're right. Filter passes the array index as the second parameter to the callback, and semver.valid has an undocumented second argument `loose`. So all except the first element, `loose` will be truthy.

@iarna I looked at testing this, but it looked very hard to test in a clean way. Any ideas how? I would guess the most important is that I don't break anything for "normal" npm-users, and hoped you had good enough test-coverage for that already.

iarna:
@gregersrygg I would not rely on our coverage, but _especially_ for something where you're adding a feature that won't interact with the standard npm registry.

So to do a integration-type test for this, I'd use `npm-registry-mock` something like how it's used in `test/tap/add-named-update-protocol-port.js` (starting at ~line 59) where it provides a mock registry response and a mock tarball response. You can provide invalid version numbers in this response and make sure that a valid version is selected and downloaded.  You can see examples of `common.npm` there too, which captures `stderr` so you can look for your warning.

lib/fetch-package-metadata.js
```diff
@@ -118,7 +118,15 @@ function fetchNamedPackageData (dep, next) {
     }
     function pickVersionFromRegistryDocument (pkg) {
       if (!regCache[url]) regCache[url] = pkg
-      var versions = Object.keys(pkg.versions).sort(semver.rcompare)
+      var versions = Object.keys(pkg.versions)
+
+      var invalidVersions = versions.filter(function (v) { return !semver.valid(v) })
+      if (invalidVersions.length > 0) {
+        log.warn('', 'The package %s has invalid semver-version(s): %s. This usually only happens for unofficial private registries. ' +
```
iarna: There should be an initial argument to this– something like `pickVersion` would be fine, I think.

schmod:
Looks like this can be reviewed now?

gregersrygg:
@schmod yes please :) The tests are red on Travis/AppVeyor, but looks like master is red too.

I struggled a bit to mock the npm server. Not sure it's the best way to reproduce the error, but I couldn't find a better way to have a package with an invalid version in it's list of versions. Also I didn't bother to make a tarball for the packages, so I just used mkdirp´s tarball. Is that ok?

I deleted the generated folders in node_modules in the cleanup, but should I delete the test/npm_cache folder too?

schmod:
@gregersrygg Coincidentally, I recently ran into some trouble figuring out the best way to mock things.  

If anything, it looks like you've figured out, and yours might be the example for others to follow in the future...

My only comments on your tests are:

* The test case doesn't seem like it would correctly catch a circumstance where `npm install` fails silently (ie. it installs nothing and exits with `0`).  I'm not sure how big of a deal this is, given that many other tests also lack this sort of validation step...
* Other similar test cases create a temporary directory in `test/tap/<name-of-test>`, `chdir` into that directory, run `npm install` from there, and delete it afterward.  [shrinkwrap-dev-dependency.js](https://github.com/gregersrygg/npm/blob/master/test/tap/shrinkwrap-dev-dependency.js) is one example that you could follow.

iarna:
Oh, this looks great! The only thing it needs is as @schmod said, to work from a temp directory. As that's a really small change, and in the interest of getting this landed, I've prepared a branch that's ready for merging here: https://github.com/npm/npm/tree/iarna/pr/10869

iarna:
Landed in 3.10.6!

gregersrygg:
Wow, thanks @iarna :)

PR #10903, opened 2015-12-20
Add a success message to adduser/login

### LINKED ISSUE #10775: UX: Add login success message
There's a failure message, but no success message. Would be nice to have confirmation.

ekmartin:
Basically just adds a green `Authentication successful.` message to login/adduser (fixes #10775):
image

Also refactors some of the adduser tests to have less duplicate code.

### DIFF
--- a/test/tap/adduser-always-auth.js
+++ b/test/tap/adduser-always-auth.js
@@ -68,32 +68,135 @@ test('npm login', function (t) {
   })
 })
 
-test('npm login --scope', function (t) {
-  mr({ port: common.port, plugin: mocks }, function (er, s) {
-    var scope = '@myco'
-    var runner = common.npm(
-      [
-        'login',
-        '--registry', common.registry,
-        '--loglevel', 'silent',
-        '--userconfig', outfile,
-        '--scope', scope
-      ],
-    opts,
-    function (err, code) {
-      t.notOk(code, 'exited OK')
-      t.notOk(err, 'no error output')
-      var config = fs.readFileSync(outfile, 'utf8')
-      t.like(config, /:always-auth=false/, 'always-auth is scoped and false (by default)')
-      s.close()
-      rimraf(outfile, function (err) {
-        t.ifError(err, 'removed config file OK')
-        t.end()
-      })
-    })
+test('npm login --scope <scope> uses <scope>:registry as its URI', function (t) {
+  var port = common.port + 1
+  var uri = 'http://localhost:' + port + '/'
+  var scope = '@myco'
+  common.npm(
+    [
+      'config',
+      '--userconfig', outfile,
+      'set',
+      scope + ':registry',
+      uri
+    ],
+  opts,
+  function (err, code) {
+    t.notOk(code, 'exited OK')
+    t.notOk(err, 'no error output')
 
-    var message = 'Logged in as u to scope ' + scope + ' on ' + common.registry + '/.'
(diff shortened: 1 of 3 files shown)


othiym23:
Thanks for putting this together, and thanks for cleaning up the tests!

Two bits of feedback:

1. Anything using color must observe the `--color` / `--no-color` configuration option used elsewhere in the code base. Also, I'm not sure that color is super useful in this case, as it's a simple single line of output. Consider ditching that.
2. Take a look at the resolution to @zkat's resolution of https://github.com/npm/npm/issues/10775#issuecomment-171468535 – it would be useful to have something like 

    ```
    Logged in as othiym23 on https://registry.npmjs.org/.
    ```

    or 

    ```
    Logged in as othiym23 to scope @nothingness on https://registry.npmjs.org/.
    ```

    which is a nice visual confirmation of what the user actually did.

If you can work those bits in, we'll get this merged in the next following release. Thanks again for your time!

ekmartin:
That seems better indeed.
Without a scope it now prints:
`Logged in as ekmartin on https://registry.npmjs.org/.`

With a scope it prints:
`Logged in as ekmartin to scope @myco on https://registry.npmjs.org/.`

Also removed the color.

othiym23:
Fantastic! Pending Travis being green, this can be merged to both `npm@master` and `npm@2.x`.

test/tap/adduser-always-auth.js
```diff
@@ -46,19 +63,37 @@ test('npm login', function (t) {
...
-
-      var label = chunk.toString('utf8').split(':')[0]
-      runner.stdin.write(responses[label])
+    var message = 'Logged in as u on ' + common.registry + '/.'
+    runner.stdout.on('data', verifyStdout(runner, message, t))
+  })
+})
 
-      if (remaining === 0) runner.stdin.end()
+test('npm login --scope', function (t) {
```
ekmartin: @othiym23, maybe the `--scope` test should check a few other things as well? 
Examples:
* That `@scope:registry` is set to the registry URI in the config upon completion.
* That the correct registry URI is set when a scope is given.

I.e. this block of code: https://github.com/npm/npm/blob/master/lib/adduser.js#L133-L140 (by the way, what does the multiple argument `npm.config.get` do here (`npm.config.get('registry', 'cli')`)?

Could test the `@` prefixing of scopes too.

othiym23: > what does the multiple argument `npm.config.get` do here (`npm.config.get('registry', 'cli')`)?

npm uses a package called [`config-chain`](http://npm.im/config-chain) to set up the precedence hierarchy between the various configuration files, environment variables, and configuration options set as command-line arguments. It's not really widely-known, but you can query one of those sources in particular by passing its name to `config.get()` as the second parameter, as you see here. The reason for this is that for the specific case of `npm login`, we don't want a `scope` set in a config file to trigger the scope-specific behavior for login.

othiym23: Also, those tests sound like great ideas. You have a few days until we're going to be in a position to merge this change, so if you want to try other approaches and refine the tests, that would be greatly appreciated. Just let us know, so we can re-review before merging. Thanks for your help!

ekmartin:
Added a few tests for `--scope` in a7f9576.

iarna:
This was merged to 3.6.0 & 2.14.16! =)

ekmartin:
Great!

Issue #9696, opened 2015-09-21
npm install EPERM renaming package.json in cache


bryanmacfarlane:
installing ...
```
npm ERR! Windows_NT 6.3.9600
npm ERR! argv "C:\\Program Files\\nodejs\\\\node.exe" "C:\\Program Files\\nodejs\\node_modules\\npm\\bin\\npm-cli.js" "install" "--loglevel" "error"
npm ERR! node v0.12.7
npm ERR! npm  v2.11.3
npm ERR! path C:\Users\vsoagentuser\AppData\Roaming\npm-cache\clone-stats\0.0.1\package\package.json
npm ERR! code EPERM
npm ERR! errno -4048

npm ERR! Error: EPERM, rename 'C:\Users\vsoagentuser\AppData\Roaming\npm-cache\clone-stats\0.0.1\package\package.json'
npm ERR!     at Error (native)
npm ERR!  { [Error: EPERM, rename 'C:\Users\vsoagentuser\AppData\Roaming\npm-cache\clone-stats\0.0.1\package\package.json']
npm ERR!   errno: -4048,
npm ERR!   code: 'EPERM',
npm ERR!   path: 'C:\\Users\\vsoagentuser\\AppData\\Roaming\\npm-cache\\clone-stats\\0.0.1\\package\\package.json',
npm ERR!   parent: 'vinyl' }
npm ERR!
npm ERR! Please try running this command again as root/Administrator.

npm ERR! Please include the following file with any support request:
npm ERR!     C:\test\npm-debug.log
```

npm-debug.log located here 
https://gist.github.com/bryanmacfarlane/cc7f49bfd0c651465121

How I Reproduced...  Our customers (visual studio online builds on azure VMs) also frequently reproduce

  1. Create Azure Windows 2012 Server VM (can't repro with fast local SSD dev box with Win10)
  2. Clear npm cache
  3. Use a package.json with a graph of many duplicated dependencies (bunch of gulp tasks is typical repro).  My package.json repro below

```json
{
  "name": "NpmRepro",
  "version": "1.0.0",
  "description": "NpmRepro",
  "main": "index.js",
  "author": "NpmReproer",
  "dependencies": {
    "del": "*",
    "gulp": "*",
    "gulp-angular-templatecache": "^1.7.0",
    "gulp-autoprefixer": "*",
    "gulp-concat": "*",
    "gulp-cli": "*",
    "gulp-if": "*",
    "gulp-less": "*",
    "gulp-minify-css": "*",
[...]

Issue #7885, opened 2015-04-08
Multiple npm install fails (EPERM error)


artemsvirskyi:
Im trying to install multiple packages in different child processes. And all the packages have some common dependency(browserify in my case).

	example
	|
	|_pkg1
	|   |_browserify
	|
	|_pkg2
	|   |_browserify
	|
	|_pkg3
	|   |_browserify
	|
	|_run.js

run.js:

	var packages = ['pkg1', 'pkg2', 'pkg3'];

	packages.forEach(function(pkg){
		exec('npm install ./' + pkg);
	});

And `node run.js` fails with log

		npm ERR! Windows_NT 6.3.9600
		npm ERR! argv "c:\\Program Files\\nodejs\\\\node.exe" "c:\\Program Files\\nodejs\\node_modules\\npm\\bin\\npm-cli.js" "i
		nstall" "./pkg3"
		npm ERR! node v0.12.2
		npm ERR! npm  v2.7.4
		npm ERR! path C:\Users\Artem\AppData\Roaming\npm-cache\acorn\0.9.0\package\package.json
		npm ERR! code EPERM
		npm ERR! errno -4048

		npm ERR! Error: EPERM, rename 'C:\Users\Artem\AppData\Roaming\npm-cache\acorn\0.9.0\package\package.json'
		npm ERR!     at Error (native)
		npm ERR!  { [Error: EPERM, rename 'C:\Users\Artem\AppData\Roaming\npm-cache\acorn\0.9.0\package\package.json']
		npm ERR!   errno: -4048,
		npm ERR!   code: 'EPERM',
		npm ERR!   path: 'C:\\Users\\Artem\\AppData\\Roaming\\npm-cache\\acorn\\0.9.0\\package\\package.json',
		npm ERR!   parent: 'syntax-error' }
		npm ERR!
		npm ERR! Please try running this command again as root/Administrator.

Example: https://github.com/SakerONE/multiple-npm-install

Ann-Dederkal:
+1 for this issue
It's very critical for our project. Please, fix it.

tamirlancomas:
gyp ERR! build error 
gyp ERR! stack Error: `make` failed with exit code: 2
gyp ERR! stack     at ChildProcess.onExit (/usr/local/lib/node_modules/npm/node_modules/node-gyp/lib/build.js:267:23)
gyp ERR! stack     at ChildProcess.emit (events.js:98:17)
[...]

PR #10940, opened 2015-12-23
cache: ignore failures replacing package.json

orangemocha:
As reported in https://github.com/npm/npm/issues/9696 , `npm install` fails frequently on slow Windows machines.

This happens because npm uses `write-file-atomic` to overwrite `package.json` in the cache, and it is not possible to write atomically to a file in Windows as it is done in UNIX. To replace a file, it cannot be open at the same time, either for reading or for other calls to `write-file-atomic`.

Failures writing to `package.json` are already ignored in [another place](https://github.com/npm/npm/blob/v2.14.13/lib/cache/caching-client.js#L183). Not saving cache information here does not seem to be an issue, but can there be a reason why we might not want to do this? @othiym23

Fixes: https://github.com/npm/npm/issues/7885
Fixes: https://github.com/npm/npm/issues/9696

### DIFF
--- a/lib/cache.js
+++ b/lib/cache.js
@@ -337,6 +337,7 @@ function afterAdd (cb) { return function (er, data) {
 
   // Save the resolved, shasum, etc. into the data so that the next
   // time we load from this cached data, we have all the same info.
+  // Ignore if it fails.
   var pj = path.join(cachedPackageRoot(data), "package", "package.json")
 
   var done = inflight(pj, cb)
@@ -347,7 +348,7 @@ function afterAdd (cb) { return function (er, data) {
     if (er) return done(er)
     writeFileAtomic(pj, JSON.stringify(data), {chown : cs}, function (er) {
       if (!er) log.verbose("afterAdd", pj, "written")
-      return done(er, data)
+      return done(null, data)
     })
   })
 }}


iarna:
I wonder if, instead of just ignoring this, that `write-file-atomic` should retry for a little bit if it gets EPERM errors during rename (in the same sort of way that `graceful-fs` provides for other fs calls). Or maybe we should just add rename to `graceful-fs` (which `write-file-atomic` already takes advantage of).

othiym23:
Alexis, could you rebase this against the latest `2.x` head so we can see how it's doing with CI? Thanks! (Also, sorry for letting this languish for a bit. We've been busy!)

joaocgreis:
@othiym23 last commit on `2.x` was on [December 11](https://github.com/npm/npm/tree/2.x@%7B2016-01-15%7D), so this is still up to date.

As suggested by @iarna, submitted a PR to `graceful-fs`: https://github.com/isaacs/node-graceful-fs/pull/55 . That should solve https://github.com/npm/npm/issues/9696 , but not when multiple npm instances run simultaneously as in https://github.com/npm/npm/issues/7885 . So this should still land (unless saving the cache here is essential for some reason).

othiym23:
Hmm... I guess the changes @zkat has made to get the tests passing for `2.x` haven't been merged to the branch yet; I'd like to wait for a passing build before we merge this, as this is a significant change. That should happen as part of next week's release, so we'll circle back to this after that's gone out.

iarna:
This was merged to 3.6.0 & 2.14.16! =)

orangemocha:
Awesome, thank you!

Issue #8770, opened 2015-06-30
ERR 404 while installing private scoped packages


mrcrgl:
Hey there,

I actually have a problem and cannot explain it on my own. Maybe someone of you.

I have a project which depends on scoped (private) packages and public ones. For continuous integration, I've put a `.npmrc` in the project root with contents like:

```
//registry.npmjs.org/:_authToken=(secret)
```

This works on my local maschine (npm 2.7.4, node 0.12.2) fine (cannot reproduce) but on codeships shell (tried many versions, but at least those from my local maschine) i'll get a 404 on my scoped packages when using `npm install`. The clou: If I try `npm install @account/package` it works. After installing ~2 packages manually, `npm install` works fine and installs the rest without errors. This is weird.

Hope someone would help me.

Thanks in advance!
Cheers

hapontemloreto:
It looks like scoped packages being installed using the tarball are failing because the url is not being encoded: @scope/package -> @scope%2fpackage.

I was able to repro this on my dev setup where we have our own private couchdb.

0 info it worked if it ends with ok
1 verbose cli [ 'C:\\Program Files\\nodejs\\node.exe',
1 verbose cli   'C:\\Users\\huaponte\\AppData\\Roaming\\npm\\node_modules\\npm\\bin\\npm-cli.js',
1 verbose cli   'install',
1 verbose cli   '@aria/commands',
1 verbose cli   '-g' ]
2 info using npm@3.5.2
3 info using node@v4.1.2
4 silly loadCurrentTree Starting
5 silly install loadCurrentTree
6 silly install readGlobalPackageData
7 silly fetchPackageMetaData @aria/commands
8 silly fetchNamedPackageData @aria/commands
9 silly mapToRegistry name @aria/commands
10 silly mapToRegistry scope (from package name) @aria
11 silly mapToRegistry registry http://arianpmprivate.cloudapp.net:5984/
12 silly mapToRegistry uri http://arianpmprivate.cloudapp.net:5984/@aria%2fcommands
13 verbose request uri http://arianpmprivate.cloudapp.net:5984/@aria%2fcommands
14 verbose request no auth needed
15 info attempt registry request try #1 at 10:40:00 AM
16 verbose request id a790a5b7e56f4878
17 verbose etag "9FGSLN3X9B5GJSOYL7GGVFAM8"
18 http request GET http://arianpmprivate.cloudapp.net:5984/@aria%2fcommands
[...]

PR #10947, opened 2015-12-23
Encoding package tarball url for scoped packages

hugoaponte:
We have our private couchdb registry to keep our scoped packages but also continuously syncing with the official registry.

I hit this issue: https://github.com/npm/npm/issues/8770

I worked in a fix and committed it to this pull request.

Please bear with me as this is my first time contributing to npm. I tried to validate the tests, but several tests were broken even before my changes.

Please share any feedback.

### DIFF
--- a/test/tap/enconding-tar-ball-for-scoped-packages.js
+++ b/test/tap/enconding-tar-ball-for-scoped-packages.js
@@ -1,0 +1,80 @@
+'use strict'
+var test = require('tap').test
+var npm = require('../../lib/npm')
+var stream = require('readable-stream')
+
+var moduleName = 'xyzzy-wibble'
+var testModule = {
+  name: moduleName,
+  'dist-tags': {
+    latest: '1.3.0'
+  },
+  versions: {
+    '1.3.0': {
+      name: moduleName,
+      version: '1.3.0',
+      dist: {
+        shasum: 'da39a3ee5e6b4b0d3255bfef95601890afd80709',
+        tarball: 'https://registry.npmjs.org/xyzzy-wibble/-/xyzzy-wibble-1.3.0.tgz'
+      }
+    }
+  }
+}
+
+var scopedModuleName = '@scope/xyzzy-wibble'
+var scopedTestModule = {
+  name: scopedModuleName,
+  'dist-tags': {
+    latest: '1.8.0'
+  },
+  versions: {
+    '1.8.0': {
+      name: scopedModuleName,
+      version: '1.8.0',
+      dist: {
+        shasum: 'da39a3ee5e6b4b0d3255bfef95601890afd80709',
+        tarball: 'https://registry.npmjs.org/@scope/xyzzy-wibble/-/@scope/xyzzy-wibble-1.8.0.tgz'
+      }
+    }
+  }
+}
+
+var lastTarball
+
+test('setup', function (t) {
+  npm.load({loglevel: 'silly'}, function (a, b) {
+    npm.registry = {
+      fetch: function (u, opts, cb) {
+        lastTarball = u
+
+        setTimeout(function () {
+          var empty = new stream.Readable()
+          empty.push(null)
+          cb(null, empty)
+        })
+      }
+    }
+    t.end()
+  })
+})
+
+var addNamed = require('../../lib/cache/add-named.js')
+test('verify_regular_module', function (t) {
+  t.plan(2)
+
+  // tarball for regular module is not encoded
+  addNamed('xyzzy-wibble', '*', testModule, function (err, pkg) {
+    t.error(err, 'Succesfully resolved regular module')
+    t.is(lastTarball, 'https://registry.npmjs.org/xyzzy-wibble/-/xyzzy-wibble-1.3.0.tgz')
+  })
+})
+
+test('verify_scoped_module', function (t) {
+  t.plan(2)
+
+  // tarball for scoped module is encoded
+  addNamed('@scope/xyzzy-wibble', '*', scopedTestModule, function (err, pkg) {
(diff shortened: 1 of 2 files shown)


iarna:
Thank you for this! This does need a test to reproduce the bug found in #8770.

(CI passes for this branch https://travis-ci.org/npm/npm/builds/98592627 so other than that this looks good.)

hapontemloreto:
Hi Rebecca

I put in place a couple of tests to cover this bug. Builds and tests are succeeding. It is my first time with tap test and npm tests, could you take a look?

Thanks in advance,

Hugo

hapontemloreto:
Hi Rebecca

Kindly ping. Thanks.

Hugo

test/tap/enconding-tar-ball-for-scoped-packages.js
```diff
@@ -0,0 +1,81 @@
...
+        tarball: 'https://registry.npmjs.org/@scope/xyzzy-wibble/-/@scope/xyzzy-wibble-1.8.0.tgz'
+      }
+    }
+  }
+}
+
+var lastTarball
+
+test('setup', function (t) {
+  npm.load(function (a, b) {
```
iarna: This and the next line can be:
```
npm.load({loglevel: 'silly'}, function (a, b) {
````

Comment (hapontemloreto) [inline test/tap/enconding-tar-ball-for-scoped-packages.js]: Thanks. Done.

Comment (hapontemloreto) [inline test/tap/enconding-tar-ball-for-scoped-packages.js]: Hi

Thanks so much for the feedback. It is done.

Hugo

**Comment by hapontemloreto** [comment]
Hi

Thanks so much for the feedback. It is done.

Hugo

**Comment by iarna** [comment]
Test looks great! I'm scheduling this for this week.

**Comment by iarna** [comment]
So when I was writing the changelog entry for this I became concerned.

I'd been looking at these changes in the small, but now that I'm looking at the bigger picture I'm no longer convinced that this is something we can accept. 

Let me see if I understand correctly– I might be wrong here:

You have a private registry that returns distribution URLs that can't be fetched as is– they require some of the URL that they return to be escaped? Is that correct?

Because if that's the case, that's _definitely_ a bug in your registry and should be fixed there. `npm` should NOT be mangling the URLs it gets back from registries, it should be handling them as-is.

**Comment by iarna** [comment]
The reason I would insist that you fix the bug in your registry, rather than working around it here, is that this potentially could break other third party registries that return it unescaped on purpose. We support all registry products, so it needs to not just break the `npm` registry but also the many varied third party ones. (There are several open source ones and several proprietary ones.)

**Comment by hapontemloreto** [comment]
Thanks for the feedback. I will review in detail. I am not sure the issue is on the registry. I am using the package https://github.com/npm/npm-registry-couchapp, which is actually being deprecated.

I see what you meant about the distribution URL. I am not sure the json that goes under .versions[X.Y.Z] in the registry is created by npm when publishing the package or by the registry when receiving it. Any hint on that would be appreciated.

I will take a look again this week.

Thanks

Hugo

**Comment by iarna** [comment]
I'm definitely surprised that `npm-registry-couchapp` would be doing that, and despite it's kind of deprecated status, if you can find a reproducer for this I'd be happy to work with you in fixing this over there.

**Comment by a-hegerath** [comment]
Sorry for jumping into this discussion, but I recently ran into this issue, too, and I am not sure if it is really only a bug in the registry. Saying that, I should add that I am quite new to npm so I might be wrong.

I got the error with the wrong URLs for scoped packages in private registries with a different registry (Sonatype Nexus). As far as I can see, for the scoped package "@angular/compiler" npm generates a URL that ends with "@angular%2fcompiler":

>npm install --verbose --always-auth true
npm verb cli [ 'D:\\ah\\nodejs\\node.exe',
npm verb cli   'D:\\ah\\nodejs\\node_modules\\npm\\bin\\npm-cli.js',
npm verb cli   'install',
npm verb cli   '--verbose',
npm verb cli   '--always-auth',
npm verb cli   'true' ]
npm info using npm@3.9.5
npm info using node@v6.2.2
npm verb mapToRegistry no registry URL found in name for scope @angular
npm verb mapToRegistry no registry URL found in name for scope @angular
npm verb mapToRegistry no registry URL found in name for scope @angular
npm verb mapToRegistry no registry URL found in name for scope @angular
npm verb mapToRegistry no registry URL found in name for scope @angular
npm verb mapToRegistry no registry URL found in name for scope @angular
npm verb request uri https://<host>/nexus/content/groups/npm_public_group/@angular%2fcompiler

While the public registry deals with this fine, my private registry returns for a GET request some package information with invalid tarball URLs, which leads to the 404 error. This seems like a bug in the registry. But for a GET with the proper URL (that ends with '@angular/compiler' instead), the returned package information contains valid tarball URLs. 

To me the question is, why does npm not apply URL encoding for the slash in the scoped package name in any case? Would this break compatibility or raise some other issues?

Thanks,
Andre

**Comment by iarna** [comment]
@a-hegerath It looks to me like that isn't your problem. Rather something is off about your registry URL because:

```
https:///nexus/content/groups/npm_public_group/@angular%2fcompiler
```
Is not a valid URL. There's no hostname there... it should be `https://HOSTNAME/path/path/path`.

> To me the question is, why does npm not apply URL encoding for the slash in the scoped package name in any case? Would this break compatibility or raise some other issues?

As I said before, if YOUR registry told us that it shouldn't be encoded, we aren't going to fiddle with that. The registry you're using needs to return URLs that can be fetched. How tarball paths are built is entirely up to the registry hosting them and npm has no insight into them. As far as npm is concerned those are opaque strings with no meaning at all. They could be shasums without package names in them for all npm knows.

iarna:
(wrong button =D)

iarna:

PR #10958, opened 2015-12-25
Change working directory for `npm run scriptname` in pkg.json

eljefedelrodeodeljefe:
This commit adds the above mentioned feature by checking whether the package
JSON includes a script in the form of...

```
{
  "scripts": {
    "change-cwd-test": {
       "cwd": "test/fixtures/",
       "run": "node -e \"console.log(process.cwd())\""
    }
   }
}
```

It then attaches a un-enumerable array to the `scripts` objects, which the `npm`
client can check for during execution.

From the `npm` client's POV, this seems a little odd since it needs to transform
a lot. But in order to pass hard checks here and w/ `npm` and not to break, this
seemed to be necessary for the author.

**Background:**
This described in greater detail the referenced issue. Fundamentally though, 
this for enabling `npm` to be more of a build tool. For example some programs 
such as make don't allow non-cwd runs. And in order to not `cd` to it, this is 
supposed to greatly help.

Credit: @eljefedelrodeodeljefe
Reviewed-By:
ISSUE: npm#10957

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -343,3 +343,4 @@ ekmartin <mail@ekmartin.com>
 Rafał Pocztarski <r.pocztarski@gmail.com>
 Ashley Williams <ashley666ashley@gmail.com>
 Mark Reeder <mreeder@uber.com>
+Robert Jefe Lindstaedt <robert.lindstaedt@gmail.com>
--- a/doc/cli/npm-run-script.md
+++ b/doc/cli/npm-run-script.md
@@ -30,6 +30,21 @@ environment variables that will be available to the script at runtime. If an
 "env" command is defined in your package it will take precedence over the
 built-in.
 
+It is also possible to run certain scripts in other working directories. This is
+enabled by `node`s `child_process` `cwd`-option. For that you need specify a
+run-script in the `scripts` objects as follows:
+
+```
+{
+  "scripts": {
+    "change-cwd": {
+      "cwd": "test/fixtures/",
+      "run": "node -e \"console.log(process.cwd())\""
+      }
+  }
+}
+```
+
 In addition to the shell's pre-existing `PATH`, `npm run` adds
 `node_modules/.bin` to the `PATH` provided to scripts. Any binaries provided by
 locally-installed dependencies can be used without the `node_modules/.bin`
--- a/lib/run-script.js
+++ b/lib/run-script.js
@@ -54,6 +54,22 @@ function runScript (args, cb) {
 
   readJson(path.resolve(pkgdir, 'package.json'), function (er, d) {
     if (er) return cb(er)
+
+    // alter pkgdir here
+    if (d.scripts && d.scripts.hasOwnProperty('_cwdChanges')) {
+      var script = []
+      if (d.scripts._cwdChanges.length > 0) {
+          // there can be only one element here. So filter it...
+        script = d.scripts._cwdChanges.filter(function (el) {
+          return el.cmd === cmd
+        })[0]
+        if (script) {
+          // ...so get it here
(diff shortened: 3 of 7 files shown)


iarna:
I'm not sure I wholly understand the problem you're trying to solve. In particular, what does your solution get you that the following doesn't?

```
{
  "scripts": {
    "change-cwd-test": "cd test/fixtures && node -e \"console.log(process.cwd())\""
   }
}
```

iarna:
I mean, I see you say "in order to not cd" but I don't understand why you find that undesirable?

eljefedelrodeodeljefe:
It's not portable to Windows for example. I first encountered problems and the need for it, when I wanted to do `gyp` related work.

eljefedelrodeodeljefe:
I guess it might render `npm` scripts a better build runner. This might have gained some traction [here](http://stackoverflow.com/questions/30286498/change-working-directory-for-npm-scripts).

Whole thing is not overwhelming I agree :)

iarna:
But it _is_ portable to windows– windows can cd into directories with forward slashes, AND it can combine commands to run with `&&`

eljefedelrodeodeljefe:
Just took some reading on this and you're quite right. It's really inconsistent on NT, since `/[word]` is reserved for command-line argument parsing, but if this particular "false flag" is not available it falls back to interpreting it as a path.

Would moving this to close.

eljefedelrodeodeljefe:
@iarna just a quick follow-up: my [question on SO](http://stackoverflow.com/questions/30286498/change-working-directory-for-npm-scripts/35802161#35802161) still has quite the traffic. So maybe you want to reconsider in the future and/or in broader scope: "npm as build tool". Best.

tjwebb:
Would also love this feature.

Pomax:
Has this work died off? This would still be a valuable feature to have, especially for tasks that need to run as part of preinstall and postinstall.

antony:
Just wanted to add that I'm trying to do exactly this with a new monorepo I'm building, and travis.

Ruby/Bundler can do it - and all the docs pertain to that, but I'm struggling with node:

```
language: node_js
env:
  - TEST_DIR=packages/core
  - TEST_DIR=packages/transports
  - TEST_DIR=packages/widget-ci
  - TEST_DIR=packages/widget-gauge
  - TEST_DIR=packages/widget-progress
  - TEST_DIR=packages/widget-statistic
  - TEST_DIR=packages/widget-time
```

Being able to put $TEST_DIR as the cwd for a script would be excellent here.

Pomax:
you can, but the way it works "out of the box" is just kind of hacky:
```
  "scripts": {
    ...
    "dothing": "cd subdir && dothing && cd ..",
    "start": "npm run dothing && node app",
    ...
  }
```

zxqfox:
> is just kind of hacky:

Feels like it should restore CWD after finish. Should we make an issue for that?

Pomax:
Did a quick test:
```
  "scripts": {
    "test": "npm run indir && touch testfile2",
    "indir": "cd dir1 && touch testfile"
  },
```
and results suggest it doesn't need the `cd ..`

zxqfox:
In npm v2 it still does

Pomax:
If you're still on npm 2, you kinda need to upgrade.

zxqfox:
@Pomax I can't. I'm using node v4.5.0 with bundled npm and can't update it atm. Sorry.

Pomax:
fair enough - note that Node 4.x has gone into maintenance-only and will be out of support in less than a year - given that the current version of npm is v4 (bundled with Node 7) I doubt any updates to 2 will still happen, but that'll be someone else's call.

PR #11036, opened 2016-01-05
INFO is missing

alunenko:
cmd INFO is missing


KenanY:
@alunenko `-d` is shorthand for `--loglevel info` so I do not think this change is proper. Do you agree?

alunenko:
+

PR #11037, opened 2016-01-05
Correct the name property max length constraint verbiage

scottaddie:
The verbiage describing the maximum length of the `package.json` file's `name` property was incorrect, as it led one to believe that the maximum length is 213 characters. I've updated it to be consistent with what's documented [here](https://github.com/npm/validate-npm-package-name#naming-rules).


iarna:

iarna:
This was merged to 2.14.16! =)

PR #11042, opened 2016-01-06
cache: measure cache age from process start time

rmg:
The default cache-min time is 10 seconds, which is shorter than the
average install time. This means that if a package has multiple
dependencies that depend on the same package there is a good chance that
a default install will result in the sub-dependencies being fetched
multiple times instead of using the copy that was put in to the cache
earlier _in the same install_.

Extend the cache-min to be at least the current age of the node process
so that objects fetched more than cache-min seconds ago are not
considered stale unless they are also older than the process.

This should save a few seconds off any moderately complex package
installation.

### DIFF
--- a/lib/cache/caching-client.js
+++ b/lib/cache/caching-client.js
@@ -100,6 +100,7 @@ function get_ (uri, cachePath, params, cb) {
   var etag
   var lastModified
 
+  timeout = Math.max(timeout, process.uptime())
   timeout = Math.min(timeout, npm.config.get('cache-max') || 0)
   timeout = Math.max(timeout, npm.config.get('cache-min') || -Infinity)
   if (process.env.COMP_CWORD !== undefined &&
@@ -115,10 +116,12 @@ function get_ (uri, cachePath, params, cb) {
     if (stat && timeout && timeout > 0) {
       if ((Date.now() - stat.mtime.getTime()) / 1000 < timeout) {
         log.verbose('get', uri, 'not expired, no request')
+        log.silly('get', 'cache hit with timeout', timeout)
         delete data._etag
         delete data._lastModified
         return cb(null, data, JSON.stringify(data), { statusCode: 304 })
       }
+      log.silly('get', 'cache miss with timeout', timeout)
 
       if (staleOk) {
         log.verbose('get', uri, 'staleOk, background update')
--- a/test/tap/cache-process-uptime.js
+++ b/test/tap/cache-process-uptime.js
@@ -1,0 +1,67 @@
+var npm = require.resolve('../../')
+var test = require('tap').test
+var path = require('path')
+var rimraf = require('rimraf')
+var mkdirp = require('mkdirp')
+var mr = require('npm-registry-mock')
+var common = require('../common-tap.js')
+var cache = path.resolve(__dirname, 'cache-process-uptime')
+var execFile = require('child_process').execFile
+var pkg = path.resolve(__dirname, '..', 'packages', 'npm-test-test-package')
+var nm = path.resolve(pkg, 'node_modules')
+var server
+
+test('mock reg', function (t) {
+  rimraf.sync(cache)
+  mkdirp.sync(cache)
+  mr({ port: common.port }, function (er, s) {
+    server = s
+    t.pass('ok')
+    t.end()
+  })
+})
+
+test('npm install with cache-min shorter than install', function (t) {
+  rimraf.sync(nm)
+  execFile(process.execPath, [
+    npm,
+    '--cache=' + cache,
+    '--registry=' + common.registry,
+    '--loglevel=silly',
+    // has to be > 0 or caching is disabled, but less than uptime, and since
+    // node startup time is 10s of ms and this is 0.1ms, it should always be
+    // less than process.uptime()
+    '--cache-min=0.00001',
+    '--cache-max=Infinity',
+    'install', 'underscore@1.5.1'
+  ], {cwd: pkg}, function (err, stdout, stderr) {
+    t.ifErr(err, 'Should not error')
+    t.match(stderr, /^npm sill get cache (hit|miss) with timeout \d/m)
+    t.notMatch(stderr, /^npm sill get cache (hit|miss) with timeout 0\.00001$/m)
+    t.end()
+  })
+})
+
+test('npm install with cache-min longer than install', function (t) {
+  rimraf.sync(nm)
+  execFile(process.execPath, [
+    npm,
+    '--cache=' + cache,
+    '--registry=' + common.registry,
+    '--loglevel=silly',
+    '--cache-min=60',
+    '--cache-max=Infinity',
+    'install', 'underscore@1.5.1'
+  ], {cwd: pkg}, function (err, stdout, stderr) {
+    t.ifErr(err, 'Should not error')
+    t.match(stderr, /^npm sill get cache (hit|miss) with timeout 60$/m)
+    t.end()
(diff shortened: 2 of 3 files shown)


rmg:
This has been stewing in the back of my mind since a conversation on twitter a month ago: https://twitter.com/feross/status/673343126149324800

/cc @feross @ceejbot @sindresorhus

feross:
wow, nice find! :+1:

rmg:
:blush: just realized I needed to add this line before the min/max config checks, otherwise they are basically being ignored.

sindresorhus:
:+1:

iarna:
This looks great! I would like to see the documentation updated too to reflect this new behavior.

rmg:
@iarna @othiym23 docs update added in latest push.

othiym23:
After some discussion with @rmg on `#npm`, I think that it would be a mistake to land this without a test around the caching behavior. Given the lack of observability of the timeout value with the current code, the test is a little baroque, but I think this would work:

1. Add a new log message, at `silly` level, that gets printed when the request timeout isn't satisfied. Have it include `'with timeout', timeout` in the message so the timeout can be searched for.
2. Modify the existing log message for when the cache timeout isn't hit to include the timeout, with similar formatting
3. Write a test to run npm via child processes set up to run two installs against a package provided by `npm-registry-mock` with `--loglevel=silly --cache-min=-1 --cache-max=Infinity`.
4. Use `t.match` to verify that the second install's request timeout is neither -1 nor `Infinity`. (Don't try to pin it to specific values, because if things are working, it's going to be an arbitrary, small number.)

Does that seem doable?

rmg:
@othiym23 that sounds like it _should_ be reasonable, just not sure I'll have enough contiguous time to learn enough of the npm internals to do it myself.

othiym23:
If you root around in `test/tap` and look for tests that require `npm-registry-mock` and / or use `common.npm()`, it should be relatively straightforward to figure out how to write the tests. Aside from those two small changes to logging, you shouldn't need to touch npm's internals.

rmg:
in case I forget to add it to a commit message:
```
before:
real	7m6.014s
user	3m38.635s
sys	0m24.925s

after:
real	6m38.291s
user	3m37.446s
sys	0m23.839s
```

rmg:
@othiym23 not _quite_ what you described, but I think it does the job. PTAL.

lib/cache/caching-client.js
```diff
@@ -115,10 +116,12 @@ function get_ (uri, cachePath, params, cb) {
     if (stat && timeout && timeout > 0) {
       if ((Date.now() - stat.mtime.getTime()) / 1000 < timeout) {
         log.verbose('get', uri, 'not expired, no request')
+        log.silly('cache hit with timeout', timeout)
```
othiym23: Add a prefix of `'get'` in this call so the `npmlog` output looks consistent.

test/tap/cache-process-uptime.js
```diff
@@ -0,0 +1,67 @@
...
+    // has to be > 0 or caching is disabled, but less than uptime, and since
+    // node startup time is 10s of ms and this is 0.1ms, it should always be
+    // less than process.uptime()
+    '--cache-min=0.00001',
+    '--cache-max=Infinity',
+    'install', 'underscore@1.5.1'
+  ], {cwd: pkg}, function (err, stdout, stderr) {
+    t.ifErr(err, 'Should not error')
+    t.match(stderr, /^npm sill cache (hit|miss) with timeout \d/m)
+    t.notMatch(stderr, /^npm sill cache (hit|miss) with timeout 0\.00001$/m)
```
othiym23: There's a (nigh-infintesimal) chance that an ordinary run could end up with the timeout set on the CLI, but I think this is reasonable for now and we'll come up with something else if it turns out to be flappy in practice.

othiym23:
@rmg That's a great start! The only thing that's missing is verifying that setting the timeout to `process.uptime()` is actually having an effect on installs in the test. That's why I suggested running the install twice in the first place – that the tests against the second install command can be guaranteed to get a cache hit. Maybe try swapping the order of the two tests? And seeing if things can be set up such that the test is guaranteed to get either a test or a failure? Since the test has its own cache, this shouldn't be too tough to make deterministic.

rmg:
@othiym23 I've added the `'get'` prefixes to the log messages.

I'm still trying to figure out a reasonable (by my own standards) way of testing the effect on cache usage during install.

zkat:
This seems to be failing on `node@<=0.10`. Could you take a look at the failure?

iarna:
@rmg The test _should_ be passing– the output has the string the regexp is looking for. I'm suspicious that it may be a change/bug in the earlier v8's regexp engine, so perhaps that log scan could be done differently? May have to involve looping over the lines of the output… =/

iarna:
(This does fail consistently on 0.10 and 0.8– both of which npm supports, so those will have to be corrected before we can merge this.)

lib/cache/caching-client.js
```diff
@@ -115,10 +116,12 @@ function get_ (uri, cachePath, params, cb) {
     if (stat && timeout && timeout > 0) {
       if ((Date.now() - stat.mtime.getTime()) / 1000 < timeout) {
         log.verbose('get', uri, 'not expired, no request')
+        log.silly('get', 'cache hit with timeout', timeout)
         delete data._etag
         delete data._lastModified
         return cb(null, data, JSON.stringify(data), { statusCode: 304 })
```
montanafox: N

iarna:
Rebased on to master this patch now fails on all versions of node.

jdalton:
Related https://twitter.com/kdzwinel/status/788132692063817729 \@kdzwinel

kdzwinel:
When yarn first came out I've been comparing it's network activity with npm's to see what's going on under the hood:

`npm i yo` - 5.3MB, 589 requests
npm

`yarn add yo` - 9.9MB, 564 requests
yarn

Besides the obvious "parallel vs sequential" thing it shows two oddities: yarn not using compression for metadata (it's fixed now) and npm is doing much more requests (as described by @rmg here).

After fixing yarn's issue I started investigating npm's but never figured out the root reason for these duplicate calls. My debugging process, that @jdalton linked above (thanks!), came to the same solution that @rmg describes here - increasing the timeout. That does work, but doesn't seem like a solid fix - we are depending on a magic number. IMO same metadata, once fetched from network/disc, should be used through the whole install process. That brings us to the second problem: increasing the timeout doesn't stop npm from reading this data from disc multiple times during install ( https://github.com/npm/npm/blob/6b6bb08af661221224a81df8adb0b72019ca3e11/lib/cache/caching-client.js#L85 ). Why npm is not storing all these data in memory? Well, it does: https://github.com/npm/npm/blob/2cdd7132abddcc7f826a355c14348ce9a5897ffe/lib/fetch-package-metadata.js#L105 . And that's as close as I could get to the actual problem. `regCache` does seem to work just fine, but there is some other thread that keeps fetching metadata from disk (and if it fails, e.g. due to said timeout, from network) for some reason. Maybe someone here will have a better idea what might be a problem. Hope this helps!

rmg:
npm v5 handles this so much better than I'm just going to close this.

zkat:
@rmg @kdzwinel I'm really curious what this flame graph looks like in `npm@5` now. I suspect we're still a bit iffy on the `fetchPackageMetadata` phase, but the tarball fetching section is, I bet, basically a solid block of concurrency. If it's not, that'd be interesting!

PR #11044, opened 2016-01-06
Update default value documentation for color variable

scottaddie:
The docs state that the default value of `color` is `false` on Windows. After executing the following command on 3 different Windows machines, I've found that the default value is actually `true`:

``` bat
npm config ls -l
```


iarna:

iarna:
This was merged to 2.14.16! =)

PR #11087, opened 2016-01-08
Consistently try git+https before git+ssh

gagern:
This change concludes what 394c2f5a1227232c0baf42fbba1402aafe0d6ffb started (and bd6919729f9535f4e3698e78eeb2331236dcdeea slightly improved): for hosted git services, try git first, then https, then ssh. Without this fix, the situation is somewhat messy if not all urls are supported. If there is no https url, ssh won't even be attempted. Conversely, if https fails and there is no ssh url, we'd end up in an infinite loop trying https over and over again.

### DIFF
--- a/lib/cache/add-remote-git.js
+++ b/lib/cache/add-remote-git.js
@@ -61,7 +61,7 @@ function addRemoteGit (uri, _cb) {
       return tryClone(from, parsed.toString(), false, cb)
     }
 
-    // try git:, then git+ssh:, then git+https: before failing
+    // try git:, then git+https:, then git+ssh: before failing
     tryGitProto(from, parsed, cb)
   } else {
     // verify that this is a Git URL before continuing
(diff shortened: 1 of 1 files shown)


gagern:
See also #9470 which questions the order chosen here. I still think that the current state should be fixed along the lines of this PR even if in the long run it might get replaced by something more flexible.

othiym23:
As @zkat said in #9470, it makes sense to change the priority order to `https:` → `ssh:` → `git:`. _Most_ users have access via HTTPs, _some_ have `ssh` set up (particularly in deployment environments)`, and _few_ need plaintext git access, but that needs to be there because of cases where users have conflicts between the auth deployed and protected access, but unauthenticated git still works.

If there are situations in which the absence of an HTTPs URL prevents the fallback to `git+ssh:`, those situations should be addressed. Swapping the precedence order again is going to cause more problems for users (every time we touch this code it generates a flurry of issues as formerly-working git workflows break).

For that reason, I'm not going to take this PR. Thanks for putting it together, and thanks for all of your attention to npm's git support. If you want to put together a patch to follow the precedence order above, it would be welcome, as long as it included tests and an update to the docs to make clear how that new precedence order works (as well, probably, as describing what it used to be, to help people who do run into the inevitable breakage understand what's happening).

othiym23:
In rereading things, I misread what this patch does, but I stand by my statement at the top of my last comment: the precedence order can change at most one more time, and that should be to  `https:` → `ssh:` → `git:`, for the reasons stated. My apologies for the confusion, and I do encourage you to rework what you've already started to incorporate the above (also you should probably reorder the functions in the file if you do so, so it's clearer to later maintainers what the precedence order is).

Issue #10594, opened 2015-11-29
npm install git+ssh://... --save


lc60005457:
When I use `npm install git+ssh://... --save`, the `package.json`

```
"dependencies": {
    "...": "ssh://...",
}
```

then, `npm install`, will error:

```
npm ERR! Unsupported URL Type: ssh://...
```

and, I need edit the `package.json`

```
"dependencies": {
    "...": "git+ssh://...",
}
```

KenanY:
@lc60005457 Is this a duplicate of https://github.com/npm/npm/issues/9679?

lc60005457:
@KenanY Yeah, it is. And I update npm to `v3.5.0` with mac. It also.

KenanY:
@lc60005457 Could you close this issue in favor of https://github.com/npm/npm/issues/9679 then? This way we can keep the issue tracker just a little cleaner :)

Issue #10284, opened 2015-11-06
npm writes git based ssh urls with incorrect url type


brodrigu:
when installing a package like so:

`npm install -S git+ssh://git@github.rp-core.com:UICore/css-roboto-font.git`

npm write to my package.json the following line:

`"css-roboto-font": "ssh://git@github.rp-core.com:UICore/css-roboto-font.git",`

then complains the next time I try to install something:

```bro:~/Code/Rubicon/CoreRepos/rx-property-modal (master)$ npm install -S git+ssh://git@github.rp-core.com:UICore/css-font-awesome.git
npm ERR! Darwin 15.0.0
npm ERR! argv "/Users/brodriguez/.nvm/versions/node/v4.2.2/bin/node" "/Users/brodriguez/.nvm/versions/node/v4.2.2/bin/npm" "install" "-S" "git+ssh://git@github.rp-core.com:UICore/css-font-awesome.git"
npm ERR! node v4.2.2
npm ERR! npm  v3.3.10

npm ERR! Unsupported URL Type: ssh://git@github.rp-core.com:UICore/css-roboto-font.git
npm ERR!
npm ERR! If you need help, you may report this error at:
npm ERR!     <https://github.com/npm/npm/issues>

npm ERR! Please include the following file with any support request:
npm ERR!     /Users/brodriguez/Code/Rubicon/CoreRepos/rx-property-modal/npm-debug.log```


no npm-debug.log was created.

KenanY:
Sounds like a duplicate of https://github.com/npm/npm/issues/9679. What do you think?

brodrigu:
Yep, didn't see that when I searched.

Issue #9873, opened 2015-10-07
npm install --save git+https://* add wrong entry to package.json


Moeriki:
`npm --version`

3.3.5

STEPS

``` bash
npm install --save git+https://myrepo.com/stash/scm/myproject/my-project.git
```

EXPECTED `package.json`

``` json
{
  "dependencies": {
    "my-project": "git+https://myrepo.com/stash/scm/myproject/my-project.git"
  }
}
```

ACTUAL `package.json`

``` json
{
  "dependencies": {
    "my-project": "https://myrepo.com/stash/scm/myproject/my-project.git"
  }
}
```

Could this possibly a bug?

I noticed ACTUAL when I ran `npm install`afterwards again for some other reason, and it failed contacting this repo.

KenanY:
Might be a dupe of https://github.com/npm/npm/issues/9679, https://github.com/npm/npm/issues/9732 (`git+` being removed).

Issue #9732, opened 2015-09-23
`git+` is dropped from `git+ssh`


donaldpipowitch:
It looks like calling `$ npm i --save-dev git+ssh://git@git.foo.lan:bar/baz.git#some-branch` will add `"baz": "ssh://git@git.foo.lan:bar/baz.git#some-branch"` to `package.json` (without `git+`). I haven't seen this with npm@2. I'm using npm@3.3.3 now.

We need `git+` or our CI server complaints with the following error:

````
npm ERR! typeerror Error: Missing required argument #3
npm ERR! typeerror     at exports.findRequirement (.../lib/node_modules/npm/lib/install/deps.js:475:3)
````

fischmax:
I can confirm this with node 4.1.1, npm 3.3.3 and Ubuntu 14.04.

KenanY:
Might be a dupe of https://github.com/npm/npm/issues/9679.

ehsalazar:
Closing as a duplicate of #9679. If you believe it's something different, please let us know. We'll be happy to reopen this issue.

PR #11091, opened 2016-01-08
Save rawSpec for modules from git

### LINKED ISSUE #9679: npm install git+ssh://.... --save problem
i use npm 3.3.3

when i  run  

```
npm install git+ssh://....  --save
```

the package.json file add this dependencies is "ssh://...."  git+ is disappeared

then i run npm update or npm install when the dependencies is update. npm say the ssh://... url is  unsupported.

gagern:
Since npm-package-arg strips a preceding `git+` from urls, and the stripped urls will cause trouble when loading the package the next time, we have to make sure that in this case at least we save the `rawSpec` not the `spec`.

I don't know enough about the difference between these two to judge whether we should _always_ be returning `rawSpec`. If I made the right choice, feel free to merge. If you think it should always be `rawSpec`, I can modify my PR.

Fixes #9679 and its many duplicates, namely #9732 #9873 #10284 #10594 (and perhaps some other unidentified ones as well).

### DIFF
--- a/lib/install/save.js
+++ b/lib/install/save.js
@@ -180,6 +180,8 @@ function computeVersionSpec (child) {
         pathname: relativePath
       })
     }
+  } else if (requested.type === 'git') {
+    return requested.rawSpec
   } else {
     return requested.spec
   }
--- a/test/tap/add-remote-git-file.js
+++ b/test/tap/add-remote-git-file.js
@@ -45,6 +45,29 @@ test('cache from repo', function (t) {
   })
 })
 
+test('save install', function (t) {
+  process.chdir(pkg)
+  fs.writeFileSync('package.json', JSON.stringify({
+    name: 'parent',
+    version: '5.4.3'
+  }, null, 2) + '\n')
+  var prev = npm.config.get('save')
+  npm.config.set('save', true)
+  npm.commands.install('.', [cloneURL], function (er) {
+    npm.config.set('save', prev)
+    t.ifError(er, 'npm installed via git')
+    var pj = JSON.parse(fs.readFileSync('package.json', 'utf-8'))
+    var dep = pj.dependencies.child
+    t.equal(
+      url.parse(dep).protocol,
+      'git+file:',
+      'npm didn\'t strip the git+ from git+file://'
+    )
+
+    t.end()
+  })
+})
+
 test('clean', function (t) {
   cleanup()
   t.end()


othiym23:
This needs a test, but this is a valuable fix that's been needed for a while, as you note. Thanks for putting this together!

gagern:
Amended commit with a test. Hope it's OK that I added this to an existing file.

Is my case distinction to only return `rawSpec` for git correct, or should we *always* be returning `rawSpec`?

iarna:
LGTM– and to your question… 

Let's review what's being specially handled already:
  tags, versions, ranges, directory & local
And you're adding git.

Missing from that are:
  tarball urls – (type=remote) these are the same either way
  hosted shortcuts – (type=hosted) eg, how you can do `foo/bar` to refer to a project on github– these get normalized to always have a prefix– that is, github shortcuts like `foo/bar` become `github:foo/bar` to match `bitbucket:foo/bar` and `gitlab:foo/bar`.

I think the behavior is good now, but if you wanted to add an `else if` for `type='hosted'` that used spec and used `rawSpec` as the fall through, I think that might be better?

gagern:
@iarna Since you ended your post with a question mark, I decided to see for myself whether your list is really exhaustive. I grepped for `(rawS|\.s)pec *=[^=]` and found the following:

* [view.js:86](https://github.com/npm/npm/blob/8dd78bbe6003cd377bf51bf24a1c5f8fd656265d/lib/view.js#L86): `nv.rawSpec = pkg.split('@')[pkg.indexOf('@')]` is for `view` so it doesn't end up in `save`.
* [realize-package-specifier/:32](https://github.com/npm/npm/blob/8dd78bbe6003cd377bf51bf24a1c5f8fd656265d/node_modules/realize-package-specifier/index.js#L32): `dep.rawSpec = dep.name` and [index.js:36](https://github.com/npm/npm/blob/8dd78bbe6003cd377bf51bf24a1c5f8fd656265d/node_modules/realize-package-specifier/index.js#L36): `if (dep.type == "local" || dep.type == "directory") dep.spec = specpath` does `local` & `directory`.
* [msvs_emulation.py:192](https://github.com/npm/npm/blob/8dd78bbe6003cd377bf51bf24a1c5f8fd656265d/node_modules/node-gyp/gyp/pylib/gyp/msvs_emulation.py#L192): `self.spec = spec`, [MSVSNew.py:132](https://github.com/npm/npm/blob/8dd78bbe6003cd377bf51bf24a1c5f8fd656265d/node_modules/node-gyp/gyp/pylib/gyp/MSVSNew.py#L132): `self.spec = spec` and [xcode_emulation.py:161](https://github.com/npm/npm/blob/8dd78bbe6003cd377bf51bf24a1c5f8fd656265d/node_modules/node-gyp/gyp/pylib/gyp/xcode_emulation.py#L161): `self.spec = spec` are likely false positives in gyp and unrelated.
* [npa.js:60](https://github.com/npm/npm/blob/8dd78bbe6003cd377bf51bf24a1c5f8fd656265d/node_modules/npm-package-arg/npa.js#L60): `res.rawSpec = arg` and [npa.js:61](https://github.com/npm/npm/blob/8dd78bbe6003cd377bf51bf24a1c5f8fd656265d/node_modules/npm-package-arg/npa.js#L61): `res.spec = arg` make them the same by default.
* [npa.js:90](https://github.com/npm/npm/blob/8dd78bbe6003cd377bf51bf24a1c5f8fd656265d/node_modules/npm-package-arg/npa.js#L90): `res.spec = version` for `version`
* [npa.js:93](https://github.com/npm/npm/blob/8dd78bbe6003cd377bf51bf24a1c5f8fd656265d/node_modules/npm-package-arg/npa.js#L93): `res.spec = range` for `range`
* [npa.js:99](https://github.com/npm/npm/blob/8dd78bbe6003cd377bf51bf24a1c5f8fd656265d/node_modules/npm-package-arg/npa.js#L99): `res.spec = arg` for `tag`
* [npa.js:106](https://github.com/npm/npm/blob/8dd78bbe6003cd377bf51bf24a1c5f8fd656265d/node_modules/npm-package-arg/npa.js#L106): `res.spec = "latest"` and [npa.js:107](https://github.com/npm/npm/blob/8dd78bbe6003cd377bf51bf24a1c5f8fd656265d/node_modules/npm-package-arg/npa.js#L107): `res.rawSpec = ""` are also `tag`
* [npa.js:125](https://github.com/npm/npm/blob/8dd78bbe6003cd377bf51bf24a1c5f8fd656265d/node_modules/npm-package-arg/npa.js#L125): `res.spec = path.resolve(arg)` is `local`
* [npa.js:132](https://github.com/npm/npm/blob/8dd78bbe6003cd377bf51bf24a1c5f8fd656265d/node_modules/npm-package-arg/npa.js#L132): `res.spec  = gitHost.toString(),` is `hosted`
* [npa.js:154](https://github.com/npm/npm/blob/8dd78bbe6003cd377bf51bf24a1c5f8fd656265d/node_modules/npm-package-arg/npa.js#L154): `res.spec = arg.replace(/^git[+]/, "")` for `git`, the relevant one for this issue
* [npa.js:160](https://github.com/npm/npm/blob/8dd78bbe6003cd377bf51bf24a1c5f8fd656265d/node_modules/npm-package-arg/npa.js#L160): `res.spec = arg` for `remote`
* [npa.js:165](https://github.com/npm/npm/blob/8dd78bbe6003cd377bf51bf24a1c5f8fd656265d/node_modules/npm-package-arg/npa.js#L165): `res.spec = urlparse.pathname` for `local`
* [npa.js:182](https://github.com/npm/npm/blob/8dd78bbe6003cd377bf51bf24a1c5f8fd656265d/node_modules/npm-package-arg/npa.js#L182): `Result.prototype.spec   = null` just sets the default

The hit that had me troubled the most was [npa.js:160](https://github.com/npm/npm/blob/8dd78bbe6003cd377bf51bf24a1c5f8fd656265d/node_modules/npm-package-arg/npa.js#L160) since it handles type `remote` which you said should not have a difference between `spec` and `specRaw`. But indeed that assignment there seems to be a no-op, overwriting a value with that same value again. So yes, at the moment the list is exhaustive.

I'd say merge as is.

However, for most of the protocols above, I get the feeling that `spec` is somehow more canonical than `rawSpec`. The `resolve` call for `local` is a prime example of this, and the handling of `hosted` as well. In that sense, I think it is better to keep the default as it is, referring to `spec`, and maintain `git` as the odd exception where the `rawSpec` contains more actually useful information than the `spec`.

iarna:
`rawSpec` is what the user entered and is always safe to save. By contrast, `spec` is used by our underlying network layer in conjunction with the other metadata in the object npa returns. So in git's case, it needed to return something git could understand, despite that being insufficient for npm to disambiguate git urls from other kinds, which is why this patch is needed. This is why I think `rawSpec` should be the fall through– we know it will work even if we know nothing else.

iarna:

PR #11092, opened 2016-01-08
Run install in git checkouts before adding them

### LINKED ISSUE #3055: run `prepublish` for git url packages

gagern:
With this change, adding a module from a git location is essentially the same as
- cloning and checking out the module into a temporary directory
- running `npm install` in there (so all devDependencies are there, and the prepublish script gets run)
- running `npm pack` and adding the resulting tarball to your project

First suggested by me in https://github.com/npm/npm/issues/3055#issuecomment-139665209, this idea saw a lot of positive feedback. The main benefits are:
- The repository is treated as a _source repository_, which means you'll have to assume that all dev dependencies and all prepublish scripts might be needed.  For every point and purpose, building from that temporary directory is just like what you'd do prior to publishing, which should work if the package is sufficiently portable.
- The permanently installed part is like a _published package_, though, with everything those have, but without anything `.npmignore` got rid of, and without unneccessary dev dependencies.  For every point and purpose, the installed package behaves just like what you'd get from the registry if it were installed there.

Fortunately, the “clone to temporary directory” part was already there. Furthermore, `npm pack` already involves creating a complete cache entry, and there is no need to copy the resulting tarball anywhere else the way pack does.  So we actually do less than `npm pack` would.  The main new addition is the call to `npm install` which was missing before.

This fixes #3055.

### DIFF
--- a/lib/cache/add-remote-git.js
+++ b/lib/cache/add-remote-git.js
@@ -346,6 +346,20 @@ function checkoutTreeish (from, resolvedURL, resolvedTreeish, tmpdir, cb) {
       }
       log.verbose('checkoutTreeish', from, 'checkout', stdout)
 
+      installAndPack(from, resolvedURL, tmpdir, cb)
+    }
+  )
+}
+
+// do the equivalent of "npm install && npm pack" for the package
+// https://github.com/npm/npm/issues/3055
+function installAndPack (from, resolvedURL, tmpdir, cb) {
+  log.info('addRemoteGit', 'running install in', tmpdir)
+  npm.commands.install(
+    tmpdir, [],
+    function (er) {
+      if (er) return cb(er)
+
       // convince addLocal that the checkout is a local dependency
       realizePackageSpecifier(tmpdir, function (er, spec) {
         if (er) {


rom1504:
:+1: 

I hope this PR finally fix that 3 years old issue that's breaking git dependencies !

gagern:
Browsing through various issues I just found that d0033645da55288688ef0fe2a9baa258130b6844 which addresses #7040 should fix this issue here as well. It includes the `install` step in `addLocalDirectory`, which is already on the codepath for git remotes. I just merged it into master and ran my current test setup, and it looks as well as with my changes here.

@iarna, do you plan to see that one merged any time soon, or should we take my approach here for the time being and remember to avoid the double install once your change (or some variant of that) makes it into master?

iarna:
My PR for that was here: https://github.com/npm/npm/pull/8725

iarna:
My concern there and here has to do with the global config object, for instance, if someone ran `npm install -g some-module-with-git-deps` then each of those deps would be installed globally. And this is just one flag as an example, there are many that change npm's behavior in subtle ways that we wouldn't want set for dependencies.

Obviously mucking about with the global config object during an install isn't safe, because other things are happening concurrently.

Now, the installer tries to only look at the global config object during initial setup, but that's currently an ambition and not a reality. Unfortunately fixing that is a rather large undertaking, thus my closing of my original PR until such time as we were ready to do it.

This is the sort of thing that should probably be done incrementally and not as a single thing. So if you'd like to help us work toward this, I would be interested in seeing PRs that rip out calls to `npm.config` from anything other than default setting in the top-level commands.

gagern:
@iarna Thanks for the explanation, I hadn't thought of these issues. For `-g` I'd maintain that the problem with not having this feature is still more serious, but that's just personal opinion, and I haven't considered all other flags.

@othiym23 What does the needs-bikeshedding tag mean here? Is it explained anywhere?

gagern:
Some more thoughts on config handling.

Do we really want to pass the `Installer` along to the cache, as [the roadmap](https://github.com/npm/npm/wiki/Roadmap-area-of-focus:-api) suggests, or would it be better to pass a `config` object? Since all we'd do with the `Installer` is accessing defaults read from the config, the latter seems more appropriate, more cleanly designed. It probably could be any object with a suitably defined `get` method.

If we pass configs, then we'd probably establish some way to clone, wrap or proxy them. In that case, should we inherit all settings, except for those blacklisted (i.e. reset to a fixed value) in a given scenario, or should we start with defaults only except for settings whitelisted for inheritance? Or should we somehow have some kind of metadata (e.g. by splitting the defaults object literal in two), indicating for each setting whether it should be inherited or not?

Splitting the list of config settings into two has some interesting semantic effect, I think. At least in my perception, those marked for inheritance would indicate settings which apply to `npm` as a whole, and therefore most suited to config file usage, whereas those not intended to be inherited are closer linked to the current command, and most likely passed as command line options. It might make sense to eventually document these in different places. The lack of command line option documentation associated with the individual commands has been bothering me for some time.

rustyconover:
:+1:

iarna:
We're not actually planning to pass the installer object into the cache. Rather, we want to rewrite the cache such that it doesn't have to know how to fetch resources. There's a cache rewrite that's been planned for a long time now, that we're hopeful we'll get to in 2016.

> The lack of command line option documentation associated with the individual commands has been bothering me for some time.

We would welcome PRs! Unfortunately _anything_ that can be a command line option can be a config file entry, and vice versa. So the one true source of knowledge is https://docs.npmjs.com/misc/config

othiym23:
There's consensus that applying the full install lifecycle to Git dependencies should happen, so something like this should land. However, this PR is stale, and depends on some fairly major redesign work that isn't going to happen in the immediate term, so I'm going to close this while at the same time encouraging someone (perhaps @gagern?) to pick up some of the work that @iarna describes and file either feature requests or new PRs including those pieces of work.

ELLIOTTCABLE:
(If any of you does so migrate this discussion and effort, please /cc me, if you would … there's no mechanism to continue to remain subscribed, and I'd really like to follow the progress of these problems / know when a concrete plan exists to wrap them up. Thanks! <3)

gagern:
Closing this PR is fine by me. I don't see me finding time to work on [the problems @iarna described](https://github.com/npm/npm/pull/11092#issuecomment-172062682) in the forseeable future, so I wouldn't at all mind if someone else were to pick this up.

I also must confess that I don't have a clear idea of what form that “planned cache rewrite” should be taking. Even my understanding of the cache as a whole is pretty sketchy at the moment. So perhaps someone who was involved in that planning can either see to its implementation, or open a feature request outlining what they had in mind, so others can work on it in the anticipated direction.

@ELLIOTTCABLE: the main feature request in #3055 is still open, and I'd guess that any major progress would be announced there as well. So I suggest you subscribe to that if you haven't already.

PR #11094, opened 2016-01-08
Recursively initialize and update submodules for git remotes

### LINKED ISSUE #1876: npm should init/update submodules when installing from git url
it would be nice if installing from a git url expanded the git submodules before running the install scripts.

gagern:
This is a fairly simple approach, which does not leverage the git caching mechansim to cache submodules. We might want to address that in some future enhancement, but doing so may take considerable time so we should get this one here landed first.

This fixes #1876. It technically (but not conceptually because the correct order is obvious) conflicts with #11092. If either one gets merged it should be fairly easy to rebase the other one, resolving the conflict manually, and I intend to do so in a timely fashion.

### DIFF
--- a/doc/cli/npm-install.md
+++ b/doc/cli/npm-install.md
@@ -169,9 +169,12 @@ after packing it up into a tarball (b).
 
           <protocol>://[<user>[:<password>]@]<hostname>[:<port>][:][/]<path>[#<commit-ish>]
 
-    `<protocol>` is one of `git`, `git+ssh`, `git+http`, or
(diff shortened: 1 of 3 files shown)


othiym23:
This needs a test, and it could use a small doc tweak to say that submodules will indeed be pulled down during the install process, but this is otherwise a good change. Somebody needs to do something similar for LFS support one of these days, too. 😈

Thank you for putting this together!

gagern:
Added test case and documentation.

iarna:
This was merged into 3.7.0!

PR #11112, opened 2016-01-11
add npm registry autocompletion

kolodny:
This uses https://constructor.io/ which is what https://www.npmjs.com/ uses for their autocomplete in their dropdown. This is mindful to abort any pending requests to the api if another autocomplete request happens, and can degrade nicely.

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -28,6 +28,7 @@ install.usage = '\nnpm install (with no args, in package dir)' +
                 '\n\nalias: npm i' +
                 '\ncommon options: [--save|--save-dev|--save-optional] [--save-exact]'
 
+var lastCompletionRequest
 install.completion = function (opts, cb) {
   validate('OF', arguments)
   // install can complete to a folder with a package.json, or any package.
@@ -82,9 +83,22 @@ install.completion = function (opts, cb) {
     })
   }
 
-  // FIXME: there used to be registry completion here, but it stopped making
-  // sense somewhere around 50,000 packages on the registry
-  cb()
+  var requestCallback = function (error, response, body) {
+    if (!error && response.statusCode === 200) {
+      try {
+        var suggestions = JSON.parse(body).sections.packages.map(function (item) {
+          return item.value
+        })
+      } catch (e) {
+        cb()
+      }
+      cb(null, suggestions)
+    } else {
+      cb()
+    }
+  }
+  if (lastCompletionRequest) lastCompletionRequest.abort()
+  lastCompletionRequest = request('https://ac.cnstrc.com/autocomplete/' + opts.word + '?autocomplete_key=2BhWUxtsm80gVBQ1jtTW', requestCallback)
 }
 
 // system packages
@@ -93,6 +107,7 @@ var path = require('path')
 
 // dependencies
 var log = require('npmlog')
+var request = require('request')
 var readPackageTree = require('read-package-tree')
 var chain = require('slide').chain
 var asyncMap = require('slide').asyncMap


othiym23:
Thanks for putting this together! Here's my thoughts:

- npm shouldn't hardcode dependencies on third-party services. If something like this were to be included, it would need to be configurable via an option, and at least until / unless constructor.io gave us an explicit go-ahead, would need to be something configured by the end-user.
- As much as possible, we're trying to keep the CLI decoupled from the bits that access the registry. Most of that functionality goes in `npm-registry-client` for now (which may be broken down into smaller components in the near future).

So:
- Add a configuration parameter for the completion URL (probably done as a format string, so that e.g. `'https://ac.cnstrc.com/autocomplete/%s?autocomplete_key=2BhWUxtsm80gVBQ1jtTW'` is possible) to `lib/config/defaults.js` (and with corresponding documentation in `doc/misc/npm-config.md`, which a test verifies is there).
- Add a new method (`autocomplete`?) to `npm-registry-client`, along with tests.
- Once we publish the new version of `npm-registry-client`, include that in the PR.
- Add tests to verify that the autocompletion is working (autocompletion is badly in need of tests, so we can probably budget some time to help figure out how to write them).

How does that sound?

kolodny:
Sounds good!

I have a question on the first suggestion point

The api currently returns an object that looks like

```json
{
  "sections": {
    "packages": [
      { "value": "item1" },
      { "value": "item2" },
      { "value": "item3" },
      { "value": "item4" }
    ]
  }
}
```
Should I also have a config value for the path to the autocomplete items:
```js
var autocompletePath = npm.config.get('autocompletePath') // 'sections.packages'
lodash.get(autocompletePath).map(item => item.value || item)
```
I would also need to add `lodash.get` as a dependency in that case

Or can I assume that any other api endpoint will conform to this response structure?

kolodny:
I've created a branch on [`npm-registry-client` on my own fork with this mostly done](https://github.com/kolodny/npm-registry-client/commit/b78520388854227d18325264d8c320c493d7f032). The only thing remaining is figuring out the best way to abstract the response-to-array method. I'm thinking of having a default key called `"autocomplete-api-to-array"` key with a default value of the function:  

```js
function (response) { return response.sections.packages.map(function (item) { return item.value }) }
```

similar to how it's done https://github.com/npm/npm/blob/master/doc/misc/npm-config.md#logstream and just pass it into the `autocompete` method

@othiym23 thoughts?

EliFinkelshteyn:
I'm Eli from Constructor.io, and I just wanted to mention explicitly we're happy to have our API used for this.

kolodny:
@othiym23 was there any update on this?

othiym23:
@kolodny The team's been running on a light release process lately, and this has fallen through the cracks as a result. The CLI team is trying to get npm's test suite improved enough that we can set up a Windows continuous integration instance. My apologies for not getting back to you sooner.

I'd suggest having a separate function that figures out how to map from the responses to the completion list, as something that could be turned into a plugin if necessary. That should probably live in `npm` rather than `npm-registry-client`, so that the thing that's doing the fetching of the data is decoupled from the thing that figures out how to present it to the end user. So, _two_ configurable / modular pieces: one to determine the pattern for the URLs to request, and one to take the return values and make them consumable completion targets.

The next tricky piece is going to be figuring out how to meaningfully test this, which will require you to take a look at how npm currently uses `npm-registry-mock` to set up mocked response paths (which is also a good reason to make the request URL configurable). The tests shouldn't talk to the network by default (although a separate integration test that verifies that `constructor.io` and the CLI are communicating properly with each other would be very useful).

Is that enough to get you unblocked?

othiym23:
I've removed the `needs-discussion` label and added `needs-docs`, because this is something we'll want to document as part of adding it.

iarna:
I'm closing this due to inactivity (It's been over a year!), but the idea is still sound. Things have changed a bit architecturally and so if someone want's try to pick this up a modern take would put this in a stand alone module that's built on `make-fetch-happen` and then implement around that. I'd love to see that in a new PR. If anyone does want to pick this up and needs help, you can find the cli team over on https://package.community in the #npm channel.

PR #11127, opened 2016-01-12
test: fix race condition in git tests

zkat:
Every git-related test should use a different port for its spawned git server. While we can listen to a git server's process shutting down, there's no guarantee that the socket will have been returned to the OS by then, so the next server might still get an EADDRINUSE.

### DIFF
--- a/test/tap/add-remote-git-fake-windows.js
+++ b/test/tap/add-remote-git-fake-windows.js
@@ -20,7 +20,7 @@ var pjParent = JSON.stringify({
   name: 'parent',
   version: '1.2.3',
   dependencies: {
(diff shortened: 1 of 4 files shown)


othiym23:
If we start to see weirdness around this, especially when we get to banging on the tests in Windows CI, we should go ahead and write a utility function to give each test a fresh git daemon port, but I suspect that that's more tricky than it seems (i.e. how do we ensure that we're not reusing a port over the course of a test run?) and I'm hopeful that this will solve the problem permanently.

:ship::shaved_ice::sunrise_over_mountains:

iarna:
This was merged to 3.6.0! =)

PR #11128, opened 2016-01-12
Add link to Local Paths section

orangejulius:
This section was referenced, so there might as well be a direct link!


othiym23:
Tiny but useful! Thank you! We'll get this out in the next releases of both `npm@3` and `npm@lts`.

iarna:

Issue #11088, opened 2016-01-08
Broken npm after updating to node 5.4.0 (Cannot find module 'brace-expansion')


franzip:
I have installed Node and NPM installed with Homebrew.
After updating to Node 5.4.0 I have run: ```npm install npm -g; npm update -g```
This seems to have broken NPM:

```
$ node -v
v5.4.0

$ npm -v
3.5.4

$ npm install
npm ERR! Darwin 14.5.0
npm ERR! argv "/usr/local/Cellar/node/5.4.0/bin/node" "/usr/local/bin/npm" "install"
npm ERR! node v5.4.0
npm ERR! npm  v3.5.4
npm ERR! code MODULE_NOT_FOUND

npm ERR! Cannot find module 'brace-expansion'
npm ERR!
npm ERR! If you need help, you may report this error at:
npm ERR!     <https://github.com/npm/npm/issues>

npm ERR! Please include the following file with any support request:
npm ERR!     /Users/franzip/Sandbox/npm-debug.log

$ cat /Users/franzip/Sandbox/npm-debug.log
0 info it worked if it ends with ok
1 verbose cli [ '/usr/local/Cellar/node/5.4.0/bin/node',
1 verbose cli   '/usr/local/bin/npm',
1 verbose cli   'install' ]
2 info using npm@3.5.4
3 info using node@v5.4.0
4 verbose stack Error: Cannot find module 'brace-expansion'
4 verbose stack     at Function.Module._resolveFilename (module.js:326:15)
4 verbose stack     at Function.Module._load (module.js:277:25)
4 verbose stack     at Module.require (module.js:354:17)
4 verbose stack     at require (internal/module.js:12:17)
4 verbose stack     at Object.<anonymous> (/usr/local/lib/node_modules/npm/node_modules/minimatch/minimatch.js:10:14)
4 verbose stack     at Module._compile (module.js:398:26)
4 verbose stack     at Object.Module._extensions..js (module.js:405:10)
4 verbose stack     at Module.load (module.js:344:32)
4 verbose stack     at Function.Module._load (module.js:301:12)
4 verbose stack     at Module.require (module.js:354:17)
4 verbose stack     at require (internal/module.js:12:17)
5 verbose cwd /Users/franzip/Sandbox
6 error Darwin 14.5.0
7 error argv "/usr/local/Cellar/node/5.4.0/bin/node" "/usr/local/bin/npm" "install"
8 error node v5.4.0
9 error npm  v3.5.4
[...]

PR #11129, opened 2016-01-13
The repo, as its checked in, has `npm ls` errors:

iarna:
```
npm ERR! missing: glob@^5.0.3, required by init-package-json@1.9.1
npm ERR! missing: has-unicode@^1.0.0, required by gauge@1.2.2
npm ERR! missing: has-unicode@^1.0.0, required by gauge@1.2.2
npm ERR! missing: has-unicode@^1.0.0, required by gauge@1.2.2
npm ERR! missing: glob@^5.0.3, required by read-package-json@2.0.2
```

This happened because of the `glob` and `has-unicode` major release updates, and went undetected because the actions necessary to run the tests (which check for this sort of thing) resolved the missing modules.

The version published to the registry is also missing these things, however if you're updating from a previous 3.x release, it will fix up the install when you do the update, installing the versions listed above in addition to the bundled deps. If you're upgrading from a 2.x release, it won't, but this won't break anything because the breaking changes in `glob` & `has-unicode` don't have negative effects for us, and node doesn't care about version numbers when loading modules.

This PR fixes this problem by updating `init-package-json` and `read-package-json` to use the new glob, and updating `gauge` to use the new `has-unicode`, and then, since `gauge` isn't a direct dependency of npm, reinstalling the existing versions of `npmlog` & `node-gyp` which were the two top level dependents that used `gauge`.


othiym23:
Rubberstamp :sheep: because the logic sounds good and the diff is too large for anything other than a double-check of `package.json`.

iarna:
Yeah, the diff is 90% removing duplicate copies of readable stream by weight.

iarna:
It looks like this breaks installs from 3.3.12 (https://github.com/npm/npm/issues/11088#issuecomment-172931543)

iarna:
This was merged to 3.5.5! =)

PR #11130, opened 2016-01-13
nuke "using npm programmatically" section

ljharb:
The programmatic `npm` API is unsupported, and is not guaranteed not to break in non-major versions.

Removing this section so newcomers aren't encouraged to discover or use it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,52 +138,6 @@ must remove them yourself manually if you want them gone.  Note that
 this means that future npm installs will not remember the settings that
 you have chosen.
 
-## Using npm Programmatically
-
-Although npm can be used programmatically, its API is meant for use by the CLI
-*only*, and no guarantees are made regarding its fitness for any other purpose.
-If you want to use npm to reliably perform some task, the safest thing to do is
-to invoke the desired `npm` command with appropriate arguments.
-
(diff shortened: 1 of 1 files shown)


othiym23:
This only really needs to be on `master` since it's mostly only seen on www.npmjs.com and GitHub, but we might as well remove it from both places because we don't support the API under `npm@lts` either.

iarna:

ljharb:
Thanks!

jenius:
So just out of curiosity, if you need to use npm programmatically, what is the current recommendation? To run it through the command line in a child process?

othiym23:
@jenius Yes. Eventually we'd like to have the CLI reduced to components that are independently usable to do all of the CLI's various tasks, but that's a very long-term project blocked on reworking how configuration is bootstrapped and managed within the current CLI implementation. Funnily enough, this is also part of what makes the current API unsuitable for third-party use.

jenius:
@othiym23 Totally understand how that would be a pretty large challenge. I'm sure whenever that does happen, the organization of the API will be much better, looking forward to it. Child process it is for now then. Thanks for the quick response :raised_hands:

PR #11142, opened 2016-01-13
install: fix race condition correcting cache directory ownership

Jimbly:
Previously, correctMkdir() would cache the results early within its operation (after the stat, before calling chownr) meaning that any second call would immediately return the cached results, before the earlier chownr finishes.
This causes a race condition where the initial chownr would fail with ENOENT (or similar errors) trying to scan files that were being actively created/deleted/etc by the install process.
The fix guards the whole correctMkdir function with inflight() so that multiple calls do not do chownr simultaneously nor return early.  It would also be reasonable to not cache the results until after the chownr has finished, however that would still lead to an even more subtle race condition, so the guard on the whole call is required.

This code only fires if the npm cache directory is not owned by the current user, so probably doesn't reproduce very often, though in one of my environments it was reproducing frequently.

I was able to get a relatively consistent failure with a large cache directory, using the Yahoo-wrapped NPM.  In vanilla NPM, I was unable to consistently reproduce this, so in that case it's possible some serial process makes a single call to correctMkdir before continuing with anything that might happen in parallel that might hit this race condition.  Regardless, I made a unit test which reproduces this on the old code and is resolved by the new code.

@bengl

### DIFF
--- a/lib/utils/correct-mkdir.js
+++ b/lib/utils/correct-mkdir.js
@@ -10,6 +10,13 @@ var stats = {}
 var effectiveOwner
 module.exports = function correctMkdir (path, cb) {
   cb = dezalgo(cb)
+  cb = inflight('correctMkdir:' + path, cb)
+  if (!cb) {
+    return log.verbose('correctMkdir', path, 'correctMkdir already in flight; waiting')
+  } else {
+    log.verbose('correctMkdir', path, 'correctMkdir not in flight; initializing')
+  }
+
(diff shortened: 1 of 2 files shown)


Jimbly:
Fixed coding style complaints from travis_ci about the new test, this is now ready for review.

othiym23:
This looks great! We'll get this into `npm@3` and `npm@lts` in their next respective releases. Thank you!

iarna:

Issue #10736, opened 2015-12-08
Slightly nicer way of doing a recursive package update?


halhenke:
Perhaps related to https://github.com/npm/npm/issues/7234

Basically a feature request for a flag that offers a nicer way of doing an old fashioned recursive update.

More or less equivalent to: 
`npm uninstall -S whatever`
`npm install -S whatever`

I know there is `npm --depth 9999 update` but it feels a bit hack-ish and requires one remembering/looking up the maximum depth - 9/10 I'll just end up going the uninstall root. 

So something thats more or less an alias for this
- Maybe a flag like `--recursive` or even a top level command...`npm deepUpdate whatever`?

zkat:
This doesn't seem any different from doing `npm update --depth Infinity`, which works just fine, and that's right in the docs. (see the description in `npm help update`, which also has some interesting information on the issues surrounding this usage).

Thanks for the submission!

othiym23:
(Also, with `npm@3`, the documentation is a little outdated – it should be perfectly safe to run `npm update --depth=Infinity`, and if somebody wanted to amend the docs to that effect, it would be much appreciated.)

halhenke:
But the difference is that `update --depth Infinity` will still respect semver in your package.json no? My usage scenario is that i often run `npm outdated` and see some sort of version bump that I want but would require me to either edit `package.json` by hand or to uninstall/reinstall. ~~As I understand it even specifically doing `npm i -S package@latest` wont override package.json.~~

OK i just tried out `npm i -S package@latest` & it seems i was wrong :disappointed: 

Still the `npm update --depth Infinity` thing is handy as `npm help update` only mentions the much less memorable `npm --depth 9999 update`. Am happy to change docs to reflect this. Thanks guys.

Aourin:
@othiym23 I am not sure if the above doc from Infinity to 9999 is the same as what you meant (by safe)

Here is a PR just in case to remove the issue.

PR #11150, opened 2016-01-14
doc: advise use of `--depth Infinity` instead of `--depth 9999`

halhenke:
Re #10736 - minor documentation change for `npm update`


iarna:

PR #11158, opened 2016-01-15
windows: fix custom node-gyp env var quoting

orangemocha:
npm sets node_config_node_gyp without quotes, so its usage should be
quoted. Indeed, it is a better practice to define environment variables
that contain paths without quotes and quote their usage.


othiym23:
Indeed. The current state of that line appears to be a simple oversight, if you look at the commit history touching that file. This will get landed on both the LTS and master branches soon. Thanks, Alexis!

orangemocha:
:smile:

iarna:

orangemocha:
Great news, thanks!

/cc @jianchun

Issue #11014, opened 2016-01-03
Error When Uninstalling Bluebird


albertreed:
I'm new to javascript, node and just about every tool I'm using so please know the appearance of ignorance is real. :-)

Running "npm uninstall bluebird --save" is asking me to post the error here - so that's what I'm doing. 

I've tried "npm cache clean"

COMMAND:
```
$ npm uninstall bluebird --save
npm ERR! Darwin 15.2.0
npm ERR! argv "/usr/local/bin/node" "/usr/local/bin/npm" "uninstall" "bluebird" "--save"
npm ERR! node v5.3.0
npm ERR! npm  v3.3.12
npm ERR! code EMISSINGARG

npm ERR! typeerror Error: Missing required argument #3
npm ERR! typeerror     at replaceModule (/usr/local/lib/node_modules/npm/lib/install/deps.js:427:3)
npm ERR! typeerror     at /usr/local/lib/node_modules/npm/lib/install/deps.js:236:7
npm ERR! typeerror     at Array.forEach (native)
npm ERR! typeerror     at exports.removeDeps (/usr/local/lib/node_modules/npm/lib/install/deps.js:232:8)
npm ERR! typeerror     at Array.<anonymous> (/usr/local/lib/node_modules/npm/node_modules/slide/lib/bind-actor.js:15:8)
npm ERR! typeerror     at LOOP (/usr/local/lib/node_modules/npm/node_modules/slide/lib/chain.js:15:14)
npm ERR! typeerror     at chain (/usr/local/lib/node_modules/npm/node_modules/slide/lib/chain.js:20:5)
npm ERR! typeerror     at Uninstaller.loadAllDepsIntoIdealTree (/usr/local/lib/node_modules/npm/lib/uninstall.js:72:3)
npm ERR! typeerror     at Array.<anonymous> (/usr/local/lib/node_modules/npm/node_modules/slide/lib/bind-actor.js:15:8)
npm ERR! typeerror     at LOOP (/usr/local/lib/node_modules/npm/node_modules/slide/lib/chain.js:15:14)
npm ERR! typeerror This is an error with npm itself. Please report this error at:
npm ERR! typeerror     <http://github.com/npm/npm/issues>

LOG FILE:
0 info it worked if it ends with ok
1 verbose cli [ '/usr/local/bin/node',
1 verbose cli   '/usr/local/bin/npm',
1 verbose cli   'uninstall',
1 verbose cli   'bluebird',
1 verbose cli   '--save' ]
[...]

PR #11171, opened 2016-01-16
removeDeps: remove package even if rmed

davidvgalbraith:
If you rm -rf a package before you uninstall --save it, it doesn't
get into tree.children. This causes removeDeps to fail with a stack
trace. Instead, removeDeps should create a module-like object with
the name of the removed package.

Fixes https://github.com/npm/npm/issues/11014

### DIFF
--- a/lib/install/deps.js
+++ b/lib/install/deps.js
@@ -231,7 +231,11 @@ exports.removeDeps = function (args, tree, saveToDependencies, log, next) {
     var pkgName = moduleName(pkg)
     if (saveToDependencies) {
       var toRemove = tree.children.filter(moduleNameMatches(pkgName))
-      replaceModule(tree, 'removed', toRemove[0])
+      var pkgToRemove = toRemove[0] || {
+        name: pkgName,
+        save: saveToDependencies
+      }
+      replaceModule(tree, 'removed', pkgToRemove)
       toRemove.forEach(function (parent) {
         parent.save = saveToDependencies
       })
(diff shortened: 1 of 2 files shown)


othiym23:
@iarna should take a look at this, because she's been the one hammering on this code the most recently. Thanks for putting this together, and thanks for writing a test!

lib/install/deps.js
```diff
@@ -231,7 +231,11 @@ exports.removeDeps = function (args, tree, saveToDependencies, log, next) {
     var pkgName = moduleName(pkg)
     if (saveToDependencies) {
       var toRemove = tree.children.filter(moduleNameMatches(pkgName))
-      replaceModule(tree, 'removed', toRemove[0])
```
iarna: I would replace this whole section with…
```
var pkgToRemove = toRemove[0] || createChild({package: {name: pkgName}})
replaceModule(tree, 'removed', pkgToRemove)
pkgToRemove.save = saveToDependencies

iarna:
The test looks good though and I'm happy to just fixup the commit part as I suggested above. So given that, LGTM.

davidvgalbraith:
Hey @othiym23 @iarna, anything else I need to do to get this one across the finish line?

iarna:
@davidvgalbraith Nope, once an issue has a `next` or `next-next` milestone, it means it's going into either that week or the following week's release, respectively, and it just has to make its way through our process.

davidvgalbraith:
Got it, thanks!

iarna:

Issue #10938, opened 2015-12-23
Empty .bin directories are not removed and result in ENOENT errors


SystemParadox:
With npm@3 I am regularly getting ENOENT errors.

When doing `npm remove` and seemingly also when doing `npm install` (perhaps because npm is moving things around in the tree?), I end up with empty `.bin` directories like this:

```
node_modules/foobar/node_modules/.bin
```

Where foobar and everything under it is empty, but the .bin wasn't removed. Any other `npm install` then reports ENOENT for `node_modules/foobar/package.json` (as per #10471).

Removing all the empty directories fixes it:

```
rmdir node_modules/foobar/node_modules/.bin
rmdir node_modules/foobar/node_modules/
rmdir node_modules/foobar
```

Expected behaviour: npm should have removed these directories already.

npm version 3.5.3

Thanks.

chrisirhc:
Likely related/same as #10887

PR #11181, opened 2016-01-17
Fix uninstall leaving nested modules directories with .bin

### LINKED ISSUE #10887: `npm uninstall` fails to remove some directories from node_modules resulting in ENOENT errors
Steps to reproduce:

top-level package.json:
```
{
  "dependencies": {
    "pkg-that-just-requires-gulp": "^1.0.0",
    "gulp": "^3.9.0"
  }
}
```

`pkg-that-just-requires-gulp` is a test package that just requires `gulp` as its only dependency.

Install the dependency tree:

```
$ npm install
```

Now, let's uninstall `gulp`:

```
$ npm uninstall --save gulp
```

We now **expect** `npm ls` to return this:
```
├─┬ dep-2@2.0.0
│ └── UNMET DEPENDENCY gulp@^3.9.0

npm ERR! missing: gulp@^3.9.0, required by pkg-that-just-requires-gulp@1.0.0
```

However, this is what we **actually** see:

```
$ npm ls
├─┬ pkg-that-just-requires-gulp@1.0.0
│ └── UNMET DEPENDENCY gulp@^3.9.0
└──  extraneous error: ENOENT, open '/Users/aleksey/my-project/node_modules/vinyl-fs/package.json

npm ERR! extraneous: vinyl-fs /Users/aleksey/my-project/node_modules/vinyl-fs
npm ERR! error in /Users/aleksey/my-project//node_modules/vinyl-fs: ENOENT, open '/Users/aleksey/my-project/node_modules/vinyl-fs/package.json'
npm ERR! missing: gulp@^3.9.0, required by pkg-that-just-requires-gulp@1.0.0
```

Examining the tree reveals that an `vinyl-fs` directory, containing nothing but `node_modules` is retained in `node_modules` even if it's been uninstalled along with gulp:

```
1$ tree node_modules/vinyl-fs
node_modules/vinyl-fs
└── node_modules

1 directory, 0 files
```

That's what causes the erroneous `ENOENT` error and the extraneous message.

`npm install`ing gulp again, does not fix the modules tree.

chrisirhc:
Fixes #10887 #10938

I haven't added a test for the main modules' `node_module/.bin`, this removes `node_module/*/node_modules/.bin` so that empty modules can be removed.

### DIFF
--- a/lib/unbuild.js
+++ b/lib/unbuild.js
@@ -80,11 +80,12 @@ function rmBins (pkg, folder, parent, top, cb) {
     } else {
       gentlyRm(path.resolve(binRoot, b), true, folder, cb)
     }
-  }, function (err) {
-    if (err) return cb(err)
-    if (!top) return gentlyRm(binRoot, true, parent, cb)
-    return cb()
-  })
+  }, gentlyRmBinRoot)
+
+  function gentlyRmBinRoot (err) {
+    if (err || top) return cb(err)
+    return gentlyRm(binRoot, true, parent, cb)
+  }
 }
 
 function rmMans (pkg, folder, parent, top, cb) {
--- a/lib/unbuild.js
+++ b/lib/unbuild.js
@@ -80,7 +80,11 @@ function rmBins (pkg, folder, parent, top, cb) {
     } else {
       gentlyRm(path.resolve(binRoot, b), true, folder, cb)
     }
-  }, cb)
+  }, function (err) {
+    if (err) return cb(err)
+    if (!top) return gentlyRm(binRoot, true, parent, cb)
+    return cb()
+  })
 }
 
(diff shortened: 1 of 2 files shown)


othiym23:
Hmm... we may want to hoist and name that error handler to a function declaration, and @iarna should review this because she's the expert on the installer, but this looks good to me aside from those concerns. Thanks for including a test!

chrisirhc:
Sounds good. I attempted to use a function declaration and cleaned up the duplicate cb() call logic. I'll await more reviewing before I squash the changes.

lxe:
:+1:

iarna:
This was merged to 3.5.5! =)

chrisirhc:
Hooray! :smile:

Issue #11119, opened 2016-01-11
Cannot read property 'length' of undefined on dedupe


STRML:
We have a large tree with webpack, babel, etc,. that we shrinkwrap before deduping.

Dedupe crashes on npmv2 (I blame babel's move off peerdeps - v6 ballooned our node_modules folder by about 400MB!), so I'm evaluating 3.

3 crashes with:

```
npm verb stack TypeError: Cannot read property 'length' of undefined
npm verb stack     at /usr/local/lib/node_modules/npm/lib/utils/lifecycle.js:56:56
npm verb stack     at /usr/local/lib/node_modules/npm/lib/utils/lifecycle.js:136:12
npm verb stack     at FSReqWrap.oncomplete (fs.js:82:15)
npm ERR! Darwin 15.2.0
npm ERR! argv "/usr/local/bin/iojs" "/usr/local/bin/npm" "--registry=http://x.x.x.x:4873" "dedupe" "--verbose"
npm ERR! node v5.4.0
npm ERR! npm  v3.5.4
npm ERR! Cannot read property 'length' of undefined
npm ERR!
```

The line is [here](https://github.com/npm/npm/blob/8dd78bbe6003cd377bf51bf24a1c5f8fd656265d/lib/utils/lifecycle.js#L56):

```js
    if ((wd.indexOf(npm.dir) !== 0 ||
          wd.indexOf(pkg.name) !== wd.length - pkg.name.length) &&
```

I inserted some debugging:

```js
    console.log({wd: wd, pkg: pkg});
    if ((wd.indexOf(npm.dir) !== 0 ||
          wd.indexOf(pkg.name) !== wd.length - pkg.name.length) &&
```

which prints:

```js
{ wd: '/projects/f/node_modules/fsevents/node_modules/node-pre-gyp',
  pkg:
   { dependencies: {},
     devDependencies: {},
     optionalDependencies: {},
     _requiredBy: [],
     _phantomChildren: {},
     version: '',
     _location: '/fsevents/node-pre-gyp',
```

Hope that helps.

iarna:
So it crashing because it thinks that `/projects/f/node_modules/fsevents/node_modules/node-pre-gyp` is missing a package name, either in an existing `package.json` or maybe more likely, missing a `package.json` entirely?

It looks like that last bit of debuggin you shared isn't complete (the trailing comma implies there were more lines)? Is there more that you could share?

[...]

PR #11185, opened 2016-01-17
band-aids issue 11119 when there is an empty skeleton of a module in …

cchamberlain:
This is the code I changed to get dedupe to pass -> https://github.com/npm/npm/issues/11119

Probably not the best solution but I am not aware of everything that's going on in this module.


othiym23:
That looks good to me (although @iarna should double check, and that line probably needs a line comment explaining why that logic works the way it does, so the next person to read it doesn't have to share your confusion). Could you write a test based on the failure in #11119? Thanks for putting this together!

cchamberlain:
@othiym23 - Yeah, I'd be happy to but working through some deploy issues at work so may be a day.

iarna:
We're landing #10582 which touches this same code (and makes it easy to test this logic) so you may want to wait for that and then base your tests on what adds to `lib/test/tap/lifecycle.js`

cchamberlain:
Sounds good, I will wait on that and circle back.

iarna:
It's now merged and ready for you any time. =)

cchamberlain:
With this corrupt structure:

image

Now prints this instead of the error that was occurring previously (I verified the Length error was still occurring with #10582):

image

Confirmed that removing the corrupt directory and rerunning dedupe works as expected.  I made an attempt at bypassing this directory as well as removing it but couldn't get those situations working right.  Seems like this isn't the location to do a rimraf in anyways, if there is a better area that I can intercept the package name and skip processing or purge, let me know.  I attempted skipping in filter-invalid-actions, but pkg name was not set and the hook did not work correctly

cchamberlain:
Ugh.. looks like I squashed poorly.. I'll close this and reopen.

PR #11188, opened 2016-01-17
doc: briefly explain what's included in `npm publish`

beaugunderson:
In [this Twitter thread](https://twitter.com/beaugunderson/status/688544393918156801) I walk through how after publishing versions from many modules (38) I managed to not know that `npm publish` includes all files that are not explicitly ignored by a local `.gitignore` or `.npmignore` (or disallowed by `.gitignore` but allowed by `.npmignore`), and found that I'd managed to include security credentials (since revoked) in published versions of one of my modules, and junk files in many others. I think it would be great to briefly note what's included in the `npm publish` documentation itself (my suggested description of the behavior may be too simplistic, but I wanted to have a starting point).

### DIFF
--- a/doc/cli/npm-publish.md
+++ b/doc/cli/npm-publish.md
@@ -11,9 +11,12 @@ npm-publish(1) -- Publish a package
 
 ## DESCRIPTION
 
-Publishes a package to the registry so that it can be installed by name. See
-`npm-developers(7)` for details on what's included in the published package, as
-well as details on how the package is built.
+Publishes a package to the registry so that it can be installed by name. All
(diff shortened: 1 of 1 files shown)


othiym23:
This is a great tweak, and I think a further sentence or two summarizing what happens when one or both of `.gitignore` or `.npmignore` drawn from `npm-developers` would be very useful to help prevent other people encountering this footgun. Thanks for putting this together, Beau!

beaugunderson:
@othiym23 no prob! (and sorry for making you respond on the weekend!)

I added a second sentence as requested. :)

iarna:

Issue #11175, opened 2016-01-16
Prefer current node in run-script


segrey:
Running an npm script using a fully specified path to node doesn't result in spawning child node processes using this node interpreter. Instead, a node interpreter from `PATH` is used. Consider this example:

```JSON
{
  "name": "sample-project",
  "scripts": {
    "print-version": "echo $PATH"
  }
}
```
Commands in terminal:
```
$ node --version
v0.12.9
$ /home/segrey/.nvm/versions/node/v5.4.0/bin/npm run-script print-version

> sample-project@ print-version /home/segrey/sample-project
> node --version

v0.12.9
```

However, v5.4.0 was expected, as it was specified explicitly.

Is it possible to prepend directory of current node interpreter to `PATH`, like it's already done with `node_modules/.bin`?

othiym23:
I agree with you that this is the common-sense way that this should work, but I also know what the CLI team's workload looks like, and we're not going to have time to get to this within the next 6-12 months (it's useful, but it's not as urgent as the other things we have scheduled). Therefore, I'm going to close this feature request.

You _can_ work around this by running

```sh
$ PATH=/home/segrey/.nvm/versions/node/v5.4.0:$PATH /home/segrey/.nvm/versions/node/v5.4.0/bin/npm run-script print-version
```

(which is more or less what nvm and other version managers do internally), but I agree that npm's current default behavior is kind of a surprise / disappointment.

However! Take a look at [the code that sets up the package script environment](https://github.com/npm/npm/blob/master/lib/utils/lifecycle.js#L83-L94), which is what sets up the path handed off to child scripts. Either `require('path').dirname(process.execPath)` would need to be prepended to the path at the very end, or (less disruptively), pushed onto the array after the list of `node_modules.bin` folders but before the existing `${PATH}` gets appended. If you want to put together a patch that does this (with a test), we'd be happy to land the patch.

Thanks for your time!

segrey:
@othiym23 Thank you for the detailed reply! Will try to come up with a patch. Btw, just curious, would it be more convenient to keep low priority (but valid) issues open?

othiym23:
Given the number of open issues, no. It is incredibly hard to use the issue tracker to figure out where the team's attention is right now, at least in part because there are hundreds of open feature requests, the vast majority of which will never be addressed by the maintainers. This leads to (yet more) duplicate requests and also contributes to the write-only nature of the issue tracker, as issue search gets decreasingly useful as the number of open issues grows.

[...]

Issue #9253, opened 2015-08-12
Use node version from current process when run node-gyp


ikokostya:
I have two node (and npm) versions on my machine:

```
$ /opt/nodejs/0.12/bin/node -v
v0.12.7
$ /opt/nodejs/0.12/bin/npm -v
2.11.3
$ node -v
v0.8.26
$ npm -v
1.4.15
```

When I try install `libxmljs` package using `node 0.12`, `npm` uses global `node 0.8` for `node-gyp`

```
$ /opt/nodejs/0.12/bin/npm i libxmljs --verbose
...
npm info install libxmljs@0.14.2

> libxmljs@0.14.2 install /home/ikokostya/proj/test/node_modules/libxmljs
> node-gyp rebuild

gyp info it worked if it ends with ok
gyp verb cli [ 'node',
gyp verb cli   '/opt/nodejs/0.12/lib/node_modules/npm/node_modules/node-gyp/bin/node-gyp.js',
gyp verb cli   'rebuild' ]
gyp info using node-gyp@2.0.1
gyp info using node@0.8.26 | linux | x64
```
After that `libxmljs` package doesn't work properly https://github.com/polotek/libxmljs/issues/312

I suggest to use same nodejs version.

othiym23:
Reclassifying this as a bug, which will with any luck be closed when #11204 is included in a release (which is going to `npm@next` later today).

PR #11204, opened 2016-01-19
run-script: prefer current node in run-script (Fixes: #11175)

segrey:


iarna:
See also: https://github.com/npm/npm/issues/11175

iarna:
Looks good– scheduling this for next week.

othiym23:
This will fix #9253 once it ships.

iarna:

segrey:
Thanks!

pesho:
@othiym23 @iarna @segrey I believe this patch is harmful and should be reverted.

For example in my case, I run a npm script inside a Python [virtualenv](https://pypi.python.org/pypi/virtualenv). Inside the `virtualenv` a custom PATH is set, which points to a different Python interpreter. However, after upgrading to npm 3.7.x, npm started prepending `/usr/bin` to the PATH in lifecycle scripts because of this PR, which caused bad breakage.

It shouldn't be npm's call to prepend `/usr/bin` to the PATH, users may have valid reasons to put something else first. The issue which prompted this (#9253) should be fixed by the user himself, by setting PATH correctly.

othiym23:
Hmm. In effect, this is due to two `virtualenv`-like tools colliding. Both `virtualenv` and npm want to control `PATH` (& many other things, most of which fortunately don't collide) in the child environments they control (and Node version managers like `nvm` and `nave` complicate this even more, as #11175 demonstrates). The question for me is whether we can make both work at the same time without too much conflict or custom configuration.

I think this patch better satisfies most users' expectations that the Node that was used to run npm will be used to run package scripts invoked within the context of that npm operation. I'm reluctant to revert this change without first taking a hard look at the problem space as a result. 

Because package scripts are invoked via shebang lines or command shims that read the path, npm has to do this indirectly by manipulating the `PATH` passed to child processes. Reverting this patch forces people into really nonergonomic corners for some common use cases, so my questions for you are how does this break things for you? What have you tried already to work around this? Are there other tools, like setting `PYTHON` in the virtual env, that will around this?

pesho:
> Both virtualenv and npm want to control PATH

There should be some sane limits to that. There is a difference between `npm/bin/node-gyp-bin` and `node_modules/.bin` on one side, and `/usr/bin` on the other side. The first two are either closely-related to, or controled directly by `npm`. It makes sense to put them in the front of `PATH`. The third one is just too general, `npm` has no idea what dangerous viruses it might contain (:smile:). 

> so my questions for you are how does this break things for you? What have you tried already to work around this? Are there other tools, like setting PYTHON in the virtual env, that will around this?

For now, I worked around this by entering the `virtualenv` from inside the run-script, like this:

```json
  "scripts": {
    "local": "env `cat .env` bash -c 'source venv/bin/activate && exec python2 main.py'"
  }
```

I don't think this should be necessary. It's better to document that run-scripts will respect the contents of `PATH` and execute the `npm` it points to, and the user should have it set correctly if having multiple `npm` versions installed.

pesho:
Alternative idea: Put symlinks to the expected `node` and `npm` versions inside `node_modules/.bin`. That way the expected ones will always be executed by run-scripts, without dangerously messing with `PATH`.

segrey:
Agree that changing `PATH` can be dangerous. The idea to put symlinks inside `node_modules/.bin` looks reasonable, but it has minor problems:
* two npm scripts may be running concurrently using different node interpreters
* some package may contribute `node_modules/.bin/node` or `node_modules/.bin/npm`.

How about another fix: keep `PATH` unchanged if `which node` on current `PATH` returns `process.execPath`? It doesn't solve the issue completely, but it reduces the probability of colliding.

pesho:
> The idea to put symlinks inside node_modules/.bin looks reasonable, but it has minor problems:

Indeed. Here is a small modification which sidesteps these problems: 

* Create a temporary directory on each `run-script` invocation, e.g. `/tmp/npm/run/<random_id>`, and put the symlinks there. Then prepend that temporary directory in front of `PATH`.
* The symlink to `npm` could even be safely put in `npm/bin/node-gyp-bin` as part of the package, as this directory already gets prepended to `PATH`. This alone wouldn't be enough though, as we also want a symlink to `node`.

The `/tmp` variant would probably leave Windows users behind, but I don't think it's too common to have multiple `npm`/`node` installations in Windows.

> How about another fix: keep PATH unchanged if which node on current PATH returns process.execPath? It doesn't solve the issue completely, but it reduces the probability of colliding.

Not ideal, but definitely an improvement. At least in my case it would solve the issue I encountered.

segrey:
Yep, that's a safer approach (a bit complicated, but seems it can't be simplified).
The question I have is should we care about globally installed modules and make them available in `/tmp/npm/run/<random_id>` too?

It's possible to use `require('os').tmpdir()`. On Windows it will probably return `C:\Users\<username>\AppData\Local\Temp`, but I didn't test.

pesho:
> The question I have is should we care about globally installed modules and make them available in /tmp/npm/run/<random_id> too?

I think not. If a `run-script` relies on a given module, it should be installed locally. It shouldn't rely on global modules.

> It's possible to use require('os').tmpdir(). On Windows it will probably return C:\Users\<username>\AppData\Local\Temp, but I didn't test.

Nice :) Although on Windows we'd probably have bigger problems - I think administrator privileges are required in order to create a symlink there.

narqo:
@othiym23 @iarna is it possible to have this fixes in `npm@2x`? There is the `2.14.17` milestone here, but looks like this fixes still haven't been ported to `2.x`.

iarna:
milestones relate to which release day it was merged. labels tell the story of what it was merged into.

segrey:
@pesho JFYI there is an attempt to mitigate possible harm caused by this change https://github.com/npm/npm/pull/12968

PR #11215, opened 2016-01-20
Do not call magic license values SPDX expressions

kemitchell:
This tiny PR corrects a false statement in the `license` section of the `package.json` doc page.

I must have missed this when we introduced the `SEE LICENSE IN <file>` magic value for `license`. The doc page says that such string values are valid SPDX expressions. They aren't. Rather, they're npm-specific magic values.


zkat:
This was merged and included in npm@3.7.0 and npm@2.14.17. Thanks!

kemitchell:
Thank _you_, Kat! All my best to the team.

PR #11232, opened 2016-01-21
docs: update included/excluded files

jscissr:
This updates #9002.

While looking through the code, I noticed that there is some duplication: [this](https://github.com/npm/npm/blob/8dd78bbe6003cd377bf51bf24a1c5f8fd656265d/lib/utils/tar.js#L64-L136) and [this](https://github.com/npm/fstream-npm/blob/d57b6b24f91156067f73417dd8785c6312bfc75f/fstream-npm.js#L91-L183) is basically the same with slight differences. Shouldn't something be done about that?

### DIFF
--- a/doc/files/package.json.md
+++ b/doc/files/package.json.md
@@ -186,9 +186,12 @@ works just like a `.gitignore`.
 Certain files are always included, regardless of settings:
 
 * `package.json`
-* `README` (and its variants)
-* `CHANGELOG` (and its variants)
+* `README`
+* `CHANGES` / `CHANGELOG` / `HISTORY`
 * `LICENSE` / `LICENCE`
+* The file in the "main" field
+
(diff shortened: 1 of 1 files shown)


iarna:
While you're tightening up the documentation you might mention that CHANGES/CHANGELOG & HISTORY can have any case and any extension. (The latter always burns me because my editor makes backup files named as `filename~` =D)

Regarding the duplication– yeah, that was an conscious time-saving measure that I'd like to get time to revisit sometime. It may be that `fstream-npm` should just be merged back into npm.

jscissr:
Done :smiley:

zkat:
This was merged and included in npm@2.14.17. It'll be merged into npm@3 next week (I misread the label so it made it in sooner 😁)

zkat:
This was included in v3.7.2. Thanks!

Issue #10820, opened 2015-12-15
Unexpected behavior in npm ls --production 


sjparsons:
Hello, I have come across some unexpected behavior with `npm ls --production`. Here's a fragment of a `package.json` file. 

```
{
  "name": "example-project",
  "version": "0.0.0",
  "description": "Example project",
  "dependencies": {
    "grunt": "^0.4.5"
  },
  "devDependencies": {
    "grunt": "^0.4.5"
  }
}
```

The output of `npm ls --json --dev` is expected:

```
{
  "name": "example-project",
  "version": "0.0.0",
  "problems": [
    "missing: grunt@^0.4.5, required by example-project@0.0.0"
  ],
  "dependencies": {
    "grunt": {
      "required": "^0.4.5",
      "missing": true
    }
  }
}
```

However, the output of `npm ls --json --production` is incorrect: 

```
{
  "name": "example-project",
  "version": "0.0.0"
}
```

In this case, `grunt` has been added to the `dependencies` object and because it's also present in the `devDependencies` object it is removed from the production dependencies that `npm ls` reports. This is undocumented behavior, but it also seems like a bad idea. The `package.json` is a interface for developers and by having unexpected consequences of actions makes developers lose trust. Simplicity is best and most understandable. 

**Summary**

If this behavior is intended then the documentation should be updated to reflect this, but if not, a patch should be submitted to correct the behavior. I'm happy to help with either of the tasks, but I'm seeking some input from the community about what is expected from this command.

PR #11245, opened 2016-01-22
ls: don't subtract dev from production deps

davidvgalbraith:
If a package is in dependencies and devDependencies,
filterByEnv removes it from ls --production. This doesn't
make sense, since the package is still a prod dependency.

Fixes https://github.com/npm/npm/issues/10820


iarna:

Issue #10347, opened 2015-11-11
put try / catch around loading onload-script


othiym23:
I'm talking about [this dusty corner](https://github.com/npm/npm/blob/master/lib/npm.js#L288-L292) of `lib/npm.js`. If the `require(onload)` fails, it should bleat helpfully at `WARN` log level.

othiym23:
And not explode with an unhelpful `MODULE_NOT_FOUND`.

othiym23:
See also #10345 for context about unhelpful explosions.

othiym23:
Adding `supportability` because [`autoinit`](http://npm.im/autoinit) has dramatically increased the usage of this "feature" and we've seen a few issues related to it as a result.

PR #11289, opened 2016-01-26
npm: don't bail if require() onload-script throws

evanlucas:
instead, log the error using the warn log level and also log a notice
that the require failed.

Fixes: https://github.com/npm/npm/issues/10347


zkat:
This was merged and included in npm@3.7.0 and npm@2.14.17. Thanks!

PR #11305, opened 2016-01-27
lifecycle: remove early exit callback if we cannot write to working directory. Also fix log message

joeferner:
Currently if the working directory is not writable by the current user and you run `sudo npm install`. The warning message and callback referenced in this PR get invoked resulting in a `WARN` message but not a failure of the install (exit code 0), because npm reduces privileges while running.

This PR changes this behavior by continuing to try and run and possibly failing the install (exit code 1) if it cannot run to completion.

This PR also fixes the existing log message which currently prints `cannot run in wd %s %s (wd=%s)...` instead of performing the proper substitutions.

### DIFF
--- a/lib/utils/lifecycle.js
+++ b/lib/utils/lifecycle.js
@@ -56,9 +56,8 @@ function lifecycle (pkg, stage, wd, unsafe, failOk, cb) {
           wd.indexOf(pkg.name) !== wd.length - pkg.name.length) &&
         !unsafe && pkg.scripts[stage]) {
       log.warn('lifecycle', logid(pkg, stage), 'cannot run in wd',
-        '%s %s (wd=%s)', pkg._id, pkg.scripts[stage], wd
+        pkg._id, pkg.scripts[stage], '(wd=' + wd + ')'
       )
-      return cb()
     }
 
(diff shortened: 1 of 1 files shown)


iarna:
> Currently if the working directory is not writable by the current user

I'm a bit confused because the code you modified doesn't have anything to do with checking to see if a folder is writable…

joeferner:
@iarna The problem is that when running as root `unsafe` is false by default (https://docs.npmjs.com/misc/config#unsafe-perm) which in turn makes this if statement true causing the behavior I described.

iarna:
Again though, if I understand the code you're modifying correctly, it's verifying that it's running from the folder its supposed to. It's saying is, warn if:

1. If the working directory isn't inside of the package directory OR
2. The working directory name does not match the package name.

Is it not a bug if those things aren't true? It seems to me that the caller must not be setting the working directory correctly?

brendo:
I really the part of this PR where is corrects the log message and stops this rabbit hole:

- What's a `wd`?
- What is `%s`, why didn't it get populated? How do I get that to work? Did I do something wrong?
- Where do I even start to debug this?

iarna:

PR #11306, opened 2016-01-27
Performance: Use JSON clone instead of lodash.cloneDeep

STRML:
Been profiling `npm` to suss out the `npmlog` slowness; while watching the flame graphs, I found a lot of this:

screen shot 2016-01-27 at 10 21 10 am

In the 'Heavy' list, `cloneDeep` is nearly at the top (below only npmlog and GC), taking nearly 6 seconds on a 2 minute install:

screen shot 2016-01-27 at 10 20 49 am

This defensive cloning can be sped up dramatically by just using `JSON.parse(JSON.stringify(o))`. Note the very small, light blue `clone` calls in the flame graph:

screen shot 2016-01-27 at 10 21 23 am

Heavy view:

screen shot 2016-01-27 at 10 21 35 am

On the same install, this clone has dropped from `6294ms` to `731ms`, a savings of about 5.5 seconds for the same result.


samccone:
cc @jdalton for input

jdalton:
If the objects being cloned are JSON compatible then `JSON.stringify` + `JSON.parse` would be the way to go. Lodash's `_.cloneDeep` has to care about things like circular references, date objects, regexes, sets, maps, etc. so is more complicated and more costly.

jdalton:
If these packages are cloned a lot you could also create a function pretty easily to generate a new JSON object based on the string:
```js
var p = fs.readFileSync('./package.json');
var f = Function('return' + p);
f(); // create the object each invocation
```

STRML:
That's a nice trick @jdalton. I did some benchmarking and it appears JSON parse/serialize is faster for <= 2 iterations, but beyond that, the function approach is an order of magnitude faster.

Since these clones are only done once, it appears JSON is the way to go.

jdalton:
I wonder if one of the loads could be eaten though. When Node requires a `.json` file via `require` it auto parses it with `JSON.parse`, instead it could be loaded with the `fs` API and turned into a creation function which would eat the cost and allow for easier clones in the future.

STRML:
That'd be nice. At this point we're sitting at about 750ms in an overall 2min build; it's substantial CPU time but I think we have lower-hanging fruit once this is merged, such as progress throttling (which is stalled in npmlog right now and shaves nearly 60s), but there are many more.

iarna:
This was merged in 3.7.0– although it turns out that this doesn't work in all situations as cycles can happen in the datastructure being cloned, so we'll be reverting this later today.

jdalton:
More on the issue https://github.com/npm/npm/issues/11349.

STRML:
Yep - still an unanswered question, re: whether or not the clone is even necessary. I ran the test suite with it removed and all tests passed; of course, this PR itself passed all the tests, so that doesn't mean it works.

PR #11348, opened 2016-01-30
Update CHANGELOG.md

rafek:
Updated the date to avoid confusion.


iarna:
Er, I dunno how I got that wrong =D

Historically we've used the US Pacific Time date though, which 01-28 (release tag is: Thu, 28 Jan 2016 17:47:32 -0800)

zkat:
This was merged as part of 7ef6793 and was included as part of v3.7.2. Cheers!

PR #11366, opened 2016-02-01
doc: Further ways the `dependencies` list in package.json can be supplied.

damncabbage:
Includes the as-an-array and as-a-string methods.

It'd be nice to write down some of the (discouraged, but possible) ways packages can be specified. I only recently become aware of these `dependencies` options when I stumbled across [this tweet](https://twitter.com/sebmck/status/691027859981365252); I'm working on something that reads parts of `package.json` files, and I was surprised to see this was possible.

(Happy to take feedback, or leave these out of the docs.)

### DIFF
--- a/doc/files/package.json.md
+++ b/doc/files/package.json.md
@@ -430,6 +430,12 @@ For example, these are all valid:
       }
     }
 
+The `dependencies` list can also be specified in the following ways, but these are **not** recommended:
+
+* As a list, eg. `{ "dependencies": ["foo-bar@^6.0.0", "dyl-thr@2.1.5"] }`
+* As a single comma-separated string, eg. `{ "dependencies": "foo-bar@^6.0.0,dyl-thr@2.1.5" }`
+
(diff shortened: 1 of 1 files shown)


iarna:
Well, TIL!

zkat:
Hey cool. I guess we do this.

As you say, it's really probably not recommended -- it's kinda obscure and probably an artifact of The Old Days. A *lot* of that stuff stays because we put a lot of emphasis on making sure the ecosystem continues to work, even if folks to advantage of sins of the past.

So the best solution the team agreed with after discussing this is to simply leave the behavior for now, but not document it. There's a chance we'll start phasing out stuff like this through formal deprecations and semver-major changes in the future, but we certainly don't want anyone new thinking this is something we think is peachy to do.

tl;dr I'm closing this because it sheds light on npm's deep dark secrets.

damncabbage:
Understood; thanks. :smile:

PR #11394, opened 2016-02-03
checks whether package is corrupt and throws better error with tests

cchamberlain:
Unmessed up version of https://github.com/npm/npm/pull/11185 to fix cryptic error when deduping and a package name cannot be resolved due to corrupted node module.

### DIFF
--- a/lib/utils/lifecycle.js
+++ b/lib/utils/lifecycle.js
@@ -2,6 +2,7 @@ exports = module.exports = lifecycle
 exports.cmd = cmd
 exports.makeEnv = makeEnv
 exports._incorrectWorkingDirectory = _incorrectWorkingDirectory
+exports._validPackage = _validPackage
 
 var log = require('npmlog')
 var spawn = require('./spawn')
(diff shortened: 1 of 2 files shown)


cchamberlain:
TIL: do not attempt PR when exhausted.

Sorry for the multiple commits, if you need them squashed I think I'll need to open a new PR or use a different strategy as `git reset --soft HEAD~2 && git commit` does not work with pushed commits. I'm a bit reliant on the travis build since on Windows many tests fail for me. I've now started running the tap test for lifecycle by itself.

cchamberlain:
Figured out the squash and got the final failing tests to pass. I had to tack an 'ELIFECYCLE' code onto the error, let me know if that's reserved for something else.

iarna:
I think we need some sort of guard (just crashing is unhelpful and not ok), but I heard some grumblings from my fellow team members, so I'd like to chat with them more before we land this, to see if we can't address their issues.

iarna:
The cli team convened and there are two things at play:

1. The bug that causes lifecycle to crash without a name. (Newly created as specifically https://github.com/npm/npm/issues/11870) 
2. That there should be some validation of name & version before trying to do certain operations. (Specifically, before the module gets added to the cache, which happens prior to installs and publishes.) This happens in `lib/cache/add-local-tarball.js`.

So I'm going to close this PR, but we would eagerly accept PRs that implement either or both of the above.


We would love

PR #11403, opened 2016-02-03
doc: removes "--depth Infinity" warning

TheBubbleDuck:
Removes warning from the doc as per a request from @othiym23 in #10736


othiym23:
@iarna, is it safe to remove this warning in `npm@3`?

iarna:
@othiym23, I believe so. `npm@3` takes all of the updates to be done and groups them by what their parent module is. For each group, it upgrades all the modules in the group with a single (internal) `npm install`. It runs them with `chain` which means it doesn't try to run them concurrently. This means that only one instance of `npm install` will be running at a time, and that in turn means that they won't step on each other.

zkat:
This was merged and included in `npm@3.7.3`. Thanks!

Issue #10966, opened 2015-12-28
doc: npm logout#scope documentation is confusing, refs `adduser`


ashleygwilliams:
from https://github.com/npm/docs/issues/466 cc/@joedragons

https://docs.npmjs.com/cli/logout

Was reviewing the above page and the scope section (https://docs.npmjs.com/cli/logout#scope) seems to be directly from npm adduser including the e.g.
From what I can tell scope on logout only defines the registry to be logged out and does not remove the scope (see npm/npm#10529). This is relayed in "If --scope is provided, this will find the credentials for the registry connected to that scope, if set." in the main section but not reiterated in the subsection.
I would have made a PR but I'm not 100% sure this is what should be being done.
But at least I don't think the scope section should reference a "npm adduser" command.

KenanY:
I believe this was addressed by https://github.com/npm/npm/pull/11416, but I guess it simply has not been deployed to the actual site yet.

PR #11416, opened 2016-02-04
doc: Remove confusing ref of adduser

wyze:
This addresses #10966. As noted in that issue, when scope is specified, it doesn't actually remove the scope config setting. I have that fixed and will submit a different PR for it.


zkat:
This was merged and included in `npm@3.7.4` and `npm@2.14.20`

Issue #9720, opened 2015-09-23
Error: Missing required argument #1 for npm update -g


siutsin:
Hi, I am not able to run `npm update -g`, following is the error log.

```
0 info it worked if it ends with ok
1 verbose cli [ '/usr/local/Cellar/node/4.1.0/bin/node',
1 verbose cli   '/Users/simonli/.node/bin/npm',
1 verbose cli   'update',
1 verbose cli   '-g' ]
2 info using npm@3.3.3
3 info using node@v4.1.0
4 verbose stack Error: Missing required argument #1
4 verbose stack     at module.exports (/Users/simonli/.node/lib/node_modules/npm/lib/install/logical-tree.js:12:3)
4 verbose stack     at /Users/simonli/.node/lib/node_modules/npm/lib/outdated.js:77:16
4 verbose stack     at /Users/simonli/.node/lib/node_modules/npm/lib/outdated.js:61:20
4 verbose stack     at /Users/simonli/.node/lib/node_modules/npm/node_modules/read-package-tree/rpt.js:188:9
4 verbose stack     at f (/Users/simonli/.node/lib/node_modules/npm/node_modules/once/once.js:17:25)
4 verbose stack     at /Users/simonli/.node/lib/node_modules/npm/node_modules/read-package-tree/rpt.js:154:20
4 verbose stack     at f (/Users/simonli/.node/lib/node_modules/npm/node_modules/once/once.js:17:25)
4 verbose stack     at then (/Users/simonli/.node/lib/node_modules/npm/node_modules/read-package-tree/rpt.js:123:22)
4 verbose stack     at /Users/simonli/.node/lib/node_modules/npm/node_modules/read-package-tree/rpt.js:83:20
4 verbose stack     at fs.js:1579:23
5 verbose cwd /Users/simonli/Desktop
6 error Darwin 14.5.0
7 error argv "/usr/local/Cellar/node/4.1.0/bin/node" "/Users/simonli/.node/bin/npm" "update" "-g"
8 error node v4.1.0
9 error npm  v3.3.3
10 error code EMISSINGARG
11 error typeerror Error: Missing required argument #1
11 error typeerror     at module.exports (/Users/simonli/.node/lib/node_modules/npm/lib/install/logical-tree.js:12:3)
11 error typeerror     at /Users/simonli/.node/lib/node_modules/npm/lib/outdated.js:77:16
11 error typeerror     at /Users/simonli/.node/lib/node_modules/npm/lib/outdated.js:61:20
[...]

PR #11433, opened 2016-02-05
src: Handle errors from `andRecalculateMetadata`

dignifiedquire:
In `outdated.js` the errors from `andRecalculateMetadata`
were swallowed resulting in the cryptic errors seen in #9720.
This fixes this and handles the error so that the npm logs
the actual error.

### DIFF
--- a/lib/outdated.js
+++ b/lib/outdated.js
@@ -75,6 +75,7 @@ function outdated (args, silent, cb) {
   if (npm.config.get('depth') === Infinity) npm.config.set('depth', 0)
 
   readPackageTree(dir, andRecalculateMetadata(function (er, tree) {
+    if (er) return cb(er)
     mutateIntoLogicalTree(tree)
     outdated_(args, '', tree, {}, 0, function (er, list) {
       list = uniq(list || []).sort(function (aa, bb) {


paulirish:
Thumbs up. This error message definitely pointed me to the fix.

Before and after: 
image

iarna:
Hi, just to be clear as nothing was said previously, but this needs a test.

paulirish:
sounds good. @dignifiedquire can you hook it up?

dignifiedquire:
Yes will take a look. @iarna For the future please leave a comment as gh does not send notifications for label changes and I had no idea this was reviewed :/

dignifiedquire:
@iarna could you point me in a direction on how to test this best? I'm not sure how to make `andRecalculateMetadata` throw an error without injecting a mocked version of it, but I haven't seen any setup/packages for doing that in the tests so far, so my guess is you don't want to go this route during testing. But I'm afraid I need some pointers on how to make it fail at this specific point then.

iarna:
@dignifiedquire Sure…

So `andRecalculateMetadata` is just a tiny wrapper around `recalculateMetadata`– we could inject that, but I think maybe injecting `readPackageTree` would be better, as `andRecalculateMetadata` passes on the error if `readPackageTree` produced one. Regardless, we use [`require-inject`](https://www.npmjs.com/package/require-inject) to do our dep injection for tests.

In a test that might look like this:
```
var test = require('tap').test
var npm = require('../../lib/npm') // the ../../ is because tests live in test/tap
var requireInject = require('require-inject')

test('setup', function (t) {
  // because of unfortunate early architecture decisions we can't require any module
  // until `npm.load` has completed.
  npm.load({},t.end)
})

test('outdated', function (t) {
    var outdated = require('../../lib/outdated.js', {
        'read-package-tree': function (_, next) {
            next(new Error('breaking on purpose'))
        }
    })
    outdated(…)
})
```

dignifiedquire:
Thanks @iarna will try and get the test done tomorrow

iarna:

Issue #7614, opened 2015-03-12
only save a certain number of npm-debug.log files


othiym23:
This depends on #6744, and is a followup.

In order to better support `npm report`, while at the same time keeping debug logs from filling up the user's cache / disk / solar system, a cap should be put on the number of debug logs in cache, preferably as a LIFO queue. My strawman suggestion for the number of logs is 20, although if we kept track of the size of the logs, it would also be nice to have a cap on the amount of disk space used by logs as well.
- [ ] only save `x` debug logs (default: 20)
- [ ] only use `y` MiB of disk space for debug logs
- [ ] make the sizes of `x` and `y` configurable
- [ ] expose a way to enumerate and browse the debug logs from the CLI
- [ ] make it easy to attach a specific crash log to [`npm report`](https://github.com/npm/npm/issues/5214)

smikes:
Also ref to #6766 , which has some cleanup logic that may be salvageable.

othiym23:
This is something that's completely within the scope of the larger audit of the user experience that the CLI team would like to in the not-too-distant future (realistically, probably towards the end of 2016).

jsdevel:
@othiym23 is it acceptable then to submit a PR to add a flag in `.npmrc` to disable `npm-debug.log`?

othiym23:
@jsdevel No, the team probably wouldn't merge such a change. From a supportability point of view, that's a dangerous option to support, because if people configure that out of habit, the team (and anyone else trying to troubleshoot) loses their single most valuable form of postmortem debugging.

jsdevel:
Right, but shouldn't that be something the user can take control of?  You can always enable `npm-debug.log` if you need it for support.

jsdevel:
I imagine it being enabled by default.

othiym23:
> shouldn't that be something the user can take control of?

I know this is a contentious position, but in this case, I believe the answer is "no." It genuinely isn't in the best interests of npm's maintainers to add features that make our jobs harder.

I wouldn't object to there being a way on a per-install basis to explicitly disable leaving an `npm-debug.log`, but only explicitly via the command line.

jsdevel:
I guess I just don't see the point.

`console.warn('Errors were found, but you\'ve disabled npm-debug.log.  Re-enable this in .npmrc to see the errors that occurred.');`

To me that feels very reasonable, and wouldn't offend any users who decide to disable `npm-debug.log`.

tarikjn:
[...]

PR #11439, opened 2016-02-05
error-handler: put npm-debug.log in cache folder

kenany:
This is, like, the fourth PR to implement this.

I see more and more users now attaching their `npm-debug.log`s to issues as `.docx`, `.pdf`, and `.zip`. Frankly I don't want to have to download any of these just to see someone's error log. I would like to see more users upload their logs to `gist.github.com` instead. `npm report` will help a lot with this, so I would like to get this prerequisite landed.


iarna:
@KenanY My main concern here is that it not fill up user's hard disks. This is particularly a concern for CI environments that could have hundreds of failing runs a day.

To land this, I'd like to see some mechanism that limits the number of logs kept in that directory to something reasonable (say, 10). This could be achieved by, if 10 or more logs exist at startup, removing the oldest until only 9 remain. Doing this at startup means that we don't have to worry about trying to do a bunch of things in the midst of an exit handler.

KenanY:
@iarna Can't believe I somehow missed your comment after all this time :(

I'll look into implementing https://github.com/npm/npm/issues/7614 within this PR as well then.

legodude17:
Something kidof related that would also help a `npm report` is to write to a `npm-debug.json` that includes parsable output so `npm report` already knows the error.

KenanY:
Added `logs-max` config (name is akin to `cache-max`). Log deletion needs to happen after config is loaded so I simply implemented that within `npm.load()` itself. Might be a better place for it elsewhere; open to suggestions.

Just needs a test now.

iarna:
Yeah, this seems otherwise ready. I'm going to see if I can't get tests written for it.

(Note from work in progress: I did end up tweaking the filename to be a bit more human readable.)

iarna:
I fixed a bug (the reported log file name was incorrect) and added a test, so this'll land this week!

iarna:
And it's merged into `release-next` as 04fca22

PR #11442, opened 2016-02-05
refusing to link when link souce resolves to link target

antialias:


antialias:
I can't find a call stack in the error output. Also, when I `npm test` locally, they always fail and it takes about 10 minutes to run. Any way I can debug this with less pain?

antialias:
bump

iarna:
@antalias If a specific test is failing you can run just that test with `npm run tap -- test/tap/name-of-test.js`

iarna:
The error in CI is because to `fs.readlinkSync` throwing an error, because you tried to read `from` as if it were a symlink (and it almost never will be one)– from the subject it's clear that you want to ignore that error. Further, please don't use `sync` calls outside of tests.

antialias:
Thanks for the tip.
Added specs, using async in non-specs.

lib/utils/link.js
```diff
@@ -25,6 +25,24 @@ function linkIfExists (from, to, gently, cb) {
...
+function resolveIfSymlink (maybeSymlinkPath, cb) {
+  fs.lstat(path.resolve(maybeSymlinkPath), function (err, stat) {
+    if (err) return cb.apply(this, arguments)
+    if (!stat.isSymbolicLink()) return cb(null, maybeSymlinkPath)
+    fs.readlink(path.resolve(maybeSymlinkPath), cb)
+  })
+}
+
+function ensureFromIsNotSource (from, to, cb) {
+  resolveIfSymlink(from, function (err, from) {
```
iarna: Please give the second from a different name, so it's clear what we're talking about here. Maybe `fromDestination`? or something like that?

lib/utils/link.js
```diff
@@ -25,6 +25,24 @@ function linkIfExists (from, to, gently, cb) {
   })
 }
 
+function resolveIfSymlink (maybeSymlinkPath, cb) {
+  fs.lstat(path.resolve(maybeSymlinkPath), function (err, stat) {
```
iarna: The `path.resolve` calls aren't doing anything here. `fs.lstat` and `fs.readlink` both accept relative arguments.

lib/utils/link.js
```diff
@@ -25,6 +25,24 @@ function linkIfExists (from, to, gently, cb) {
...
+    if (err) return cb.apply(this, arguments)
+    if (!stat.isSymbolicLink()) return cb(null, maybeSymlinkPath)
+    fs.readlink(path.resolve(maybeSymlinkPath), cb)
+  })
+}
+
+function ensureFromIsNotSource (from, to, cb) {
+  resolveIfSymlink(from, function (err, from) {
+    if (err) return cb.apply(this, arguments)
+    if (from === path.resolve(to)) {
```
iarna: `from` here is the destination of the symlink, which might be relative and so probably _should_ be passed through `path.resolve` too, but it should be relative to where the symlink was, not cwd. So something like `path.resolve(path.dirname(from), fromDestination)`.

iarna:
@antialias Looking good, just a few more comments =)

antialias:
Implemented your feedback. Build for node 0.8 seems flappy, but I don't think it's related to this branch.

iarna:
LGTM!

antialias:
@iarna Should I close this pr because it's been cherry-picked into master?

zkat:
This was merged into master as of `npm@3.8.5`! Thanks a bunch! :dancers:

antialias:
I'd like to get this into the v2 branch as well since we're still using it at work. If that's ok, can you tell me which branch I should open the PR against?

PR #11461, opened 2016-02-07
Clarify homepage url

stevemao:


iarna:
So all this does is remove:

> This is *not* the same as "url".  

From the docs? I'm don't see how that clarifies things?

stevemao:
Sorry, didn't give enough information.

>This is *not* the same as "url"

People just think this sentence is confusing and @othiym23 said this could be removed.

In fact, we could link `url` so people know what `url` we are talking about. Reading the sentence again, I'm actually not quite sure what this means:

> If you put a "url" field, then the registry will think it's a redirection to your package that has been published somewhere else, and spit at you.

iarna:
TBH, I think that entire warning, spitting and all can go. It's a warning that you shouldn't use the `url` field in the `package.json` for you homepage, which would be fair enough except that it's in the section about the `homepage` field. If folks are reading that section they're in the right place and don't need warnings.

iarna:
I'm honestly not sure what it's even referring to. I suspect that it's maybe some earlier version of what's now the `dist` field. Regardless, mentioning an unrelated field here just muddies the water.

zkat:
This was already merged as 2b57606852a2c2a03e4c4b7dcda85b807619c2cf

PR #11475, opened 2016-02-09
Fix: show settings from per-project .npmrc file in "npm config list"

### LINKED ISSUE #11472: "npm config list" ignores per-project .npmrc
I created an empty directory and executed this:

```
> echo foo=1234 > .npmrc
> npm -v
3.7.1
> npm config get foo
1234
```

which looks ok. But if I execute `npm config list` in the same directory, the `foo` setting does not show up. Same on Windows and Linux.

Is this expected behavior or a bug?

mjomble:
Fixes #11472

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -195,6 +195,25 @@ function list (cb) {
     msg += '\n'
   }
 
+  // project config file
+  var project = npm.config.sources.project
+  var pconf = project.data
+  var ppath = project.path
+  var pconfKeys = getKeys(pconf)
+  if (pconfKeys.length) {
+    msg += '; project config ' + ppath + '\n'
+    pconfKeys.forEach(function (k) {
+      var val = (k.charAt(0) === '_')
+              ? '---sekretz---'
+              : JSON.stringify(pconf[k])
+      if (pconf[k] !== npm.config.get(k)) {
+        if (!long) return
+        msg += '; ' + k + ' = ' + val + ' (overridden)\n'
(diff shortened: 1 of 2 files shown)


iarna:
I'd like to see a test for this.

mjomble:
Ok, I can look into adding some

mjomble:
I've added a test now.

I noticed there were no tests at all for `npm config list` and although it would be nice to also have tests for other config types (cli, env, user, global), I felt they were outside the scope of this PR. So I only implemented a test for per-project conf for now and added a TODO comment for implementing the others.

The Travis build failed in 0.8, but it seems like a generic configuration issue that is not related to this PR. I noticed a similar failure in a different PR as well. For >= 0.10, all builds are passing.

iarna:
The 0.8 failure is, as you assume, not related to your PR. (Ordinarily 0.8 does pass– I've flagged it for a rebuild. It's an issue we've visited earlier, and is related to a bug in 0.8's `http` module that we've papered over as best as we were able, unfortunately it still seems to show up, albeit much less often.)

iarna:
(I also agree that other tests are out of scope here. They'd be fantastic, but are hardly required for this PR.)

mjomble:
Maybe I'll look into the other tests once this PR gets merged :)

iarna:
Thank you for putting this all together, LGTM

iarna:
This landed in 3.8.2!

PR #11490, opened 2016-02-10
Document the --registry flag for `npm search`

plumlee:
We use an internal private module server and I was trying to search the public registry. Tried this and it worked. I looked for a default CLI page that might be the better place to indicate which flags (if there are others?) work with all CLI commands, but the docs aren't laid out that way.


iarna:
Thanks for this! This landed in 3.8.1 & 2.14.22!

PR #11496, opened 2016-02-11
Release 3.7.3

zkat:


svnpenn:
@zkat the change to cmd-shim/index.js:

http://github.com/npm/npm/commit/29536f4#diff-ab197d6181d37fe17f926ced7746fefa

is causing some people problems:

http://stackoverflow.com/q/37643385/pipe-is-returning-empty-string-in-bash-in-git-for-windows

Why are you using sed instead of cygpath?

PR #11525, opened 2016-02-12
deps: dedupe deeply nested bundled modules

MylesBorins:
The current tree includes a bunch of nested modules which are causing
the CI build for Node.js to not work on windows.

A bit of forensics found that the tree in
`npm/node_modules/npm-install-check/node_modules/npmlog/node_modules/gauge`
is much larger in the latest release due to duplicated modules.

This commit used npm@3 to dedupe the offending folder.


TheAlphaNerd:
So with this patch we are able to build on CI... although I just checked and realized that `npm ls` is not clean on this... :cry:

TheAlphaNerd:
Closing for different fix

PR #11526, opened 2016-02-12
fix(mingw): Prefer locally installed npm in Git Bash.

### LINKED ISSUE #11524: Prefer user-installed npm for Git Bash users
Similar to the enhancement made to the cmd and shell scripts in 89320ba9d27e25ea3f8c88d7089d9fcf0bf0d440, the `bin/npm` script could be enhanced further to prefer a user-installed instance of npm when running within a Git Bash environment.

To would resolve the inconsistency when working on a Windows machine with a globally installed npm instance (which came with the Node installation), and a user-installed npm instance, between cmd and Git Bash.

In the case of cmd, cmd will prefer the user-installed npm. Git Bash still runs the globally installed version.

A small change to the npm shell script installed globally corrected the inconsistency. 

```bash
diff --git i/bin/npm w/bin/npm
index 45e8e41..5acd6fb 100755
--- i/bin/npm
+++ w/bin/npm
@@ -15,6 +15,13 @@ fi
 NPM_CLI_JS="$basedir/node_modules/npm/bin/npm-cli.js"

 case `uname` in
+  *MINGW*)
+    NPM_PREFIX=`"$NODE_EXE" "$NPM_CLI_JS" prefix -g`
+    NPM_PREFIX_NPM_CLI_JS="$NPM_PREFIX/node_modules/npm/bin/npm-cli.js"
+    if [ -f "$NPM_PREFIX_NPM_CLI_JS" ]; then
+      NPM_CLI_JS="$NPM_PREFIX_NPM_CLI_JS"
+    fi
+    ;;
   *CYGWIN*)
     NPM_PREFIX=`"$NODE_EXE" "$NPM_CLI_JS" prefix -g`
     NPM_PREFIX_NPM_CLI_JS="$NPM_PREFIX/node_modules/npm/bin/npm-cli.js"
```

If this seems like an acceptable change I can submit a pull request for it.

ghost:
Existing behavior invoking npm within a Git Bash session is to prefer
the globally installed instance of `npm`; the instance of npm installed
alongside Node when using the Node installer.

The cmd script, and the shell script, have been modified in the past to
pull a `prefix` setting from the global `npmrc` file, and invoke the
`npm` instance located at the `prefix` location. Because the value of
`prefix` is the user's AppData directory where npm installs new copies
of itself, the user gets the latest installed npm.

However, the setup was not extended to support Git Bash. Therefore Git
Bash users still get the version of npm installed beside Node.

Add support for using the same prefix for invoking the user installed
npm in Git Bash sessions.

Fixes #11524

### DIFF
--- a/bin/npm
+++ b/bin/npm
@@ -15,6 +15,13 @@ fi
 NPM_CLI_JS="$basedir/node_modules/npm/bin/npm-cli.js"
 
 case `uname` in
+  *MINGW*)
+    NPM_PREFIX=`"$NODE_EXE" "$NPM_CLI_JS" prefix -g`
+    NPM_PREFIX_NPM_CLI_JS="$NPM_PREFIX/node_modules/npm/bin/npm-cli.js"
+    if [ -f "$NPM_PREFIX_NPM_CLI_JS" ]; then
+      NPM_CLI_JS="$NPM_PREFIX_NPM_CLI_JS"
+    fi
+    ;;
   *CYGWIN*)
     NPM_PREFIX=`"$NODE_EXE" "$NPM_CLI_JS" prefix -g`
     NPM_PREFIX_NPM_CLI_JS="$NPM_PREFIX/node_modules/npm/bin/npm-cli.js"


iarna:
LGTM

iarna:

PR #11292, opened 2016-01-27
Rewrite legacy tests

iarna:
The great legacy test rewrite. All of the things in `test/packages` have been mechanically converted into stand alone tap tests in `test/tap/legacy-*.js`. Remaining work to be done is:

Port tests to tap:

@zkat will do these:
- [x] test/tap/legacy-array-bin.js
- [x] test/tap/legacy-blerg.js
- [x] test/tap/legacy-blerg3.js
- [x] test/tap/legacy-dir-bin.js
- [x] test/tap/legacy-env-reader.js
- [x] test/tap/legacy-files.js
- [x] test/tap/legacy-ignore-nested-nm.js
- [x] test/tap/legacy-ignore.js

@iarna will do these:
- [x] test/tap/legacy-missing-bindir.js
  - [x] test/tap/legacy-no-auth-leak.js (new test)
- [x] test/tap/legacy-optional-deps.js
- [x] test/tap/legacy-platform-all.js
- [x] test/tap/legacy-platform.js
- [x] test/tap/legacy-private.js
- [x] test/tap/legacy-shrinkwrap.js
- [x] test/tap/legacy-test-package.js
- [x] test/tap/legacy-url-dep.js

See 4238ea9 for an example of what these ports look like.

When one of us is done with our list above we'll check back in and probably move on to the following two (or split up the remaining legacy tests above, depending on progress).

We also need to write tests that:
- [x] Ensure that the artifact created by `npm pack` is able to install itself.

PR #11547, opened 2016-02-15
./ syntax should be valid in .npmignore

zkat:
Currently, `./` does not ignore files when used in `.npmignore`. This fixes it so it works as expected

Also note that the glob of patches is because this is rebased on top of #11292


zkat:
Replaced this with #11570 so we only see patches related to that particular branch

PR #11548, opened 2016-02-15

zkat:

This depends on #11292


zkat:
Replaced this with #11569 so that we can just see the changes on top of the test rewrite. I'll remake against `master` after that's merged.

PR #11549, opened 2016-02-15
npm run should use shell defined in npm config to spawn new processes

xalopp:
Due to a feature/bug in the default /bin/sh of **debian** / **ubuntu** (`dash(1)`), `npm run` will fail, if the current working directory contains a percent symbol ("%").
`npm run` will add to environment variable `PATH`  `$PWD/node_modules/.bin` as described in [npm-run-script.md](https://github.com/npm/npm/blob/master/doc/cli/npm-run-script.md) and triggers the [feature/bug in DASH](http://www.mail-archive.com/dash%40vger.kernel.org/msg00970.html) because the current working directory contains a "%".
On systems, where /bin/sh is a bash (like OpenSuSE) or another not affected shell (like KSH), this is not an issue.
# how to reproduce

```
mkdir /tmp/%
cd /tmp/%
cat <<_EOF > package.json
{
  "name": "test",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "dependencies": {},
  "devDependencies": {
    "karma": "^0.13.21"
  },
  "scripts": {
    "test": "karma start --single-run"
  },
  "author": "",
  "license": "ISC"
}
_EOF
npm install
npm test
```

Output of `npm test`:

```
/tmp/%$ npm test

> test@1.0.0 test /tmp/%
> karma start --single-run

sh: 1: karma: not found
npm ERR! Test failed.  See above for more details.
```
# solution

Instead of spawning a hard coded `'sh` subprocess [lifecycle.js line 216](https://github.com/npm/npm/blob/94929cf05854886b1aa9b611ad3bbb6dbbb575f3/lib/utils/lifecycle.js#L216), it would be more consistent to use the `npm config` value `shell` like [explore.js line 21](https://github.com/npm/npm/blob/94929cf05854886b1aa9b611ad3bbb6dbbb575f3/lib/explore.js#L21)

With this PR, `npm test` will work properly unless your configured shell is dash with `npm config set /bin/dash`.

As `npm run` support setting configuration as parameter,

```
npm run test --shell /bin/dash
```

will still fail,

```
npm run test --shell /bin/bash
```

will work fine.

This PR also enables user to easily use a shell which might not be in located in default PATH environment or a alternative shell. This might be helpful for Jenkins instances on exotic UN*X systems.

### DIFF
--- a/doc/cli/npm-run-script.md
+++ b/doc/cli/npm-run-script.md
@@ -43,6 +43,16 @@ instead of `"scripts": {"test": "node_modules/.bin/tap test/\*.js"}` to run your
 If you try to run a script without having a `node_modules` directory and it fails,
 you will be given a warning to run `npm install`, just in case you've forgotten.
 
+## CONFIGURATION
+
+### shell
+
+* Default: SHELL environment variable, or "bash" on Posix, or "cmd" on
+  Windows
+* Type: path
+
+The shell to run for the `npm run-script` command.
+
 ## SEE ALSO
 
 * npm-scripts(7)
--- a/lib/utils/lifecycle.js
+++ b/lib/utils/lifecycle.js
@@ -213,7 +213,7 @@ function runCmd_ (cmd, pkg, env, wd, stage, unsafe, uid, gid, cb_) {
     conf.gid = gid ^ 0
   }
 
-  var sh = 'sh'
+  var sh = npm.config.get('shell')
   var shFlag = '-c'
 
   if (process.platform === 'win32') {
@@ -224,6 +224,7 @@ function runCmd_ (cmd, pkg, env, wd, stage, unsafe, uid, gid, cb_) {
 
   log.verbose('lifecycle', logid(pkg, stage), 'PATH:', env[PATH])
   log.verbose('lifecycle', logid(pkg, stage), 'CWD:', wd)
+  log.verbose('lifecycle', logid(pkg, stage), 'SHELL:', sh)
   log.silly('lifecycle', logid(pkg, stage), 'Args:', [shFlag, cmd])
 
   var progressEnabled = log.progressEnabled
--- a/test/tap/lifecycle-path.js
+++ b/test/tap/lifecycle-path.js
@@ -44,7 +44,7 @@ test('setup', function (t) {
 })
 
 test('make sure the path is correct', function (t) {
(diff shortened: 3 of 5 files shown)


iarna:
I think the reason it's currently hard coded to `sh` is to ensure scripts are run with a bourne-type shell, instead of whatever shell the user has selected. Note that the default for `--shell` defaults to `SHELL` from the environment which is the user's current shell, and that's not guaranteed to be bourne  or POSIX compatible (eg `tcsh`).

So I'm not totally comfortable with this solution, but I do agree that it needs _some_ solution.

xalopp:
@iarna I agree that using `$SHELL` to default the shell in  `npm run` can be an issue, mostly when the scripts contains some bourne-shell specific shell-scripting, like in fsevent or popsicle  (see https://github.com/strongloop/fsevents/blob/master/package.json#L18 or https://github.com/blakeembrey/popsicle/blob/master/package.json#L19).

Unfortunately [npm-run-script(1)](https://github.com/npm/npm/blob/master/doc/cli/npm-run-script.md), [npm-scripts(7)](https://github.com/npm/npm/blob/master/doc/misc/npm-scripts.md) and [package.json(5)](https://github.com/npm/npm/blob/master/doc/files/package.json.md) doesn't say if some "advanced" BASH-scripting is allowed/tolerated, or is rather discouraged, as it won't work on other platforms (like WIN32).

If bourne-shell scripting is de facto a standard for npm package scripts, how about introduce a new config variable `script_shell` which defaults to "sh"? This will be 100% backward compatible and user that might  need a different shell (bash, ksh,...) are free to use it.

iarna:
I feel a bit better about a config option like you describe, with docs warning that setting it to something not broadly `sh` compatible could give you a bad day.

I'm gonna draw in more of the cli team to comment here too, because I think this may be contentious.

xalopp:
@iarna  @othiym23 
Being able to set a specific shell for `npm run`, like a shell with sandboxing/auditing feature, would also help to audit packages. This can be helpful to deal with
 [Vulnerability Note VU#319816 "npm fails to restrict the actions of malicious npm packages"](https://www.kb.cert.org/vuls/id/319816)

iarna:
A split out version of this landed with #16687 as a09c1a69d!

PR #11569, opened 2016-02-16

zkat:

This depends on #11292


iarna:
wow this one is gonna kinda suck to get right from a UX perspective =/

'cause yeah, if I have folders in my `files` section then I def want `.npmignore` to help me out. But if I have explicit files listed in my `files` section and they get skipped due to a `.npmignore` I'm gonna be _really_ surprised.

files: ```
docs/README.txt
```
npmignore:
```
*.txt
```

Ok that example is extremely contrived. My gut instinct here is that this is gonna burn folks, but I'm going to have to think more about what a real world example would look like.

zkat:
Closing again and remaking against master.

PR #11600, opened 2016-02-18
install: Make install reports only include versions installed

iarna:
This makes install select modules to display by install path.  Previously it
was using name@version, which worked as long as you didn't have duplicates
in your tree, but was basically wrong forever if you did (as in npm's own
tree).


iarna:
Landed in 3.8.1

PR #11611, opened 2016-02-19
travis: save the cache between runs

iarna:
See if this speeds things up any =D


iarna:
It's not faster. :sob:

PR #11613, opened 2016-02-20
2.x rewrite legacy tests

zkat:
This is a port of #11292 to the 2.x LTS.

Pretty much the only difference is a small patch for the optional deps test.


iarna:
Release in 2.14.21! =)

Issue #11349, opened 2016-01-30
"circular structure" error with v3.7.0


benoitvallon:
I have the following error: `Converting circular structure to JSON` with npm v3.7.0 (node 5.5.0). I think I get the idea behind the "circular structure" message, but when there are a lot of dependencies in a project I wonder how it is possible to avoid that error. To me it seems impossible to check manually the dependencies of dependencies of my project. 

You can see [here on travis](https://travis-ci.org/benoitvallon/react-native-nw-react-calculator/builds/105833713) that previous npm versions work well. The project is there [https://github.com/benoitvallon/react-native-nw-react-calculator](https://github.com/benoitvallon/react-native-nw-react-calculator)

[npm-debug.log.zip](https://github.com/npm/npm/files/110685/npm-debug.log.zip)

Since v3.3.10 I had this issue https://github.com/npm/npm/issues/10985 and I had to run `npm install` 2 times but now that's an other error.

4kochi:
I get the same when trying to install the module `npm-check-updates`.

```
chuck:~ chuck$ sudo npm i -g npm-check-updates
npm ERR! Darwin 15.3.0
npm ERR! argv "/usr/local/bin/node" "/usr/local/bin/npm" "i" "-g" "npm-check-updates"
npm ERR! node v5.5.0
npm ERR! npm  v3.7.0

npm ERR! Converting circular structure to JSON
npm ERR! 
npm ERR! If you need help, you may report this error at:
npm ERR!     <https://github.com/npm/npm/issues>

npm ERR! Please include the following file with any support request:
npm ERR!     /Users/chuck/npm-debug.log
chuck:~ chuck$ sudo npm i -g npm-check-updates
npm ERR! Darwin 15.3.0
npm ERR! argv "/usr/local/bin/node" "/usr/local/bin/npm" "i" "-g" "npm-check-updates"
npm ERR! node v5.5.0
npm ERR! npm  v3.7.0

npm ERR! Converting circular structure to JSON
npm ERR! 
npm ERR! If you need help, you may report this error at:
npm ERR!     <https://github.com/npm/npm/issues>

npm ERR! Please include the following file with any support request:
npm ERR!     /Users/chuck/npm-debug.log
```

4kochi:
Might be caused by 1d1ea7eeb958034878eb6573149aeecc686888d3

othiym23:
[...]

PR #11614, opened 2016-02-20
test: `clone()` with circular dependencies. Fix #11370

### LINKED ISSUE #11370: Add failing test for JSON clone
My commit 1d1ea7eeb958034878eb6573149aeecc686888d3 broke some packages under 3.7.0 that had circular references by the time they reached this clone. Currently no tests cover this issue. We need to add one to ensure it can't happen again.

dirceu:
Add regression test so issues like #11349 won't happen again.

Fixes: #11370

### DIFF
--- a/test/tap/circular-dep-clone.js
+++ b/test/tap/circular-dep-clone.js
@@ -1,0 +1,47 @@
+var fs = require('graceful-fs')
+var path = require('path')
+
+var mkdirp = require('mkdirp')
+var osenv = require('osenv')
+var rimraf = require('rimraf')
+var test = require('tap').test
+
+var common = require('../common-tap.js')
+
+var pkg = path.join(__dirname, 'circular-dep-clone')
+
+var EXEC_OPTS = { cwd: pkg }
+
+var json = {
+  name: 'circular-dep-clone',
+  description: 'fixture',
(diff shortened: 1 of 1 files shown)


iarna:
I tried this with 25f297a and it doesn't fail, so it seems this test doesn't reproduce the problem we encountered. It's not circular dependencies that cause issues, we have those all the time, it's package metadata that ends up with cycles in it. (It's not even clear to me _how_ this happens, tracking that down too would probably be useful.)

dirceu:
Thanks for the feedback! This test fails for me against the version of `clone` from https://github.com/npm/npm/commit/1d1ea7eeb958034878eb6573149aeecc686888d3; for instance, it fails if I edit `lib/fetch-package-metadata.js` like this:

```diff
- var clone = require('lodash.clonedeep')
+ var clone = function (obj) { return JSON.parse(JSON.stringify(obj)) }
```

... and passes with the current version of clone (from `lodash.clonedeep`).

iarna:
Ah yeah, I apparently pasted the wrong hash in 🙃

iarna:
Ideally I'd like to see a version of this that doesn't use the network (that uses `npm-registry-mock` as a fake registry).

PR #11626, opened 2016-02-21
lifecycle: prefer current node in run-script

narqo:
This is a backport of #11204 by @segrey to 2.x branch which I believe was lost.

Fixes: #11175
Fixes: #9253

As mentioned in https://github.com/npm/npm/pull/11204#issuecomment-186705334 current behaviour of `npm@latest-2x` is really annoying in cases where one can't have nodejs + npm installed globally. 

For example our internal build farms have different nodejs versions installed to `/opt/nodejs/<ver>`, so we run `npm` with `/opt/nodejs/<N>/bin/node /opt/nodejs/<N>/bin/npm` from our build scripts. Everything works fine for projects that run nodejs-0.10 or node-5 that have `npm@1x` and `npm@3x` accordingly and which don't have the bug. But for projects that use nodejs-4 builds might fail if one uses an npm module that needs `#!/usr/bin/env node` to run.

### DIFF
--- a/lib/utils/lifecycle.js
+++ b/lib/utils/lifecycle.js
@@ -83,6 +83,9 @@ function lifecycle_ (pkg, stage, wd, env, unsafe, failOk, cb) {
   // the bundled one will be used for installing things.
   pathArr.unshift(path.join(__dirname, "..", "..", "bin", "node-gyp-bin"))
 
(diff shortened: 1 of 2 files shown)


iarna:
Release in 2.14.21! =)

PR #11644, opened 2016-02-23
doc: Add directories.test doc

lewiscowper:
I raised npm/docs#476 when I discovered what I thought was unusual behaviour with the directories parameter after `npm init`-ing, and then I discovered it was just an undocumented feature. So I did some digging over in that issue, and this PR is the end result.

Given the `directories.test` is similarly unexposed as with a few of the other `directories.*` properties, I figured keeping a similar writing style when mentioning it was 👍.

Thanks! 💖


iarna:
LGTM

iarna:

PR #11645, opened 2016-02-23
Document `--save-bundle` option in main install page.

### LINKED ISSUE #11636: Document `--save-bundle` option in main install page
There are 5 ways to save an installation record:

* save
* save-deps
* save-optional
* save-exact
* save-bundle

4 of these are documented in the main install page at https://docs.npmjs.com/cli/install so we should probably include `save-bundle` in there too.

yayudev:
Fixes #11636.

### DIFF
--- a/doc/cli/npm-install.md
+++ b/doc/cli/npm-install.md
@@ -13,7 +13,7 @@ npm-install(1) -- Install a package
     npm install <folder>
 
     alias: npm i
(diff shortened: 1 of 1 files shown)


iarna:
This looks good to me. Plus it turns out that `--save-bundle` doesn't work the way I'd assumed it did. =D (It really is an option to --save/--save-dev/--save-optional… it won't do anything without one of those.)

iarna:

PR #11666, opened 2016-02-24
Add maxsockets config setting for registry client

misterbyrne:
This provides a means of passing the `maxsockets` configuration setting down from `npm` to the `npm-registry-client`.

Related to https://github.com/npm/npm-registry-client/pull/130


lib/config/defaults.js
```diff
@@ -165,6 +165,7 @@ Object.defineProperty(exports, 'defaults', {get: function () {
     loglevel: 'warn',
     logstream: process.stderr,
     long: false,
+    maxsockets: 50,
```
misterbyrne: Again, to me a lower value like 5 or 10 would seem a reasonable enough default, but since the current effective value is `Infinity` I erred on the side of caution.

My own tests were on a decent wifi network - albeit geographically far away from the registry - ~20-30ms ping time to registry.npmjs.org and a throughput of around 7MB/s from registry.npmjs.org.

iarna:
LGTM

iarna:
Release in 3.8.0! =)

misterbyrne:
Oooh. Yay! :smiley:

PR #11668, opened 2016-02-24
./ syntax must be valid in .npmignore

zkat:
Currently, `./` does not ignore files when used in `.npmignore`. This fixes it so it works as expected


zkat:
Related PR: https://github.com/npm/fstream-ignore/pull/12

zkat:
This apparently is not at all how git works. Not sure why I thought that. So I'm closing this.

Issue #8791, opened 2015-07-01
npm pack with unexistant files


radare:
The `package.json` "files" entry can contain files that does not exist in the filesystem. The `npm pack` does not shows any warning or error when constructing the tarball.

KenanY:
https://github.com/npm/npm/issues/6582?

zkat:
I've been poking at this (and other related issues), and there's some classes of problems that seem to be really hard/not-possible to implement without a significant amount of refactoring of the various modules involved in the process of adding/ignoring files.

In the case of this one, it turns out [specific information about what specific file was being required is obfuscated from the time the ignore file is read](https://github.com/npm/fstream-npm/blob/master/fstream-npm.js#L234-L238).

Reading further, we find that [all ignore rules are applied, as a batch, on a per-file basis](https://github.com/npm/fstream-ignore/blob/master/ignore.js#L214-L250).

Kinda a fundamental thing, which affects not just this, but #11669, is that the `files` array is treated as an "inverted" ignore file, as a flat set of (negated) globs, and the only place in the code that is able to make decisions about this sort of warning is also [completely unaware of `package.json` as a concept](https://github.com/npm/fstream-npm/blob/master/fstream-npm.js#L202-L204).

Yes, there's ways around this, which involve continuing to do more of the [copypasting](https://github.com/npm/npm/blob/master/lib/utils/tar.js#L64-L136) and [overriding](https://github.com/npm/fstream-npm/blob/master/fstream-npm.js#L91-L183) that has gotten us fairly deep in the hole by now and is already causing major maintainability/readability problems. The more I crawl through this code though, the more I'm convinced that the only way to get stuff like this working without making *everything* worse is to just reevaluate the way we're dealing with the fstream family of libraries.

I'm going to remove the `bug` label from this, because I really don't consider a value-added warning to be a bug, as nice as I agree this particular feature would be.

zkat:
(Update: #12191 addresses this 🎉 -- it took some changes related to other stuff to make this reasonably possible)

radare:
Cool!

iamstarkov:
@zkat is there any chance that this will be merged and deployed with next npm release?

zkat:
So, because of #5082, there was recently a lot of work and attention on fstream and company, and I wanted to lay off additional features like these.

[...]

PR #11669, opened 2016-02-24

zkat:

Related: https://github.com/npm/fstream-npm/pull/20

### DIFF
--- a/test/tap/files-and-ignores.js
+++ b/test/tap/files-and-ignores.js
@@ -190,32 +190,34 @@ test('.gitignore should have identical semantics', function (t) {
   })
 })
 
-test('.npmignore should always be overridden by files array', function (t) {
+test('.npmignore should always override files array', function (t) {
   var fixture = new Tacks(
     Dir({
       'package.json': File({
         name: 'npm-test-files',
         version: '1.2.5',
         files: [
-          'include',
+          'other',
           'sub'
         ]
       }),
       '.npmignore': File(
-        'include\n' +
+        'other\n' +
         'ignore\n' +
-        'sub/included\n'
+        'sub/other\n'
       ),
-      include: File(''),
+      other: File(''),
       ignore: File(''),
       sub: Dir({
+        other: File(''),
         include: File('')
       })
     })
   )
   withFixture(t, fixture, function (done) {
     t.notOk(fileExists('ignore'), 'toplevel file excluded')
-    t.ok(fileExists('include'), 'unignored file included')
+    t.notOk(fileExists('other'), 'file included in `files` excluded')
+    t.notOk(fileExists('sub/other'), 'nested ignored file excluded')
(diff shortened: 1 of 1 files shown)


zkat:
Relevant: https://github.com/npm/npm/issues/8791#issuecomment-197724983

phated:
Is this considered a breaking change?  Will this also fall back to the gitignore?  I think falling back to gitignore would be very breaking in my projects.

zkat:
@phated We've been talking about it in context of it being a bug, since it was kind of a ??? feature.

Thanks for letting us know about it potentially breaking projects -- I think we definitely need to evaluate how our fallback to `.gitignore` works, for sure.

phated:
Cool, simplest example I can think of is:
```
// package.json
{
  files: [ 'dist' ]
}
// .gitignore
/dist/
```

zkat:
@phated I started a separate ticket as a feature request for considering removing the fallback (and also kinda a handy place to collect some of the thinking about how all these different parts interact). See #12331

zkat:
It's gonna be a while before we get back to stuff like this, specially since this is a breaking change we may never in fact incorporate, so I'm unassigning myself and leaving this PR floating around until a bright future where we can do more files stuff

Turbo87:
I would very much consider this a bugfix since the [official documentation](https://docs.npmjs.com/files/package.json#files) states:

> You can also provide a ".npmignore" file in the root of your package or in subdirectories, which will keep files from being included, even if they would be picked up by the files array.

jbcpollak:
I guess I'll be in the minority, but we _rely_ on this behavior to help us elegantly exclude _dist_ from our git repository, but include it in our published tarballs. To achieve this, all we need to do is have a `.gitignore` that includes `dist` (along with many other things), and then a `files` block that includes `dist` and whatever else we explicitly need.

If this issue was resolved, we would need to create a `.npmignore` file that is a duplicate of `.gitingore`, minus `dist`. Maintaining that accurately would get pretty messy.

tomscholz:
@mitar Why did you gave a thumb down?

eight04:
Probably because of the reason written by @jbcpollak.

However, it can be solve if we completely separate how `.gitignore` and `.npmignore` work, so that `files` array would whitelist `.gitignore` but not `.npmignore`: https://github.com/npm/npm/issues/12331

mitar:
Because I think array should have precedence over npmignore.

mitar:
(I also upvoted the comment from @jbcpollak.)

iarna:
I'm closing this because this PR is no longer actionable (too great a distance between the changes and today) and because the relevant code is about to be entirely replaced with something new and less horrible, which will make future changes easier to grok.

If @zkat decides they want this again, they know where to find me. =p

Also: `.npmignore`/`.gitignore` in subdirectories DO override the top level `files`.

Issue #11688, opened 2016-02-25
Travis: test fails regularly on node 0.8/npm 2.x


zkat:
This has been a bit of a thorn on our side while we've been trying to rely more on Travis, and it would be good to take some time to look at it and see if there's any way to resolve whatever race condition is causing it, even if at the test level:

```
test/tap/peer-deps-invalid.js .........................
  installing dependencies that have conflicting peerDependencies
  not ok should be equal
    +++ found
    --- wanted
    -1.3.3
    +1.3.1
    compare: ===
    at:
      line: 84
      file: test/tap/peer-deps-invalid.js
      column: 13
    stack: |
      test/tap/peer-deps-invalid.js:84:13
      cb (lib/install.js:119:20)
      lib/install.js:131:16
      lib/install.js:268:7
      node_modules/read-installed/read-installed.js:142:5
      node_modules/read-installed/read-installed.js:263:14
      cb (node_modules/slide/lib/async-map.js:47:24)
      node_modules/read-installed/read-installed.js:263:14
      cb (node_modules/slide/lib/async-map.js:47:24)
      node_modules/read-installed/read-installed.js:263:14
      asyncMap (node_modules/slide/lib/async-map.js:27:18)
      next (node_modules/read-installed/read-installed.js:234:5)
      node_modules/read-installed/read-installed.js:179:7
      LOOP (fs.js:1143:14)

  installing dependencies that have conflicting peerDependencies
  not ok should be equal
    +++ found
    --- wanted
    -The package underscore@1.3.3 does not satisfy its siblings' peerDependencies requirements!
    +The package underscore@1.3.1 does not satisfy its siblings' peerDependencies requirements!
    compare: ===
    at:
      line: 85
      file: test/tap/peer-deps-invalid.js
      column: 13
    stack: |
      test/tap/peer-deps-invalid.js:85:13
      cb (lib/install.js:119:20)
      lib/install.js:131:16
      lib/install.js:268:7
      node_modules/read-installed/read-installed.js:142:5
      node_modules/read-installed/read-installed.js:263:14
      cb (node_modules/slide/lib/async-map.js:47:24)
      node_modules/read-installed/read-installed.js:263:14
      cb (node_modules/slide/lib/async-map.js:47:24)
      node_modules/read-installed/read-installed.js:263:14
      asyncMap (node_modules/slide/lib/async-map.js:27:18)
      next (node_modules/read-installed/read-installed.js:234:5)
[...]

PR #11691, opened 2016-02-26
Fix flappy test/tap/peer-deps-invalid.js due to race in peer-dep handling

iarna:
This is a fix for #11688, by making the test resilient to a race in npm@2– the race means that in the case of a peer dependency conflict, we can't say in advance which of the conflicting peers will produce the error message. All we can say is that one of them WILL.


zkat:
LGTM :tada: :sheep:

zkat:
Merged because those intermittent failures were driving me up the wall. @iarna is basically the best.

Issue #11699, opened 2016-02-26
npm install -production failing since 3.7.4 for all modules (OSX and Linux)


ixalon:
I've been having an error when using `npm install` using the `-production` flag since updating npm to 3.7.5, I've rolled back to 3.7.4 and still see the issue (tested on both OSX and Ubuntu). No debug log is written. 3.7.3 is working correctly.

npm correctly installs the module but immediately before exiting (with return code 1) logs: `npm ERR! code 1`.

I've setup my environment and run a test installing lodash (the same occurs with all modules I've tried) as such:
```bash
rm -rf /tmp/npm-test ~/.npm
mkdir /tmp/npm-test
cd /tmp/npm-test
npm init -y
npm install -production -verbose lodash 2> install.log
```

My `~/.npmrc` contains only:
```
//registry.npmjs.org/:_authToken=<token>
```

The log from 3.7.3: https://gist.github.com/ixalon/bbba9547787f4d356f5a
The log from 3.7.4: https://gist.github.com/ixalon/260011abf364889a2e76

The diff between the resulting two logs shows nothing significant other than the final outcome:
```diff
diff v3.7.3.log v3.7.4.log
8c8
< npm info using npm@3.7.3
---
> npm info using npm@3.7.4
12c12
< npm info attempt registry request try #1 at 6:33:43 PM
---
> npm info attempt registry request try #1 at 6:32:33 PM
14c14
< npm verb request id 41cc55263c6b90a8
---
> npm verb request id 1bc96515d810d857
23c23
< npm verb headers   date: 'Fri, 26 Feb 2016 18:33:45 GMT',
---
> npm verb headers   date: 'Fri, 26 Feb 2016 18:32:34 GMT',
26c26
< npm verb headers   'x-served-by': 'cache-lhr6322-LHR',
---
> npm verb headers   'x-served-by': 'cache-lhr6325-LHR',
29c29
< npm verb headers   'x-timer': 'S1456511624.593941,VS0,VE813',
---
> npm verb headers   'x-timer': 'S1456511554.377427,VS0,VE202',
43,44c43,44
< npm info retry fetch attempt 1 at 6:33:45 PM
< npm info attempt registry request try #1 at 6:33:45 PM
---
> npm info retry fetch attempt 1 at 6:32:34 PM
> npm info attempt registry request try #1 at 6:32:34 PM
48c48
[...]

PR #11703, opened 2016-02-26
ls: Do not set process.exitCode when ls is executed during the unbuild phase

ixalon:
Prevents invalid error status when `npm install --production <module>` is used.
Fixes: #11699

My thinking behind this is that `ls` is purely informational when executed as part of an `npm install` (in the unbuild phase). It shouldn't be responsible for deciding if the install should fail (i.e. shouldn't set the process exit code).

When you execute it via `npm ls <package>` (etc.) then it's valid that it may exit with a failure code if `<package>` is not found (the existing behaviour).

### DIFF
--- a/lib/ls.js
+++ b/lib/ls.js
@@ -95,7 +95,17 @@ var lsFromTree = ls.fromTree = function (dir, physicalTree, args, silent, cb) {
   }
   console.log(out)
 
-  if (args.length && !data._found) process.exitCode = 1
+  // The 'ls' command can be called directly from the command line or for
+  // providing informational output during an 'npm install' (in the unbuild phase).
+  // Executing 'npm ls <module>' for a module that is not installed results
+  // in a exit code of 1. Executing 'npm ls --production <module>' for a
+  // module that is not listed as a production dependency will also exit with 1.
+  // If this same behaviour is followed during 'npm install --production <module>'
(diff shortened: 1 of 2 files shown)


lib/ls.js
```diff
@@ -95,7 +95,7 @@ var lsFromTree = ls.fromTree = function (dir, physicalTree, args, silent, cb) {
   }
   console.log(out)
 
-  if (args.length && !data._found) process.exitCode = 1
+  if (args.length && !data._found && npm.command !== 'unbuild') process.exitCode = 1
```
iarna: I'd like a comment here explaining why we're doing this, because I just know that when we go to rewrite things to not have a global npm object, that this bit is going to be super confusing. =p

ixalon: Thanks - agreed. I'll add.

ixalon:
@iarna Comment added - ready for re-review :-)

iarna:
I'd like to solve this structurally differently (by changing how install calls ls) but it's an easy change so I'll fold that into this PR and make use of your test as is.

iarna:
Good news, this landed in 3.8.1– you can see in e40d4572cc98db06757df5b8bb6b7dbd0546d3d7 how I changed the implementation approach (moving it to `lib/install.js`).

PR #11719, opened 2016-02-29
Fix Alias name from 'verison' to 'version'

zeg-io:
Nothing major here, but version is misspelled in the help text, this fixes it.


othiym23:
Like `isntall` and a few other cases, this is an intentional affordances for the fat-fingered among us. Since the current behavior is intentional, we won't be taking this patch. Thanks for your time!

PR #11729, opened 2016-03-01
Fix #11003 remove-verison-alias-from-cli

### LINKED ISSUE #11003: help should not show aliases
I believe it's quite confusing to get a list of all the aliases, intermingled with true commands, when entering `npm -?` on the command line.

This is particularly true for mistyped aliases like `verison` appearing in the list of commands.

Moreover, the list of commands becomes overwhelmingly long due to the high amount of feign commands, making the programmer believe a oversized long learning curve is required to handle `NPM`.

So I suggest to only list true commands on `npm help` and omit the aliases. I believe it's sufficient to have aliases quoted along with the description of the command in the manual.

danielkman:
I also saw `isntall` alias there but with a some code and a comment saying "  // we have our reasons "
Should this remove it to?

### DIFF
--- a/lib/npm.js
+++ b/lib/npm.js
@@ -91,8 +91,7 @@
     't': 'test',
     'find-dupes': 'dedupe',
     'ddp': 'dedupe',
-    'v': 'view',
-    'verison': 'version'
+    'v': 'view'
   }
 
   var aliasNames = Object.keys(aliases)


othiym23:
These affordances are here to provide simple help for the fat-fingered among us, and are an important element of the CLI's character. As such, the team isn't interested in removing them. Thanks for your time!

danielkman:
Hey @othiym23 thanks for the quick response, here "[help should not show aliases](https://github.com/npm/npm/issues/11003)" you commented that you wuld welcome a PR for removing the typo from the help output, if you changed your mind then maybe you should close that issue?

PR #11748, opened 2016-03-02
doc: add aliase command to cli docs

watilde:
Some cli docs have an aliases but some other docs doesn't have it.

e.g.
https://docs.npmjs.com/cli/search#synopsis

### DIFF
--- a/doc/cli/npm-adduser.md
+++ b/doc/cli/npm-adduser.md
@@ -5,6 +5,8 @@ npm-adduser(1) -- Add a registry user account
 
     npm adduser [--registry=url] [--scope=@orgname] [--always-auth]
 
+    aliases: login, add-user
+
 ## DESCRIPTION
 
 Create or verify a user named `<username>` in the specified registry, and
--- a/doc/cli/npm-bugs.md
+++ b/doc/cli/npm-bugs.md
@@ -5,6 +5,8 @@ npm-bugs(1) -- Bugs for a package in a web browser maybe
 
     npm bugs [<pkgname>]
 
+    aliases: issues
+
 ## DESCRIPTION
 
 This command tries to guess at the likely location of a package's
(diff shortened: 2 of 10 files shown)


othiym23:
This is a solid addition. Thanks as always, @watilde!

doc/cli/npm-help.md
```diff
@@ -5,6 +5,8 @@ npm-help(1) -- Get help on npm
 
     npm help <term> [<terms..>]
 
+    aliases: apihelp
```
othiym23: we should probably nuke `apihelp` as an alias / leave this undocumented, as there currently is no meaningful API documentation (on purpose).

doc/cli/npm-docs.md
```diff
@@ -8,6 +8,8 @@ npm-docs(1) -- Docs for a package in a web browser maybe
     npm home [<pkgname> [<pkgname> ...]]
     npm home .
 
+    aliases: home
```
othiym23: Out of scope for this PR, but `npm docs` and `npm home` are semi-different conceptually, so we may want to document them separately at some point. Also, for this PR, it might make sense to nuke the two `npm home` examples, in much the same way that you did with `npm tst` below.

watilde: I agreed it basically, but it's same case with `npm tst` and I deleted it in [npm-test.md](https://github.com/watilde/npm/blob/patch/add-aliases/doc/cli/npm-test.md#synopsis). If they are same case, I will keep `npm tst` again.

doc/cli/npm-help.md
watilde: Sure thing! Many different kind of alias are there in npm actually.

watilde:
Ok, I updated:
+ Keep no change on `npm-docs.md` and `npm-help.md` like you commented

seldo:
🎉

iarna:

zkat:
This landed in LTS with 2.15.0

PR #11761, opened 2016-03-03
help: display all commands except for affordance commands

watilde:
Fixes https://github.com/npm/npm/issues/11003. There are many aliases in npm.js, but actually they involve 2 different concepts, "shorthands" and "affordances". I put a `cmd.js` to the config directory to reuse the list as a module and separate shorthands and affordance from aliases. Then eventually call the cmd.js from `npm.js` and `help.js` appropriately.

config/cmd.js
- add a command list file as cmd.js
- separate shorthands and affordance from aliases in cmd.js

npm.js
- replace hard corded command list with cmd.js

help.js
- display all commands except for affordance commands

dependency
- add lodash.assign@4.0.6

### DIFF
--- a/lib/config/cmd.js
+++ b/lib/config/cmd.js
@@ -1,0 +1,109 @@
+// short names for common things
+var aliases = {
+  'rm': 'uninstall',
+  'r': 'uninstall',
+  'un': 'uninstall',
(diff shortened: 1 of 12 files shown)


iarna:
This landed in 3.8.2!

Issue #11495, opened 2016-02-10
npm ls --parseable doesn't seem to respect depth argument


glenpike:
If I run

```
npm ls --depth=0 --parseable
```
I get a lot of cruft about nested dependencies - I should only see the top level ones listed.

zacdoe:
I have written a potential fix for this bug, however in order to properly test my changes I would have to alter the underlying `npm-registry-mock`. I have a failing test and fix for `npm ls --depth=0 --parseable --long` and am working on the test for `npm ls --depth=1 --parseable --long` but the mocked out packages only go one dependency deep, so there is no way to verify whether the depth has stopped at the correct level. There needs to be one more level of depth in the mocked packages to properly test the distinction between `--depth=1` and `--depth=Infinity`. Can anyone advise me on the best strategy for achieving this?

PR #11773, opened 2016-03-04
added fix for npm ls --depth=0 --parseable bug and added related tests

zacdoe:
This is a fix for the bug #11495. The change stops the recursion when the depth has gone too far, where previously it always traversed all levels. This change should ideally include one additional test that makes a distinction between `--depth=1` and `--depth=Infinity` but that would require changing the underlying _npm-registry-mock_ with another level of dependency.


iarna:
This landed in 3.8.2!

zacdoe:
So excited ! This was my first open source contribution!! Thank you!

PR #11782, opened 2016-03-05
Use simple progress bar when unicode option is false

### LINKED ISSUE #11781: Should use simple progress bar when unicode option is false
I guess this issue are happened frequently on a environment which is using east asian characters. When user sets a east asian character width as a double width, the progress bar will be broken like the following screen shot I attached:
screen shot 2016-03-05 at 15 10 42

The npmlog actually has a nice theme to solve this problem and it will be used when the method [disableUnicode](https://github.com/npm/npmlog/blob/6a64249cc74a59189cd5f071d2c16a3ee3b6ac64/log.js#L39-L42
) is executed. In npm, currently the themes [are handled by kinds of OS](https://github.com/iarna/gauge/blob/master/themes.js#L10-L12) automatically, but this issue can happen on any OS. That's the reason why I could reproduce this issue easily on my OS X.

Then `npm install` should use the ascii theme when a part of the npm config `unicode` is false. Also, please take a look at [this test code](https://github.com/npm/npm/blob/master/test/tap/install-cli-unicode.js), this case has already expected but currently the `unicode` option does not be used for the progress bar. It's a bug.

related issue: https://github.com/watilde/npm/issues/3

watilde:
Fixes #11781 and https://github.com/watilde/npm/issues/3, referenced [the commit](https://github.com/npm/npm/commit/423d8f7e8a291ecceab121b9d2ea207c809e17fd) that was also providing an option to the npmlog.

Docs has already existed, it means this PR is like a bug fix in spite of providing an option to npmlog, I guess. https://github.com/npm/npm/blob/master/doc/misc/npm-config.md#unicode

### DIFF
--- a/lib/npm.js
+++ b/lib/npm.js
@@ -371,6 +371,12 @@
           log.disableProgress()
         }
 
+        if (config.get('unicode')) {
+          log.enableUnicode()
+        } else {
+          log.disableUnicode()
+        }
+
         // at this point the configs are all set.
         // go ahead and spin up the registry client.
         npm.registry = new CachingRegClient(npm.config)
--- a/test/tap/progress-config.js
+++ b/test/tap/progress-config.js
@@ -11,6 +11,10 @@ var requireInject = require('require-inject')
 // Make sure existing environment vars don't muck up the test
 process.env = {}
 
+function hasOnlyAscii (s) {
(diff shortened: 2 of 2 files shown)


iarna:
Thank you, this is much appreciated! LGTM

iarna:
This landed in 3.8.2.

Issue #11121, opened 2016-01-12
Don't list dependencies after install (it's very spammy!)


stevage:
(From [StackOverflow](http://stackoverflow.com/questions/34734035/dont-show-complete-npm-dependency-list-after-installation)) I recently upgrade from ancient to new NPM. After `npm install` completes in the package that I work on mostly, NPM spews out hundreds of lines of output, similar to the result of `npm list`.

![screenshot 2016-01-12 12 14 56](https://cloud.githubusercontent.com/assets/678344/12256578/5d6bdb5e-b952-11e5-8b83-afb9aaf7f055.png)

This seems unwarranted.  Output of that size isn't useful, and it hides the output of the `postinstall` phase, where I had hoped to include a user-friendly message.

Could this behaviour be changed, to skip the dependency listing? Or at least to make it optional somehow?

The package is [here](github.com/nicta/nationalmap).

The full output of `npm install` is [here](https://gist.github.com/stevage/9289124b8d9d935daf51) - it's 923 lines, of which ~900 are the dependency list.

KenanY:
@stevage Sounds like you want https://github.com/npm/npm/issues/10732.

stevage:
Thanks for that pointer. Thinking about this more, I think the current behaviour is simply broken, and adding options to turn broken behaviour into acceptable behaviour doesn't seem right.

Put it this way: in what situation is seeing 900 lines of package listing after installation desirable?

- New user: no, they just get overwhelmed
- Experienced user: no, their eyes will glaze over and they'll probably use some workaround to disable it
- Package developer: no way are they going to read 900 lines of output looking for something out of the ordinary

But agree with comments in #2040 that `npm install` should produce some output - and forcing the user to hide *all* output (including errors and warnings) with `--loglevel silent` would be wrong.

Could I suggest replacing all that output with:

>`900 packages installed. Run 'npm list' for details.`

bazineta:
I think @stevage hits it on the head here; there is no circumstance I can think of in which an amount of output that typically overflows the scroll buffer is actually helpful.  In general, we're now just running the install a few times to see if it took, frankly.

zkat:
This is a duplicate of #10732, so I'm closing this.

Issue #10732, opened 2015-12-07
Add option to hide summary output from npm install


ChrisPearce:
When executing npm install or npm uninstalll, the commands output a summary of all the package dependencies that were installed/uninstalled.

This seems to occur even if the log level is explicitly set to --silent.

This output can be quite verbose. When dealing with large build flows, this output can quickly become unwieldy - making important warnings/errors harder to find quickly.

It would be very helpful to have an option where this summary output could be disable.

A suggestion made as part of recent discussions in #2040 was to provide a --no-summary flag (or something similar). I've created this issue so that the suggestion can be tracked/discussed separately.

zarko-tg:
You're the man @ChrisPearce!

xtoddx:
Making `--silent` work as expected would be preferable to adding yet another flag.

It seems that currently `--silent` is only a proxy for `--loglevel silent`, but making silent also set progress=false and disabling the summary output would be fan-tastic!

uglow:
+1
I'm running tests on TravisCI and 800 lines of my 900-line log are the summary info. I'm using Node 5.4.0. I'd prefer that `--verbose` turns on the tree-summary.

zkat:
Leaving this open for now as we very much intend to change this output of `npm install`, but we don't know what that looks like yet or when we'll do it.

stevage:
Ok, since my other issue was closed, I just want to make clear that I don't think the solution is "add an option to hide summary output", but to never show that spam. I suggested:

> `900 packages installed. Run 'npm list' for details.`

IMHO, the current situation is pretty badly broken: 900 lines of output for me, with no way to disable it, and shown even when `--loglevel silent` is enabled.

othiym23:
I've gone back and forth on this a number of times due to the historical legacy of this command and its output being part of the "CLI API contract", but recently we've been more in a mode of "set fire to the past and all who sailed it" and I think that especially given the complexity of the average install, it no longer makes sense for us to be printing the full logical tree or set of changes on install. Instead, we'd like to have a simple summary line like `apt-get`'s where we provide statistics like the number of packages added and removed, the size change of the install, and the duration of the install (even though that's probably just going to make people yell at us until we get more work done on `npm@3`'s performance).

So! The npm CLI team will take this on sometime within the next year, but if somebody wants to take a stab at patching this before we get to it, start the PR _early_, because the team are likely to have strong opinions about the user experience.

billinghamj:
[...]

PR #11784, opened 2016-03-06
"Silent" disables printing dependency tree

Symbitic:
I know there are several requests to make `npm install` stop printing the dependency tree (See #2040, #10732, and #11121). This pull request disables printing the dependency tree after installing is finished, if `loglevel` is set to `silent`.

Includes documentation (a note added to `npm-install.md` explaining the change in behavior) and a tap test.

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -272,8 +272,10 @@ Installer.prototype.run = function (cb) {
         [this, this.saveToDependencies])
     }
   }
-  postInstallSteps.push(
-    [this, this.printInstalled])
+  if (npm.config.get('loglevel') !== 'silent') {
+    postInstallSteps.push(
+      [this, this.printInstalled])
+  }
 
   var self = this
   chain(installSteps, function (installEr) {
--- a/doc/cli/npm-install.md
+++ b/doc/cli/npm-install.md
@@ -296,6 +296,10 @@ The `--only={prod[uction]|dev[elopment]}` argument will cause either only
 See `npm-config(7)`.  Many of the configuration params have some
 effect on installation, since that's most of what npm does.
 
+By default, after `npm install` finishes installing successfully it prints
+a dependency tree, similar to `npm ls`. If `loglevel` is set to `silent`, this
+behavior is disabled.
+
 ## ALGORITHM
 
 To install a package, npm uses the following algorithm:
--- a/test/tap/install-cli-silent.js
+++ b/test/tap/install-cli-silent.js
@@ -1,0 +1,61 @@
+var fs = require('graceful-fs')
+var path = require('path')
+
+var mkdirp = require('mkdirp')
+var mr = require('npm-registry-mock')
+var osenv = require('osenv')
+var rimraf = require('rimraf')
+var test = require('tap').test
+
+var common = require('../common-tap.js')
+var server
+
+var pkg = path.resolve(__dirname, 'install-cli-silent')
+
+var EXEC_OPTS = { cwd: pkg }
+
+var json = {
+  name: 'install-cli-silent',
+  description: 'fixture',
+  version: '0.0.1'
+}
+
+test('setup', function (t) {
+  rimraf.sync(pkg)
+  mkdirp.sync(pkg)
+  fs.writeFileSync(
+    path.join(pkg, 'package.json'),
+    JSON.stringify(json, null, 2)
+  )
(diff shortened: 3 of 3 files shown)


iarna:
I'm hesitantly supportive of this. Having some way to suppress that seems desirable, at the same time, it's not logging. It's sent to stdout not stderr, and historically we've treated things sent to stdout as part of the npm contract, which would make this a breaking change. (Which wouldn't stop us from taking this mind you.) Stuff to stdout is a part of the contact because other tools built on top of npm may consume it.

As this is, it ALSO suppresses the output in these cases:

```
npm install --loglevel=silent --json
npm install --loglevel=silent --parseable
```

And I don't think that's desirable. (Those are produced as an alternative to the tree, via the same route as the install tree.)

e-cloud:
perhaps providing another level to totally suppress all stuff to stdout?

othiym23:
It's past time for us to at least make it possible to suppress the tree output on install, but since this is part of the API contract of the CLI, that makes this a breaking change. Never fear, though, because there are a number of things waiting for a breaking change, and so there will probably be a new major version of npm within the next 2-3 months. We'll just hold onto this patch until then.

othiym23:
Thanks for your patience, in the meantime!

lib/install.js
```diff
@@ -272,8 +272,10 @@ Installer.prototype.run = function (cb) {
         [this, this.saveToDependencies])
     }
   }
-  postInstallSteps.push(
-    [this, this.printInstalled])
+  if (npm.config.get('loglevel') !== 'silent') {
```
deckar01: Based on https://github.com/npm/npm/pull/11784#issuecomment-195034724 this should be reverted, and the conditional should take into consideration `--json` and `--parseable` flags somewhere closer to [install.js#L687](https://github.com/npm/npm/blob/2005f4ce11f6cdf142f8a77f4f7ee4996000fb57/lib/install.js#L687).

deckar01: [ls.js#L95](https://github.com/npm/npm/blob/b2a46693add048fc1468905b6afa254d75702308/lib/ls.js#L95) seems to be the offending line.

@iarna At first I thought those wrote to a file, but seeing that they are logged with `console.log()`, I think the issue is that they are not properly using the `npmlog` utility.

hunterloftis:
Even of loglevel is "error" I wouldn't expect to see the npm tree (on "silent" I would expect to not see error messages).

deckar01:
@iarna @othiym23 What if a new logging level was introduced to preserve the existing behavior?

| level | number |
|--------|-----------:|
| silly | -Infinity |
| verbose | 1000 |
| info | 2000 |
| http | 3000 |
| warn | 4000 |
| error | 5000 |
| **tree**| :wink: Infinity - 1 |
| silent | Infinity |

`npm install -s --silent` and `npm config set loglevel silent` would internally use the "tree" level to preserve backwards compatibility.

A new flag like `npm install -ss --super-silent` and config level `npm config set loglevel super-silent` could then be used to access the true "silent" level.

deckar01:
I think I see part of the issue now. `npmlog` prefixes the log with `npm <level> `, so this parsable output can't be assigned a log level without updating `npmlog.log()` to allow disabling the default prefix.

This makes me lean more towards a solution like a `--no-tree` flag.

lib/install.js
iarna: @deckar01 By those you mean `--json` and `--parseable`? Those _have_ to go to stdout as their output is intended for consumption by other programs. Or do you mean the archy tree that shows up without those flags? I can see the argument that it should go to the log…

deckar01: @iarna, those are the flags I was confused about. [I mentioned below](https://github.com/npm/npm/pull/11784#issuecomment-210131905) that `npmlog` would have to support disabling the line prefix to allow outputting parsable data. I also [devised a scheme](https://github.com/npm/npm/pull/11784#issuecomment-210120127) to force true silence without breaking backwards compatibility.

billinghamj:
Is this going to be going into npm@4? I think it's really important to get this fixed **before** npm@4 is released, so this doesn't have to wait until npm@5...

iarna:

(Also we don't print the dependency tree on install any more so maybe this is no longer an issue?)

deckar01:
This didn't make it into npm 4....................

Issue #11735, opened 2016-03-01
Installing devDependencies does not update shrinkwrap


sheepsteak:
I've created a shrink-wrap using `npm shrinkwrap --dev`. Installing new dependencies (`npm i -S express`) correctly updates the `npm-shrinkwrap.json`. However, installing a devDependency (`npm i -D eslint@1`) does **not** update the `npm-shrinkwrap.json`.

This seems counterintuitive as we've specified we want everything shrinkwrapped so you would expect `npm-shrinkwrap.json` to always be updated. To get around this at the moment we have to always remember to do `npm shrinkwrap --dev` after installing a new devDependency. Also, when updating a dependency with something like `npm i -D eslint@2` the `npm-shrinkwrap.json` is not updated either, which is probably even worse as it's already specified in the `npm-shrinkwrap.json`! :fearful:

This is with `npm@3.8.0`.

othiym23:
Yep, this seems like a pretty straightforward bug. The CLI team is in the middle of a significant rewrite of its test suite, so it may be a little while until we get around to fixing this, but in the meantime, a patch that ensure that the behavior with shrinkwraps on `--save-dev` matches the behavior on `--save` would be very welcome. Thanks for your time!

JohanNordlinder:
My team also suffer from this bug and I agree with sheepsteak, `npm install -[-S|--save|-D|` should also update the shrinkwrap we've created with `npm shrinkwrap --dev`, this is the behavior you would expect after reading the documentation.

mkscrg:
We hit this as well. It makes managing shrinkwrapped devDependencies pretty difficult!

szimek:
@othiym23 If shrinkwrap file was generated without `--dev` option, should `npm install --save-dev` update it as well?

PR #11785, opened 2016-03-06
Installing dev dependencies updates npm-shrinkwrap.json

doug-wade:
Fix https://github.com/npm/npm/issues/11735

### DIFF
--- a/lib/install/save.js
+++ b/lib/install/save.js
@@ -13,8 +13,9 @@ var deepSortObject = require('../utils/deep-sort-object.js')
 var parseJSON = require('../utils/parse-json.js')
 var moduleName = require('../utils/module-name.js')
 
-// if the -S|--save option is specified, then write installed packages
-// as dependencies to a package.json file.
+// if the -S|--save or -D|--save-dev options are specified, then write installed
+// packages as dependencies or devDependencies to a package.json file (and
+// npm-shrinkwrap.json file, if applicable).
 
 exports.saveRequested = function (args, tree, andReturn) {
   validate('AOF', arguments)
@@ -56,7 +57,7 @@ function saveShrinkwrap (tree, next) {
     })
 
     if (!saveOptional && saveDev && !hasDevOnlyDeps) return next()
-    if (saveOptional || !save) return next()
+    if (saveOptional || (!save && !saveDev)) return next()
 
     if (hasDevOnlyDeps) {
       var dev = npm.config.get('dev')


doug-wade:
I'd kinda prefer to have a test for this, but as best I can tell, none of the existing install tests verify the behavior of npm-shrinkwrap.json, and I'm not entirely sure how I should structure them.  Should I add a new test class, something like test/tap/install-save-dev-with-shrinkwrapped-dev-deps-adds-new-dep.js in which I:

1.) set up a new package
2.) install a dep and a dev dep
3.) shrinkwrap
4.) install --save-dev other-dev-dep@version
5.) verify npm-shrinkwrap.json contains other-dev-dep at version

Or are we okay with saying that since there aren't tests for install's effects on npm-shrinkwrap.json or install --save-dev, that the community (me) shouldn't be setting precedent on how to test those?

sheepsteak:
I made the original #11735 issue last week and had chance to look at this over the weekend too. I also tracked it down to that line but also noticed some edge cases.

If you do the following with your fix:
1. `npm init`
2. `npm i -S is-object`
3. `npm shrinkwrap --dev`
4. `npm i -D is-array`

You'll notice the shrinkwrap doesn't add `is-array` even though you specified `--dev`. This is because there's nothing in the shrinkwrap to tell npm what the original intent was and because it can't see any current devDependencies in there.

I guess the drastic solution would be to add an extra key to the shrinkwrap stating what the original user intent was but that's going to only be forward compatible and I have no idea how older npm will handle it. But the short-term solution might be to output a message when someone does `npm shrinkwrap --dev` but with **no** devDependencies that they need to do it again after installing their first devDependency.

Not sure if this should be another issue though as this PR does fix the original issue :)

iarna:
@doug-wade I too would love to see tests for this. Our test coverage is still sketchy, so some features are currently just missing. Tests are everyone's responsibilty, so the cli team LOVES community provided tests. =)

I would do as you describe, create a new test script (and those series of steps sound reasonable to me).

Further testing hints:

1. You'll find that `npm-registry-mock` provides a handful of modules that are installable.
2. For setting up your initial environment, `tacks` aims to make this easier, see `test/tap/install-report-just-installed.js` for an example of its usage.

Bockit:
Would it make sense for the key live in package.json?

michaelnisi:
@doug-wade In step 3.) of your test outline, you mean `npm shrinkwrap --dev`, right?

michaelnisi:
@iarna Maybe a new flag isn't needed. `npm shrinkwrap --dev` could add an empty `'devDependencies'` array to `npm-shrinkwrap.json` to remember user intent in later dev dep installs.

michaelnisi:
@doug-wade If you don't mind, I'd start writing tests for this.

iarna:
@michaelnisi Something along those lines is what I was thinking, yes. I wouldn't do it as an array, however. Just a boolean in the `npm-shrinkwrap.json`.

iarna:
This is a thiiiing now! (That is, this PR is obsolete 'cause we implemented it.)

Issue #10968, opened 2015-12-28
doc: scope page should mention required version


ashleygwilliams:
from: https://github.com/npm/docs/issues/417 cc/@danpaz @KenanY 

"The [scope page](https://docs.npmjs.com/misc/scope) mentions:

> (As of 2015-04-19, the public npm registry does support scoped packages)

I'm not sure exactly when scoping became supported, though that date does line up with [v2.0.0-beta3 in the changelog](https://github.com/npm/npm/blob/master/CHANGELOG.md#v200-beta3-2014-09-04).

If publishing/installing scoped packages depends on npm 2 maybe the documentation can clarify that? I know I've had some heartache trying to install private modules and getting a 404 only to realize I was on npm 1.4."

(at least until versioned docs happen)

PR #11786, opened 2016-03-06
Add npm required version to scoped package docs

doug-wade:
Fix https://github.com/npm/npm/issues/10968


iarna:
I like actually calling out the versions that things became available in! We need to do that more.

LGTM

iarna:

iarna:
This landed in lts in 2.15.1.

PR #11820, opened 2016-03-09
Doc: fix description of npm-link

rhgb:
The command `npm link` should be linking package from local folder to global,
and `npm link package-name` should be from global to local. The current description
is reversed.


iarna:

iarna:
This landed in lts in 2.15.1.

PR #11838, opened 2016-03-09
publish: don't publish unless package.json is correct strictly

### LINKED ISSUE #11837: don't publish unless `version` in package.json passes strict semver validation
It appears that the CLI will currently publish a package even if the version isn't one that it would accept at install time. The primary registry has been normalizing published packages, so as long as they're parseable with `semver` in loose mode, the resulting package will be installable. The registry team recently rolled out a change to the registry that (unintentionally) removed this normalization, and it caused immediate installation failures for end users. `npm publish` should refuse to publish any package that doesn't have a strictly conformant value in `version` (even though registries will need to continue normalizing version fields because of the large number of iffy version fields in the wild).

watilde:
Fixes #11837.

### DIFF
--- a/lib/publish.js
+++ b/lib/publish.js
@@ -42,8 +42,9 @@ function publish (args, isRetry, cb) {
   }
 
   var arg = args[0]
+  var strict = !npm.config.get('force')
   // if it's a local folder, then run the prepublish there, first.
-  readJson(path.resolve(arg, 'package.json'), function (er, data) {
+  readJson(path.resolve(arg, 'package.json'), strict, function (er, data) {
     if (er && er.code !== 'ENOENT' && er.code !== 'ENOTDIR') return cb(er)
 
     if (data) {
--- a/test/tap/publish-invalid-data.js
+++ b/test/tap/publish-invalid-data.js
@@ -1,0 +1,150 @@
+var common = require('../common-tap.js')
+var test = require('tap').test
+var npm = require('../../lib/npm.js')
+var mkdirp = require('mkdirp')
+var rimraf = require('rimraf')
+var path = require('path')
+var fs = require('fs')
+var mr = require('npm-registry-mock')
+
+var osenv = require('osenv')
+
+var PKG_DIR = path.resolve(__dirname, 'publish-invalid-data')
+var CACHE_DIR = path.resolve(PKG_DIR, 'cache')
+
+var DEFAULT_PKG = {
+  'name': 'examples',
+  'version': '1.2.3'
+}
+
+var mockServer
+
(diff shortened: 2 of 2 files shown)


watilde:
This ci error has happened by the test cases of the `npm-registry-couchapp` which are executed in `npm/test/tap/registry.js` . The following patch is also need to fix the issue: https://github.com/npm/npm-registry-couchapp/pull/255

iarna:
I think `read-package-json`'s strict flag does quite a bit more than just strict version checking, which makes me a little hesitant about this.

Let's trace it through… it's not used by `read-package-json` directly, just passed through to `normalize-package-data`. It's only used by the fixer's (`lib/fixer.js`):

* `fixVersionField`: Passed through inverted to `semver.valid`, will throw if the version is invalid.
* `fixNameField`: If the name is not a string, throws. If the name is invalid (including leading/trailing spaces, which ordinarily are trimmed) it throws. **If the name has mixed case it throws.**

Ok, so I'm ok with everything except that last bit, because there _are_ legacy modules in the registry that still have mixed case and I we've broken them before. =/ @othiym23 maybe allow loose mode with `--force`?

watilde:
Indeed. I just added `--force` option handling to allow loose mode in publish.js and added test cases for the following cases you wrote.
+ fixVersionField: Passed through inverted to semver.valid, will throw if the version is invalid.
+ fixNameField: If the name is not a string, throws. If the name is invalid (including leading/trailing spaces, which ordinarily are trimmed) it throws. If the name has mixed case it throws.

iarna:
This landed in 3.9.0 =)

watilde:
Oops, I couldn't find referring commits in v3.9.0...? https://github.com/npm/npm/pull/12596

The related PR is still open as well:
https://github.com/npm/npm-registry-couchapp/pull/255

iarna:
Ah no, you're right, I pulled that at the last minute'cause of test failures 'cause I didn't notice the `npm-registry-couchapp` change that also needed to land. =D I'm reopening this and we'll get this in this week.

iarna:
(FYI: If you look at https://github.com/npm/npm/commit/fa888d6b97526d0a7c0375e1dcc287bd99819209 you'll also see that I rewrote your tests to A) not use the api, use common.npm instead and B) use a static set of fixtures w/ Tacks instead of mutating a package.json each time.)

watilde:
Ahh yeah that's true. Let me update my test codes later today as first time to observe the new house rule :)

watilde:
Ok I updated and rebased the test codes followed your update :) @iarna

PR #11884, opened 2016-03-13
Fix #11863: Mention node_modules ignore [DOCS]

### LINKED ISSUE #11863: node_modules ignore not mentioned
In the [files section of the npm website](https://docs.npmjs.com/files/package.json#files), there is no mention of `node_modules` being ignored when publishing to npm.

Jameskmonger:
This fixes #11863 by mentioning that the `node_modules` folder is always ignored.

### DIFF
--- a/doc/files/package.json.md
+++ b/doc/files/package.json.md
@@ -206,6 +206,7 @@ Conversely, some files are always ignored:
 * `._*`
 * `npm-debug.log`
 * `.npmrc`
+* `node_modules`
 
 ## main
 


iarna:
Thank you for this, while there's a little greater nuance than them being entirely ignored (bundled dependencies are included), I think including `node_modules` in this list won't cause any confusion.

zkat:
This was merged into mainline as of `npm@3.8.5`. Soon to be in LTS, too.

zkat:
This was included in LTS with `npm@2.15.3`. Thanks for the patch!

PR #11912, opened 2016-03-15
test: Fix npm-self-install test to install to /tmp

iarna:
Previously this was installing into a temp subdir in `test/tap`, which wouldn't catch the case where a module was installed in the local `node_modules` folder but not in dependencies, as node would look up
the tree and use the copy from the version of npm being tested.

This solves that by moving the install folder to `/tmp` (which is where it was in the original legacy test version).


othiym23:
LGTM! @doug-wade, you may want to take a look at this too – `osenv.tmpdir()` is a more robust way of figuring out where the temporary directory is cross-platform. :sheep::rocket::fire:

iarna:
This landed in 2.15.1 & 3.8.3.

PR #11931, opened 2016-03-17
test: remove disabled tests

othiym23:
And also remove `test/update-test.sh`, which hasn't been used for a pretty long time now.

Steps along the way to hoisting the tap tests up to the root `test/`.

**r**: @iarna, with whom I was discussing this on Monday.


iarna:
This landed in latest in 3.8.3 and will land in lts this week.

zkat:
This landed in LTS with `npm@2.15.2`.

PR #11938, opened 2016-03-17
improve npm search results with this workaround

jasonkhanlar:
Fixes last search result to appear, missing newline caused last result to not appear.

Implements workaround to improve search result response time, throws error if more than 10,000 search results to avoid waiting indefinitely for error timeout, though up to 50,000 or so results may work without error as well, but response time is not acceptable

### DIFF
--- a/lib/search.js
+++ b/lib/search.js
@@ -2,6 +2,7 @@
 module.exports = exports = search
 
 var npm = require('./npm.js')
+var log = require('npmlog')
 var columnify = require('columnify')
 var updateIndex = require('./cache/update-index.js')
 
@@ -63,6 +64,9 @@ function search (args, silent, staleness, cb) {
     // prettify and print it, and then provide the raw
     // data to the cb.
     if (er || silent) return cb(er, data)
+    if (Object.keys(data).length > 10000) {
+        throw new Error("More than 10,000 matches")
+    }
     console.log(prettify(data, args))
     console.log()
     cb(null, data)
--- a/lib/search.js
+++ b/lib/search.js
@@ -64,6 +64,7 @@ function search (args, silent, staleness, cb) {
     // data to the cb.
     if (er || silent) return cb(er, data)
     console.log(prettify(data, args))
+    console.log()
     cb(null, data)
   })
 }


lib/search.js
```diff
@@ -63,7 +64,11 @@ function search (args, silent, staleness, cb) {
     // prettify and print it, and then provide the raw
     // data to the cb.
     if (er || silent) return cb(er, data)
+    if (Object.keys(data).length > 10000) {
+        throw new Error("More than 10,000 matches")
+    }
     console.log(prettify(data, args))
+    console.log()
```
watilde: i guess this line was for debugging :)

iarna: `console.log` already adds newlines, so I'm wondering if the one above was hitting some line length limit? Regardless, it seems to me that this line isn't really the right way to approach that problem.

iarna:
@jasonkhanlar Before we can take this, the tests are going to need to be passing. Currently our linter is complaining about a few things:

```
  /home/travis/build/npm/npm/lib/search.js:5:5: "log" is defined but never used
  /home/travis/build/npm/npm/lib/search.js:68:9: Expected indentation of 6 space characters but found 8.
  /home/travis/build/npm/npm/lib/search.js:68:25: Strings must use singlequote.
```

InfiniteStyles:
Could anyone clarify what is going on here with the All checks have failed? It seems like there is a library conflict and the merge has not been accepted? Or is that the linter issue? I've been following NPM trying to better understand.

othiym23:
@InfiniteStyles Our AppVeyor builds are flaky as heck, but as of now Travis builds should be green for any new pull requests, as we've recently gotten all those tests passing. This PR is pretty old, though, and precedes a lot of the recent changes to the tests.

On that note, this change has been superseded by 2b8057be2e1b51e97b1f8f38d7f58edf3ce2c145, which is included in `npm@4.0.0` (`npm@4` is still going through its bake-in phase, but you can install it with `npm install -g npm@next`). As such, I'm going to close this PR. Thank you, @jasonkhanlar for your time and the contribution!

PR #11947, opened 2016-03-17
lib: check parent is there before checking target

thefourtheye:
if the parent is not available, this would bomb at runtime with the
error,

```
TypeError: Cannot read property 'target' of null
```


iarna:
Just tracing through in my head when this would happen:

This code path _should_ only be able to happen if the top level module is flagged as a link, because only the top level module won't have a parent.

But actions are never taken against the top level module, so this indicates to me that there's a deeper bug at play here. =/

Still, I'm inclined to take this in the name of resiliency, as I think fixing this will be best addressed by the `read-package-tree` (and thus tree node object) refactor, to bring all that logic into a single place.

PR #11952, opened 2016-03-18
doc: fixed a issue with the are-we-there-yet link

dozoisch:
- The link was pointing to nasdaq private market instead of npm


mathieumg:
Seems like this was fixed in https://github.com/npm/npm/commit/f2ae3baf9e4b48282bcfceb94ce861b9221450d0

Issue #11983, opened 2016-03-20
`npm search` out of memory


SimenB:
Unsure what other info I can provide. no `npm-debug.log` was produced.

```shell
$ node -v && npm -v
v4.4.0
3.8.2
```

```shell
$ npm s
npm WARN Building the local index for the first time, please be patient
▄ ╢░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░╟

<--- Last few GCs --->

  468599 ms: Scavenge 1397.4 (1456.3) -> 1397.4 (1456.3) MB, 0.5 / 0 ms (+ 1.2 ms in 1 steps since last GC) [allocation failure] [incremental marking delaying mark-sweep].
  469717 ms: Mark-sweep 1397.4 (1456.3) -> 1388.5 (1456.3) MB, 1117.5 / 0 ms (+ 1.2 ms in 1 steps since start of marking, biggest step 1.2 ms) [last resort gc].
  470825 ms: Mark-sweep 1388.5 (1456.3) -> 1391.1 (1456.3) MB, 1108.5 / 0 ms [last resort gc].


<--- JS stacktrace --->

==== JS stack trace =========================================

Security context: 0x2c2503bb4629 <JS Object>
    1: /* anonymous */(aka /* anonymous */) [/Users/simen/.nvm/versions/node/v4.4.0/lib/node_modules/npm/node_modules/columnify/columnify.js:~216] [pc=0x1eb8edc272d6] (this=0x2c2503b041b9 <undefined>,item=0x2af6d8a3c6f1 <an Object with map 0xe4662f76099>)
    2: arguments adaptor frame: 3->1
    3: InnerArrayMap(aka InnerArrayMap) [native array.js:~1001] [pc=0x1eb8edce39a1] (this=0x2c2503b041b9...

FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - process out of memory
[1]    18752 abort      npm s
```

[...]

PR #11984, opened 2016-03-20
Exit early if no arguments provided to search

SimenB:
#11983

Probably not the solution you want, but anyways


iarna:
Maybe not ideal, but it's instantly better than what we currently do. =D Thank you for putting this together.

SimenB:
No problem! As this has been merged manually, I'll close this PR

PR #11987, opened 2016-03-20
Fix failing test

yodeyer:
This test fails because `--registry` option is missing (the real npm registry is used instead of the mocked one).


iarna:
Thank you! LGTM!

zkat:
This landed in LTS  with `npm@2.15.2` and mainline with `npm@3.8.5`

PR #11995, opened 2016-03-21
Improve handling of special cases for files included in packages

glenjamin:
This PR replaces #10445 which is a follow up to #9642

This matches the behavior in fstream-npm@1.0.6 from d82ff81403e906931fac701775723626dcb443b3. Most notably, the package 'history' will not longer be included in the packages of anything which depends on it.

I wasn't sure what the convention around helper methods in tests was as I couldn't find any examples, let me know if this approach isn't on-brand.


glenjamin:
The test that is failing on this branch is also failing on master for me - but the last travis build of master seems ok - could be something time based? :sob:

iarna:
Regarding the failing test, it's fixed by #11987. There was a test relying on outdated output that was apparently looking at the live registry, so when one of the modules in the test was updated the test broke. =D

test/tap/pack-file-inclusion.js
```diff
@@ -0,0 +1,82 @@
...
+var test = require('tap').test
+var common = require('../common-tap')
+var fs = require('graceful-fs')
+var join = require('path').join
+var mkdirp = require('mkdirp')
+var rimraf = require('rimraf')
+var tar = require('tar')
+var zlib = require('zlib')
+
+var pkg = join(__dirname, 'pack-age')
```
iarna: I prefer to name these after the test– if the test crashes, it'll be clear which one made it, plus it ensures tests don't stomp on each other. (Relatedly, someday I'd like to be able to run tests concurrently. Currently tests often setup stuff listening on fixed port #s, so they can't, but…)

test/tap/pack-file-inclusion.js
```diff
@@ -0,0 +1,82 @@
...
+
+test('setup', function (t) {
+  rimraf.sync(pkg)
+  mkdirp.sync(pkg)
+  fs.writeFileSync(join(pkg, 'package.json'), JSON.stringify({
```
iarna: You might be interested in looking at `tacks` module, currently used in the various `legacy-*` tests for setting up fixtures like this.

iarna:
I had a couple of minor comments about the tests, but overall it looks good. @zkat has been working heavily in this part of code recently, so I'm going to leave it to her to LGTM.

zkat:
Hi, @glenjamin! This functionality is already tested (in significant detail) at https://github.com/npm/npm/blob/master/test/tap/files-and-ignores.js

That said, those tests probably still have gaps, and if there's stuff that your PR intended to test that isn't already covered over there, feel free to adapt that file, though I'm not sure about having a completely new file for this.

You may also want to look at https://github.com/npm/npm/milestones/Files%20and%20Ignores, which includes a number of things that are still left to do regarding this particular area.

zkat:
Oh! I overlooked the actual change: I thought your patch to `tar.js` was just some reformatting. Sorry I missed that.

I assume you pulled this straight from [`fstream-npm`](https://github.com/npm/fstream-npm/blob/master/fstream-npm.js#L92-L104), which already has the directory check: There's some really awkward duplication of code here, and I think it might be nice to fix that, but I'm still feeling cautious about anything resembling a refactor of that part of the code.

A patch that tests that directories with those names are not automatically included would be A+, though. Take a look at https://github.com/npm/npm/blob/master/test/tap/files-and-ignores.js#L446-L495 for that specific part of the tests :)

glenjamin:
> This functionality is already tested (in significant detail) at https://github.com/npm/npm/blob/master/test/tap/files-and-ignores.js

Aha, that's good - I was a bit concerned when I couldn't find any existing tests. It does suggest there's a bit of a discoverability issue though. Is there a knack to knowing where to look for these sorts of things?

Anyway, I'll this this PR updated with the additional cases folded into the existing test file.

zkat:
> a bit of a discoverability issue

Probably! We have a lot of tests right now, and we're kinda focusing on getting them all to pass on all our platforms, and making sure we have decent coverage. I think it's worth thinking about _how_ we organize them so it's easier to find what's tested where. `files-and-ignores.js` is a relatively new test, too (converted from old legacy tests that only did a subset of this). :)

Thanks a lot again! And I look forward to an updated PR :D

glenjamin:
Ok, so I tried updating the test you linked, and found the following:

* It's slightly annoying that I can't run individual tests through tap
* The test for these additional cases passes on `master`
```
test('unconditional inclusion does not capture modules' , function (t) {
  var fixture = new Tacks(
    Dir({
      'package.json': File({
        name: 'npm-test-files',
        version: '1.2.5'
      }),
      'node_modules': Dir({
        'readme': Dir({ 'file': File('') }),
        'README': Dir({ 'file': File('') }),
        'licence': Dir({ 'file': File('') }),
        'license': Dir({ 'file': File('') }),
        'history': Dir({ 'file': File('') }),
        'History': Dir({ 'file': File('') }),
        'changelog': Dir({ 'file': File('') }),
        'ChangeLOG': Dir({ 'file': File('') }),
      })
    })
  )
  withFixture(t, fixture, function (done) {
    t.notOk(fileExists('node_modules/readme/file'), 'readme module not included')
    t.notOk(fileExists('node_modules/README/file'), 'README module not included')
    t.notOk(fileExists('node_modules/licence/file'), 'licence module not included')
    t.notOk(fileExists('node_modules/license/file'), 'license module not included')
    t.notOk(fileExists('node_modules/history/file'), 'history module not included')
    t.notOk(fileExists('node_modules/History/file'), 'History module not included')
    t.notOk(fileExists('node_modules/changelog/file'), 'changelog module not included')
    t.notOk(fileExists('node_modules/ChangeLOG/file'), 'ChangeLOG module not included')

    done()
  })
})
```
* The `pack` command, when run on the same fixture, would fail the test
```
Fri 25 Mar 09:57:16
~/Development/GitHub/npm/test/tap/files-and-ignores/npm-test-files (master) glen ○ retina
> find .
.
./node_modules
./node_modules/changelog
./node_modules/changelog/file
./node_modules/history
./node_modules/history/file
./node_modules/licence
./node_modules/licence/file
./node_modules/license
./node_modules/license/file
./node_modules/readme
./node_modules/readme/file
./package.json

Fri 25 Mar 09:57:17
~/Development/GitHub/npm/test/tap/files-and-ignores/npm-test-files (master) glen ○ retina
> npm pack
npm-test-files-1.2.5.tgz

Fri 25 Mar 09:58:33
~/Development/GitHub/npm/test/tap/files-and-ignores/npm-test-files (master) glen ○ retina
> tar tf npm-test-files-1.2.5.tgz
package/package.json
package/node_modules/changelog/file
package/node_modules/history/file
package/node_modules/licence/file
package/node_modules/license/file
```

I'm struggling to understand how `fstream-npm` fits in here, it has the required logic, but then `utils/tar.js` overrides it?

I inherited this PR after running into this issue myself and discovering that there was a pending fix that was just missing tests - so I'm not entirely clear on what's going on here.

zkat:
You need to move all those files up and out of `node_modules`, since we want to test that those matches aren't picked up by our toplevel logic.

That said, it's a bit weird that things inside `node_modules` aren't being ignored. They're supposed to be managed by the (extensive) bundling logic.

So, try it more like:
```
  var fixture = new Tacks(
    Dir({
      'package.json': File({
        name: 'npm-test-files',
        version: '1.2.5'
      }),
      'readme': Dir({ 'file': File('') }),
      'README': Dir({ 'file': File('') }),
      'licence': Dir({ 'file': File('') }),
      'license': Dir({ 'file': File('') }),
      'history': Dir({ 'file': File('') }),
      'History': Dir({ 'file': File('') }),
      'changelog': Dir({ 'file': File('') }),
      'ChangeLOG': Dir({ 'file': File('') })
     })
```

glenjamin:
The Node modules thing is the bug I'm trying to fix. https://github.com/npm/npm/issues/9642

The forced includes end up bundling packages with those names, but only with pack - not a filesystem-to-filesystem install.

zkat:
Oh no. I misunderstood the problem! Totally my fault. I'm really sorry for any time wasted due to my mistake. :(

Anyway: I get what you mean now, and that *is* really weird. I'd like to have the `files-and-ignores` part of this test, even though it doesn't test the specific bug you ran into -- which means it's probably better to keep your original tests, since I have no better home for them right now, and they wouldn't actually share the logic in `files-and-ignores`.

The patch _does_ fix the problem for me, though, so it's just a matter of making sure we have test that proves it both ways.

I like @iarna's suggestion of using `tacks` for mocking up the module you're going to test with `pack` though, and otherwise I'll just wait for you to push a new version of this that includes both tests, and I'll probably :+1: soon after.

Thanks for your patience <3

glenjamin:
How about if I extend `files-and-ignores` to run each scenario with install **and** pack?

Then if the logic can be collapsed to a single path, the test can be simplified.

zkat:
Ordinarily, I'd say meh because that would make the tests run much slower.

In light of having these strange differences, though, that seems reasonable to me.

@iarna you have opinions on doubling up on commands like this?

glenjamin:
I'm going to dig into this a bit more and see if can tell what's going on.

So far

* `npm install pkg` correctly omits the files
* `npm pack pkg` puts the files into the tarball
* `npm install pkg-1.0.0.tgz` when the tarball has the files omits them

So it looks like there's some alternative logic being used when installing, although the only place I can see `fstream-npm` being used is in `lib/utils/tar.js`.

glenjamin:
Ok, I now know why `install` works.

At the end of the `extract` action `bundledModules` are staged out of the way, and then the `node_modules` dir is `gentlyRm`ed. So the modules aren't ignored, they're just deleted after unpack by some other code.

The installer code when installing from a directory internally uses the same packaging code to create the tarball and then unpacks it into the target.


Based on this summary, how would you feel about changing this test to use `pack` + listing the tarball to do its assertions, and not use `install` at all?

I think this is more along the spirit of the test, as files and ignores are really part of tarball creation.

zkat:
@glenjamin This sounds like a great idea. We have other tests, apparently, that test the actual extraction. I think leaving the responsibility of testing the extraction to those tests (or writing more tests for the way `install` extracts things) is better.

So yes! It would be awesome if the tests in `files-and-ignores` just tested the behavior of `pack` itself. :+1: Hopefully without changing any of the actual existing tests (so I don't have to fix conflicts with my other work in `files-and-ignores.js` ;) )

glenjamin:
Rather than updating the copy-pasted code, removing the duplication passes all the tests, and appears to be the correct thing to do.

This is the diff between the deleted method, and the one that now runs via the inheritance.

```diff
--- tar.js	2016-03-30 14:31:17.000000000 +0100
+++ fstream.js	2016-03-30 14:31:24.000000000 +0100
@@ -1,4 +1,5 @@
-BundledPacker.prototype.applyIgnores = function (entry, partial, entryObj) {
+Packer.prototype.applyIgnores = function (entry, partial, entryObj) {
+  if (!entryObj || entryObj.type !== 'Directory') {
   // package.json files can never be ignored.
   if (entry === 'package.json') return true
 
@@ -10,21 +11,30 @@
 
   // changelogs should never be ignored.
   if (entry.match(/^(changes|changelog|history)(\.[^\.]*)?$/i)) return true
+  }
 
   // special rules.  see below.
   if (entry === 'node_modules' && this.packageRoot) return true
 
+  // package.json main file should never be ignored.
+  var mainFile = this.package && this.package.main
+  if (mainFile && path.resolve(this.path, entry) === path.resolve(this.path, mainFile)) return true
+
   // some files are *never* allowed under any circumstances
+  // (VCS folders, native build cruft, npm cruft, regular cruft)
   if (entry === '.git' ||
-      entry === '.lock-wscript' ||
-      entry.match(/^\.wafpickle-[0-9]+$/) ||
       entry === 'CVS' ||
       entry === '.svn' ||
       entry === '.hg' ||
+      entry === '.lock-wscript' ||
+      entry.match(/^\.wafpickle-[0-9]+$/) ||
+      (this.parent && this.parent.packageRoot && this.basename === 'build' &&
+       entry === 'config.gypi') ||
+      entry === 'npm-debug.log' ||
+      entry === '.npmrc' ||
       entry.match(/^\..*\.swp$/) ||
       entry === '.DS_Store' ||
-      entry.match(/^\._/) ||
-      entry === 'npm-debug.log'
+      entry.match(/^\._/)
     ) {
     return false
   }
@@ -65,9 +75,19 @@
     }
 
     // only include it at this point if it's a bundleDependency
-    return this.isBundled(entry)
+    var bd = this.package && this.package.bundleDependencies
+
+    if (bd && !Array.isArray(bd)) {
+      throw new Error(this.package.name + '\'s `bundledDependencies` should ' +
+                      'be an array')
+    }
+
+    var shouldBundle = bd && bd.indexOf(entry) !== -1
+    // if we're not going to bundle it, then it doesn't count as a bundleLink
+    // if (this.bundleLinks && !shouldBundle) delete this.bundleLinks[entry]
+    return shouldBundle
   }
   // if (this.bundled) return true
 
-  return Packer.prototype.applyIgnores.call(this, entry, partial, entryObj)
+  return Ignore.prototype.applyIgnores.call(this, entry, partial, entryObj)
 }
```

zkat:
p.s. I added it to the Files and Ignores milestone cause basically super relevant :)

glenjamin:
Anything else you need from me on this?

zkat:
LGTM! 🎉 

Thanks a bunch for the patch and patience with the back-and-forth. This is great! 👯

taion:
Can this PR be closed, given https://github.com/npm/npm/commit/c615182c8b47e418338eb1317b99bb66987cda54?

zkat:
This was included in `npm@3.8.7`. Thank you SO MUCH again, @glenjamin -- this was super handy :)

PR #12000, opened 2016-03-22
deps: remove async-some that is no longer used

watilde:
async-some was used only in `gently-rm.js`, but now it has already been removed

History
- Added: replace some.js with (Zalgo-safe) async-some@1.0.0: https://github.com/npm/npm/commit/9ef8fe4d6ead3acb3e88c712000e2d3a9480ebec
- Removed: https://github.com/npm/npm/commit/56533663a74fcf9fb04e7d27abab155437105b9b#diff-5b73c13931031bd42f6cbaee4de149dd
- Current: https://github.com/npm/npm/blob/master/lib/utils/gently-rm.js


watilde:
The failed test will be fixed by https://github.com/npm/npm/pull/11987

iarna:
Good catch! Thank you!

zkat:
This was included in `npm@3.8.5`. Thanks, @watilde!

PR #12003, opened 2016-03-22
test: fix test in tap/outdated-symlink

watilde:
async@2.x was published
https://github.com/caolan/async/releases


watilde:
Oops, duplicated #11987

PR #12019, opened 2016-03-23
generate a consistent shasum for git dependencies

stevecrozz:
Formerly, the shasum was generated from a tar.gz file which was
generated from a cloned git tree. Because git doesn't keep track of file
metadata such as timestamps, the tar.gz file generated from the cloned
git tree has a different shasum during every npm install.

So instead of generating a shasum in this way, we use the sha1 from the
git tree where the source was obtained. The result is a shasum that is
always the same for a given git tree. This allows systems like docker
build to make use of the build cache for build steps that involve
running `npm install` with git-based dependencies.

### DIFF
--- a/lib/cache.js
+++ b/lib/cache.js
@@ -281,7 +281,7 @@ function add (args, where, cb) {
     switch (p.type) {
       case 'local':
       case 'directory':
-        addLocal(p, null, cb)
+        addLocal(p, null, null, cb)
         break
       case 'remote':
         // get auth, if possible
--- a/lib/cache/add-local.js
+++ b/lib/cache/add-local.js
@@ -18,7 +18,7 @@ var iferr = require('iferr')
 
 module.exports = addLocal
 
-function addLocal (p, pkgData, cb_) {
+function addLocal (p, pkgData, shasum, cb_) {
   assert(typeof p === 'object', 'must have spec info')
   assert(typeof cb === 'function', 'must have callback')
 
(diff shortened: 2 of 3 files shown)


iarna:
This is really interesting, although I think I'm going to have to spend some more time thinking through the implications of having a shasum that can't be computed from the tarball in cache. (I'll do that thinking out load here in the issue when I do.)

This is definitely going to need some tests. Relatedly, I see that this patch is causing some test failures:

`test/tap/cache-add-localdir-fallback.js` &`test/tap/outdated-symlink.js`

stevecrozz:
Sounds good @iarna. I can sign up to write tests and fix existing tests that are broken, but I figured I should put that off until after we decide if this is something we want. I can chime in here if you want to discuss anything.

stevecrozz:
I had initially taken another approach with this where I would set the metadata in the tar archive to a static timestamp of anything, like the unix epoch. An approach like that would allow the declared shasum to match the contents of the archive and for the shasum to remain consistent since it would disregard the timestamps.

That could still be the right approach and we could explore that some more, but I found it pretty difficult when I was trying to do it that way.

RobAtticus:
As a developer who uses git for some internal libs, this would be greatly appreciated. It is a bit annoying that each `npm install` reinstalls the lib even when there are no changes.

iarna:
@RobAtticus I don't believe this will help you with that…

stevecrozz:
What can I do to help move this along. There are many possible ways to solve this issue and it seems like some discussion is needed in order to decide on the best course of action. Is there someone I can talk to about it?

stevecrozz:
Would it be easier to get this in if I could find a way to compute a shasum from the source files, excluding file metadata that causes each git checkout to end up with a different shasum? I tried doing that, but found it difficult. I could take another crack at it if that's going to be an easier path for review...

This has the potential to save a lot of internet bandwidth for people pushing and pulling docker images. Any docker layer that naively adds a node_modules folder that includes a git dependency always breaks the layer cache, forcing clients to download (in my case hundreds of megabytes) of otherwise cacheable data.

iarna:
When tar@3 lands we'll have this as tar.gz's will not vary when built from the same components. I'm closing this as soon-to-be-obsolete. (But also because this patches code that doesn't exist in npm any more, git code now lives in `pacote`.)

PR #12041, opened 2016-03-24
Update npm-disputes.md

Raynos:
Hi @izs 

I tried updating the disputes document with an "proof of concept" paragraph.

Based on recent feedback it seems like the disputes process might benefit from more "personal" discussions with the parties involved. A little human touch and an opportunity to allow parties to come to an "agree to disagree" conclusion with npm Inc might make the process more pleasant for everyone involved.

I know human problems don't always work out but making people feel like they have some kind of input even if ultimately futile can reduce frustration in the long run.

I suspect that if you do want to adjust this document, you would use different more appropriate phrasing.
Jake.

### DIFF
--- a/doc/misc/npm-disputes.md
+++ b/doc/misc/npm-disputes.md
@@ -67,6 +67,12 @@ regardless of the justification.  When humans solve their problems by
 talking to other humans with respect, everyone has the chance to end up
 feeling good about the interaction.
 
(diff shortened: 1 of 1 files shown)


isaacs:
This is a downstream doc from https://github.com/npm/policies, where a PR already exists on this subject.  (Haven't reviewed this language, but I'll take a look while reviewing that other one.)

https://github.com/npm/policies/pull/42

PR #12075, opened 2016-03-26
doc: add --ignore-scripts flag to install doc page

paulirish:
I was beyond confused that this worked, but I didn't see documentation for it on the web or in the `npm help install` page.

Turns out it's documented [over here](https://docs.npmjs.com/misc/config#ignore-scripts) but I think all runtime flags to `npm install` should be readable in that commands docs.


zkat:
LGTM! :)

zkat:
This was included with `npm@3.8.6` and `npm@2.15.3`. Thanks a bunch, @paulirish!! :tada: :sheep:

Issue #6024, opened 2014-08-23
npm should not cache tarballs from Github alongside registry tarballs


grncdr:
**Using `npm@2.0.0-beta.0`**

Tarballs from github should not be added to the npm cache, since they may (often) not be valid packages before their "prepublish" script is run.

For example:
``` bash
npm install petkaantonov/bluebird.git#v2.3.1
node -e 'require("bluebird")'
```
This will fail because bluebird needs a build step. What's worse though is if you install again from the registry, you get the cached github tarball:

``` bash
npm install bluebird@2.3.1
node -e 'require("bluebird")'
```

othiym23:
Closing as duplicate of #5941 – this is something we intend to fix via a cache rewrite. Thanks for your time, and patience!

Issue #2568, opened 2012-06-26
feature request: force install from cache.


dominictarr:
```
npm install module --cache
```

would behave as if there is not network connection.

this would be very useful when the connection is slow, 
and you know the modules you are installing well.

it would also create the possibility for a background process that tails npm
and keeps your cache up to date. making installs super fast!

ipuustin:
It would be great to get an update to this issue. Problem with `--cache-min 9999999` is that installing dependencies in a tree does not work: it expects packages to be installed in a flat hierarchy. In a perfect world I could add several packages in the cache and then install only one; the rest would be automatically used as dependencies to the top-level package.

despairblue:
Using `--cache-min 999999` will also fail `npm install` whenever you install a package that depends on a newer version of a package you installed in the passed. Took me a while to figure out why that happened:

![screen shot 2015-12-18 at 23 05 38](https://cloud.githubusercontent.com/assets/927609/11908549/5a2b5f92-a5dd-11e5-9b95-a5f54e3fe5be.png)

timoxley:
@despairblue Yep it's painful. Tracking that issue here https://github.com/npm/npm/issues/8581

nicgirault:
Does someone understand what is the main purpose of the npm cache directory if running twice 'npm install package' generate two http requests?

othiym23:
@nicgirault When packages are in the cache, the initial metadata request sends the cached etag for the package, and in the vast majority of cases, the registry will return a 304 and the package tarball won't need to be downloaded again.

nicgirault:
@othiym23 thank you, your answer is clear.

zachborboa:
This may now be relevant.
https://github.com/strongloop/loopback/issues/1079

othiym23:
@zachborboa That's strictly an issue with StrongLoop's packages, and is unrelated to this issue.

zachborboa:
@othiym23 I still think it is related. If the optional dependency were truly cached, they would only ever receive a single (analytics) ping for the specified version.

othiym23:
That's not how the cache works, @zachborboa – it's always going to fire off a request to see if it has the latest version of a resource. Having an "offline" or "use the cache only" mode would prevent those network checks, and, barring what changes would come from making the cache content-addressable / read-through by fetching tarballs from the cache by their shasums, not the default mode for the cache, now or in the planned future. Discussion about how StrongLoop uses `optionalDependencies` belongs over there, and is not a part of this discussion.

[...]

PR #12080, opened 2016-03-26
install: Add --from-cache to force using the cache

Zirak:
When passed, the registry client will favour taking modules from cache instead of hitting the network.

Begins to address #2568.

This does not fully address #2358 (`--offline`) since git urls and github shorthands are as of yet not handled (and may even complicate our lives further, as per #5941 and #6024). This is why I haven't documented the flag yet: does this flag along help enough to warrant its existence, or should it be dismissed until a real `--offline` flag is brought forth?

Another question: I'm a bit unsure how to go about testing this. I poked around [npm-registry-mock](https://github.com/npm/npm-registry-mock) and the tests a bit but haven't found what I need to test whether the registry was hit or not. Is there any precedent that I missed?

### DIFF
--- a/lib/cache/caching-client.js
+++ b/lib/cache/caching-client.js
@@ -99,6 +99,7 @@ function get_ (uri, cachePath, params, cb) {
   var stat = params.stat
   var etag
   var lastModified
+  var fromCache = npm.config.get('from-cache')
 
   timeout = Math.min(timeout, npm.config.get('cache-max') || 0)
   timeout = Math.max(timeout, npm.config.get('cache-min') || -Infinity)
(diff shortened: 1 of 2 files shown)


Zirak:
Looking around a bit I stumbled on [this comment by @othiym23](https://github.com/npm/npm/pull/10331#issuecomment-155506782) which made me reconsider my approach. Initially I went with subclassing `CachingClient` but realised I only wanted to amend that `if`. I'll think about this for a bit.

iarna:
We have `--offline` now, and `--prefer-offline` so this is now obsolete. (Logic for the new implementation is now in `pacote` and `cacache`)

Zirak:
That's good news! Thanks for following up.

Issue #12082, opened 2016-03-26
npm run-script doesn't run scripts (automatically exits with 0)


bfred-it:
Until a few hours ago, running `npm run build` would produce an output similar to this:

<img src="https://raw.githubusercontent.com/dsernst/npm-run-help/master/screenshot.png">

Now it just exits without any output unless I use `--verbose`, which still doesn't fix it.

<img width="500" alt="npm info it worked if it ends with ok
npm verb cli [ '/usr/local/bin/node',
npm verb cli   '/usr/local/bin/npm',
npm verb cli   'run',
npm verb cli   'build',
npm verb cli   '--verbose' ]
npm info using npm@3.7.3
npm info using node@v5.8.0
npm verb run-script [ 'prebuild', 'build', 'postbuild' ]
npm info lifecycle my-module@1.0.0~prebuild: my-module@1.0.0
npm info lifecycle my-module@1.0.0~build: my-module@1.0.0
npm info lifecycle my-module@1.0.0~postbuild: my-module@1.0.0
npm verb exit [ 0, true ]
npm info ok" src="https://cloud.githubusercontent.com/assets/1402241/14060307/5d82592e-f31d-11e5-8c4c-ab08136011a8.png">

`package.json` is read correctly because if I remove the `build` script it tells me it's missing.

```json
{
  "name": "my-module",
  "version": "1.0.0",
  "description": "my-module",
  "scripts": {
    "build": "echo 1"
  }
}
```

It doesn't matter what version of npm and node I try, even 0.10 and 2.x.x. It always behaves the same. Currently on:

```
$ npm -v
3.7.3
$ node -v
v5.8.0
```

I tried

* rebooting
* cleaning cache
* uninstalling npm completely and reinstalling it but nothing changed. 

## Edit 1

Currently debugging this with:

```sh
node-inspector &
node --debug-brk /usr/local/bin/npm run build
```

I also found the log level "silly" which gives some important information:

<img width="728" alt="screen shot 2016-03-26 at 23 38 10" src="https://cloud.githubusercontent.com/assets/1402241/14061159/dfaa6e94-f3ab-11e5-90ce-93c4235526e4.png">

Will keep updating

bfred-it:
Okay. This is resolved. I had followed the [recent suggestion to `ignore-scripts`](http://blog.npmjs.org/post/141702881055/package-install-scripts-vulnerability) globally:

```sh
npm config set ignore-scripts true
```

[...]

PR #12083, opened 2016-03-26
lifecycle: inform users of why scripts are being ignored

fregante:
A global setting of `ignore-scripts:true` can cause `npm run-script` to silently fail to run scripts and cause issues like #12082.

This will add an info-level log. It goes from this:

screen shot 2016-03-26 at 23 57 25

to this:

screen shot 2016-03-26 at 23 57 07

### DIFF
--- a/lib/utils/lifecycle.js
+++ b/lib/utils/lifecycle.js
@@ -46,7 +46,12 @@ function lifecycle (pkg, stage, wd, unsafe, failOk, cb) {
   if (!pkg) return cb(new Error('Invalid package data'))
 
   log.info('lifecycle', logid(pkg, stage), pkg._id)
-  if (!pkg.scripts || npm.config.get('ignore-scripts')) pkg.scripts = {}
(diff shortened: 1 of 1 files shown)


bfred-it:
Re-pushed because the tests failure was unrelated to my code

iarna:
There's talk of having `npm run-script` still work when `ignore-scripts` is true, but this of course won't carry over to times that scripts are being invoked implicitly (as that's the whole point of the option), so this still seems valuable to me, particularly at the info level.

jdalton:
I actually punted on `ignore-scripts` because things like `npm test` broke.

zkat:
This got released along with `npm@3.8.7` -- thanks, @bfred-it! :tada:

bfred-it:
Cool! Thanks for the merge :)

I was wondering if we should avoid some repetition from all those logs (in red)

f6fc3c5a-f3ae-11e5-9349-6e21721d154f

PR #12084, opened 2016-03-26
Correct possibly stale cache from network

### LINKED ISSUE #8581: cache-min & unknown version

Zirak:
When a package is fetched from the cache which cannot satisfy the version requirements, an attempt to fetch it fron the network is made.

Fix #8581.

I'm in the middle of figuring out how to properly test this: it requires setting up the cache and  publishing a new version (one way or another). I tested it logically using _science_*.

\* Science in this case being manually editing the `.cache.json` file.


timoxley:
@Zirak nice, thank you for looking into this.

lib/fetch-package-metadata.js
```diff
@@ -149,6 +149,16 @@ function fetchNamedPackageData (dep, next) {
         }
       }
 
+      // We didn't manage to find a compatible version
+      // If this package was requested from cache, force hitting the network
+      if (pkg._cached) {
+        log.silly('fetchNamedPackageData', 'No valid target from cache, forching network')
```
ralt: Typo ("forching")

Zirak: @ralt Thanks, fixed.

iarna:
Thanks for putting this all together, we landed this today in 3.9.0!

iarna:
FYI, I rewrote your test a bit to not use a fixture folder (we're moving away from that). You can see the rewritten test here: https://github.com/npm/npm/blob/111ae3ec366ece7ebcf5988f5bc2a7cd70737dfe/test/tap/retry-on-stale-cache.js

PR #12092, opened 2016-03-27
cli: remove unused file cli.js

watilde:
this file is not used currently:
https://github.com/npm/npm/commits/master/cli.js


watilde:
oops, it worked on my env but I need to re-read the codes... Let me close this.

PR #12093, opened 2016-03-27
package-json: update bugs url to https from http

watilde:


zkat:
LGTM! Thanks, @watilde! :)

zkat:
This shipped with `npm@3.8.6` and `npm@2.15.3`. Thanks for the lovely patches as always, @watilde!

PR #12096, opened 2016-03-27
Replace cli.js with bin/npm-cli.js and remove cli.js

watilde:
The first of the `cli.js` was used for the bin field in package.json, but now it is used only by internal build codes. Also, some docs pointed it as an example instead of `bin/npm-cli.js`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,23 +53,23 @@ latest:
 	@echo "Installing latest published npm"
 	@echo "Use 'make install' or 'make link' to install the code"
 	@echo "in this folder that you're looking at right now."
-	node cli.js install -g -f npm ${NPMOPTS}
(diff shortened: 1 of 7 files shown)


zkat:
This looks fine to me. LGTM! Thanks, @watilde

watilde:
@iarna Hi! Was this patch added into v4, even this PR doesn't have the milestone?

zkat:
@watilde I started an npm@5 release branch for ongoing npm@5 work. This has been merged in. Thank you for your patience, and thank you for the patch and the helpful work, as always. <3

npm@5 will become `latest` sometime in the next month or two, hopefully. In the meantime, people can play with that branch. :)

PR #12107, opened 2016-03-28
doc: Add 'login' as alias to 'adduser' cli command

gnerkus:


iarna:
Seems reasonable to me. We have a few other aliases in the short-help for other commands, and this one is common enough that I think it should be there.

zkat:
This got released with `npm@3.8.7`. Thanks for the doc patch, @gnerkus!

I'm gonna take off the `merge-to-latest` tag, and include this in LTS next week. I'll close it after that. :+1:

iarna:
Merged into v2.15.4!

PR #10547, opened 2015-11-25
remove faq

ashleygwilliams:
fixes #10546

PR #12143, opened 2016-03-30
remove faq link and command as it's removed at #10547

watilde:
The faq docs was already removed at #10547, but related codes and links are still there.
Also, the build system as a prepublish script in npm  to create man and html directory does not clear old docs currently. Let me create a PR for it as well later.

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,7 +141,6 @@ you have chosen.
 ## More Docs
 
 Check out the [docs](https://docs.npmjs.com/),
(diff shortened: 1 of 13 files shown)


zkat:
LGTM! Thanks!

zkat:
This was merged into `npm@3.8.6`. Thanks! :dancer: I omitted it from `npm@2.15.3`.

PR #12146, opened 2016-03-30
makefile: add doc-clean to `make publish`

watilde:
from #12143. The official [release process](https://github.com/npm/npm/wiki/Release-Process#thursday-cutting-the-release) has a release command `make publish` and the command does not do clean docs before releasing new version. It means there is a possibility that some compiled docs files which is removed a long time ago exist in release codes, depends on the local of the release manager.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -166,7 +166,7 @@ ls-ok:
 gitclean:
 	git clean -fd
 
-publish: gitclean ls-ok link doc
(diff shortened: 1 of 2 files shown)


zkat:
LGTM! Thanks! :)

zkat:
This shipped with `npm@3.8.6` and `npm@2.15.3`. Thanks again for the patches, as always. :dancer:

PR #12150, opened 2016-03-30
Ensure that 'npm cache ls' outputs actual filenames

isaacs:
Hi, cli team.  I don't know if there's a bug open for this already, but I noticed it the other day, and had a moment to write some code, so here you go :D

### DIFF
--- a/lib/utils/completion/file-completion.js
+++ b/lib/utils/completion/file-completion.js
@@ -1,7 +1,6 @@
 module.exports = fileCompletion
 
 var mkdir = require('mkdirp')
-var path = require('path')
 var glob = require('glob')
 
 function fileCompletion (root, req, depth, cb) {
@@ -18,8 +17,7 @@ function fileCompletion (root, req, depth, cb) {
     glob(pattern, opts, function (er, files) {
       if (er) return cb(er)
       return cb(null, (files || []).map(function (f) {
-        var tail = f.substr(root.length + 1).replace(/^\//, '')
-        return path.join(req, tail)
+        return f.substr(root.length + 1).replace(/^\/|\/$/g, '')
       }))
     })
   })
--- a/test/tap/builtin-config.js
+++ b/test/tap/builtin-config.js
@@ -84,6 +84,7 @@ test('use first npm to install second npm', function (t) {
           bin,
           'install', npm,
           '-g',
+          '--ignore-scripts',
           '--prefix=' + folder + '/second',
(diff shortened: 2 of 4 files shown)


iarna:
Hmm, this test passes w/o the patch so... I'm not actually clear on what the bug was to start with?

iarna:
We have observed that `npm cache ls` can end up with a doubled up portion, which maybe this was meant to address. I'll double back and see if this doesn't fix that problem (and fix up the test if so).

isaacs:
@iarna Weird!  The test definitely didn't pass on my system without the patch when I wrote it, but does pass now

The root problem still exists, though, so the test is probably just not reproducing it properly or something.

```
$ npm cache ls rimraf
~/.npm/rimraf/rimraf/
~/.npm/rimraf/rimraf/2.1.4/
~/.npm/rimraf/rimraf/2.1.4/package.tgz
~/.npm/rimraf/rimraf/2.1.4/package/
~/.npm/rimraf/rimraf/2.1.4/package/package.json
~/.npm/rimraf/rimraf/2.2.8/
~/.npm/rimraf/rimraf/2.2.8/package.tgz
~/.npm/rimraf/rimraf/2.2.8/package/
~/.npm/rimraf/rimraf/2.2.8/package/package.json
~/.npm/rimraf/rimraf/2.5.0/
~/.npm/rimraf/rimraf/2.5.0/package.tgz
~/.npm/rimraf/rimraf/2.5.0/package/
~/.npm/rimraf/rimraf/2.5.0/package/package.json
~/.npm/rimraf/rimraf/2.5.1/
~/.npm/rimraf/rimraf/2.5.1/package.tgz
~/.npm/rimraf/rimraf/2.5.1/package/
~/.npm/rimraf/rimraf/2.5.1/package/package.json
~/.npm/rimraf/rimraf/2.5.2/
~/.npm/rimraf/rimraf/2.5.2/package.tgz
~/.npm/rimraf/rimraf/2.5.2/package/
~/.npm/rimraf/rimraf/2.5.2/package/package.json

$ ls -laF ~/.npm/rimraf/rimraf/
ls: /Users/isaacs/.npm/rimraf/rimraf/: No such file or directory
```

isaacs:
Updated test to fail on master properly.  It was only passing if no path arg was passed to cache ls.

isaacs:
Travis fails are now only due to timeouts in Node 0.x.

iarna:
This landed in 3.9.0 =)

PR #12213, opened 2016-04-05
config: fix a shorthand value of `-n`

watilde:
In docs, a shorthand `-n` means `--yes false`, but the code has `--no-yes`. I di grep `--no-yes` in all codes of npm cli, but I couldn't find it anywhere so I think the code is little bit old and the docs is correct.

refs: [npm-config.md#shorthands-and-other-cli-niceties](https://github.com/npm/npm/blob/master/doc/misc/npm-config.md#shorthands-and-other-cli-niceties)

### DIFF
--- a/lib/config/defaults.js
+++ b/lib/config/defaults.js
@@ -385,7 +385,7 @@ exports.shorthands = {
   E: ['--save-exact'],
   O: ['--save-optional'],
   y: ['--yes'],
-  n: ['--no-yes'],
+  n: ['--yes', false],
   B: ['--save-bundle'],
   C: ['--prefix']
 }


othiym23:
The two are equivalent in `nopt`'s topsy-turvy world, so in the interests of keeping churn around config to a minimum, I'm going to reject this change. Thanks for putting it together, though, @watilde!

watilde:
Hi Forrest! Thanks for your comment. I didn't know that `--no` means adding `false` as an argument for an option. Seems interesting!

Well, I just noticed that npm config can have a `yes`, but there is no docs in `misc/npm-config.md`.

othiym23:
> I just noticed that npm config can have a `yes`, but there is no docs in `misc/npm-config.md`. Is there any special meaning on it...?

You'd have to [take a look at `nopt`](https://github.com/npm/nopt#readme) to know if it's generally usable as an override, but `--yes` on its own (and `-y`) are used by `npm init`, among other subcommands.

watilde:
Yes it is, `--yes` are used by init and also can be overwritten: https://gist.github.com/watilde/6711e517ceb176f92992837e0c5292d7#file-use-case-md

Isn't it a similar case with `if-present` config as a one that is used by only one command and can be overwritten...?

watilde:
^I meant it like this commit ;)
https://github.com/watilde/npm/commit/c9e08b30ced03003d4100ad768197c9ed27485fa

PR #12221, opened 2016-04-06
[WIP] Fix lifecycle signal test

### LINKED ISSUE #12220: Broken Lifecycle Signal Tests
The tests in `test/tap/lifecycle-signal.js` are skipped on Windows and Travis, but are failing on Linux.
- Ubuntu 15.04
- node v5.10.0

**lifecycle signal abort**

This test fails because regardless of the signal used to kill the process in `preinstall` the child's exit code is `1` and exit signal is `null`. There seems to be an error handler stripping the signal and changing the code to 1. I tested that `node -e` wasn't eating the error, so it must be somewhere around the install action or the lifecycle.

[test/tap/lifecycle-signal.js#L45](https://github.com/npm/npm/blob/545434766eb3681d3f40b745f9f3187ed63f310a/test/tap/lifecycle-signal.js#L45)

Originally added by @TooTallNate in https://github.com/npm/npm/commit/d885f81ec67c9ac583a830f4dfa7ad8fb0e9c21f

**lifecycle propagate signal term to child**

This test fails, because the test assumes killing a child process kills its subprocess, but that is incorrect.

> I think you have the wrong idea about processes and process groups. On exit, child processes get reparented, they're not forcibly killed by io.js or the kernel.
> https://github.com/nodejs/node/issues/2098#issuecomment-118322915

http://stackoverflow.com/questions/33325301/killing-a-bash-script-does-not-kill-child-processes

[test/tap/lifecycle-signal.js#L67](https://github.com/npm/npm/blob/545434766eb3681d3f40b745f9f3187ed63f310a/test/tap/lifecycle-signal.js#L67)

Originally added by @daniel-pedersen in https://github.com/npm/npm/commit/545434766eb3681d3f40b745f9f3187ed63f310a.

---

```
ubuntu@ubuntu:~/npm$ DEPLOY_VERSION=testing npm run tap -- "test/tap/lifecycle-signal.js" 

> npm@3.8.6 tap /home/ubuntu/npm
> tap --timeout 240 "test/tap/lifecycle-signal.js"

test/tap/lifecycle-signal.js .......................... 5/9 4m
  lifecycle signal abort
  not ok should be equal
    +++ found                                                          
    --- wanted                                                         
    -[null]                                                            
    +1                                                                 
    compare: ===
    at:
      line: 44
      column: 7
      file: test/tap/lifecycle-signal.js
      type: ChildProcess
    stack: |
      ChildProcess.<anonymous> (test/tap/lifecycle-signal.js:44:7)
      Pipe._onclose (net.js:477:12)

  lifecycle signal abort
  not ok should be equal
    +++ found                                                          
    --- wanted                                                         
    -"SIGSEGV"                                                         
    +[null]                                                            
    compare: ===
    at:
      line: 45
      column: 7
      file: test/tap/lifecycle-signal.js
      type: ChildProcess
    stack: |
      ChildProcess.<anonymous> (test/tap/lifecycle-signal.js:45:7)
      Pipe._onclose (net.js:477:12)

  lifecycle propagate signal term to child
  not ok expected to throw
    at:
      line: 68
      column: 7
      file: test/tap/lifecycle-signal.js
      type: ChildProcess
    stack: |
      ChildProcess.<anonymous> (test/tap/lifecycle-signal.js:68:7)

  not ok timeout!
    at:
      line: 3
      column: 11
      file: node_modules/tap/lib/root.js
    signal: SIGTERM
    handles:
      - type: Socket
        events:
          - end
          - finish
          - _socketEnd
          - close
      - type: Socket
        events:
          - end
          - finish
          - _socketEnd
          - close
          - data

total ................................................. 5/9


  5 passing (4m)
  4 failing
```

deckar01:
~~Fixes~~ Investigates #12220


deckar01:
Fixing the tests to match the current behavior might not be the desired fix. If the implementation needs to be fixed rather than the tests, this test should probably just be skipped by all platforms so that there is a breadcrumb to follow later.

zkat:
This probably needs a closer look: the tests pass on OSX for me, which means we're seeing some platform differences.

deckar01:
@zkat I am going to build on a different Linux platform to see if the issue is Ubuntu/Debian specific.

deckar01:
The tests pass on OpenSUSE Tumbleweed using both node `v4.3.1` and `v5.10.0`.

The tests fail on Ubuntu 15.10 using both `v0.10.25` and `v5.10.1`.

iarna:
Yeah, the travis exclusion in the test seems extremely sketchy to me. It was added in 7d0aec2, but the commit description does not seem to justify the  change.

iarna:
I'd like to see any patch to this remove the travis exclusion and pass in our matrix of npm versions on Travis, in addition to whatever distro specific testing we end up doing.

Travis is Ubuntu of some vintage, IIRC?

deckar01:
@iarna I agree. When I remove the Travis exclusion (with the changes in this PR) I get [inconsistency between node versions](https://travis-ci.org/deckar01/npm/builds/121044031).

node v0.12 an iojs seem to be simplifying the error signal and code on Ubuntu 12.04...

I got the failure on (Ubuntu 15.04 + node v5.10.1) and (Ubuntu 15.10 + node v0.10.25, v5.10.1)

 - [ ] I will remove the Travis exclusions from master and see if the results are inverted.

Travis uses Ubuntu 12.04 SE by default.

 > STANDARD: OS - Ubuntu 12.04 LTS Server Edition 64 bit
https://docs.travis-ci.com/user/ci-environment/#Virtualization-environments

 - [ ] I will also run Travis on Ubuntu 14.04 (Trusty) to see if anything changes.

deckar01:
- I removed the Travis exclusions from master and the results were not inverted.
   - node [v5](https://travis-ci.org/deckar01/npm/jobs/121566046#L1962), [v4](https://travis-ci.org/deckar01/npm/jobs/121566047#L2096), [iojs](https://travis-ci.org/deckar01/npm/jobs/121566048), [v0.12](https://travis-ci.org/deckar01/npm/jobs/121566049#L2094), [v0.10](https://travis-ci.org/deckar01/npm/jobs/121566050#L2101), & [v0.8](https://travis-ci.org/deckar01/npm/jobs/121566051#L2211).
 - I also ran Travis on Ubuntu 14.04 (Trusty) the results were the same as Ubuntu 12.04.
   - node [v5](https://travis-ci.org/deckar01/npm/jobs/121566846#L1962), [v4](https://travis-ci.org/deckar01/npm/jobs/121566847#L2095), [iojs](https://travis-ci.org/deckar01/npm/jobs/121566850#L2184), [v0.12](https://travis-ci.org/deckar01/npm/jobs/121566852#L2094), [v0.10](https://travis-ci.org/deckar01/npm/jobs/121566854#L2101), & [v0.8](https://travis-ci.org/deckar01/npm/jobs/121566856#L2211).

These results indicate that on Ubuntu the error response from child processes are consistently simplified across node versions.

The [previous results](https://travis-ci.org/deckar01/npm/builds/121044031) indicate that some versions of node the lifecycle logic is [not killing the child process](https://travis-ci.org/deckar01/npm/jobs/121044032#L1964) properly.

This means that `lifecycle.js` may need to [start child processes in a process group so that they can be killed in a way that propagates](http://stackoverflow.com/a/33367711). I'm not sure this is a cross platform solution, so it may be a good idea to think about using a library that normalizes this behavior across platforms.

An [archived node issue](https://github.com/nodejs/node-v0.x-archive/issues/1811) had a mention in [another issue](https://github.com/szwacz/electron-boilerplate/issues/12) which lead me to [`tree-kill`](https://www.npmjs.com/package/tree-kill).

 - [ ] I will give `tree-kill` a shot and see if it propagates error signals more consistently.

daniel-pedersen:
When investigating now I discovered that lifecycle does not spawn child processes directly, instead going via `sh`. This makes the tests dependent on which shell is used for spawning the child. In OS X, `bash` is used, which seems to kill its process tree upon exit, which is why the tests pass.

My commit was hasty and I just happened to copy the travis exclusion from the previous test. A proper recursive-kill solution in lifecycle is needed to guarantee expected behavior across environments.

deckar01:
Using `tree-kill` for the `spawn.kill()` implementation fixes the `'lifecycle propagate signal term to child'` test, which was failing with a timeout due to the fact that the inner child was never being killed.

It does not fix the `'lifecycle propagate signal term to child'` test, which is failing due to the error signal being simplified.

I'm not sure what to make of this:

```js
// test.js

var child_process = require('child_process')

child = child_process.spawn(
  'node', ['-e', 'process.kill(process.pid, "SIGSEGV")']
)

// This is basically what lifecycle.js is doing on line 234.
child.on('close', function(code, signal) {
  console.log('CLOSE (' + code + ') (' + signal + ')')
  process.kill(process.pid, signal)
})
```

```
$ node test.js; echo $?
CLOSE (null) (SIGSEGV)
Segmentation fault (core dumped)
139
```

Node seems to return the right signal.

I put the same debug statement in the close handler on line 234, then added an npm script to test it:

```json
"crash": "node -e \"process.kill(process.pid,'SIGSEGV')\""
```

```
$ npm run crash; echo $?

> npm@3.8.6 crash /home/jared/src/npm
> node -e "process.kill(process.pid,'SIGSEGV')"

Segmentation fault (core dumped)
CLOSE (139) (null)

npm ERR! ...
...

1
```

It looks like when npm tries to kill its process, `signal` is `null`, so it ends up exiting with code `1`.

If I add a line to fill in the missing signal value:

```js
if(code == 139) signal = 'SIGSEGV'
```

I get:

```
$ npm run crash; echo $?

> npm@3.8.6 crash /home/jared/src/npm
> node -e "process.kill(process.pid,'SIGSEGV')"

Segmentation fault (core dumped)
CLOSE (139) (null)
Segmentation fault (core dumped)
139
```

And that code to signal mapping allows the tests to pass.

I'm not sure why pure node only has the signal and the npm run script only has the code, but whatever is causing it is platform dependent. I am still curious what the core issue is, but I have a feeling that it might be external and a code to signal mapping might be the solution.

iarna:
Some in-air discussion between @othiym23 and I:

We think that we should change `npm` to use bash explicitly and not `/bin/sh`, because there are just too many places `npm` depends on, sometimes subtle, sometimes not so subtle behavior of bash.

deckar01:
`dash` is the offending shell on Ubuntu and Debian, which is a minimal POSIX implementation, so you shouldn't have to worry about loosing features when switching from `dash` to `bash` [0].

There is a possibility that there are systems configured to use an even more featureful shell than `bash` that could cause scripts to break when switching to `bash`.

`/bin/sh` is guarunteed to be aliased on all POSIX systems, but `bash` is not [1].

I don't think switching to `bash` is a bad idea, but it does mean relying on an extension specific behavior.

On one hand, adding a shim to kill process trees across all platforms guarantees backwards compatibility.

On the other hand, standardizing and documenting the npm script execution environment seems like a step towards a more stable deploy environment.

[0] https://wiki.ubuntu.com/DashAsBinSh
[1] http://stackoverflow.com/questions/5725296/difference-between-sh-and-bash

deckar01:
Closing due to inactivity. I will leave #12220 open until the tests are passing on Ubuntu locally.

PR #12241, opened 2016-04-06
gitignore: add /test/npm_cache

zkat:
This file is just kind of annoying to have in git status all the time, and I figure we may as well ignore it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@ npm-debug.log
 /npm-*.tgz
 /node_modules/npm-registry-client/test/fixtures
 /test/fixtures/config/userconfig-with-gc
+/test/npm_cache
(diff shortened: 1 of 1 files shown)


othiym23:
The problem with just ignoring the directory is that it makes it harder to notice when a test may be failing due to a busted `test/npm_cache` hanging around from a previous test run. The real fix for this is to switch the tests to use a version of the tap runner that comes out of `test/common-tap.js` so that we can ensure that each test suite gets a clean cache.

iarna:
Removing it is what `zz-clieanup.js` does, incidentally. (And currently `test/common-tap.js` is what defines this folder and injects this config into the env– everything using `common-tap` uses this folder.)

zkat:
Closing this. Thanks for the pointers 👯

Issue #11976, opened 2016-03-19
material-design-icons failing to install with "ERR! cb() never called!"


shyndman:
Hi there,

I'm a maintainer of the [google/material-design-icons](https://github.com/google/material-design-icons) repo. We've received a report that our 2.2.2 release fails to install with `"ERR! cb() never called!"` ([issue](https://github.com/google/material-design-icons/issues/342)) . I've been able reproduce myself, with the two versions npm 3.7.3 and 3.8.2. Version 2.2.1 works fine.

I'd assumed that I failed to publish properly, so went ahead and cut a new 2.2.3 version, but I'm still seeing the same problem.

Any sense of what's going on? Nothing stands out when I run `npm -d install`.

The only thing I did notice is that package.json downloaded for 2.2.1 is quite different than 2.2.2's.

pYr0x:
hi,
same problem with https://github.com/canjs/canjs
using:
* npm 3.7.3
* node 5.9.0

only-style:
how to solve the problem guys?

- nodejs - 5.9.0
- npm - 3.7.3

erykpiast:
I got the same, versions like @pYr0x has. But cannot say what package exactly, it just fails when I try to install dependencies for my app.

Klathmon:
@erykpiast I figured out which package it was by doing the old fashioned "delete half my deps and try again, if it works, switch which half i removed, repeat until i'm down to 1 package" 

In my case it ended up being the google/material-design-icons package but only versions after 2.2.1

erykpiast:
So, in my case it's `typescript 0.9.1-1` (as a `devDependency`).

fzaninotto:
Same problem with `npm install ng-admin@1.0.0-alpha1`

arman-mukatov:
That should work:

`npm cache clean`

simply remove the .npm folder from our home directory.

`rm -rf ~/.npm`

and try again. It worked for me!

xorgy:
I'm also seeing this issue.

unindented:
node 5.9.0
npm 3.7.3

TravisCI is alerting me of this issue across a bunch of projects, and I'm seeing it locally too.

`npm cache clean` didn't fix it for me.

greyhwndz:
have the same error on my end:
`npm ERR! cb() never called!`
node: v5.9.0
npm: 3.8.2

Tried `npm cache clean` & `rm -rf ~/.npm` to no effect

pornel:
```sh
[...]

PR #12281, opened 2016-04-08
Do not swallow errors, when reading invalid tgz files

inthemill:
When Errors occur reading tgz files (why ever) npm says
`npm ERR! cb() never called!`
instead of printing a meaningful error.
This PR fixes it.

### DIFF
--- a/lib/fetch-package-metadata.js
+++ b/lib/fetch-package-metadata.js
@@ -277,8 +277,8 @@ function hasTarHeader (c) {
 
 function untarStream (tarball, cb) {
   validate('SF', arguments)
-  
-  var cbOnceForError= once(cb)
+
+  var cbOnceForError = once(cb)
   var stream
   var file = stream = fs.createReadStream(tarball)
   var tounpipe = [file]
--- a/lib/fetch-package-metadata.js
+++ b/lib/fetch-package-metadata.js
@@ -277,15 +277,15 @@ function hasTarHeader (c) {
 
 function untarStream (tarball, cb) {
   validate('SF', arguments)
-  cb = once(cb)
-
+  
+  var cbOnceForError= once(cb)
   var stream
   var file = stream = fs.createReadStream(tarball)
   var tounpipe = [file]
   file.on('error', function (er) {
     er = new Error('Error extracting ' + tarball + ' archive: ' + er.message)
     er.code = 'EREADFILE'
-    cb(er)
+    cbOnceForError(er)
   })
   file.on('data', function OD (c) {
     if (hasGzipHeader(c)) {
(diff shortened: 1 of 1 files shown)


iarna:
Could you write a test for this?

My attempts at reproducing the problem you described aren't working. Further, looking at the code you changed I don't get why allowing the callback in https://github.com/npm/npm/pull/12281/files#diff-d3763bf2eddb70ce9fcd3df658b6193cL304 to happen more than once (or to happen if an error already happened) would actually help this situation?

iarna:
(Even just a tarball that reproduces this problem would be fantastic!)

inthemill:
Looking at line 304 you see, that cb is always when start reading the file, an error which occurs after that is swallowed.

inthemill:
It happens for me with a tarball which I can not provide for legal reasons. The problem seems only to occur on certain node or npm versions. I can imagine that issue #11976 covers the same issue.

I tried to write a unit test, but failed, as the code has too many dependencies and I do not know how you would mock them. There is no test in the first place for this method, which could have helped me.

inthemill:
I will try to provide you with more information to reproduce the problem.

iarna:
If you can just find a tarball that you can share to reproduce the problem with, we can put together a test for this pretty trivially. (Not a unit test, mind, but a test just the same.)

iarna:
Ah, I see, it's calling back before its done! That's not right. I think we might be able to solve this by moving 304 down to 339. That is to say, don't call it w/o error args until we get a close event from the final stream.

iarna:
Given that, I think I may see how to trigger this error–how big is the package file that's causing the error for you?

iarna:
Ok, so refreshing my brain as to why this code works this way:

That early exit is correct. It's exiting early because what the consumer of this function does is work with the returned stream, and we want to be able to stream the results without having to buffer the entire file in memory.

What those error handlers should be doing is passing the error event along to the final stream, eg, line 312 should read:
```
      stream.emit('error', er)
```

Neither `doUntar` nor `doGunzip` should be calling `cb` under any circumstances.

This means that `addShrinkwrap` (which is the consumer of `untarStream`) should also provide error handlers on its stream.

inthemill:
The problem was not the package, but i guess a bug in the zlib part of node 5.9.  Node 5.10 works great again.
To write a test, I guess you have to fake gunzip or untar  to produce an error 'mid-stream'.

iarna:
Alright, since this isn't even repeatable for you with recent nodes I'm going to close this PR. I _am_ making a branch that closes the loophole I identified here however.

inthemill:
Thanks, i'm sure your PR will fix the problem I encountered with swallowed errors much cleaner.

Issue #12276, opened 2016-04-08
Extra peer dependencies are installed when shrinkwrap is present


robrichard:
Using npm 2.15.3

If I set up a project with three dependencies: 

``` json
  "dependencies": {
    "react": "^0.14.8",
    "react-bootstrap": "^0.28.5",
    "react-dom": "^0.14.8"
  }
```

Note that react-bootstrap@0.28.5 has a peer dependency on react >=0.14.0 and react-dom >=0.14.0.

Starting from an empty node_modules folder, running npm install will install these packages:

```
react-dom@0.14.8 node_modules/react-dom

react-bootstrap@0.28.5 node_modules/react-bootstrap
├── uncontrollable@3.2.3
├── classnames@2.2.3
├── react-prop-types@0.3.0
├── keycode@2.1.1
├── dom-helpers@2.4.0
├── react-overlays@0.6.3 (react-prop-types@0.2.2)
├── warning@2.1.0 (loose-envify@1.1.0)
├── invariant@2.2.1 (loose-envify@1.1.0)
├── lodash-compat@3.10.2
└── babel-runtime@5.8.38 (core-js@1.2.6)

react@0.14.8 node_modules/react
├── fbjs@0.6.1 (whatwg-fetch@0.9.0, ua-parser-js@0.7.10, loose-envify@1.1.0, promise@7.1.1, core-js@1.2.6)
└── envify@3.4.0 (through@2.3.8, jstransform@10.1.0)
```

So far so good. If I run `npm shrinkwrap`, a shrinkwrap file will be produced (attached). The shrinkwrap seems fine and matches what is currently installed.

Now if I remove `node_modules` and run `npm install` again (with the shrinkwrap present), the following packages are installed

```
react-dom@0.14.8 node_modules/react-dom

react@0.14.8 node_modules/react
├── envify@3.4.0 (through@2.3.8, jstransform@10.1.0)
└── fbjs@0.6.1 (whatwg-fetch@0.9.0, ua-parser-js@0.7.10, loose-envify@1.1.0, promise@7.1.1, core-js@1.2.6)

react-bootstrap@0.28.5 node_modules/react-bootstrap
├── uncontrollable@3.2.3
├── classnames@2.2.3
├── react-prop-types@0.3.0
├── keycode@2.1.1
├── dom-helpers@2.4.0
├── warning@2.1.0 (loose-envify@1.1.0)
├── invariant@2.2.1 (loose-envify@1.1.0)
[...]

PR #12290, opened 2016-04-09
Peer shrinkwrap fix

### LINKED ISSUE #5135: peerDependencies and npm shrinkwrap
If I have an app[1] that requires a module, in this case `express@~3.4.8` and it also has a module[2] that requires that same version. That module[2] also requires another module[3] that requires modules (`express-state` & `express-map`) with a peerDependency on `express@3.x`.

If I install this, as is, then it will work perfectly:
https://gist.github.com/davglass/11225979#file-no-shrinkwrap-txt

However, if I add a shrinkwrap file:
https://github.com/davglass/peer-app/blob/master/npm-shrinkwrap.json

Then try to install it:
https://gist.github.com/davglass/11225979#file-shrinkwrap-txt

It will not recognize that `express@3.4.8` was installed at a higher level and it will pull in the latest `express@3.5.1` which will break the application. 

What I think is happening here is that when processing the shrinkwrap file, the peerDependencies logic kicks in and says "hey, I need `express@3.x` and it's not at this level, go fetch it" when it should check to see if it's available in the tree above it. This is what "I think" `npm install` does for us. But since that logic is skipped for the shrinkwrap, it's actually doing the wrong thing.

Links:
[1]app: https://github.com/davglass/peer-app
[2]module1: https://github.com/davglass/module1
[3]module2: https://github.com/davglass/module1

Possibly related to #5080 

I'm happy to dive in and see if I can find the root cause of this issue but I wanted to confirm first that this is an issue.

If #5080 removes the installation of peerDependencies and just warns, then I think this issue will go away since the install won't happen.

ghost:
This should fix #5135.  

I read the guidelines for contributing which recommended squashing to a single commit.  I intentionally left two commits so that you can verify the added test is failing first.  I was just bitten by not doing this, as I was comparing two undefined values output by the rmFrom function.  There was also another test that was fixed, that was also comparing two undefined values.

In response to [CERT vulnerability note VU#319816](https://www.kb.cert.org/vuls/id/319816) we started requiring the use of npm-shrinkwrap.json files in our builds. As this issue circumvents this precaution while also breaking applications, we are anxious to see this issue resolved in the next lts release.


majgis:
The following provides some insight into what the problem was:

```bash
$ grep -r -A 3 "\"peerDependencies\"" node_modules/
node_modules//react-bootstrap/node_modules/react-dom/package.json:  "peerDependencies": {
node_modules//react-bootstrap/node_modules/react-dom/package.json-    "react": "^15.0.1"
node_modules//react-bootstrap/node_modules/react-dom/package.json-  },
node_modules//react-bootstrap/node_modules/react-dom/package.json-  "scripts": {
--
node_modules//react-bootstrap/node_modules/react-overlays/package.json:  "peerDependencies": {
node_modules//react-bootstrap/node_modules/react-overlays/package.json-    "react": ">=0.14.0",
node_modules//react-bootstrap/node_modules/react-overlays/package.json-    "react-dom": ">=0.14.0"
node_modules//react-bootstrap/node_modules/react-overlays/package.json-  },
--
node_modules//react-bootstrap/node_modules/uncontrollable/package.json:  "peerDependencies": {
node_modules//react-bootstrap/node_modules/uncontrollable/package.json-    "react": ">=0.11.0"
node_modules//react-bootstrap/node_modules/uncontrollable/package.json-  },
node_modules//react-bootstrap/node_modules/uncontrollable/package.json-  "release-script": {
--
node_modules//react-bootstrap/package.json:  "peerDependencies": {
node_modules//react-bootstrap/package.json-    "react": ">=0.14.0",
node_modules//react-bootstrap/package.json-    "react-dom": ">=0.14.0"
node_modules//react-bootstrap/package.json-  },
--
node_modules//react-dom/package.json:  "peerDependencies": {
node_modules//react-dom/package.json-    "react": "^0.14.8"
node_modules//react-dom/package.json-  },
node_modules//react-dom/package.json-  "scripts": {
```

test/common.js
```diff
@@ -1,7 +1,10 @@
-
```
iarna: This file isn't, as far as I can tell, used at all any more. It should really go away, I wouldn't extend it. Just put your function directly into `common-tap` above. All the tests should really be loading `common-tap` anyway, as it does some important environment configuration.

test/packages/npm-test-shrinkwrap/test.js
```diff
@@ -1,4 +1,4 @@
-var path = require("path")
```
iarna: This file is a part of the legacy tests and... as best as I can tell shouldn't actually be in the branch at all? Regardless, the one that's used now is [test/tap/legacy-shrinkwrap.js](https://github.com/npm/npm/blob/lts/test/tap/legacy-shrinkwrap.js)

test/common.js
```diff
@@ -1,7 +1,10 @@
...
-// whatever, it's just tests.
-;["util","assert"].forEach(function (thing) {
-  thing = require("thing")
-  for (var i in thing) global[i] = thing[i]
-}
-
+exports.rmFrom = function rmFrom (obj) {
```
iarna: While I think `rmFrom` is fine when the function is declared near its context (as it was previously), as a globally available tool I'd like the function name to be more descriptive. `rmFromInShrinkwrap` or something like that?

majgis: I will change the name to rmFromInShrinkwrap.

test/common.js
majgis: I will remove test/common.js

test/packages/npm-test-shrinkwrap/test.js
majgis: test/tap/legacy-shrinkwrap.js below had it's behavior changed

I will delete this file after confirming it is dead code.

ericdfields:
Just caught this too in our CI environment. Thanks for addressing it.

joelzimmer:
As someone who is affected by it, any ETA on when this would be merged into npm2?

majgis:
@joelzimmer If you or anyone else could confirm that this change works, without causing adverse effects, it might help it along:

To install npm with this fix:
```
npm i -g git+https://github.com/majgis/npm.git#366c0728a1a8865d265d5a9f06389fe0b5dd8183
```

To restore to the latest official version:
```
npm i -g npm@2
```

joelzimmer:
I was able to confirm that this solves the issue we were experiencing (referenced in #5135.)

robrichard:
I can confirm that this also solves https://github.com/npm/npm/issues/12276

kevinold:
@majgis I can confirm that this solves #12276

jvivs:
Any guidance on when this would make it into an npm LTS release, and then the subsequent node LTS release?

majgis:
@iarna Thank you for your earlier review.  Is there anything else I can do on this PR?

iarna:
LGTM. As for timing, I would expect it to go out in the next `lts` release– maybe next week? @zkat is the owner of that.

baens:
@zkat & @iarna: looks like there was just a release of npm & node but this didn't make the cut. Any idea if this can be in the next minor bump?

iarna:
@baens npm has releases every week, FYI. Looks like I'm responsible for both releases this week, so I'm scheduling this for the lts release. =)

baens:
Awesome @iarna, thank you!

iarna:
And this landed in 2.15.6! =)

joelzimmer:
@iarna thanks so much! When will this be tagged as part of `lts`?

baens:
HI @iarna, any ETA on when `2.15.6` will land on the node `v4.x-staging` branch with this fix in it?

Issue #11692, opened 2016-02-26
I want to translated chinese for npm project，convenient for china developers。I hope join npm github  organization。


CoderBOBO:

othiym23:
The npm CLI team is still coming up with a plan for localizing npm. It will probably involve [intl.js](https://www.npmjs.com/package/intl) and standardization of npm's error messages, and is something that the current core team may begin work on this year, but only towards the end of the year.

We're also working on documenting the way that users can become collaborators on the project. Until that plan is ready, the easiest way to get involved with work on the project is to just go ahead and submit pull requests. If you want to start work on translating npm for Chinese users, the team can help you figure out where to start and support you if you if you run into obstacles. Thanks for your interest in contributing!

PR #12298, opened 2016-04-10
Support i18n in npm

watilde:
Hello :wave: 
I'm trying support `i18n` in npm. Related issue is here: https://github.com/npm/npm/issues/11692
## Translation files

Like [man pages](http://www.win.tue.nl/~aeb/linux/man/), translated files should be placed outside of the core codes. My design is here to provide translation files by npm registry and support version up of npm it self, create a light library that has a concept of "version" in directory architecture for translation files. See the following process as more practical detail:

Install:

```
$ npm install -g npm-pages-ja
$ npm-pages-ja | npm config set locale-file-directory
```

Update:

```
$ npm install -g npm-pages-ja
$ npm install -g npm
```

Version:

```
locale/file/directory/
├── 3.7.0.json
├── 3.7.1.json
...
└── 3.8.7.json
```

Also `i18npm` return the original value if there is no translation data to keep it safer.
That's why I try to create a [y18n](https://github.com/yargs/y18n) based library called [i18npm](https://github.com/watilde/i18npm#i18npm-).
## Intl.js

Currently the `Intl` Object(ECMA-402) @othiym23 mentioned before has only 3 method(`Collator`, `NumberFormat`, `DateTimeFormat`) and it can't handle locale messages.
## gettext or the other libs

[GNU gettext](https://www.gnu.org/software/gettext/), [Jed](https://github.com/SlexAxton/Jed) and [i18next](https://github.com/i18next/i18next) are also good choices depend on what kind of architecture we should have. Let me put `needs-discussion` on here.

Thanks!


iarna:
I've only glanced at this so far, but it looks like a reasonable approach. Tagging @aredridel as I'm sure she'll be interested too.

iarna:
As there hasn't been any movement here in over a year I'm gonna close this, though this shouldn't be taken as a lack of interesting in doing I18N. We may end up borrowing the approach from [npx](https://github.com/zkat/npx).

PR #12329, opened 2016-04-12
test: Fix progress config testing to ignore local user configs

iarna:
Previously, _any_ local setting would cause the tests to fail as
they were trying to test what the default values for the progress
bar would be in different environments and any explicit setting
overrides those defaults.

### DIFF
--- a/test/tap/progress-config.js
+++ b/test/tap/progress-config.js
@@ -1,6 +1,9 @@
 'use strict'
+var path = require('path')
 var test = require('tap').test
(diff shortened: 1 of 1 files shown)


iarna:
This has historically only shown up when node contributors have tried to run the npm tests w/ local configs. For instance, https://github.com/nodejs/node/pull/6153#issuecomment-209116394

iarna:
Merged into v3.8.8!

Issue #12164, opened 2016-04-01
`npm run something` miss setting the `cwd` when running the command in a project sub-folder.


xieranmaya:
For example I have a project which installed a `babel-cli` locally, and added a `script` entry in the `package.json` to run `babel` by `npm run`, then I `cd` to a sub-folder(i.e. `/foo/` folder) of my project and run `npm run babel a.jsx -o a.js`, the script will fail by not found the file `a.jsx`.

I should run the command `npm run foo/a.jsx -o foo/a.js` for the right behavior.

Which means `npm run` sets the `cwd` to the project root path but not the path which running the command, I think that is not an expected behavior, cause git(or most other command) will setting the correct `cwd` when you run git command in sub-folder of a repo:

/foo/ > `git add a.js` will work, not need to run `git add foo/a.js`(and in fact this command will fail)

PR #12356, opened 2016-04-14
Add INIT_CWD environment variable when npm run scripts

MichaelQQ:
As the issue #12164 mentioned, npm run scripts sets the cwd to the project root path. 
When we need to run scripts in subdir, we should use the script like "cd subdir && babel src -d lib"

In some cases, we want to use the same script in different subdir, so we should add different scripts like:
"cd subdir1 && babel src -d lib", "cd subdir2 && babel src -d lib", etc.

Therefore, I add a INIT_CWD environment variable, which is add when npm run script. In the same situation, we can have only one script "cd $INIT_CWD && babel src -d lib" or "babel $INIT_CWD/src -d $INIT_CWD/lib". Type npm run script in the subdir which I want to do it and all subdir can run the same script.

### DIFF
--- a/doc/cli/npm-run-script.md
+++ b/doc/cli/npm-run-script.md
@@ -40,6 +40,20 @@ you should write:
 
 instead of `"scripts": {"test": "node_modules/.bin/tap test/\*.js"}` to run your tests.
 
+Also, `npm run` adds current working directory to `INIT_CWD`. You can run scripts in subdirectory of project. For example, you want to run babel in specific subdirectory. You should write:
+
+    "scripts": {
+      "build": "babel $INIT_CWD/src -d $INIT_CWD/lib"
+    }
+    
+or you can write:
+
+    "scripts": {
+      "build": "cd $INIT_CWD && babel src -d lib"
+    }
+    
+Now, you can run script, `npm run build` in any subdirectory you want to compiler JavaScript.
+
 `npm run` sets the `NODE` environment variable to the `node` executable with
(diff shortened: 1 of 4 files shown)


iarna:
We're not opposed to this on principle. The change should go in `lib/utils/lifecycle.js` however. (See where it sets `env.npm_lifecycle_event`, as that seems like the right place for this sort of thing).

This will also need tests and documentation before it can land.

iarna:
I would stay w/ the original env var (all caps), but that's the right place, thank you. Now we just need a test and a patch to documentation so folks know that this is available. =)

MichaelQQ:
I've added the test and rename it back to INIT_CWD.

skiano:
Is this still being considered? I am not sure I understand why the integration tests failed.

MichaelQQ:
Just rebase to npm:latest

skiano:
Thanks! Any idea if/when it will make it to the main branch?

iarna:
@skiano It still needs tests and docs.

@MichaelQQ You said "I've added the test" but I don't see any tests in your PR? We also need documentation for this feature.

MichaelQQ:
I forgot to push the test :sweat_smile:
I've added the documentation for INIT_CWD. Please check it. Thank you. :bow:

iarna:
Looks great, thank you! I've landed this as 20c46228d in `release-next` with a bit of a rewrite to the documentation. It'll be out in a week or two in our next `npm` release.

Issue #12372, opened 2016-04-15
npm install ignores nested shrinkwrapped dependencies


misterbyrne:
Related issue? https://github.com/npm/npm/issues/6928

During an `npm install`, the shrinkwrap is inflated, but only one level deep. As a result the versions of these nested sub-dependencies are resolved the non-shrinkwrapped way.

This only seems to occur during multi-stage installs (i.e. `rm -rf node_modules && npm install` will result in the correct versions being installed). When it does happen it can result in different versions of sub-dependencies being installed than specified in the shrinkwrap file. This is probably more likely to happen where people rely on multi-stage installs (i.e. restoring a 'cached' `node_modules` directory from a previous build followed by `npm install` - pretty common in hosted CI - e.g. CodeShip / CircleCI).

### Reproduction steps

(there's a similar test [here](https://github.com/npm/npm/issues/12373) that fails on the master branch)

1. `npm init`
2. `npm install repeating@1.0.0 --save`
3. `npm install redent@1.0.0 --save`
4. `npm shrinkwrap` - notice that there's a sub-dependency nested within the shrinkwrap

    ![image](https://cloud.githubusercontent.com/assets/1167425/14575786/fe4b8a8a-035b-11e6-911c-e6b22315d8d8.png)

5. Edit the shrinkwrap file to change the `repeating@2.0.1` sub-dependency to `repeating@2.0.0`, e.g.

    `sed -i.bak "s/2.0.1/2.0.0/g" npm-shrinkwrap.json`

    I know manually editing a shrinkwrap file is kinda crazy, but this is just to get a quick repro in place - in practise this happens frequently whenever the sub-dependency has a patch release version bump [albeit from a lower version number to a higher one].

6. `npm install`

### Expected outcome

The `repeating@2.0.1` sub-dependency is downgraded to `2.0.0`

### Actual outcome

Nothing happens - the 'ideal' tree and the current tree are identical.

Incidentally, this was made apparent because ember cli runs a [dependency checker](https://github.com/quaertym/ember-cli-dependency-checker) before running and flags any version mismatches between what's in `npm-shrinkwrap.json` and `node_modules`.

iarna:
Yeah, this is awkward because we currently allow partial shrinkwraps (which is necessary to install `hapi`), which means that you're allowed to have modules in your tree that aren't in the shrinkwrap as long as the match a `package.json` entry, and in this case it does. =/

[...]

PR #12373, opened 2016-04-15
Inflate nested dependencies within a shrinkwrap

misterbyrne:
Seems to fix https://github.com/npm/npm/issues/12372

### DIFF
--- a/lib/install/inflate-shrinkwrap.js
+++ b/lib/install/inflate-shrinkwrap.js
@@ -32,7 +32,7 @@ var inflateShrinkwrap = module.exports = function (tree, swdeps, finishInflating
                   child.package.version === sw.version)) {
       if (!child.fromShrinkwrap) child.fromShrinkwrap = spec
       tree.children.push(child)
-      return next()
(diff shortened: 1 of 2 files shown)


iarna:
This change looks good to me, but it's going to need a test.

misterbyrne:
Cool! I've already included a test - which fails without my change and passes with the change. I'm not sure whether you mean that you would prefer my test to be extracted into its own file, whether you would prefer a unit test rather than an integration test. Or something else :/

iarna:
So you did! I don't know how I overlooked that!

iarna:
I think I may rewrite it to be a separate test file (as I think it's outside of the scope of the existing file) but I'm happy to do that as part of landing this.

misterbyrne:
Cool - Thanks! - I'm pretty excited about this one! 😄

misterbyrne:
https://github.com/npm/npm/issues/12702 😞

iarna:
This was merged in 3.9.0– I'll see if I can't do another small patch release this week for #12702.

Issue #12155, opened 2016-03-31
Failing to parse package.json must be an error, not warning


dmitriid:
Given 

``` json
{
  'name': 'some-name',
  'dependencies': {}
}
```

Running

```
> npm --version
3.3.12

> npm install --progress=false --loglevel=error
> 

> npm install
npm WARN EJSONPARSE Failed to parse json
npm WARN EJSONPARSE Unexpected token '\'' at 2:3
npm WARN EJSONPARSE   'name': 'some-name',
npm WARN EJSONPARSE   ^
npm WARN EPACKAGEJSON campaign No description
npm WARN EPACKAGEJSON campaign No repository field.
npm WARN EPACKAGEJSON campaign No README data
npm WARN EPACKAGEJSON campaign No license field.
```

Failure to parse `package.json` must be a fatal error, not a warning

lholmquist:
@dmitriid @othiym23 should this be closed? looks like it was fixed here: https://github.com/npm/npm/pull/12406

PR #12406, opened 2016-04-19
install: failing to parse package.json should be error

watilde:
Fixes https://github.com/npm/npm/issues/12155. We can stop install process when the `EJSONPARSE` error happened, because the process never finish if npm can't parse `package.json`.

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -578,6 +578,9 @@ Installer.prototype.readLocalPackageData = function (cb) {
     readPackageTree(self.where, iferr(cb, function (currentTree) {
       self.currentTree = currentTree
       self.currentTree.warnings = []
+      if (currentTree.error && currentTree.error.code === 'EJSONPARSE') {
+        return cb(currentTree.error)
+      }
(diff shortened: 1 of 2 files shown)


iarna:
I'm merging this as 2010d18. I did two things:
  1. I squashed the test and change commits–I like to see patches in the same commit as the patch they support. Having them together makes later searching through the commit log easier because if `git blame` leads me to a commit id, `git show` will show me everything relevant.
  2. I rewrote the test to match our newer house style. We haven't documented this much anywhere, but I encourage you to take a look, the primary changes are using [`tacks`](https://npmjs.com/tacks) to setup and cleanup our test case and throwing if the `err` arg from `common.npm` exists. We do the latter because that is only set if `common.npm` couldn't exec `npm` for some reason, usually an OS problem. It's never an error in the software being tested, but a problem with the machine running the tests.

watilde:
Thanks for letting me know them! The newer house style seems neat. Will try to read some the other tests and use it next time.

iarna:
Merged into v3.8.8!

PR #12415, opened 2016-04-19
Cert key doc

rvedotrc:
https://github.com/npm/npm/issues/7672#issuecomment-83073296

"I'd love it if somebody tweaked the config documentation to make it clearer that cert, key, and ca are for file contents, not paths."

"ca" already seemed to have that explanation.  I've applied that same explanation to "cert" and "key".


zkat:
LGTM!

iarna:
Merged into v3.8.8 & v2.15.4!

Issue #12472, opened 2016-04-25
npm run appends doublequote when last character of arg is backslash


midniteio:
## When running directly from node:

> λ node node_modules\tcli test -f path\to\feature\
> 
> Selected server: "Selenium"
> ───────────────────────────────────────────────────────────────────────────────────────────────────────────────
> 
> > Skipping download. Selenium exists at ./bin
> > Skipping download. Chromedriver exists at ./bin
> > Skipping unzip. Chromedriver already exists in ./bin
> > Selenium started successfully
> > [ 'path\to\feature\' ]
> 
> Cucumber
> ───────────────────────────────────────────────────────────────────────────────────────────────────────────────
> 
> Feature: X

The above continues on successfully and as expected.
## When running from `npm run <cmd>`:

> λ npm run tcli -- test -f  path\to\feature\
> 
> > repo@x.y.z tcli C:\path
> > tcli "test" "-f" "path\to\feature\"
> 
> Selected server: "Selenium"
> ───────────────────────────────────────────────────────────────────────────────────────────────────────────────
> 
> > Skipping download. Selenium exists at ./bin
> > Skipping download. Chromedriver exists at ./bin
> > Skipping unzip. Chromedriver already exists in ./bin
> > Selenium started successfully
> > [ 'path\to\feature**"**' ]
> 
> Cucumber
> ───────────────────────────────────────────────────────────────────────────────────────────────────────────────
> 
> Error: ENOENT: no such file or directory, lstat 'path\to\feature**"**'

PR #12473, opened 2016-04-25
Handle trailing unescaped backslash in arg

midniteio:
This is a small fix in reference to https://github.com/npm/npm/issues/12472, where if an arg is passed that ends in a backslash (such as an auto-completed path), joinArgs will escape the doublequote it appends to each arg, and the value returned is not what it is expected to be.

### DIFF
--- a/lib/run-script.js
+++ b/lib/run-script.js
@@ -168,9 +168,13 @@ function run (pkg, wd, cmd, args, cb) {
 
 // join arguments after '--' and pass them to script,
 // handle special characters such as ', ", ' '.
+// if arg ends in trailing backlash, ensure it is escaped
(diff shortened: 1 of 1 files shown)


iarna:
Thanks for finding this bug and putting together a fix. Could you add a test that verifies this behavior?

iarna:
This hasn't had any activity for over a year so I'm going to close it. The specific issue does remain, but I'm not entirely comfortable with this approach. It's not so much the fault of this patch as it is the fault of the thing that appended arguments on to run scripts — I think that would be better deeper in the library stack.

Our approach in the future will likely be to delegate this to `libnpx`.

PR #12474, opened 2016-04-25
Update README.md

skistaddy:
Added another fancy way to install. :P

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,8 @@ paths, etc.) then read on.
 
 ## Fancy Install (Unix)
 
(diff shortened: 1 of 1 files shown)


iarna:
The README instructions are about how to get npm if you don't have it, so I think installing npm with npm is out of scope for those. Further, doing so is not Unix specific so even if we were to mention that (in the context of "update your npm with…") it shouldn't be in that section. And `npm install npm` will just install it as a dependency of your current module, `npm install -g npm` would update it.

bjskistad:
Ok. :)

PR #12475, opened 2016-04-25
Warn if argument starts with non-ascii dash

iarna:
Options can only start with ascii dashes.  Ordinarily this isn't a problem but many web documentation tools "helpfully" convert `--` into an emdash (—).  When user's copy and paste from this documentation this makes commands mysteriously fail.

We should at least warn if someone is setting off this foot gun (which is what this patch currently does). Possibly we should just refuse to run, as I can't think of a scenario where this would be appropriate behavior.

### DIFF
--- a/lib/npm.js
+++ b/lib/npm.js
@@ -96,6 +96,14 @@
         }
         if (args.length === 1) args.unshift([])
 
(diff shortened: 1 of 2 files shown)


iarna:
This landed in 3.9.0 =)

PR #12476, opened 2016-04-25
fix npm crash when bundled dep lacks package.json

dflupu:
Until a few months ago it was possible to install bundled dependencies that did not contain a package.json. That is no longer the case. Attempting to do so with the current version of npm results in npm crashing without providing adequate information. This should make it fail with a clearer error message.

EDIT: modified the check a bit, since i was bubbling up previously ignored errors

### DIFF
--- a/lib/install/action/extract.js
+++ b/lib/install/action/extract.js
@@ -30,6 +30,11 @@ function andUpdatePackageJson (pkg, buildpath, next) {
 function andStageBundledChildren (pkg, buildpath, log, next) {
   var staging = path.resolve(buildpath, '..')
   return iferr(next, function () {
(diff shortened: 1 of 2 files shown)


iarna:
Could you add a test for this scenario? I can't seem to repro it…

dflupu:
I'll do so when I get the chance. Meanwhile, I've published a package that should help you reproduce it. Just do `npm i dflupu-test-npm-01`.

What I'm getting is along the lines of:
```
npm ERR! Path must be a string. Received undefined
npm ERR! 
```

iarna:
Thank you! That looks great!

iarna:
This landed in 3.9.1!

Issue #12347, opened 2016-04-14
'resolved' not test in 'npm shrinkwrap' (with replication case, node 4.4, npm 3.8.6)


sminnee:
I believe that this a reoccurence of #3581. I have a new case where the 'resolved' field isn't being populated in npm 3.8.6 running on node 4.4.1

Start with this package.json

    {
      "name": "tmp1",
      "version": "1.0.0",
      "description": "test",
      "main": "index.js",
      "scripts": {
        "test": "echo \"Error: no test specified\" && exit 1"
      },
      "author": "",
      "license": "ISC",
     "dependencies": {
        "blueimp-file-upload": "^6.0.3",
        "jquery-sizes": "^0.33.0"
      }
    }

Run the following commands (starting without node_modules or npm-shrinkwrap.json)

     npm install
     npm shrinkwrap

npm-shrinkwrap.json contains a sub-entry for jquery without a resolved flag

    {
      "name": "tmp1",
      "version": "1.0.0",
      "dependencies": {
        "blueimp-canvas-to-blob": {
          "version": "3.3.0",
          "from": "blueimp-canvas-to-blob@>=2.0.3",
          "resolved": "https://registry.npmjs.org/blueimp-canvas-to-blob/-/blueimp-canvas-to-blob-3.3.0.tgz"
        },
        "blueimp-file-upload": {
          "version": "6.9.7",
          "from": "blueimp-file-upload@>=6.0.3 <7.0.0",
          "resolved": "https://registry.npmjs.org/blueimp-file-upload/-/blueimp-file-upload-6.9.7.tgz"
        },
        "blueimp-load-image": {
          "version": "2.6.1",
          "from": "blueimp-load-image@>=1.2.1",
          "resolved": "https://registry.npmjs.org/blueimp-load-image/-/blueimp-load-image-2.6.1.tgz"
        },
        "blueimp-tmpl": {
          "version": "3.4.0",
          "from": "blueimp-tmpl@>=2.1.0",
          "resolved": "https://registry.npmjs.org/blueimp-tmpl/-/blueimp-tmpl-3.4.0.tgz"
        },
        "jquery": {
          "version": "2.2.3",
          "from": "jquery@>=1.6.0",
          "resolved": "https://registry.npmjs.org/jquery/-/jquery-2.2.3.tgz"
        },
        "jquery-sizes": {
          "version": "0.33.0",
          "from": "jquery-sizes@>=0.33.0 <0.34.0",
[...]

Issue #11736, opened 2016-03-01
Git dependencies and npm shrinkwrap don't work together


bouk:
Hi,

It seems npm shrinkwrap and git dependencies are fundamentally incompatible with each other. 

```bash
echo '{"private": true, "dependencies": {"jquery": "jquery/jquery"}}' > package.json
npm install
npm shrinkwrap
rm -rf node_modules
npm install
npm shrinkwrap
```

This will display an error similar to

```
npm ERR! Darwin 15.3.0
npm ERR! argv "/Users/bouke/.nvm/versions/node/v5.5.0/bin/node" "/Users/bouke/.nvm/versions/node/v5.5.0/bin/npm" "shrinkwrap"
npm ERR! node v5.5.0
npm ERR! npm  v3.3.12

npm ERR! Problems were encountered
npm ERR! Please correct and try again.
npm ERR! invalid: have jquery@3.0.0-pre (expected: github:jquery/jquery) /tmp/lolnpm/node_modules/jquery
npm ERR! extraneous: jquery@3.0.0-pre /tmp/lolnpm/node_modules/jquery
npm ERR!
npm ERR! If you need help, you may report this error at:
npm ERR!     <https://github.com/npm/npm/issues>

npm ERR! Please include the following file with any support request:
npm ERR!     /tmp/lolnpm/npm-debug.log
```

Installing from a shrinkwrap should not cause shrinkwrapping after to fail

othiym23:
This appears to be a bug in how the installer deals with hosted git dependencies once they've been cached (you can see this just by reinstalling a hosted git package once it's been added to your cache and running `npm ls`, which will give you an error similar to the one above). Added to `big-bug` because it produces inconsistent installs and blocks use of shrinkwrap. Thanks for the report!

XYliang:
Try using an administrator mode

kazkovsky:
+1

I have my github repository in shrinkwrap, but it doesn't update on any machine.

`cat npm-shrinkwrap.json | grep <my repo>`

```
 "<my repo>": {
      "from": "git+https://<token>:x-oauth-basic@github.com/<my repo>.git",
      "resolved": "git+https://<token>:x-oauth-basic@github.com/<my repo>#9a68c540902e6497855699b990e662e5ab73d3ba"
```

after running `npm install` it keeps old version:

`npm list`
```
│ └── y18n@3.2.0
[...]

Issue #11493, opened 2016-02-10
`npm ls` fails on shrink-wrapped packages that have nested subdependencies whose versions are set to "" (empty string).


lxe:
Tough one to explain but easy to reproduce:
#### Description

Some packages are published with some of their dependency versions set to `""` (empty string). For example `follow-redirects@0.0.3`, who sets 'underscore' dependency to `""`:

```
$ npm info follow-redirects@0.0.3 dependencies
{ underscore: '' }
```

If we shrinkwrap a package whose dependency chain contain such modules, and then run npm install on it (after removing / with no previously existing `node_modules` tree), `npm ls` command will fail with `invalid`.
#### Procedure

Let's create a shrink-wrap with npm v2 (deep dependencies) or just force deep tree with another version of underscore at the top of our tree:

**package.json**

```
{
  "name": "shrinkwrap-fail",
  "version": "1.0.0",
  "dependencies": {
    "follow-redirects": "0.0.3",
    "underscore": "1.0.3"
  }
}
```

Let's run `npm install`, followed by `npm ls`:

```
$ npm install
...
shrinkwrap-fail@1.0.0 /Users/aleksey/shrinkwrap-fail
├─┬ follow-redirects@0.0.3
│ └── underscore@1.8.3
└── underscore@1.0.3

$ npm ls
shrinkwrap-fail@1.0.0 /Users/aleksey/shrinkwrap-fail
├─┬ follow-redirects@0.0.3
│ └── underscore@1.8.3
└── underscore@1.0.3
```

Everything looks good!

For further debugging purpose, let's examine the generated `package.json` of `underscore@1.8.3` :

```
$ head -n10 node_modules/follow-redirects/node_modules/underscore/package.json
{
  "_args": [
    [
      "underscore@",
      "/Users/aleksey/shrinkwrap-fail/node_modules/follow-redirects"
    ]
  ],
  "_from": "underscore@latest",
  "_id": "underscore@1.8.3",
  "_inCache": true,
```

Let's now create a `npm-shrinkwrap.json`:

```
$ npm shrinkwrap
wrote npm-shrinkwrap.json
```

Now, blow away `node_modules`, and run `npm install` to regenerate the module tree from the shrinkwrap:

```
$ rm -rf node_modules
$ npm install
....
├─┬ follow-redirects@0.0.3
│ └── underscore@1.8.3
└── underscore@1.0.3
```

It succeeds, however, `npm ls` shows `underscore@1.8.3 invalid`:

```
[...]

Issue #11373, opened 2016-02-02
npm shrinkwrap is throwing error invalid error on dependencies that have "" version


zhuojloh:
attempt to npm shrinkwrap again after npm install from a dependency that have "" as version will produce error below. 

```
npm ERR! invalid: have async@1.5.2 (expected: ) ...
npm ERR! extraneous: async@1.5.2 ...
```

main dependencies

```
  "dependencies": {
    "async": "1.5.2",
    "gulp-s3": "^0.3.0",
  },
```

gulp-s3 dependencies

```
  "dependencies": {
    "async": "",
    "event-stream": "*",
    "gulp-util": "~2.2.6",
    "knox": "",
    "mime": "~1.2.11"
  },
```

holm:
Seems the same as https://github.com/npm/npm/issues/11493

sminnee:
Here's a replication case:

Test `package.json`

    {
      "name": "tmp3",
      "version": "1.0.0",
      "description": "",
      "main": "index.js",
      "scripts": {
        "test": "echo \"Error: no test specified\" && exit 1"
      },
      "author": "",
      "license": "ISC",
      "dependencies": {
        "async": "1.5.2",
        "gulp-s3": "^0.3.0"
      }
    }

Test script (`../../npm` is the folder where npm is checked out)

    #!/bin/bash
    rm -rf node_modules npm-shrinkwrap.json
    node ../../npm/bin/npm-cli.js install
    node ../../npm/bin/npm-cli.js shrinkwrap
    rm -rf node_modules
    node ../../npm/bin/npm-cli.js install
    node ../../npm/bin/npm-cli.js shrinkwrap

PR #12518, opened 2016-04-29
More _from metadata fixes for shrinkwrap reliability (#11736 and #12347)

sminnee:
One of these @iarna has already mentioned in comments in #12347.

These fixes both related to improving `npm shrinkwrap` reliability by getting more a consistent definition of `_from` values.

It still needs tests, which I'll try and get to over the next few days, but I wanted to expose my work-to-date.

### DIFF
--- a/lib/cache/add-remote-git.js
+++ b/lib/cache/add-remote-git.js
@@ -39,6 +39,14 @@ function addRemoteGit (uri, _cb) {
   assert(typeof _cb === 'function', 'must have callback')
   var cb = dezalgo(_cb)
 
+  // Set the data._from value to the package name & uri used to specify the dependency
+  // This ensures that shrinkwraps work reliably
+  function finish(er, data) {
+    if (er) return cb(er)
+    data._from = data.name + '@' + uri
+    cb(er, data)
+  }
+
   log.verbose('addRemoteGit', 'caching', uri)
 
   // the URL comes in exactly as it was passed on the command line, or as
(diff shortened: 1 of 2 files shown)


sminnee:
Note that #11373 / #11493 is a related issue still not resolved by this.

lib/cache/add-remote-git.js
```diff
@@ -375,14 +383,6 @@ function updateSubmodules (from, resolvedURL, tmpdir, cb) {
         // https://github.com/npm/npm/issues/6400
         addLocal(spec, null, function (er, data) {
           if (data) {
-            if (npm.config.get('save-exact')) {
-              log.verbose('addRemoteGit', 'data._from:', resolvedURL, '(save-exact)')
```
iarna: I think even with your patch we need this behavior if `--save-exact` was used, as it saves with a commit id.

iarna:
Status note: This is waiting on an update that restores `--save-exact` behavior.

xdissent:
Anything else I can do to help with this one?

zkat (changes requested):
Going to make a note also that this code literally does not exist anymore.

I'm not sure off the top of my head what the corresponding patch would be, though, and whether pacote would need to be patched -- pacote shouldn't have this sort of awareness about saving, imo.

iarna:
As the code this PR is patching is no longer in use (and indeed, I don't believe this is still an issue) I'm going to close this.

The original issue here, IIRC, is that subsequent installs from a shrinkwrap would end up with a `_from` pointing at the tarball URL and then that would go into the `from` field in future shrinkwraps, causing churn there.

Values for the shrinkwraps are no longer produced so naively, but more importantly, the `from` field is no longer included in shrinkwraps.

PR #12550, opened 2016-05-03
test: disable coverage in recursively-called tap

othiym23:
Testing a candidate fix of @izs's.

### DIFF
--- a/test/tap/registry.js
+++ b/test/tap/registry.js
@@ -49,7 +49,7 @@ function runTests () {
         env: env,
         stdio: 'inherit'
       }
-      common.npm(['test', '--', '-Rtap'], opts, function (err, code) {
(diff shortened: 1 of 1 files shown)


isaacs:
In my local testing, doing `--no-coverage` was still causing some weirdness.  Removing the `COVERALLS_REPO_TOKEN` from the child proc env did the trick, tho. https://github.com/npm/npm/pull/12150/commits/0bc4572c83688c4d8f45a8454ddefb8f9f0ff5d6

iarna:
Looks like this is working (the build failures are the out-of-memory errors in lcov-parse on 0.12 and 0.10).

iarna:
This landed in 3.9.1. We're going to want to port this back to lts, so I'm going to flag it as such.

PR #12585, opened 2016-05-04
doc: engineStrict is quite gone

othiym23:
Not "deprecated" so much as "extirpated". We should probably have better delineations about how to talk about "deprecated" features (which are still there, potentially warning people when they're used), and "removed" features (where the supporting code for the feature is no longer in the code base). Getting that in the developer handbook is something I'll try to do at some point.

**r**: @iarna


iarna:
LGTM

iarna:
This landed in 3.9.1! =)

PR #12586, opened 2016-05-05
[doc] Fix defining when node-gyp rebuild is called

reconbot:
It is called during install instead of preinstall stage. This now matches https://docs.npmjs.com/misc/scripts#default-values


iarna:
Huh, I didn't realize that this didn't default if there was a `preinstall`– I'd though it was `install` only. =D Thank you for this!

iarna:
This landed in 3.9.1 & 2.15.6! =)

PR #12636, opened 2016-05-09
[doc] Fix the def of the default `install` script

reconbot:
This is a followup from #12586 and it’s behavior is defined in the following code https://github.com/npm/read-package-json/blob/master/read-json.js#L146-L164

### DIFF
--- a/doc/misc/npm-scripts.md
+++ b/doc/misc/npm-scripts.md
@@ -71,7 +71,7 @@ npm will default some script values based on package contents.
 
 * `"install": "node-gyp rebuild"`:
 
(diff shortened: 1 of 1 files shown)


iarna:
Thank you, this is an excellent clarification.

iarna:
This landed in 3.9.1 & 2.15.6! =)

drifkin:
When does this make it to the docs website? I'm assuming this is the source of <https://docs.npmjs.com/misc/scripts>. A teammate ran into the behavior that this PR clarifies.

Issue #9436, opened 2015-08-28
npm ls not finding module


isaacs:
This is weird:

```
$ npm i ../deep-equal
tap@1.3.2 /Users/isaacs/dev/js/tap
└── deep-equal@1.0.0-patched-by-isaacs

$ npm ls deep-equal
tap@1.3.2 /Users/isaacs/dev/js/tap
└── (empty)

npm ERR! code 1

$ npm -v
3.3.1
```

Linked install at 2724e30b71ac79f5c6e1aa821474c70ab4f0a222.  Module is a bundled dep with a prerelease version to indicate that I messed with it.  When I change the version to `1.0.99999`, it works fine.

othiym23:
My guess is that it's something to do with the installed version no longer matching the semver range for the package in `package.json` (because I think I've seen some somewhat related behavior), but @iarna will have to take a look at this one to know for sure.

alunny:
I've been running into this too, not sure if @iarna or someone else at npm inc is looking into it.

@othiym23 is right about the root cause - when you don't specify a version range for the npm-ls command, it sets a default version range of '*' - source https://github.com/npm/npm/blob/master/lib/ls.js#L282

```
> semver = require('semver')
> semver.satisfies('1.0.0-patched-by-isaacs', '*', true)
false
```

The fix would be either (or both)
* modify the node-semver module to match any string for '*' (seems intuitive)
* change npm-ls to not call semver if there's no version number/range - something like

```diff
- found = semver.satisfies(dep.version, args[i][1], true)
+ found = args[i][1] === '*' || semver.satisfies(dep.version, args[i][1], true)
```diff

PR #12685, opened 2016-05-13
Fix npm ls for filters on prerelease packages

zkat:
treat a plain package name in ls as a versionless match. This fixes https://github.com/npm/npm/issues/9436


othiym23:
Love the tests, and the change looks sound. Could you add a few comments explaining what's going on in that chunk of the code so that in the future we know which pieces are being used for that ls, and why? I read #9436 and the tests, and it still took me a little while to understand the purpose of this patch. (Also, a little context in the commit message would be helpful.) Aside from that, LGTM.

zkat:
(for record-keeping, @othiym23 gave this a 🐑 on slack)

zkat:
@evanlucas I just didn't rebase and repush this branch properly -- `fstream@1.0.9` isn't actually included here. That's just cruft from when I was merging it (oops, sorry!)

The `npm ls`-specific patches for this were included in `npm@3.9.3`

Issue #12696, opened 2016-05-13
Add `sign-git-commit` config for GPG-signed commits during `npm version`


tribou:
Currently, `npm@3.8.9` supports a `sign-git-tag` config which enables the `git tag -s` command instead of `git tag` in order to perform GPG-signed tags.  However, the git commit performed prior to the tagging and after the package version is bumped does not support signing.  This results in a signed tag but an unsigned git commit.

For consistency, it would be nice to have an additional `git-sign-commit` boolean npm config flag which enabled `git commit -S` when creating the commit to save the new package version.

This would be a non-breaking feature addition since the default behavior could fallback to the current `git commit` when the `sign-git-commit=true` config is not present.

The change could me made [similar to this](https://github.com/npm/npm/blob/2a5977e0c65b244e92d848fcd56f2f80ba8cdf3b/lib/version.js#L261).  However, if I'm not mistaken, the `-S` flag in the `git commit` command may need to be spaced separately from the `-m` instead of being combined like the the `-sm` in the `git tag` command.

maxkoryukov:
Probably, **must be reopened**

@tribou

tribou:
Yes, it'd be great if this could be re-considered.

The PR would probably need to be re-written since it's been over a year...

tribou:
For anyone following this thread, I forgot to mention this feature was re-opened here: #17429.

PR #12697, opened 2016-05-13
lib: Add sign-git-commit config

tribou:
Adds `sign-git-commit` config referenced in #12696


tribou:
...looks like I'll have to keep `commitFlag` from being passed to `git.chainableExec()` if `git-sign-commit=false`

zkat (approved):
This LGTM! It seems reasonable enough to include, so we'll have it as part of the next release. Thanks!

tribou:
@zkat Just a heads up! I originally wrote this PR against `npm@^3` and haven't tested it at all in v5 😟 

I'm afraid it might need to be rewritten 😢

zkat:
@tribou I think you'll find the conflict to be very small! It would be great if you just resolved the conflict and pushed out a new, rebased version -- or I can do it myself when it's time to merge (although that'd be more work for me lolsob). Your patch is straightforward enough :)

tribou:
Ah, I didn't realize that! I'll try to look into it soon

zkat:
I'm doing my merges today, and it looks like this isn't ready yet, so I'm pushing it out until next time we do a PR sweep.

Issue #10379, opened 2015-11-13
preinstall execution order in npm@3.x


brianc:
Sorry if this has been brought up before, but it looks like the preinstall step for my local repo now runs _after_ dependent modules have been installed with `npm@3.x`?

With `npm@2.x` we used the preinstall hook to delete our `node_modules` folder & clone it from a git repo...so then the `install` step would execute, see that everything was already installed, and do nothing.  With `npm@3.x` we have the same `preinstall` hook, but first npm does an install into the `node_modules` folder itself _then_ it runs the `preinstall` hook.  Is this a bug?  It seems like its not really a "pre-install" step anymore if it actually fires _after_ the install...

thepont:
I'm also seeing this issue. I'm attempting to uninstall a local module, so a fresh version of the module is installed on each install.

The result is that the module is not installed at all after running an `npm install` 

I'm running npm@3.3.6

gbisset:
I've also just come across this issue on npm v3.4.0.

We were also deleting the node_modules folder in the preinstall step.  As it now runs after the packages have been installed, they are being deleted immediately after install which was causing our postinstall to fail as it depends on the installed packages

awestroke:
I am affected by this too. I am very, very curious as to why this change was made. It seems completely nonsensical and unnecessary.

ErisDS:
Also facing this issue. We're using a preinstall hook (following advice of @othiym23) to check for a compatible node version before install. Otherwise, the install fails on incompatible dependencies, and our users end up reporting issues on the dependency repos rather than realising they have an unsupported version.

alexflorisca:
Same issue. We are trying to run a script before `npm install` runs (which sets up an auth token) in order to install a package from a private repository. This is screwing our heroku deployment!

wookiehangover:
I'm also running into some pain with this bug... `preinstall` is a really useful hook and this makes it way less useful!

iarna:
I would like to return this to npm@2 style behavior, running `npm install` much earlier.

Running this before we figure out what we would do (that is, before we produce the idealized tree) is pretty easy and supports most of the use cases. Slightly more awkward is supporting the "preinstall mutates the tree in ways install should account for" use case, but it should still be doable.

ErisDS:
@iarna Did you mean running `preinstall` earlier?

holm:
[...]

PR #12705, opened 2016-05-14
Move preinstall to same section as install/postinstall

victorb:
Currently, due to #10379, the preinstall hook is run after the module is installed. This commit fixes the documentation to reflect this.


IanVS:
It might also be worth indicating that this is a bug and linking to the github issue.

VictorBjelkholm:
Seems to have been fixed so I'm gonna close this.

PR #12719, opened 2016-05-16
Handle 'shasum_unavailable' flag in registry dist metadata

TimeToogo:
This bypasses the `package lacks shasum` error if the registry has explicitly set the `shasum_unavailable` to true.
This allows registries to override the need to compute sha1 checksums for each package and version.
Which in turn means the registry does not need to host the dist tarball files for every package in the registry.

### DIFF
--- a/lib/cache/add-named.js
+++ b/lib/cache/add-named.js
@@ -207,9 +207,9 @@ function addNameVersion (name, v, data, cb) {
         }
         tb = url.format(tb)
 
-        // Only add non-shasum'ed packages if --forced. Only ancient things
-        // would lack this for good reasons nowadays.
-        if (!dist.shasum && !npm.config.get('force')) {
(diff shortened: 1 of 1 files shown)


TimeToogo:
I would love to get this included in the next release if possible. It is currently a blocking issue with implementing NPM support for https://package.studio/ as having to provide a checksum in the dist metadata itself is incompatible with generating the tarballs on the fly from the GitHub/BitBucket/GitLab api. 

Perhaps it could be an option to provide the checksum flag in a response header from the tarball download request?

zkat:
Checking shasums is a pretty core part of the contract between npm and the registry. This sort of change isn't something we will accept. Thank you, genuinely, for the time you put into getting this PR together, and for the follow-up, though the CLI team agreed to close this particular issue.

PR #12731, opened 2016-05-17
fix npm 3.9.1 crash caused by process.nextTick() recursion on Node.js 0.10

JensEggers:
## The crash

I get lots of these:

```
(node) warning: Recursive process.nextTick detected. This will break in the next version of node. Please use setImmediate for recursive deferral.
```

and then the process dies saying this:

```
node.js:0
// Copyright Joyent, Inc. and other Node contributors.
^
RangeError: Maximum call stack size exceeded
```
## The failing test

I have a project with a large number of dependencies, some private. Older versions of npm are fine, but using npm 3.9.0 or npm 3.9.1 on Node 0.10 the first 'npm install' works, and subsequent runs crash, presumably while npm is building up the installed dependency tree.
## The fix

Replacing process.nextTick() with setImmediate() fixes this. My test case has lots of private packages and is hard to share in public. I'm hoping the fix is clear enough to use as is.

### DIFF
--- a/lib/install/deps.js
+++ b/lib/install/deps.js
@@ -86,7 +86,7 @@ function childDependencySpecifier (tree, name, spec, cb) {
   if (!tree.resolved) tree.resolved = {}
   if (!tree.resolved[name]) tree.resolved[name] = {}
   if (tree.resolved[name][spec]) {
-    return process.nextTick(function () {
+    return setImmediate(function () {
       cb(null, tree.resolved[name][spec])
     })
   }
--- a/lib/install/deps.js
+++ b/lib/install/deps.js
@@ -31,6 +31,9 @@ var moduleName = require('../utils/module-name.js')
 // The export functions in this module mutate a dependency tree, adding
 // items to them.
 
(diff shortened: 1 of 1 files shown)


JensEggers:
Travis Tests are failing on Node v0.12 and I'm at a loss as to what to do about it. It appears as thought the test container runs out of memory after all the tests are complete:
```
total ........................................... 4137/4137
  4137 passing (31m)
  ok
FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - process out of memory
Aborted (core dumped)
```

Can anyone help?

zkat:
#12757 is a duplicate of this, I think, but uses `dezalgo` in a different place, which you could argue is more idiomatic. Either way, we could use some tests (and neither PR has tests, so idk if you or @lxe want to see your PRs through. So whoever's more comfortable finishing things up.

JensEggers:
I'd support going with #12757 because it's more idiomatic, and especially so now that it has a test :-)

Closing this one.

Issue #12567, opened 2016-05-03
if "UNMET PEER DEPENDENCIES" is a warning, it should be yellow and lowercase


bendavis78:
The "UNMET PEER DEPENDENCIES" output is red and all-caps, making it seem like an error, but the corresponding message below says it's just a warning. This is confusing.

![image](https://cloud.githubusercontent.com/assets/324392/15001119/8e2d9812-115a-11e6-882b-68bb72dcfffe.png)

Rahulraj9736:
Installation configuration..!

PR #12739, opened 2016-05-18
12567 - Change warning color to yellow

lbearl:
Resolves issue #12567 .

### DIFF
--- a/lib/ls.js
+++ b/lib/ls.js
@@ -361,7 +361,7 @@ function makeArchy_ (data, long, dir, depth, parent, d) {
 
   if (data.peerMissing) {
     var peerMissing = 'UNMET PEER DEPENDENCY'
-    if (npm.color) peerMissing = color.bgBlack(color.red(peerMissing))
+    if (npm.color) peerMissing = color.bgBlack(color.yellow(peerMissing))
     out.label = peerMissing + ' ' + out.label
   }
 


lbearl:
As far as justifying this change goes: as a user I would expect that an error would cause early termination of the program (as proceeding forward with one or more errors could lead to unexpected behavior). The red text indicates an error, but the code doesn't currently terminate the application, instead execution is allowed to proceed. 

While it would be fairly direct to change the behavior of the application to terminate upon experiencing this issue, that seems to have a large potential for side effects for end-users.

iarna:
@lbearl `red` means an error in the package you're running `npm ls` on. If the peer dependency is a legit dependency then that package won't work without it, so `red` seems appropriate to me. 

To me `yellow` would imply _caution_, that is, a warning that you can, if you want, ignore. Peer dependencies aren't like that. They're necessary, it's just that `npm` won't install them for you.

zkat:
I agree with Rebecca on this: this _is_ an error, even if it isn't one that npm will automatically fix for you. This absolutely does require user interaction to fix, and _will_ break unless you get lucky.

Thanks so much for taking the time to look into this and filing a patch for it. UX stuff like this does matter to us a lot, so getting PRs like this is totally a delight, even if we decided not to take this one in particular.

Have a good one! And thanks again!

Issue #12718, opened 2016-05-16
When shrinkwrap is used npm install won't update git dependencies


kossnocorp:
See the problem:

![](http://i.ncrp.co/2F1u040j3t2B/Fullscreen_16_05_16_15_46.png)

To reproduce locally, run (see https://github.com/kossnocorp/npm-shrinkwrap-bug):

```bash
git clone git@github.com:kossnocorp/npm-shrinkwrap-bug.git
cd npm-shrinkwrap-bug
npm install
```

philberg:
If you run shrinkwrap on this, it will also fail.

I am also facing this issue for the following reason.

My top level package "A" has a dep "B" which is installed via git.  Dep "B" has a dep on the chai package (v3.4.*) and package "A" has a devDep on chai (v3.3.*).  This is breaking shrinkwrap

`npm ERR! missing: chai@3.4.*, required by package B`

iarna:
For reference:
```
$ npm ls
npm-shrinkwrap-bug@ /Users/rebecca/code/npmtest/12718/npm-shrinkwrap-bug
└── left-pad@1.1.0 invalid (git://github.com/stevemao/left-pad.git#37723f75f3a2cf99ce0b64bd14de6c2818cfb506)

npm ERR! invalid: left-pad@1.1.0 /Users/rebecca/code/npmtest/12718/npm-shrinkwrap-bug/node_modules/left-pad
```
Obviously, it should be seeing that the installed version doesn't match and try to install from the shrinkwrap (which would then fail, because that committish doesn't exist).

kossnocorp:
I've successfully wrote a fix:

```diff
diff --git a/lib/install/inflate-shrinkwrap.js b/lib/install/inflate-shrinkwrap.js
index f5f53f1..dfc12fc 100644
--- a/lib/install/inflate-shrinkwrap.js
+++ b/lib/install/inflate-shrinkwrap.js
@@ -29,7 +29,7 @@ var inflateShrinkwrap = module.exports = function (tree, swdeps, finishInflating
     if (child && (child.fromShrinkwrap ||
                   (sw.resolved && child.package._resolved === sw.resolved) ||
                   (sw.from && url.parse(sw.from).protocol && child.package._from === sw.from) ||
-                  child.package.version === sw.version)) {
+                  (!child.package.gitHead && child.package.version === sw.version))) {
       if (!child.fromShrinkwrap) child.fromShrinkwrap = spec
       tree.children.push(child)
[...]

PR #12770, opened 2016-05-19
Fix missing update action for git deps when shrinkwrap is used

kossnocorp:
Prevent inflateShrinkwrap function from relying on a package version for git dependencies. That was the cause of two problems:
1. When node_modules dir contains an outdated git dependency, it won't be updated during npm install.
2. npm shrinkwrap command fails.

For more details about the bug, see https://github.com/npm/npm/issues/12718.

_I still working on tests, will appreciate some guidance on that._

### DIFF
--- a/lib/install/inflate-shrinkwrap.js
+++ b/lib/install/inflate-shrinkwrap.js
@@ -26,10 +26,7 @@ var inflateShrinkwrap = module.exports = function (tree, swdeps, finishInflating
              ? name + '@' + sw.from
              : name + '@' + sw.version
     var child = onDisk[name]
-    if (child && (child.fromShrinkwrap ||
-                  (sw.resolved && child.package._resolved === sw.resolved) ||
-                  (sw.from && url.parse(sw.from).protocol && child.package._from === sw.from) ||
-                  child.package.version === sw.version)) {
+    if (childIsEquivalent(child, sw)) {
       if (!child.fromShrinkwrap) child.fromShrinkwrap = spec
       tree.children.push(child)
       return inflateShrinkwrap(child, sw.dependencies || {}, next)
@@ -56,4 +53,12 @@ var inflateShrinkwrap = module.exports = function (tree, swdeps, finishInflating
       }))
     }))
   }, finishInflating)
+}
+
+function childIsEquivalent (child, sw) {
+  if (!child) return false
+  if (child.fromShrinkwrap) return true
+  if (sw.resolved) return child.package._resolved === sw.resolved
+  if (sw.from && url.parse(sw.from).protocol) return child.package._from === sw.from
+  return child.package.version === sw.version
 }
--- a/test/tap/install-shrinkwrapped-git.js
+++ b/test/tap/install-shrinkwrapped-git.js
@@ -1,0 +1,159 @@
+var fs = require('fs')
+var path = require('path')
+var resolve = path.resolve
+var osenv = require('osenv')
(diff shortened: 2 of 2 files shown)


kossnocorp:
@iarna @othiym23 hey! Is there any chance to get this reviewed and merged?

iarna:
@kossnocorp We evaluate PRs for inclusion sometime between Mon and Wed for inclusion the following Thursday.

lib/install/inflate-shrinkwrap.js
```diff
@@ -29,7 +29,7 @@ var inflateShrinkwrap = module.exports = function (tree, swdeps, finishInflating
     if (child && (child.fromShrinkwrap ||
                   (sw.resolved && child.package._resolved === sw.resolved) ||
                   (sw.from && url.parse(sw.from).protocol && child.package._from === sw.from) ||
-                  child.package.version === sw.version)) {
+                  (!child.package.gitHead && child.package.version === sw.version))) {
```
iarna: I'd like a more general correction here. This code is saying that a module on disk matches the shrinkwrap if ANY of these are true, and I'm pretty sure that's not actually what we want to do.

I don't know how I got my head screwed on wrong when I wrote that, but I'm ​_pretty sure_​ I intended those to be exclusive, eg, if there's a _resolved then rely on that for identity, else if there's a from with a protocol, use that for identity, and if nothing else is available use version for identity.

So I _think_ this should actually read:
```js
    if (child) {
      var useExisting
      if (child.fromShrinkwrap) {
        useExisting = true
      } else if (sw.resolved) {
        useExisting = child.package._resolved === sw.resolved
      } else if (sw.from && url.parse(sw.from).protocol) {
        useExisting = child.package._from === sw.from
      } else if (child.package.version === sw.version) {
        useExisting = true
      }
      if (useExisting) {
        if (!child.fromShrinkwrap) child.fromShrinkwrap = spec
        tree.children.push(child)
        return inflateShrinkwrap(child, sw.dependencies || {}, next)
      }
    }
```

iarna:
Regarding the test, we have been working hard to eliminate any dependencies that talk to the network, so I would much appreciate it if you could rework this to do its work locally.

You can see an example of creating a little git repo and firing up a local git server in [`add-remote-git.js`](https://github.com/npm/npm/tree/master/test/tap/add-remote.git.js), though I'd prefer new tests to write (and cleanup) their files with [`tacks`](npmjs.com/package/tacks) (see [`gently-rm-linked-module.js`](https://github.com/npm/npm/tree/master/test/tap/gently-rm-linked-module.js) as an example) instead of doing it by hand.

kossnocorp:
@iarna thank you, working on that 👍

kossnocorp:
@iarna sorry for the delay, here is the change log:

1. Rebase the branch onto master
2. Apply your solution with refactored `childIsEquivalent` function (kudos for flawless GitHub-comment-programming!)
3. Make tests work offline.

I've tried Tacks approach, but frankly code got even messier because I still have to manually handle a lot of operations in order to reproduce the bug.

kossnocorp:
😉

kossnocorp:
❤️

conf:
@iarna Any update on this? This is a blocker for using `npm-shrinkwrap` feature for us since we depend on some package forks on Github and it's a pain to update for everybody in the team.

kossnocorp:
😿

iarna:

kossnocorp:
🙏

PR #12772, opened 2016-05-19
Implement handling for 'shasum_in_headers' metadata flag.

TimeToogo:
This is a follow-up regarding my previous [pull request](https://github.com/npm/npm/pull/12719). Given your feedback I restructured my approach to not remove but only defer the checksum generation to the actual tarball download HTTP response. This approach allows both the ability to generate tarballs on-the-fly while still providing the safety of the checksum validation via a custom  `X-Shasum` header in the tarball download response. This will only be activated if the `shasum_in_headers` flag in the package metadata is set to true.

Would this approach be any more feasible?

### DIFF
--- a/lib/cache/add-named.js
+++ b/lib/cache/add-named.js
@@ -207,9 +207,9 @@ function addNameVersion (name, v, data, cb) {
         }
         tb = url.format(tb)
 
-        // Only add non-shasum'ed packages if --forced. Only ancient things
-        // would lack this for good reasons nowadays.
-        if (!dist.shasum && !npm.config.get('force')) {
+        // Only add non-shasum'ed packages if --forced or 'shasum_in_headers' is true.
+        // Only ancient things would lack this for good reasons nowadays.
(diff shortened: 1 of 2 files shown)


TimeToogo:
Sorry but is there any news on this? What are the issues that prevent it from being merged?
Like I said, this is blocking issue for me, so I am happy to work on this pull request to meet further requirements.

iarna:
We've learned a bunch on this topic since your previous PR. In particular, there's actually a W3C standard for this sort of thing, specifically the [subresource integrity](https://w3c.github.io/webappsec/specs/subresourceintegrity/) standard. We have that implemented over in [ssri](https://www.npmjs.com/package/ssri) and in fact require its use if you want to use sha2-512.

We don't currently use this header, but if you'd like to add that the place for it is now in [pacote](https://github.com/zkat/pacote).

I'm going to close this PR because it's against code that doesn't exist any more, but as I say, we'd welcome the equivalent PR made against pacote.

Issue #10343, opened 2015-11-11
`npm install` after `npm link` will "steal" dependencies from linked packages


jhenninger:
Create two packages:

``` sh
jo@t420:/tmp/npmtest$ mkdir A B
```

A depends on lodash:

``` sh
jo@t420:/tmp/npmtest$ cd A
jo@t420:/tmp/npmtest/A$ npm init --yes && npm install --save lodash
Wrote to /tmp/npmtest/A/package.json:

{
  "name": "A",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "keywords": [],
  "author": "",
  "license": "ISC"
}


A@1.0.0 /tmp/npmtest/A
`-- lodash@3.10.1 

npm WARN EPACKAGEJSON A@1.0.0 No description
npm WARN EPACKAGEJSON A@1.0.0 No repository field.
```

B depends on A which is symlinked for development:

``` sh
jo@t420:/tmp/npmtest/A$ cd ../B
jo@t420:/tmp/npmtest/B$ npm init --yes                       
Wrote to /tmp/npmtest/B/package.json:

{
  "name": "B",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "keywords": [],
  "author": "",
  "license": "ISC"
}
```

Create the link:

``` sh
jo@t420:/tmp/npmtest/B$ npm link ../A
npm WARN EPACKAGEJSON A@1.0.0 No description
npm WARN EPACKAGEJSON A@1.0.0 No repository field.
/home/jo/code/nvm/versions/node/v4.2.1/lib/node_modules/A -> /tmp/npmtest/A
/tmp/npmtest/B/node_modules/A -> /home/jo/code/nvm/versions/node/v4.2.1/lib/node_modules/A -> /tmp/npmtest/A
```

So far so good, A still has lodash in its node_modules:

``` sh
jo@t420:/tmp/npmtest/B$ tree -d ..
..
├── A
│   └── node_modules
│       └── lodash
│           ├── array
│           ├── chain
│           ├── collection
│           ├── date
│           ├── function
│           ├── internal
│           ├── lang
│           ├── math
│           ├── number
│           ├── object
│           ├── string
│           └── utility
└── B
    └── node_modules
[...]

PR #12775, opened 2016-05-20
Fix bug involving chained symlinks where modules would be pruned from the tree that shouldn't be

### LINKED ISSUE #10800: npm link breaks npm install under some conditions on 3.X
If you issue `npm install` in a module with multiple `npm link`d dependencies strange things can happen.

After a lot of bisecting, this is the smallest set of modules that I have managed to reproduce the install failures this with:

```sh
git clone https://github.com/clux/smell
git clone https://github.com/clux/ffa
git clone https://github.com/clux/tiebreaker
cd smell
npm install
npm link
cd ../ffa
npm link smell
npm install
npm link
cd ../tiebreaker
npm link smell ffa
npm install
```

The last instruction will fail with

```
npm ERR! Linux 3.13.0-24-generic
npm ERR! argv "/home/clux/local/node-v5.2.0/bin/node" "/home/clux/local/node/bin/npm" "install"
npm ERR! node v5.2.0
npm ERR! npm  v3.3.12
npm ERR! path /home/clux/testrepos/tiebreaker/node_modules/escodegen/node_modules/esprima/bin/esparse.js
npm ERR! code ENOENT
npm ERR! errno -2
npm ERR! syscall chmod

npm ERR! enoent ENOENT: no such file or directory, chmod '/home/clux/testrepos/tiebreaker/node_modules/escodegen/node_modules/esprima/bin/esparse.js'
npm ERR! enoent This is most likely not a problem with npm itself
npm ERR! enoent and is related to npm not being able to find a file.
npm ERR! enoent 

npm ERR!     /home/clux/testrepos/tiebreaker/npm-debug.log
```

And what's worse, the first module's `node_modules` folder (`smell/node_modules`) disappears during the last failed install!

By doing `npm install` inside `tiebreaker` without first issuing `npm link smell ffa` (or by just blatting `./node_modules` after), it will succeed.

This kind of mechanic of making the least dependent modules link-availble first, then installing the rest all worked flawlessly on huge module sets with npm@2, but now I pretty much have to restrict `link` to the occasional manual hack. The more modules I put in, it may still work, but there will be loads more `skippingAction Module is inside a symlinked module` warnings, and more often than not, there will something that breaks it. This example does not have any warnings, it just breaks at the last state, leaving my folders broken in the process.

iarna:
Fixes #10800 and maybe #10343.

TODO:
- [x] Integration test for #10800 

`n/a` Integration test for #10343 _The core issue ended up being unrelated, even if many of the follow on comments are probably fixed by this. But those follow on comments are covered by the test for #10800._

The core problem here was the installer's pruning phase was removing things from the tree that it shouldn't have.

Now as it happens, the code base has been moving in a direction that would allow it to not have pruning as a distinct phase. As such, I decided to bite the bullet and make the changes necessary to allow this, culminating in removing the phase entirely.

Definitions:
- _physical tree_ - A tree based on location of packages on disk. Represented by the `parent` _(one)_ and `children` _(many)_ properties.
- _logical tree_ - A tree based on the relationship between packages. Represented by the `requiredBy` _(many)_ and `requires` _(many)_ properties.

This work broke down into a couple of related chunks:

In order to remove the pruning phase I had to shore up the installers ability to maintain it's logical relationships on the fly (that is, `.requires` & `.requiredBy`)
- 192de88 deps: Make removeObsoleteDep remove from phys tree too<br>
  Previously it removed the node from the logical tree, that is, requires/requiredBy, but not the physical tree, that is parent/children.
- 700bc74 install: Use removeObsoleteDep when removing from phys tree
- 31a6f6d install: Remove failed non-optional deps from tree<br>
  This is specifically referring to failures in finding a matching version.<br><br>
  It sounds a little weird, because failing non-optional deps are ordinarily fatal and would fail out the entire tree, but what this does is ensure that transitive non-optional deps of an optional dependency get removed from the tree.  Without this, a failure in one dep of an optional dep would result in all the other deps of the optional dep still being installed.
- 242bcbb deps: Guard failedDependency against requiredBy not existing

In turn, it ended up being necessary to tell loadDeps to skip nodes we've removed from the tree. It can encounter them because it makes its list before it starts walking the tree.
- 16f2b9a install: loadDeps should skip nodes marked as removed<br>
  You might wonder how it's possible for "removed" nodes to get passed to loadDeps in the first place: Siblings of a failed dep are unbound from the tree but have a half life, in the form of the list of nodes that loadDeps is iterating over.
  Modules in `requiredBy` need to be uniqued on path, as it's totally valid to have more than one module with the same name requiring the same module.
- e5d106a deps: Rewrite replaceModule to match by name _or_ path<br>
  Previously replaceModule only matched via name, which was appropriate for updating `requires` and `children` but was not appropriate for `requiredBy`. Because `requiredBy` is a list of modules that require this module it can and will include modules that exist more than once in the physical tree but the current node as a dependency.  As such, those modules need to be uniqued on path instead.
  We had `userRequired` & `existing` properties on nodes, indicating that they were explicitly requested for this install or were on already disk when started, respectively. But we didn't always _use_ them as they were added later, sometimes we used huristics based on `_requiredBy` which was super ugly.
- 479c23e deps: Consistently use userRequired & existing<br><br>
  Also don't set them at recalculateMetadata time.  They're not reset, we don't have to recompute them.

Relatedly, we previously checked to see if a module was the root of the tree with `!node.parent`. This explicitly sets a property (`node.isTop`), which in turn makes code that uses much easier to grok.
- f5ee2c3 install: Refactor to explicitly check for top

Previously we kept loose notes on relationships in `node.package._requiredBy`, which was a text field that required some interpretation to use correctly. Now that relationships are a core part of the tree/graph nodes, we no longer need that:
- 9f59260 mutate-into-logical-tree: Rewrite to use requiredBy instead of _requiredBy
- 1a656c4 tar: Rewrite to use requiredBy instead of _requiredBy
- d6ccf7d is-dev: Rewrite to use requiredBy instead of _requiredBy
- d164fa5 diff-trees: Rewrite to use requiredBy instead of _requiredBy
- 0ed9fe8 install: Rewrite computeLinked to not use _requiredBy
- 6491996 save: Eliminate use of _requiredBy
- ada1d3f node, deps: Eliminate _requiredBy entirely

Having eliminated `_requiredBy`, `is-extraneous` needed changes which revealed how hard it was to reason about. So we replace it with something that is much more straightforward:
- d7279c5 is-extraneous: Rewrite to use the logical tree<br>
    A package is extraneous if:
  - it is disconnected from the top of the tree, UNLESS
  - There is no top level package.json, OR
  - it was requested as part of the current installation. (That is, `npm install foo` shouldn't show `foo` as extraneous in its summary, but `npm ls` after _should_.)
    It became useful to keep track of location of a node in the tree with `node.package._location`, but in the interests of keeping the `package.json` data read only, I moved it to `node.location`
- d8a32ae install: Rename node.package._location to node.location

Similarly with `_phantomChildren`. `phantomChildren` is how we stop ourselves from stepping on a hoisted children– that is every level we hoist them through we add the child to `phantomChildren`. If we end up also needing another version of the same package, we won't hoist to or through nodes that have it in their `phantomChildren`, just like we wouldn't hoist to or through if it was in their `children`.
- 3fe5bee node: Rename node.package._phantomChildren to node.phantomChildren

And finally `_bundled` which currently `fetch-package-metadata`'s `addBundled` adds. Because it's output from`read-package-tree`, it can't be saved. Instead of special casing this at save time, we just remove it ASAP. Long term I'd like to eliminate its use entirely. (A putative future version of `fetch-package-metadata` would take a node object and could annotate there.)
- 6b7bdcd install: Minimize the use of package._bundled

And with all of the `_` vars that the installer sets moved away, we need to turn around and generate them before saving because: A) they make debugging a lot easier and B) Visual Studio relies on `_requiredBy` at least.
- 959f5cc update-package-json: Update to explicitly include debugging data.<br>
  Historically we maintained `_requiredBy` and `_phantomChildren` (and others) in `node.package` and then just past them through to disk. As those npm now reads (versions) of those attributes directly from `node`, that means its no longer maintaining those package attributes.<br><br>
  But those package attributes are super useful for debugging, so we change this to explicitly generate them.<br><br>
  Also `_requiredBy` is used by the Visual Studio `npm` plugin. Yeah. =/<br><br>
  We also update sorting to be way more limited, to just fields we know we can safely sort.

And so, finally, we're ready to remove the install pruning step.
- dc850e7 install: Remove install pruning stage
- 39137fc deps: Rewrite how recalculateMetadata follows dependencies.<br><br>
  We keep track of missing dependencies and missing dev dependencies separately. This allows things like `npm ls` to make decisions about what it wants to warn about.<br><br>
  Despite having concrete lists of dependencies and dev dependencies we were guessing as to which was which. This _usually_ works fine, but in cases where dependencies and dev dependencies share a requirement but not precisely the same version, it could lead to misleading error messages.<br><br>
  Further, things in the shrinkwrap that are not in the package.json were not being
  reported at all.<br><br>
  This change does two things:<br><br>
  1. It rewrites `markDeps` to take an object with name, spec & kind (`prod` or `dev`). Previously it took a string of `name@spec` and extracted name & spec from that.
  2. When a dependency is both a prod and a dev dep, it will be correctly categorized. Previously if the dev spec didn't match it would still be flagged as a prod mismatch.
- 9ba2c79 mutate-into-logical-tree: Flag missing dev deps<br>
  This allows `ls` to, when asked to show dev dependences, eg, `npm ls --dev`, complain about missing dev deps.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "version": "3.9.6",
+  "version": "3.10.0",
   "name": "npm",
   "description": "a package manager for JavaScript",
   "keywords": [


iarna:
@othiym23 says 👍

othiym23:
:sheep::rocket::fire:

PR #12776, opened 2016-05-20
updated to current syntax

fibo:
It looks like it was not updated with current `npm run-script` syntax, since it now acceps only one parameter.
If I am not wrong, running scripts of `<pkg>` dependencies was supported in previous versions (what a pity it was removed), so I updated documentation.

### DIFF
--- a/doc/misc/npm-scripts.md
+++ b/doc/misc/npm-scripts.md
@@ -34,7 +34,7 @@ following scripts:
   stop and start scripts if no `restart` script is provided.
 
 Additionally, arbitrary scripts can be executed by running `npm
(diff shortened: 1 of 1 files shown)


zkat:
LGTM :)

zkat:
Thanks for this! This was one of the changes that was included as part of `npm@2` back in the day, and this was a straggler in the docs. This patch landed in `latest` as of `npm@3.9.4`, and it'll land in `lts` at some point in the future. I'll leave the PR open until then! 🎉

zkat:
I've merged this into the LTS branch. It'll be included with the next LTS release as well, whenever that happens. Thanks, again!

Issue #12781, opened 2016-05-20
npm markAsFailed recursion results in "Maximum call stack size exceeded"


vtravieso:
Invoking an npm command on the 'cli-color' package results in a "Maximum call stack size exceeded". 

Node: v4.4.4
npm: v3.9.2

Steps to reproduce:
1. npm install -g cli-color
2. change directory to where cli-color is installed
3. npm uninstall foo

Result: Npm throws the error "Maximum call stack size exceeded"

The error seems to be caused by the cli-color package having circular dependencies, which causes an infinite recursion in the npm code that marks a package and its reqs as failed (node_modules\npm\lib\install\actions.js:54:16) - see attached log file.

Skipping the recursion when the req has already been flagged avoids the error.

[npm-debug.zip](https://github.com/npm/npm/files/274612/npm-debug.zip)

jblotus:
+1

I had similar issue - disabling progress actually allowed my install to complete
`npm set progress=false && rm -rf ~/.npm && rm -rf node_modules && npm install`

https://github.com/npm/npm/issues/11283

PR #12784, opened 2016-05-20
Do not recurse into packages already marked as failed

vtravieso:
Fixes issue #12781 

On encountering an error, a pkg and its requires are marked as failed with a recursive function. When a package has circular dependencies this can result in an infinite recursion.


chris-grabcad:
Needs more tests? - maybe master needs to pass all the tests before you demand more tests?
#justsaying

iarna:
I like this, but as @zkat indicated, it would be very helpful if you could put together a reproducer for this bug. Due to the nature of this change and the nature of the bug, I'm inclined to advocate for taking this patch even without a test, but it would still be much preferable with one.

iarna:
I've taken this with a slightly different implementation in 55ac2fca8 which is merged into `release-next` and will be in the next npm release

Issue #12654, opened 2016-05-10
installs wrong version of modules


ColCh:
npm versions affected as well: `3.8.9`, `3.9.0`

After `npm cache clean`, `rm -r node_modules` and `npm install`, I experience following:
( I have `"lodash": "^4.11.1",` in my deps)

`npm ls`:

```
npm ERR! missing: lodash@^3.10.0, required by babel-core@6.8.0
npm ERR! missing: lodash@^3.10.1, required by babel-generator@6.8.0
npm ERR! missing: lodash@^3.10.0, required by babel-register@6.8.0
npm ERR! missing: lodash@^3.10.1, required by babel-template@6.8.0
npm ERR! missing: lodash@^3.10.1, required by babel-traverse@6.8.0
npm ERR! missing: lodash@^3.10.1, required by babel-types@6.8.1
npm ERR! missing: lodash@^3.10.0, required by babel-plugin-transform-es2015-block-scoping@6.8.0
npm ERR! missing: lodash@^3.10.0, required by babel-helper-define-map@6.8.0
npm ERR! missing: lodash@^3.10.0, required by babel-helper-builder-react-jsx@6.8.0
npm ERR! missing: lodash@^3.2.0, required by babel@5.8.38
npm ERR! missing: lodash@^3.10.0, required by babel-core@5.8.38
npm ERR! missing: lodash@^3.9.3, required by babel-plugin-proto-to-assign@1.0.4
npm ERR! missing: lodash@^3.10.1, required by grouped-queue@0.3.2
npm ERR! missing: lodash@^3.5.0, required by yeoman-generator@0.20.3
npm ERR! missing: lodash@^3.2.0, required by cheerio@0.19.0
npm ERR! missing: lodash@^3.6.0, required by yeoman-assert@2.2.1
npm ERR! missing: lodash@^3.3.1, required by inquirer@0.8.5
```

Running `npm install` second time resolves this problem:

```
➜  FOO git:(navigation) npm install
FOO@0.0.1 /FOO
├─┬ babel-core@6.8.0
│ ├─┬ babel-generator@6.8.0
│ │ └── lodash@3.10.1 
│ ├─┬ babel-register@6.8.0
│ │ └── lodash@3.10.1 
│ ├─┬ babel-template@6.8.0
│ │ └── lodash@3.10.1 
│ ├─┬ babel-traverse@6.8.0
│ │ └── lodash@3.10.1 
│ ├─┬ babel-types@6.8.1
[...]

Issue #11766, opened 2016-03-04
Dev deps that conflict with subdeps can produce invalid trees


iarna:
Using this `package.json`:

```
{
  "dependencies": {
    "aws-sdk": "^2.2.41"
  },
  "devDependencies": {
    "lodash": "^4.6.1"
  }
}
```

```
$ npm install
example@1.0.0 /Users/rebecca/code/npmtest/example/
├─┬ aws-sdk@2.2.41
│ ├── sax@1.1.5
│ ├── xml2js@0.4.15
│ └── xmlbuilder@2.6.2
└── lodash@4.6.1
$ npm ls
example@1.0.0 /Users/rebecca/code/npmtest/example
├─┬ aws-sdk@2.2.41
│ ├── sax@1.1.5
│ ├── xml2js@0.4.15
│ └─┬ xmlbuilder@2.6.2
│   └── UNMET DEPENDENCY lodash@~3.5.0
└── lodash@4.6.1
```

Running `npm install` a second time corrects the issue.

robey:
thanks for listening to me flail helplessly on this and finding a simple repro case. :)

![thanks](https://cloud.githubusercontent.com/assets/172/13514168/5847533e-e15b-11e5-88f3-813cf11c6d23.jpg)

nmccready:
Yup this just bit us in the but here where we were using object-assign as a devDep and gulp needed it as a dependency. However Heroku nodejs buildpack strips (prunes) all devDependencies. So gulp would bomb out. 

Any idea of resolution? We are seeing this on 2.14.22

munro:
I'm experiencing a very similar issue: `missing: async@^1.4.0, required by handlebars@4.0.5`

I believe what's happening is `devDependencies` is not being considered when flattening the packages, so `dependencies.handlebars` installs `1.x` to `node_modules/`, then `devDependencies.async` comes along and clobbers it with `2.x`.  So when I go to shrinkwrap, it sees this discrepancy and aborts.

Handlebars has their dependencies setup correctly, and when I poke into `node_modules/handlebars/node_modules/` I don't see an `async` package.

Immediate work around solution is moving the `devDependencies` to `dependencies`.

```json
{
  "dependencies": {
    "handlebars": "^4.0.5",
  },
  "devDependencies": {
    "async": "^2.0.0-rc.2",
  }
}
```

node v6.1.0
npm 3.8.6

schmod:
This may be similar to the transitive dependency issue described in #10727.  

[...]

Issue #11062, opened 2016-01-07
Tree not cleanly installed for torrent-stream


TheAlphaNerd:
osx 10.10
node v5.4.0
npm v3.3.12

to reproduce

`git clone https://github.com/mafintosh/torrent-stream.git`
`npm install`
`npm ls`

```
npm ERR! missing: hat@0.0.3, required by torrent-stream@1.0.2
npm ERR! missing: bencode@^0.7.0, required by parse-torrent-file@2.1.4
npm ERR! missing: bittorrent-tracker@^7.0.0, required by torrent-discovery@5.2.0
npm ERR! missing: bencode@^0.7.0, required by bittorrent-dht@6.2.0
npm ERR! missing: bencode@^0.7.0, required by k-rpc-socket@1.4.3
```

This works as expected on LTS (npm v2.x). What is even weirder is that if you checkout v1.0.1 of torrent-stream everything gets resolved. [This commit](https://github.com/mafintosh/torrent-stream/commit/b990ef1ef9c5655b32921253dd8604bd2733bba0) pruning deps broke everything. Manually install hat `npm i hat` makes `npm ls` resolve error free.

/cc @mafintosh

iarna:
I tried this w/ `npm@3.5.4`, and `https://github.com/mafintosh/torrent-stream.git#61b7bd9` from and "only" `bittorrent-tracker` fails to install:

```
npm ERR! missing: bittorrent-tracker@^7.0.0, required by torrent-discovery@5.3.0
```

Incidentally, a second run of `npm install` corrects the tree:

iarna:
For fun, I tried `npm install https://github.com/mafintosh/torrent-stream.git#61b7bd9` and bafflingly that worked just fine.

TheAlphaNerd:
we are deep into the weirdness

benoitvallon:
I mentioned missing packages on first `npm install` as well in this issue https://github.com/npm/npm/issues/10985 it seems to me that those bugs have similarities.

`git clone https://github.com/benoitvallon/react-native-nw-react-calculator.git`
`cd react-native-nw-react-calculator`
`npm install`
`npm ls`

Here is what is missing after the first `npm install`
```
npm ERR! missing: shebang-regex@^1.0.0, required by babel-core@6.4.0
npm ERR! missing: slash@^1.0.0, required by babel-core@6.4.0
npm ERR! missing: babel@^5.4.7, required by fbjs-scripts@0.4.0
```
Everything is here after the second run.

[...]

Issue #11043, opened 2016-01-06
Nested modules are missing


TerenceZ:
The npm version is v3.5.2, Here is the package dependencies:

``` json
  "devDependencies": {
    "babel": "^6.3.26",
    "babel-cli": "^6.3.17",
    "babel-core": "^6.3.26",
    "babel-preset-es2015": "^6.3.13",
    "babel-preset-stage-0": "^6.3.13",
    "rimraf": "^2.4.4"
  },
  "dependencies": {
    "babel-eslint": "^4.1.6",
    "eslint": "^1.10.3",
    "eslint-plugin-html": "^1.1.0",
    "glob": "^6.0.1",
    "parse5": "^2.1.0",
    "stylelint": "^3.1.3"
  }
```

It's worth noting that the `babel-eslint^4.1.6` requires `babel-core^5.8.33` as dependency.

After `npm install`, we can find that the `babel-core^5.8.33` hasn't been installed as nested node module in `babel-eslint`. If we type `npm ls --depth=0`, a warning **"missing: babel-core@^5.8.33, required by babel-eslint@4.1.6"** is raised up.

benoitvallon:
Same problem here issue: https://github.com/npm/npm/issues/10985

```shell
npm ERR! missing: shebang-regex@^1.0.0, required by babel-core@6.3.26
npm ERR! missing: slash@^1.0.0, required by babel-core@6.3.26
npm ERR! missing: babel@^5.4.7, required by fbjs-scripts@0.4.0
```

AgentME:
Just realized it's the same problem here too: https://github.com/npm/npm/issues/11200

marcj:
Same here. Any ideas?

AnimiVulpis:
I created a [minimal example repo](https://github.com/AnimiVulpis/npm-bug) on which a sub-dependency is missing on every clean install.

The missing dependency only occurs if `redux` is in **dependencies** and an incompatible version of `lodash` in **devDependencies**.

(no, a `npm cache clean` does not help 😉 )

Maybe this helps someone figure out why this is happening.

pwalczyszyn:
I have similar problem.
Two solutions that worked for me:

1. Moved all `devDependencies` to `dependencies`.
2. Instead of running `npm install` I run `npm update`.

schmod:
I've reproduced this on NPM v3.8.7.  It appears to have the same cause as #10727.

zkat:
[...]

PR #12811, opened 2016-05-23
Fix #10727 - Consider devDependencies when deciding whether to hoist a package

### LINKED ISSUE #10727: Must call `npm install` twice to get all the dependencies pulled down
This is a strange one.  I was updating my version of handlebars to the latest and I created a simple bash script for the rest of my team to use.  However, there's an issue whereby you need to call `npm install` twice in order to pull down all the dependencies.  So here's everything I've got:

My package.json file: https://gist.github.com/JoeCianflone/ea7ceffacd7a2aab36ab#file-package-json
My update script: https://gist.github.com/JoeCianflone/ea7ceffacd7a2aab36ab#file-update-sh
and the error I get when running gulp compile which indicates that gulp is missing a dependency: https://gist.github.com/JoeCianflone/ea7ceffacd7a2aab36ab#file-error-sh

I have tried moving the `npm install` to the end of the file, I've done this manually multiple times and I can always recreate this error.  I'm running npm 3.3.12 and I'm on OSX 10.11 using the built in terminal.  If you need any more info please let me know.

schmod:
Fixes several issues that were causing devDependencies to block certain transitive dependencies from being installed, or leave incompatible versions in place.

`deps#earliestInstallable` now considers `devDependencies` that are being installed when determining where to place a package, and will hoist or resolve conflicts where necessary, so that the `devDependencies` and `dependencies` can coexist peacefully.  

Previously, if we requested:

```
pkg@1.0.0 
├── depA@1.0.0 (dev dependency)
├─┬ libA@1.0.0
│ ├── depA@2.0.0
```

We would actually install:

```
pkg@1.0.0 
├── depA@1.0.0 (dev dependency)
├── libA@1.0.0
```

`depA@2.0.0` would first be placed in the root, as `earliestInstallable` did not consider `pkg`'s devDependencies when looking for potential conflicts.  Laster, we would process the devDependencies, and `depA@1.0.0` would overwrite `depA@2.0.0`'s position in the tree.  [Overwriting nodes like this seems somewhat bug-prone, although I'm not sure if there are any good options for detecting when it happens so we can print a warning...]

---

This fix uncovered a second condition (observed in the `shrinkwrap-shared-dev-dependency` test-case), where devDependencies are not properly considered when computing the minimal number of packages that need to be installed.  If a package is included as a `devDependency` and is also transitive dependency of a regular dependency, that package would be installed twice, leaving us with a tree that looks something like this:

```
pkg@1.0.0 
├── depA@2.0.0 (dev dependency)
├─┬ libA@1.0.0
│ ├── depA@2.0.0 (dependency of libA. redundant)
```

This patch alters `npm install`'s behavior so that the transitive dependencies of our dev + prod dependencies are computed together.  Previously, we calculated the two trees separately, and production dependencies had no knowledge of any devDependencies that were being installed in the same session.

---

The symptoms described in #10727 could be worked around by running `npm install` twice.  I honestly can't figure out why this workaround actually worked, and consider the behavior to be somewhat alarming, as it suggests there may be inconsistent behavior elsewhere. 

---

Fixes #10727 #11062 #12654 #10277 #11766 #11043


lib/install/deps.js
```diff
@@ -613,8 +635,12 @@ var earliestInstallable = exports.earliestInstallable = function (requiredBy, tr
   // if this tree location requested the same module then we KNOW it
```
schmod: @iarna: While working on this issue, this caveat bothered me somewhat, as it makes `earliestInstallable` highly dependent on the behavior of `findRequirement` (which is particularly troublesome, given that `earliestInstallable` is exported and used elsewhere)

It may be worth considering somehow merging the two methods, or adding some sort of safeguard mechanism so that we can verify that `findRequirement` indeed did its job.  

You seem to know this code better than anyone else -- I'd be interested in hearing your thoughts about how we might be able to make this more robust without hurting performance.

lib/install.js
```diff
@@ -382,11 +383,13 @@ Installer.prototype.loadAllDepsIntoIdealTree = function (cb) {
     steps.push([validateArgs, this.idealTree, this.args])
     steps.push([loadRequestedDeps, this.args, this.idealTree, saveDeps, cg.newGroup('loadRequestedDeps')])
   } else {
-    if (this.prod) {
+    if (this.prod && this.dev) {
```
iarna: Can you explain to me why `loadAllDeps` (and thus this code) needs to exist? It looks to me like the change to `earliestInstallable` would be sufficient by itself.

iarna: I assume you're viewing this as related to your second condition, but I'm not entirely comfortable with this approach. Splitting the prod and dev steps is intentional and the idea is that:
```
npm install --only=prod
npm install --only=dev
```
should give you exactly the same tree as:
```
npm install
```
And I think your change is intentionally breaking that?

lib/install/deps.js
iarna: Ok, so let's back up and talk about what `earliestInstallable` is supposed to do:

You give it a tree, the node in the tree with this requirement and the package.json data for the package and it tells you the location closest to the root of the tree that you can install the package in without producing conflicts. If it can't find any such location then it just returns the requiring node.

I think the problem here is that this comment is a bit off base. We do make one assumption:

* You already know that you need to install this module, you just don't know where yet.

Ordinarily this code is pointless given the existence of `_phantomChildren`– but `_phantomChildren` isn't populated in two circumstances:

1. If you requested the installation of just one (or a few) named modules, eg `npm install thingy`.
2. If you asked for `--only=dev` on a folder without prod deps installed.
3. Or relatedly, if you did an ordinary install and we still have a separate dev deps stage.

So having written this out it seems clear to me that it should be doing a version match like `findRequirement` does to see if the version we have in our hands is compatible with the spec in dependencies (and devDependencies) and if it isn't returning null.

I think that would look something like:
```js
var deps = tree.package.dependencies || {}
if (deps[pkg.name] && !doesPkgVersionMatch(deps[pkg.name], pkg)) return null
var devDeps = tree.package.devDependencies || {}
if (devDeps[pkg.name] && !doesPkgVersionMatch(devDeps[pkg.name], pkg)) return null
```

With a `doesPkgVersionMatch` being split from `doesChildVersionMatch` like:
```js
function doesChildVersionMatch (child, requested, requestor) {
  // we always consider deps provided by a shrinkwrap as "correct" or else
  // we'll subvert them if they're intentionally "invalid"
  if (child.parent === requestor && child.fromShrinkwrap) return true
  return doesPkgVersionMatch(requested, child.package)
}

function doesPkgVersionMatch (requested, pkg) {
  // ranges of * ALWAYS count as a match, because when downloading we allow
  // prereleases to match * if there are ONLY prereleases
  if (requested.spec === '*') return true

  var pkgReq = pkg._requested
  if (pkgReq) {
    if (pkgReq.rawSpec === requested.rawSpec) return true
    if (pkgReq.type === requested.type && pkgReq.spec === requested.spec) return true
  }
  if (!registryTypes[requested.type]) return requested.rawSpec === pkg._from
  return semver.satisfies(child.package.version, requested.spec)
}
```

iarna: With an inverted scenario from what I discussed in https://github.com/npm/npm/pull/12811/files#r64309610: ```
pkg@1.0.0 
├── DEV: depA@^1.2.0
├─┬ libA@1.0.0
│ ├── depA@~1.2.0
```
A transitive prod dep can change what dev dep version gets installed. This would mean that:
```
npm install --only=dev && npm install --only=prod
```

Would give you:
```
pkg@1.0.0 
├── depA@1.4.5
├─┬ libA@1.0.0
│ ├── depA@1.2.7
```
But:
```
npm install --only=prod && npm install --only=dev
```
would give you:
```
pkg@1.0.0 
├── depA@1.2.7
├── libA@1.0.0
```
But I'm a lot more comfortable with that lack of symmetry than the other kind.

Comment (schmod) [inline lib/install/deps.js]: > So having written this out it seems clear to me that it should be doing a version match like findRequirement does to see if the version we have in our hands is compatible with the spec in dependencies (and devDependencies) and if it isn't returning null.

Yes, that behavior seems like it helped the original bug to slip through:

My prod deps would throw `lodash@0.9.2` into my root `node_modules` folder.  During `loadDevDeps`, a copy of `lodash@2.0.0` would silently take its place, because `earliestInstallable` wasn't able to realize that the versions were incompatible.

As you mentioned, `earliestInstallable` makes a strong assumption that we _really_ need to install this module _somewhere_, and by falling back to the root `node_modules` folder, guarantees that we do so at all costs.  If `earliestInstallable` crawls all the way up to the root, and finds a conflicting version of the same package already there, it assumes that we _really need_ the package, and overwrites the existing one.  

Unfortunately, `earliestInstallable` has no way of knowing that the versions conflicted, so it had no way of raising a warning.  I don't think we would have a good way of recovering from this situation, but silent failure seems bad if we can avoid it.

File: lib/install.js (hunk 2 above)
Comment (schmod) [inline lib/install.js]: Interesting.

I initially made this change, because the other fix (considering dev dependencies) caused the `shrinkwrap-dev-depdency.js` test to fail.  

In that scenario, package.json asks for this:

```
pkg@1.0.0 
├── DEV: depA@1.0.0
├─┬ libA@1.0.0
│ ├── depA@1.0.0
```

The test case expects us to install this:
```
pkg@1.0.0 
├── depA@1.0.0
├── libA@1.0.0
```

What we actually get (without my `loadAllDeps` change) is this:
```
pkg@1.0.0 
├── depA@1.0.0
├─┬ libA@1.0.0
│ ├── depA@1.0.0
```

Unlike your example (which is indeed valid), this tree is needlessly suboptimal.  Nobody benefits from this behavior -- the tree is bigger, and looks _less_ like the prod-only tree.  

Curiously, if `loadDevDeps` is run separately, but before `loadDeps`, we get the correct tree.

This seems to relate to our discussion about `earliestInstallable` above.  When processing devDeps, `findRequirement` doesn't seem to have picked up on the copy of `depA` in the root folder.  `earliestInstallable` sees a copy of `depA` in the root folder, doesn't realize that it's a compatible version (because it trusted that `findRequrement` to have previously found that for us), and decides to hoist a new copy of `depA` into `libA`.

Comment (schmod) [inline lib/install.js]: From your example, I'm a little curious about why we got this tree:

```
pkg@1.0.0 
├── libA@1.0.0
├── depA@1.2.7
```

Is this how we would resolve a similar conflict among transitive dependencies?

Would this:

```
pkg@1.0.0 
├─┬ libA@~1.0.0
│ ├── depA@~1.2.0
├─┬ libB@~1.0.0
│ ├── depA@^1.2.0
```

also yield the same tree (prefering a shared copy of `depA@1.2.7` instead of putting `depA@1.4.5` in `libB`)?

File: lib/install/deps.js (hunk 1 above)
Comment (iarna) [inline lib/install/deps.js]: We should get warnings about conflicts at the root level due to invalidness checking, although really I'd like conflicts to be warned separately from invalid installed modules.

File: lib/install.js (hunk 2 above)
Comment (iarna) [inline lib/install.js]: Yes, your example would produce the same result. The difference is that it would _always_ produce that result no matter what options you passed into npm.

I think having it not always select the most recent version it _could_ use is ok, I just don't want different results for dev and prod environments.

Comment (schmod) [inline lib/install.js]: Agreed that this is something we want to avoid.

I'll roll back my change, and figure out if there's a better way to fix the bug, without also breaking the `shrinkwrap-dev-depdency.js` test-case.  [Suggestions welcome!]

As far as this functionality goes:

1. Are there cases where we _cannot_ guarantee that transitive devDependencies won't affect the regular/transitive dependencies that get installed?
2. Should this behavior be documented and tested?
3. What about weird stuff like this? 
    
    ```json
    "dependencies": {   
      "lodash": "2.0.0"
    },
    "devDependencies": {
      "lodash": "4.0.0"
    }
    ```
    Currently, we seem to treat this as valid (although many things don't work).  `npm install` against this package resolves `lodash@2.0.0`, and exits cleanly, only indicating that anything is run when `npm ls` is printed to the console.  `npm install --prod` and `npm ls --prod` work as expected.  `npm install --only=dev` doesn't install anything.

That being said, maybe we're starting to stray too far away from the original issue?

**Comment by schmod** [comment]
On @iarna's advice, I've restructured this so `loadDeps` and `loadDevDeps` are once again discrete stages.

`npm install` now attempts to install a valid tree of prod + dev dependencies that _most closely resembles_ the prod-only tree (possibly at the expense of reduced efficiency).  We do this by installing the production dependencies, and subsequently install the devDeps on top of the resulting tree as an additive operation.  

`devDependencies` (and their transitive deps) will attempt to re-use packages installed by the prod deps, _but not the other way around_.

There is one unavoidable exception to this rule (and the original impetus for this PR): Transitive production dependencies need to be hoisted out of the root package if the root package lists the same package as a devDependency.

If the devDependency and transitive production dependency _happen to have_ the same version, we'll re-use the same module, but won't allow this to influence which versions get installed.  I think that this offers the best tradeoff between predictability and efficiency.  

---

One caveat:

This hoisting behavior only takes place if both prod AND dev dependencies are being installed simultaneously.  

`npm install` is therefore *NOT* guaranteed to be equivalent to `npm install --only=prod && npm install --only=dev`, as the second command can potentially overwrite root-level packages installed by the first.

There are two possible fixes for this issue (if we even want to consider it an issue):

* **Simple but inefficient**:  Always assume we might install our root-level devDeps, and always hoist  transitive deps that conflict with devDeps at the root level. _Easy fix. Predictable.  Yields inefficient production-only trees._  
* **More complicated**: If `npm install` overwrites a root-level package, it should determine if the overwritten package needs to reinstalled at a higher level in the tree to meet other dependencies.  _Should guarantee that a `npm install` command won't leave with any packages in `node_modules` with unmet dependencies.  Could be computationally-intensive and produce huge trees if `npm prune` isn't run periodically.  Might contradict user expectations of what `--only` does._

IMO, this is something we can tackle separately in the future.

**Comment by schmod** [comment]
Also, the failing test appears to be unrelated to this PR (as the same test is failing on other builds)

**Comment by zkat** [comment]
I think this might be ready for another round of review?

**Comment by mourner** [comment]
Any updates on this?

**Comment by iarna** [comment]
> npm install is therefore NOT guaranteed to be equivalent to npm install --only=prod && npm install --only=dev, as the second command can potentially overwrite root-level packages installed by the first.

True, but the second run wouldn't leave a broken tree. `npm` is smart enough to hoist the transitive deps if it replaces one in a separate invocation.

File: test/tap/unit-deps-earliestInstallable.js [hunk 3]
```diff
@@ -0,0 +1,92 @@
...
+  // an incompatible verson of dep2. required by dep1
+  var dep2a = {
+    package: {
+      name: 'dep2',
+      version: '2.0.0'
+    },
+    parent: dep1
+  }
+
+  var pkg = {
```
iarna: The top of the tree should have `isTop: true` set

test/tap/unit-deps-earliestInstallable.js
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  var dep2 = {
+    package: {
+      name: 'dep2',
+      version: '1.0.0'
+    }
+  }
+
+  var pkg = {
```
iarna: Also needs `isTop`

lib/install/deps.js
```diff
@@ -643,8 +643,21 @@ var earliestInstallable = exports.earliestInstallable = function (requiredBy, tr
...
-  if (!tree.removed && requiredBy !== tree && deps[pkg.name]) {
+  var deps = Object.keys(tree.package.dependencies || {})
+
+  // if we're at the root of the tree, and are planning to install
+  // devDependencies in a future pass, don't put any conflicting
+  // packages in the root during the first pass.
+  if (!tree.parent && (npm.config.get('dev') || !npm.config.get('production'))) {
+    // if the same package exists in the root as a devDep, reuse it if the
+    // version matches exactly.
+    if (pkg.version && pkg.version === tree.package.devDependencies[pkg.name]) {
```
iarna: This can't work because `tree.package.devDependencies[pkg.name])` is a specifier, not a version. This needs to be a semver check.

lib/install/deps.js
```diff
@@ -643,8 +643,21 @@ var earliestInstallable = exports.earliestInstallable = function (requiredBy, tr
...
   // isn't compatible because if it were findRequirement would have
   // found that version.
-  var deps = tree.package.dependencies || {}
-  if (!tree.removed && requiredBy !== tree && deps[pkg.name]) {
+  var deps = Object.keys(tree.package.dependencies || {})
+
+  // if we're at the root of the tree, and are planning to install
+  // devDependencies in a future pass, don't put any conflicting
+  // packages in the root during the first pass.
+  if (!tree.parent && (npm.config.get('dev') || !npm.config.get('production'))) {
```
iarna: We also have `--also` and `--only`. This should probably be:

```
var alwaysIncludeDev = npm.config.get('dev') || /^dev(elopment)$/.test(npm.config.get('also'))
var onlyIncludeProduction = npm.config.get('production') || /^prod(uction)$/.test(npm.config.get('only'))
if (!tree.parent && (alwaysIncludeDev || !onlyIncludeProduction)) {
```

But to be honest, I'm not sure we should even gate this? Why not do this kind of hoisting even if you ask for `--only=prod`? It'd ensure that you got the same tree with `npm install --only=prod && npm install --only=dev` as you would with `npm install`

iarna (changes requested):

test/tap/unit-deps-earliestInstallable.js
iarna: (Setting this allows the assertions to fail without this patch. Without setting this, it crashes before the assertions can fail.)

lib/install/deps.js
iarna: It's not actually as simple as a semver check as not all specifiers are semver. This is what `doesChildVersionMaatch` but that takes a package node as an argument, which makes this check awkward.

test/tap/unit-deps-earliestInstallable.js
```diff
@@ -0,0 +1,92 @@
...
+    children: [dep1],
+    package: {
+      name: 'pkg',
+      dependencies: { dep1: '1.0.0' },
+      devDependencies: { dep2: '1.0.0' }
```
iarna: If you change this to `^1.0.0` then the test will fail where it should succeed.

iarna:
My version of this PR: https://github.com/npm/npm/compare/release-next...iarna/pr/12811

zkat:
@schmod thanks a ton for this patch. Sorry about the delay in getting it merged! This PR was included in `npm@4.0.5` :)

PR #12814, opened 2016-05-24
shrinkwrap: add lifecycle script

### LINKED ISSUE #10744: Feature request: `postshrinkwrap` lifecycle event
I'd find it very useful to have a `postshrinkwrap` lifecycle event that would be run *after* `npm shrinkwrap`, specified in either the `scripts` section of `package.json` or in `node_modules/.hooks`.

(While I don't need it, I suppose a `preshrinkwrap` event would make sense, too.)

While I realize I might be (ab)using the `shrinkwrap` functionality in a somewhat non-standard way, I want to do a bit of normalization on the `npm-shrinkwrap.json` file—in my case, just normalizing the `resolved` URL. While I can do this as a `pre-commit` hook in `git`, it'd be nice to do that immediately after the `npm-shrinkwrap.json` file is written.

SimenB:
Fixes #10744

Tried copying the approach in #8287, as suggested in the issue.

Works in my local testing at least 😄

### DIFF
--- a/lib/shrinkwrap.js
+++ b/lib/shrinkwrap.js
@@ -7,8 +7,10 @@ var path = require('path')
 var log = require('npmlog')
 var writeFileAtomic = require('write-file-atomic')
 var iferr = require('iferr')
+var readPackageJson = require('read-package-json')
 var readPackageTree = require('read-package-tree')
 var validate = require('aproba')
+var chain = require('slide').chain
 var npm = require('./npm.js')
 var recalculateMetadata = require('./install/deps.js').recalculateMetadata
 var validatePeerDeps = require('./install/deps.js').validatePeerDeps
@@ -16,6 +18,7 @@ var isExtraneous = require('./install/is-extraneous.js')
 var isOnlyDev = require('./install/is-dev.js').isOnlyDev
 var packageId = require('./utils/package-id.js')
 var moduleName = require('./utils/module-name.js')
+var lifecycle = require('./utils/lifecycle.js')
 
 shrinkwrap.usage = 'npm shrinkwrap'
 
@@ -30,11 +33,24 @@ function shrinkwrap (args, silent, cb) {
   }
 
   var dir = path.resolve(npm.dir, '..')
+  var packagePath = path.join(npm.localPrefix, 'package.json')
   npm.config.set('production', true)
-  readPackageTree(dir, andRecalculateMetadata(iferr(cb, function (tree) {
-    var pkginfo = treeToShrinkwrap(tree, !!npm.config.get('dev') || /^dev(elopment)?$/.test(npm.config.get('also')))
-    shrinkwrap_(pkginfo, silent, cb)
-  })))
+
+  readPackageJson(packagePath, iferr(cb, function (data) {
+    lifecycle(data, 'preshrinkwrap', function () {
+      readPackageTree(dir, andRecalculateMetadata(iferr(cb, function (tree) {
+        var pkginfo = treeToShrinkwrap(tree, !!npm.config.get('dev') || /^dev(elopment)?$/.test(npm.config.get('also')))
+
+        chain([
+          [lifecycle, tree.package, 'shrinkwrap'],
+          [shrinkwrap_, pkginfo, silent],
+          [lifecycle, tree.package, 'postshrinkwrap']
+        ], iferr(cb, function (data) {
+          cb(null, data[0])
+        }))
+      })))
+    })
+  }))
 }
 
 function andRecalculateMetadata (next) {
--- a/test/tap/shrinkwrap-lifecycle.js
+++ b/test/tap/shrinkwrap-lifecycle.js
@@ -1,0 +1,55 @@
+var fs = require('graceful-fs')
+var path = require('path')
+
+var mkdirp = require('mkdirp')
+var osenv = require('osenv')
+var rimraf = require('rimraf')
+var test = require('tap').test
+
+var common = require('../common-tap.js')
+var pkg = path.resolve(__dirname, 'shrinkwrap-lifecycle')
+
+test('npm shrinkwrap execution order', function (t) {
+  setup()
+  fs.writeFileSync(path.resolve(pkg, 'package.json'), JSON.stringify({
+    author: 'Simen Bekkhus',
+    name: 'shrinkwrap-lifecycle',
+    shrinkwrap: '0.0.0',
+    description: 'Test for npm shrinkwrap execution order',
+    scripts: {
+      preshrinkwrap: 'echo this happens first',
+      shrinkwrap: 'echo this happens second',
+      postshrinkwrap: 'echo this happens third'
+    }
+  }), 'utf8')
+  common.npm(['shrinkwrap'], [], function (err, code, stdout) {
+    if (err) throw err
+
+    var indexOfFirst = stdout.indexOf('echo this happens first')
+    var indexOfSecond = stdout.indexOf('echo this happens second')
+    var indexOfThird = stdout.indexOf('wrote npm-shrinkwrap.json')
+    var indexOfFourth = stdout.indexOf('echo this happens third')
+
+    t.ok(indexOfFirst >= 0)
+    t.ok(indexOfSecond >= 0)
+    t.ok(indexOfThird >= 0)
+    t.ok(indexOfFourth >= 0)
+
+    t.ok(indexOfFirst < indexOfSecond)
+    t.ok(indexOfSecond < indexOfThird)
+    t.ok(indexOfThird < indexOfFourth)
+
+    t.end()
+  })
+})
+
+test('cleanup', function (t) {
+  process.chdir(osenv.tmpdir())
+  rimraf.sync(pkg)
+  t.end()
+})
+
+function setup () {
+  mkdirp.sync(pkg)
+  process.chdir(pkg)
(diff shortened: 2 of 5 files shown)


lib/shrinkwrap.js
```diff
@@ -33,10 +36,32 @@ function shrinkwrap (args, silent, cb) {
...
+      chain([
+        [lifecycle, data, 'preshrinkwrap'],
+        [shrinkwrap_, pkginfo, silent],
+        [lifecycle, data, 'postshrinkwrap']
+      ], cb)
+    }))
   })))
 }
 
+function readPackage (cb) {
```
SimenB: is there a common util for this?

lib/shrinkwrap.js
```diff
@@ -33,10 +36,35 @@ function shrinkwrap (args, silent, cb) {
...
     var pkginfo = treeToShrinkwrap(tree, !!npm.config.get('dev') || /^dev(elopment)?$/.test(npm.config.get('also')))
-    shrinkwrap_(pkginfo, silent, cb)
+
+    readPackage(iferr(cb, function (data) {
+      chain([
+        [lifecycle, data, 'preshrinkwrap'],
+        [shrinkwrap_, pkginfo, silent],
+        [lifecycle, data, 'postshrinkwrap']
+      ], function(err, data)  {
+        cb(err, data[0])
```
SimenB: is there some cleaner way to unwrap the array?

lib/shrinkwrap.js
```diff
@@ -33,10 +36,34 @@ function shrinkwrap (args, silent, cb) {
...
     var pkginfo = treeToShrinkwrap(tree, !!npm.config.get('dev') || /^dev(elopment)?$/.test(npm.config.get('also')))
-    shrinkwrap_(pkginfo, silent, cb)
+
+    readPackage(iferr(cb, function (data) {
+      chain([
+        [lifecycle, data, 'preshrinkwrap'],
+        [shrinkwrap_, pkginfo, silent],
+        [lifecycle, data, 'postshrinkwrap']
+      ], iferr(cb, function (data) {
+        cb(null, data[0])
```
SimenB: is there some cleaner way to unwrap the array?

othiym23:
To land, this needs a new test that the correct events fire on `npm shrinkwrap`. Once that's landed, we can get this out, which will be the first semver-minor bump in a bit for the CLI. Thanks for your time!

SimenB:
What's the best way to test that? I see `version-lifecycle`, but that seems overkill, as it also tests the git tagging which isn't part of shrinkwrap

SimenB:
Wrote a test for it now, unsure if it's the best way to test it...

lib/shrinkwrap.js
SimenB: this still seems like something that should be a common util

test/tap/shrinkwrap-lifecycle.js
```diff
@@ -0,0 +1,42 @@
...
+test('npm shrinkwrap execution order', function (t) {
+  setup()
+  fs.writeFileSync(path.resolve(pkg, 'package.json'), JSON.stringify({
+    author: 'Simen Bekkhus',
+    name: 'shrinkwrap-lifecycle',
+    shrinkwrap: '0.0.0',
+    description: 'Test for npm shrinkwrap execution order',
+    scripts: {
+      preshrinkwrap: 'echo this happens first',
+      postshrinkwrap: 'echo this happens second'
```
zkat: Could you add the `shrinkwrap` event here, too?

SimenB: What do mean? I added an assert that `wrote npm-shrinkwrap.json` comes in the middle, but adding things like `shrinkwrap: 'hello world'` doesn't do anything

iarna: But it needs to do something, as every other lifecycle includes all three. What it would do, I dunno, usually in cases like this its called after, just before the `post` lifecycle.

lib/shrinkwrap.js
iarna: The call to `readPacakge` is unnecessary here. `tree.package` contains the same thing as `data`.

SimenB: Ah, thank you! Makes sense to not have to parse it again

test/tap/shrinkwrap-lifecycle.js
SimenB: But it's not called, unless I'm doing something wrong...

lib/shrinkwrap.js
iarna: There is! (The module `read-package-json` is how this is ordinarily done.) But see above!

test/tap/shrinkwrap-lifecycle.js
iarna: No, sorry, I was suggesting that you call it. Uh, I'll add a note above…

lib/shrinkwrap.js
```diff
@@ -33,10 +36,34 @@ function shrinkwrap (args, silent, cb) {
   npm.config.set('production', true)
   readPackageTree(dir, andRecalculateMetadata(iferr(cb, function (tree) {
     var pkginfo = treeToShrinkwrap(tree, !!npm.config.get('dev') || /^dev(elopment)?$/.test(npm.config.get('also')))
-    shrinkwrap_(pkginfo, silent, cb)
+
+    readPackage(iferr(cb, function (data) {
+      chain([
+        [lifecycle, data, 'preshrinkwrap'],
+        [shrinkwrap_, pkginfo, silent],
```
iarna: I'm suggesting you call a `shrinkwrap` lifecycle either before or after this. Which isn't super important, we just need to document it. For what it's worth, `build` does `preinstall`, _work_, `install`, `postinstall` and`unbuild` runs `preuninstall`, `uninstall`, _work_, `postuninstall`. So seems like _after_ would make more sense, to align with `build`'s behavior.

lib/shrinkwrap.js
```diff
@@ -33,10 +36,34 @@ function shrinkwrap (args, silent, cb) {
   npm.config.set('production', true)
   readPackageTree(dir, andRecalculateMetadata(iferr(cb, function (tree) {
     var pkginfo = treeToShrinkwrap(tree, !!npm.config.get('dev') || /^dev(elopment)?$/.test(npm.config.get('also')))
-    shrinkwrap_(pkginfo, silent, cb)
+
+    readPackage(iferr(cb, function (data) {
+      chain([
+        [lifecycle, data, 'preshrinkwrap'],
```
iarna: Given our experiences with other lifecycles, I wonder if `preshrinkwrap` shouldn't run _before_ we do the `readPackageTree` (this would mean reading the package.json prior to calling it. This would allow `preshrinkwrap` lifecycles to modify the tree, which might be desirable for some folks.

So what I'm suggesting is:

1. lifecycle: preshrinkwrap
2. readPackageTree
3. lifecycle: shrinkwrap
4. _shrinkwrap
5. lifecycle: postshrinkwrap

lib/shrinkwrap.js
SimenB: ahhh, ok, now I follow. Thanks!

lib/shrinkwrap.js
SimenB: ah, ok, now I'm with you 😄

SimenB:
@zkat @iarna should be good now! I think... 😆

lib/shrinkwrap.js
```diff
@@ -31,10 +34,37 @@ function shrinkwrap (args, silent, cb) {
...
+          [lifecycle, tree.package, 'postshrinkwrap']
+        ], iferr(cb, function (data) {
+          cb(null, data[0])
+        }))
+      })))
+    })
+  }))
+}
+
+function readPackage (cb) {
```
iarna: Please use `read-package-json` instead of this. (And yeah, I see that `lib/version.js` _isn't_– we must have missed that during code review. =/) Using it would look like:
```
var readPackageJson = reaquire('read-package-json')

var packagePath = path.join(npm.localPrefix, 'package.json')
readPackageJson(packagePath, iferr(cb, function (data) {
…
}))
```

SimenB: done, thank you 😄

iarna:
LGTM! Adding to this week's release.

SimenB:
Latest push is just a typo fix

```patch
diff --git i/doc/cli/npm-shrinkwrap.md w/doc/cli/npm-shrinkwrap.md
index 96c3c42..6fc7d96 100644
--- i/doc/cli/npm-shrinkwrap.md
+++ w/doc/cli/npm-shrinkwrap.md
@@ -112,7 +112,7 @@ resolution using "version" if one isn't.
 If `preshrinkwrap`, `shrinkwrap` or `postshrinkwrap` are in the `scripts` property of the
 `package.json`, they will be executed by running `npm shrinkwrap`.
 `preshrinkwrap` and `shrinkwrap` are executed before the shrinkwrap, `postshrinkwrap` is
-executed afterwards. For example to run som postprocessing on the generated file:
+executed afterwards. For example to run some postprocessing on the generated file:
 
     "scripts": { "postshrinkwrap": "node fix-shrinkwrap.js" }
 ```

SimenB:
I don't know why this wasn't closed, but it's merged into master, so yay!

PR #12828, opened 2016-05-24
Don't try to reinstall linked dependencies that don't have a version specifier

schmod:
With this `package.json`:

``` js
var pkgj = {
  'name': 'pkg',
  'version': '1.2.3',
  'dependencies': {
    'dep': '../foo'
  }
}
```

If I have installed `dep` into `pkg`'s `node_modules` folder via `npm link`, running `npm install` will always cause npm to attempt to reinstall all of `dep`'s dependencies directly into `pkg`.

---

With this PR, if a dependency that does not have a version specifier is linked in, npm now assumes that the dependency is valid (because we have no good way of proving otherwise).  

Mitigates #10343

### DIFF
--- a/lib/install/deps.js
+++ b/lib/install/deps.js
@@ -65,6 +65,8 @@ function doesChildVersionMatch (child, requested, requestor) {
   // we always consider deps provided by a shrinkwrap as "correct" or else
   // we'll subvert them if they're intentionally "invalid"
   if (child.parent === requestor && child.fromShrinkwrap) return true
+  // symlinked packages without a version specifier should also be assumed to be "correct"
+  if (child.parent === requestor && child.isLink && !registryTypes[requested.type]) return true
   // ranges of * ALWAYS count as a match, because when downloading we allow
   // prereleases to match * if there are ONLY prereleases
   if (requested.spec === '*') return true
--- a/test/tap/ignore-install-link.js
+++ b/test/tap/ignore-install-link.js
@@ -30,7 +30,7 @@ var myreg = require('http').createServer(function (q, s) {
   s.end(JSON.stringify({'error': 'forbidden'}) + '\n')
 }).listen(common.port)
 
-test('setup', function (t) {
+function setup (cb) {
   rimraf.sync(root)
   mkdirp.sync(root)
   mkdirp.sync(path.resolve(pkg, 'node_modules'))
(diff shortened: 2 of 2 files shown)


lib/install/deps.js
```diff
@@ -65,6 +65,8 @@ function doesChildVersionMatch (child, requested, requestor) {
   // we always consider deps provided by a shrinkwrap as "correct" or else
   // we'll subvert them if they're intentionally "invalid"
   if (child.parent === requestor && child.fromShrinkwrap) return true
+  // symlinked packages without a version specifier should also be assumed to be "correct"
+  if (child.parent === requestor && child.isLink && !registryTypes[requested.type]) return true
```
schmod: Point of discussion:  If my dependency *does* have a version specifier in `package.json`, we'll still traverse it if the versions do not match.

Should the presence of a symlinked package take precedence over whatever is defined in package.json? 

---

I haven't been able to reproduce the behavior described in #10343 with a versioned dependency, but if a conflict exists, we print a warning, as well as a _ton_ of spurious warning messages about redundant dependencies existing inside of the symlinked module.  

For instance, if I `npm link` `lodash@4.0.0` into a package that asks for `lodash@2.0.0` in its package.json, I'll get several hundred messages along the lines of:

```
npm WARN skippingAction Module is inside a symlinked module: not running remove fs-extra@0.30.0 node_modules/lodash/node_modules/fs-extra
npm WARN skippingAction Module is inside a symlinked module: not running remove fileset@0.2.1 node_modules/lodash/node_modules/fileset
npm WARN skippingAction Module is inside a symlinked module: not running remove xmlbuilder@3.1.0 node_modules/lodash/node_modules/xmlbuilder
...
npm WARN update-linked node_modules/lodash needs updating to 2.0.0 from 4.13.2-pre but we can't, as it's a symlink
```

Removing the `&& !registryTypes[requested.type]` from the conditional above eliminates these messages, which is _probably_ what we want here.  Right?

iarna:
So, I'm concerned by this test as it asserts that an install with a broken local dependency should succeed. I don't think that's what we want.

I'm curious if the changes in https://github.com/npm/npm/pull/12775 resolve your larger issue?

iarna:
I should note though that local dependencies (with a path specifier) are not currently compatible with symlinked deps. I would be very interesting in understanding what folks use cases are when using _both_.

They don't work together as npm has no way to confirm that a symlinked dep is equivalent to a local specifier (it does this for local deps ordinarily by looking at the _requested field in the installed copy, but this is unavailable for linked deps). Regardless, it shouldn't be removing or moving anything from inside the other dep  (as in some of the examples #10343) and _that_ behavior should largely be fixed by https://github.com/npm/npm/pull/12775.

iarna:
I've been thinking about this and I'm coming around to the idea that symlinked dependencies should just _always_ count as a match– if you symlink something in, then maybe we should just consider a name match to count as ok?

schmod:
Oops.  Apparently I started writing a reply to your comment from 4 days ago, and never got around to posting it...

My use-case centers around symlinked SCM/Github dependencies, which are a little less far-fetched than symlink FS dependencies.  

My $0.02 is that `npm link` basically says "I know what I'm doing," and trusts that you've linked the correct dependency in.  If we have access to a version specifier, it seems sensible to log a `WARN` if the versions don't match, but NPM shouldn't try to correct any discrepancies on its own.

On the other hand, I think we need to definitively answer whether or not NPM should try to resolve/install the transitive dependencies of linked modules.

schmod:
Any progress reviewing this?

iarna:
> My $0.02 is that npm link basically says "I know what I'm doing," and trusts that you've linked the correct dependency in. If we have access to a version specifier, it seems sensible to log a WARN if the versions don't match, but NPM shouldn't try to correct any discrepancies on its own.

I've come around to this. Flag it as `invalid` in `npm ls` but leave it be. (the former should happen automatically).

> On the other hand, I think we need to definitively answer whether or not NPM should try to resolve/install the transitive dependencies of linked modules.

This is a good question… our current behavior is this:

If we can resolve a missing transitive dep by installing something higher in the tree than the link, then we do that. If the transitive dep would have to be installed in the linked module then we refuse to do it.

I think I'd prefer, for the simplicity of both implementation and explaining this to folks, to make symlinked modules entirely opaque. Set the expectation that they should have their own complete set of dependencies and we're not going to try to install them for you.

iarna:
@othiym23 and I have discussed making symlinks more opaque before. My tentative plan was twofold:

1. Update `read-package-tree` to stop walking into symlinks– when it hits one it'd include it but not recurse into its children.
2. Update `install/deps.js` to not recurse into children of symlinks / worry about their deps when flagging missing/invalid dependencies.

schmod:
The `install/deps.js` changes are probably satisfied by #12865

iarna:
This is no longer relevant is because `file:` specifiers now create symlinks.

PR #12843, opened 2016-05-25
Fix unreadable text in `npm outdated` using solarized dark theme

### LINKED ISSUE #12842: NPM outdated 'dir' column unreadable with solarized dark theme
I didn't even realize there was anything in the Location column until I accidentally highlighted:

screen shot 2016-05-25 at 5 37 52 pm

I vote we complete the color wheel triad with cyan:

screen shot 2016-05-25 at 5 39 23 pm

tribou:
Fixes #12842

### DIFF
--- a/lib/outdated.js
+++ b/lib/outdated.js
@@ -142,8 +142,6 @@ function makePretty (p) {
     columns[0] = color[has === want || want === 'linked' ? 'yellow' : 'red'](columns[0]) // dep
     columns[2] = color.green(columns[2]) // want
     columns[3] = color.magenta(columns[3]) // latest
-    columns[4] = color.brightBlack(columns[4]) // dir
-    if (long) columns[5] = color.brightBlack(columns[5]) // type
   }
 
   return columns


othiym23:
This isn't something we can test in an automated way, so somebody is going to need to eyeball this in a variety of terminals / a variety of OSes to ensure that making this change doesn't cause problems for somebody _else_.

zkat:
Because any particular color usage is potentially problematic across terminals, and in the interest of not needing to do all these manual tests and guaranteeing maximum readability, it might be better off to simply omit these two color settings altogether: The column doesn't really need any color coding and doesn't really gain anything from setting any color. By not setting a color, it will simply use the default text color, which will be the same one used for the `Current` column, which we already assume works reasonably well. How about that change?

tribou:
Sure, that works well for me.  Here's the updated look on solarized dark:

screen shot 2016-06-01 at 9 42 11 pm

iarna:
LGTM! I'm adding this to this weeks release.

iarna:
Landed in 3.10.0!

PR #12845, opened 2016-05-26
Fix prograss bar section name when packing

iarna:
Without this you get `pack:[object Object]`.

**r: @zkat**


zkat:
LGTM

zkat:
This was included in `npm@3.9.4`. Thanks! 🎉

PR #12846, opened 2016-05-26
Report missing dev deps when you ask for a listing of dev deps.

iarna:
This allows `ls` to, when asked to show dev dependences, eg, `npm ls --dev`, complain about missing dev deps.

New functionality, it will need a test. We may want to take a moment to evaluate the UX for this too– `npm ls` filters these as you'd anticipate, but at the moment it reports these as just `UNMET DEPENDENCY` and I would be inclined to make that `UNMET DEV DEPENDENCY`, mirroring what we do with optional dependencies.

### DIFF
--- a/lib/install/mutate-into-logical-tree.js
+++ b/lib/install/mutate-into-logical-tree.js
@@ -70,18 +70,27 @@ function translateTree_ (tree, seen) {
   tree.children.forEach(function (child) {
     pkg.dependencies[moduleName(child)] = translateTree_(child, seen)
   })
-  Object.keys(tree.missingDeps).forEach(function (name) {
+
+  function markMissing (name, requiredBy) {
     if (pkg.dependencies[name]) {
+      if (pkg.dependencies[name].missing) return
       pkg.dependencies[name].invalid = true
       pkg.dependencies[name].realName = name
       pkg.dependencies[name].extraneous = false
     } else {
       pkg.dependencies[name] = {
-        requiredBy: tree.missingDeps[name],
+        requiredBy: requiredBy,
         missing: true,
         optional: !!pkg.optionalDependencies[name]
       }
     }
+  }
+
+  Object.keys(tree.missingDeps).forEach(function (name) {
+    markMissing(name, tree.missingDeps[name])
+  })
+  Object.keys(tree.missingDevDeps).forEach(function (name) {
+    markMissing(name, tree.missingDevDeps[name])
   })
   var checkForMissingPeers = (tree.parent ? [] : [tree]).concat(tree.children)
   checkForMissingPeers.filter(function (child) {


lib/install/mutate-into-logical-tree.js
```diff
@@ -70,18 +70,27 @@ function translateTree_ (tree, seen) {
   tree.children.forEach(function (child) {
     pkg.dependencies[moduleName(child)] = translateTree_(child, seen)
   })
-  Object.keys(tree.missingDeps).forEach(function (name) {
+
+  function markMissing (name, requiredBy) {
     if (pkg.dependencies[name]) {
+      if (pkg.dependencies[name].missing) return
```
othiym23: Forgive me for a question to which I feel I should already know the answer, but does the logical tree conflate `dependencies` and `devDependencies`? I get why there are two passes for `tree.missingDeps` and `tree.missingDevDeps` now, but I find both being flattened down into a single dependency list here confusing.

othiym23:
> I would be inclined to make that `UNMET DEV DEPENDENCY`, mirroring what we do with optional dependencies.

I agree. This is a small change, but I think developers really appreciate more explicit errors.

lib/install/mutate-into-logical-tree.js
```diff
@@ -70,18 +70,27 @@ function translateTree_ (tree, seen) {
...
       pkg.dependencies[name].invalid = true
```
othiym23: Are the `invalid`, `realName`, `extraneous`, and `missing` values documented anywhere? More generally, is there some kind of schema for what properties can be attached to a package in the logical tree? I did a fair amount of searching even to discover how `translateTree` is meant to be used, and even now am a little confused as to what distinction to draw between `invalid` and `extraneous` or `missing`.

othiym23:
With a test and maybe a few tweaks to the output along the lines you propose, I'm +1 on this change.

lib/install/mutate-into-logical-tree.js
iarna: `translateTree` is used to convert a npm@3 type tree into a `read-installed` type tree, so it's all based around the semantics that `read-installed` provides. It _is_ confusing. =/

lib/install/mutate-into-logical-tree.js
iarna: As I say above, these are all `read-installed`'s thing. IMO, the fix for this will be to rewrite `npm ls` to use the `

`invalid` is a dep whose version is incompatible with your package.json.
`extraneous` is a dep that's not in your package.json.
`missing` is a dep in your package.json that's not on disk.

I did a bit of digging for what `realName` is supposed to be. It's non-functional in npm@3 (`read-package-tree` does not currently make this information available), but in npm@2 it was used to report instances where the directory name in `node_modules` didn't match the name in `package.json`. `realName` would be the name of the directory.

In npm@2 this presented this way:
```
invalid@1.0.0 /Users/rebecca/code/npmtest/invalid
└── packagename@1.0.0 (dirname) invalid

npm ERR! invalid: packagename@1.0.0 /Users/rebecca/code/npmtest/invalid/node_modules/dirname
```
In npm@3 that looks like: 
```
invalid@1.0.0 /Users/rebecca/code/npmtest/invalid
├── UNMET DEPENDENCY dirname@1.0.0
└── packagename@1.0.0 extraneous

npm ERR! missing: dirname@1.0.0, required by invalid@1.0.0
npm ERR! extraneous: packagename@1.0.0 /Users/rebecca/code/npmtest/invalid/node_modules/dirname
```

iarna:
So it turns out this was primarily about fixing `test/tap/install-with-dev-dep-duplicate.js` with #12775, sooo, I've folded this code back in there and I'm going to close this PR.

Issue #11911, opened 2016-03-15
`npm ls` (list) sometimes overwrites package devDependencies info with linked dep's


jmm:
Say you have:
- an app called `app`
- `app` has a devDep on `dev-dep@1.0.0`
- `app` depends on `local-dep` via `npm link` or otherwise via symlink
- `local-dep` has a devDep on `dev-dep@2.0.0`

In some cases `npm ls` (in `app`'s root dir) will display the info for `local-dep`'s dependency on `dev-dep` instead of `app`'s. It's non-deterministic in the sense that it depends on the name of the linked package (e.g. results may be different if it's `a-local-dep` vs `z-local-dep`). If it comes after the name of the main package in the sort implemented by `npm`, then the linked package's devDep information will overwrite the main package's in the output.

With `node@5.6.0`, `npm@3.8.1`, on Ubuntu 12.

[Minimal repro](https://github.com/jmm/npm-ls-linked-dep-dev-dep-overwrite-repro).

Key action is happening here:
- [mutate-into-logical-tree.js#L27-L35](https://github.com/npm/npm/blob/580f88a8183375a9c045529f10bf1aaea75f7bdd/lib/install/mutate-into-logical-tree.js#L27-L35)
- [mutate-into-logical-tree.js#L71](https://github.com/npm/npm/blob/580f88a8183375a9c045529f10bf1aaea75f7bdd/lib/install/mutate-into-logical-tree.js#L71)

schmod:
I'm not sure if this is the same issue, but I've encountered similarly-spurious behavior with `npm ls`, linked packages, and devDependencies.

In particular, it seems like `npm ls` will *ALWAYS* show any installed devDependencies of a linked package in your tree.

So, if you have something like this:

```
pkg
├── async@2.0.0 (symlink)
└── underscore@1.6.0 (DEP)
```

```
$ (cd ../async && npm install --prod)
$ npm install
$ npm ls

us-link@1.0.0 /Users/schmod/source/us-link
├─┬ async@2.0.0-rc.5 -> /Users/schmod/npm/lib/node_modules/async
│ └── lodash@4.13.1
└── underscore@1.6.0
```

This seems fairly reasonable.  I personally don't think that I should be seeing the transitive dependencies of async, but this isn't exactly _unexpected_ either.  

---

I wonder what happens if I install `async`'s devDeps?

```
$ (cd ../async && npm install)
$ npm ls

us-link@1.0.0 /Users/schmod/source/us-link
[...]

PR #12847, opened 2016-05-26
When shrinkwrapping dev dependencies, missing dev dependencies should be an error

iarna:
As with the changes to `ls`, if you are explicitly are shrinkwrapping dev dependencies then it's reasonable to insist that they all be installed.

### DIFF
--- a/lib/shrinkwrap.js
+++ b/lib/shrinkwrap.js
@@ -91,6 +91,18 @@ function shrinkwrapDeps (dev, problems, deps, tree, seen) {
         (topname ? ', required by ' + topname : ''))
     }
   })
+  if (dev) {
+    Object.keys(tree.missingDevDeps).forEach(function (name) {
+      var invalid = tree.children.filter(function (dep) { return moduleName(dep) === name })[0]
+      if (invalid) {
+        problems.push('invalid dev dep: have ' + invalid.package._id + ' (expected: ' + tree.missingDevDeps[name] + ') ' + invalid.path)
+      } else if (!tree.package.optionalDependencies || !tree.package.optionalDependencies[name]) {
+        var topname = packageId(tree)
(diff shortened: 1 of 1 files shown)


lib/shrinkwrap.js
```diff
@@ -74,6 +74,18 @@ function shrinkwrapDeps (dev, problems, deps, tree, seen) {
         (topname ? ', required by ' + topname : ''))
     }
   })
+  if (dev) {
+    Object.keys(tree.missingDevDeps).forEach(function (name) {
```
othiym23: Can this be extracted into a function that can be used for both `dependencies` and `devDependencies`, similar to what you did for #12846? AFAICT, the only difference is that one says "`invalid:`", and the other says, "`invalid dev dep:`".

othiym23:
> if you are explicitly are shrinkwrapping dev dependencies then it's reasonable to insist that they all be installed.

Agreed.

lib/shrinkwrap.js
iarna: looking at this code it seems like there's an additional difference…

lib/shrinkwrap.js
```diff
@@ -74,6 +74,18 @@ function shrinkwrapDeps (dev, problems, deps, tree, seen) {
...
+      var invalid = tree.children.filter(function (dep) { return moduleName(dep) === name })[0]
+      if (invalid) {
+        problems.push('invalid dev dep: have ' + invalid.package._id + ' (expected: ' + tree.missingDevDeps[name] + ') ' + invalid.path)
+      } else if (!tree.package.optionalDependencies || !tree.package.optionalDependencies[name]) {
```
iarna: Shouldn't be checking optional dependencies, since dev deps can't be optional.

schmod:
There's some discussion in #11911 about how (or if) this should work with linked packages.

PR #12849, opened 2016-05-26
Rewrite how recalculateMetadata flags missing dependencies

iarna:
We keep track of missing dependencies and missing dev dependencies separately. This allows things like `npm ls` to make decisions about what it wants to warn about.

Despite having concrete lists of dependencies and dev dependencies we were guessing as to which was which. This _usually_ works fine, but in cases where dependencies and dev dependencies share a requirement but not precisely the same version, it could lead to misleading error messages.

Further, things in the shrinkwrap that are not in the package.json were not being reported at all.

This PR includes three things:
1. It rewrites `markDeps` to take an object with name, spec & kind (`prod` or `dev`). Previously it took a string of `name@spec` and extracted name & spec from that.
2. When a dependency is both a prod and a dev dep, it will be correctly categorized. Previously if the dev spec didn't match it would still be flagged as a prod mismatch.
3. Reminders that we don't want to call recalculateMetadata quite as often as we currently are.

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -363,6 +363,7 @@ Installer.prototype.loadIdealTree = function (cb) {
     [this, this.loadAllDepsIntoIdealTree],
     [this, this.finishTracker, 'loadAllDepsIntoIdealTree'],
 
+    // TODO: Remove this (should no longer be necessary, instead counter productive)
     [this, function (next) { recalculateMetadata(this.idealTree, log, next) }],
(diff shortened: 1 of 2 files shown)


iarna:
This, it turns out, was not just fixing some obscure corner cases, but fixing ones that came up after we ditched using _requiredBy and friends. I'm closing this as I've folded it back into #12775

Issue #5082, opened 2014-04-15
npm publish missing index.js


pgilad:
Happened to me twice already, while publishing a package I wrote (https://github.com/pgilad/gulp-angular-htmlify if it matters), the package was published without `index.js`.

I had no notice or info that anything went wrong during `npm publish` phase. But then got notified by a user that the package wasn't installing with `index.js`.

The only thing that I can think of that caused this, other than a bug, is that I had the project open in Vim, that may have caused a failure to publish the file.

Did anyone run into a similar case? Is there anyway to verify the package contents other than installing it as a dependency and seeing what's installed?

This happened with:
git version 1.8.3.2
npm version 1.4.6
node version 0.10.26
ubuntu 13.10

sbull:
I've run into this problem a few times (most recently a few weeks ago), and I've written a test script to try to reproduce the problem:

https://gist.github.com/sbull/ccf7e53a16755f5c7ecc

If nothing else, you could use something like this to determine if `npm pack` bundled all of the files you wanted in your module before publishing it.

briancavalier:
We just had a similar issue happen in cujojs/when#424, but for a file other than index.js.  AFAICT, nothing about our package layout, package.json, or npmignore changed between 3.7.0 and 3.7.1, but the file `poll.js` seems to have been silently omitted by `npm publish`.  If you `npm install when@3.7.0`, poll.js will be there, `npm install when@3.7.1` it isn't, and `npm install when@3.7.2` it is.

I was using a fairly outdated version of npm when I published 3.7.1, so I'm not sure if that could be related:

```
❯ npm --version
1.4.28
❯ node --version
v0.10.33
```

othiym23:
@briancavalier thanks for the report! This is sounding more and more like a race condition in `fstream-npm` or `node-tar`. I am incredibly short on time right now, but I'll try to set some time aside to come up with some tests that might force this to occur.

(`npm@1.4.28` has a bunch of race conditions that are remedied in `npm@2`, but I don't think much of the code around publishing has changed.)

peter-mouland:
The first time I haven't run `npm pack` in a long time and it happened again :(

Affected package: https://www.npmjs.com/package/component-helper 
npm: 2.1.11
node: v0.10.35

ljharb:
[...]

PR #12862, opened 2016-05-26
.npmignore and package.json `files` directive are broken

Yeti-or:
I've found several issues that could be proven with unit tests:
Firstly documentation says: https://docs.npmjs.com/files/package.json#files

```
Certain files are always included, regardless of settings:

package.json
README (and its variants)
CHANGELOG (and its variants)
LICENSE / LICENCE
```

but they are not:
- README could be ignored with .npmignore
- README not added if it is not included in `files` directive
- And .npmignore breaks when we add `README` in it

If somebody show me where logic of .npmignore and `files` directive are. I can fix it.

### DIFF
--- a/test/tap/files-and-ignores.bugs.js
+++ b/test/tap/files-and-ignores.bugs.js
@@ -1,0 +1,152 @@
+'use strict'
+var test = require('tap').test
+var common = require('../common-tap.js')
+var path = require('path')
+var rimraf = require('rimraf')
+var mkdirp = require('mkdirp')
+var fs = require('graceful-fs')
+var tar = require('tar')
+var zlib = require('zlib')
+var basepath = path.resolve(__dirname, path.basename(__filename, '.js'))
+var fixturepath = path.resolve(basepath, 'npm-test-files')
+var targetpath = path.resolve(basepath, 'target')
+var Tacks = require('tacks')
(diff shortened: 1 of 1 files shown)


zkat:
Great! Thanks for finding this and pointing it out: The main place where you would be able to fix stuff like this is in https://github.com/npm/fstream-npm, which is what handles the bulk of the npm-specific ignore logic.

I'd tell you to go ahead and fix it right now and help you through it, but there's some notes on touching that code right this minute: @othiym23 is currently working on another bug involving that code, and one of the very real possibilities is that we might throw away the way files/ignores are handled under the hood and write most of that code over again.

I'm gonna loop Forrest in, and let him make a call about how to move forward with this issue. Cheers!

It would also be super-helpful if you (separately) made a regular issue describing so we can track this as a `bug` separately from the PR, but I'll leave this open. Thanks again!

Yeti-or:
@zkat I've created three separate issues:
 #12875  #12876 #12877 
Could you point me to branch/pr/issue where @othiym23 is working on this?

othiym23:
I think these tests are another manifestation of #5082 under Node 6+, and as such I'm going to close this PR in favor of solving the underlying issue. See #5082 and its linked issues for deeper discussion.

Issue #9519, opened 2015-09-08
Make linked modules opaque, rather than merely immutable


iarna:
Make the dependency recursurer (`lib/install/deps.js`:`andForEachDeps`) not look at the children of linked modules.

This brings two benefits:

First it means that we won't waste time looking up dependencies inside linked deps, that should already be complete unto themselves.

Second, it ensures that we don't mess with those trees. Current behavior involves mutating the tree around linked deps, and then skipping actions that would impact them. This is error prone and complicated.

tellnes:
+1

After upgrading to 3x, npm started to mess up the dependencies for linked packages which breaks the installation when using those package standalone or linked to another project.

jewelsjacobs:
How about adding an npm config key / value you can put in the package.json of a package to indicate not to allow dependencies to be built from a parent, with true as a default? I.e.

    "build_by_parent": false

tellnes:
This is actually a blocker for me upgrading to 3x.

PR #12865, opened 2016-05-26
npm ls & friends shouldn't introspect into linked packages

schmod:
This prevents `npm ls` from listing the dependencies of linked packages, and prevents a similar (and useless) traversal from running when we run commands like `npm install` and `npm update`.

This represents a behavior-change, although the existing behavior is confusing (and often buggy) in the most-common use-case, where the linked package's devDependencies are installed, which isn't something we typically expect to see in our dependencies.

I believe that _any_ decisions or inferences that we make about how to traverse linked packages (and what to print in `npm ls`) is going to lead to ambiguous or confusing behavior.  We already don't traverse the children of extraneous packages, and I think that it makes the most sense to extend that behavior to linked packages as well.

See #11911

### DIFF
--- a/lib/install/deps.js
+++ b/lib/install/deps.js
@@ -101,6 +101,7 @@ function recalculateMetadata (tree, log, seen, next) {
   validate('OOOF', arguments)
   if (seen[tree.path]) return next()
   seen[tree.path] = true
+  if (tree.isLink) return next()
   if (tree.parent == null) resetMetadata(tree)
   function markDeps (spec, done) {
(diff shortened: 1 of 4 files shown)


othiym23:
@iarna, can you take a look at this / provide feedback?

polpo:
Would this solve #9519?

iarna:
I'm going to close this as it's now obsolete as linked modules are now first class with the new `file:` specifier behavior. We do in-fact manage deps inside linked modules.

PR #12866, opened 2016-05-26
save to bundleDependencies not bundledDependencies

iarna:
`normalize-package-json` and everything else expects the standard version to be `bundleDependencies`, `bundledDependencies` to be the alias.

I apparently misread this somewhere along the line and had been writing out `bundledDependencies` to the `package.json` files in `node_modules` folders. Of course, when we read them back in that gets transformed back to `bundleDependencies` so that was _harmless_, if a bit weird.

This patch corrects the weirdness.


iarna:
We did this elsewhere as a part of the run up to npm@5.

PR #12926, opened 2016-06-01
Remove two unused dependencies

mmalecki:
I was idly looking at the source and was surprised that we use those two, so I `ack`'d and it seems like we don't, so here's a PR.


zkat:
LGTM!

zkat:
This was included in `npm@3.9.6`. Thanks a bunch! 🎉

PR #12928, opened 2016-06-01
Change error message so that npm does not advise users to run as root

breathe:


lib/utils/error-message.js
```diff
@@ -26,7 +26,7 @@ function errorMessage (er) {
     case 'EACCES':
     case 'EPERM':
       short.push(['', er])
-      detail.push(['', ['\nPlease try running this command again as root/Administrator.'
+      detail.push(['', ['\nPlease adjust file ownership or permissions -- or run this command again as a different user.'
```
othiym23: > run this command again as a different user

This needs to be worded better – as such, it turns a clear, explicit action item (which a number of people disagree with) into something confusingly vague.

EmmaRamirez: What about something along the lines of

> Please try adjusting file ownership or permissions, or running this as a user with root permissions.

Or maybe better yet, mentioning a couple other alternatives, like using a node version manager or creating a user dedicated directory.

othiym23:
This issue hasn't been updated for a month and a half, and needs changes to be merged. As such, I'm going to close it. Feel free to submit a new patch that includes the feedback from this PR (and maybe includes some tweaks to the docs to reflect the change in emphasis here – just this change on its own is likely to result in a bunch of support requests from confused users). Thanks for your time and effort!

RyanNerd:
I've been developing software for nearly 30 years. Switched from Windows to Linux and as such needed to get NPM up and running. This _suggestion_ bit me hard in the butt with Ember. Ember uses NPM when making web apps great. When making add-ons NPM pitched a fit and suggested running as root.

This _worked_ but **all** my  files and directories were owned and had permissions as ` root`. I'm decent enough with Linux shell that I could fix this. Problem is my command line fixes didn't apply to hidden files and directories such as .git. What a mess.

PR #12929, opened 2016-06-02
Release 3.10.0

zkat:


zkat:
Turns out a feature that was gonna bump the minor isn't really going in this week. This has been replaced by #12940.

PR #12933, opened 2016-06-02
Fixed ignored by default files on publication

Jokero:
The same as on https://docs.npmjs.com/misc/developers

### DIFF
--- a/doc/files/package.json.md
+++ b/doc/files/package.json.md
@@ -202,6 +202,8 @@ Conversely, some files are always ignored:
 * `npm-debug.log`
 * `.npmrc`
 * `node_modules`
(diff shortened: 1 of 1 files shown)


doc/files/package.json.md
```diff
@@ -202,6 +202,8 @@ Conversely, some files are always ignored:
 * `npm-debug.log`
 * `.npmrc`
 * `node_modules`
+* `.npmrc`
```
KenanY: `.npmrc` is already on this list

Jokero: Yeah, you are right. As I see this page last update was made on January 29, 2016. Should be deployed

PR #12936, opened 2016-06-02
added --preid option for `npm version`

### LINKED ISSUE #12609: Add `--preid` to `version` command
Add `--preid` parameter from https://github.com/npm/node-semver#prerelease-identifiers

or second optional parameter for `npm version`:

```
npm version [<newversion> | major | minor | patch | premajor | preminor | prepatch | prerelease | from-git] [alpha | beta | rc | dev]
```

command:

``` bash
npm version prepatch rc
```

should increment `1.0.1` to `1.0.2-rc.0`
similar by

``` bash
semver 1.0.1 -i prepatch --preid rc
```

spudly:
Fixes #12609


spudly:
Not sure what to do about the CI test results here. Were these tests broken on master?

* appveyor failed on some tests (git-races) that are completely unrelated to my changes
* travis just ran out of memory while running the tests (can't say I'm surprised - that's a boatload of tests)

spudly:
Looks like all the latest pull requests are failing the same tests, so I'm going to just assume the tests are broken on master and I didn't break them. 😇

spudly:
It's been 28 days. When can I expect something to be done with this pull request?

othiym23:
The CLI team is still debating whether this is a change we want to land, and also has a decent backlog of pull requests to get through. I can't say exactly when we'll have time to get this reviewed in more detail and landed / rejected, but it's my responsibility to get back to people with feedback on PRs, and I'm sorry for the delay. Thanks for your time and patience!

spudly:
> The CLI team is still debating whether this is a change we want to land

Are there any legitimate concerns preventing this pull request from being merged? If so, please post them! Personally, I doubt you've even looked at it. I feel like this pull request has been completely ignored.

> and also has a decent backlog of pull requests to get through. I can't say exactly when we'll have time to get this reviewed in more detail and landed / rejected

It's been 2 months! How long do I have to wait? If you expect to ever get through your backlog, this one is definitely low-hanging fruit. The change could not be more simple. The only change I made was to add an option that provides one additional parameter to a single function call. The rest is all tests and documentation.

spudly:
Crickets.... :sob:

davidagee:
bump, this is part of the semver spec and not having it in `npm version` requires using vendor dependencies like https://www.npmjs.com/package/npmversion, essentially making `npm version` useless for our prerelease workflows.

jbcpollak:
also there is a problem with line endings in npmversion which makes that workaround useless (or complicated) on OS X

othiym23:
After some further consideration, and discussion within the team, I've decided that this isn't a change the CLI team is going to land. Working with prerelease versions is tricky, and this feature is enough of an edge case that I think it increases, rather than decreases, the number of opportunities for prerelease versions to be a footgun.

My concern about footguns is connected to the decision, in `semver@^4`, to make prerelease versions fall outside the range matching for `~` and `^`. Were `semver`'s behavior to change, it might make more sense to have a way to opt into (and then increment within) a given prerelease identifier. After watching the travails of the React community in using prerelease versions with `peerDependencies` and other interrelated suites of packages, it's pretty clear that the current behavior of semver with respect to prerelease versions is of pretty limited usefulness. That means there's a substantial possibility that that behavior may change.

As it stands, though, I think users who want to use prerelease versions are better off using third-party tools like `npmversion`, or build scripts that use `npm version "x.y.z-prealpha.2"` (or whatever suits your use case) instead. Thanks for your time, and my apologies for the delay!

spudly:
Wow. That was unexpected. NPM doesn't support semver? really?

othiym23:
> NPM doesn't support semver? really?

npm is probably the most conformant package manager out there when it comes to its adherence to the SemVer 2 spec. That's not the same as choosing to provide affordances for all the pieces of the spec in its tooling, in much the same way that not all the pieces of `package.json` are set up by `npm init`. I'd really like to avoid introducing features that complicate how people understand tools like `npm version` that should be as simple as possible, and the current state of prerelease version IDs is such that, in my judgement, that's tough to do.

Is my explanation of our / my reasoning for rejecting this change otherwise unclear?

jbcpollak:
How do you suggest we publish feature branches in multi-package projects
for integration testing?

For example, say I'm working on the foo project, feature branch feature-a,
and I want to publish that so bar project feature-a can depend on it and do
integration testing without disrupting developers working on other features
in both projects?

othiym23:
> How do you suggest we publish feature branches in multi-package projects
for integration testing?

@jbcpollak, that feels orthogonal to SemVer, and to this PR.

I've learned through experience that SemVer isn't a super great tool for either topic-branch-based development (because of the constraints on how [prerelease release IDs](http://semver.org/#spec-item-9) work) or Maven / Ivy-style SNAPSHOT releases (thanks to SemVer's insistence that each release have a distinct release ID, and [build IDs](http://semver.org/#spec-item-10) being stripped prior to range matching). To me, this feels like SemVer defining a release as being a finished artifact rather than a work in progress.

The simplest way to do what you're trying to do is to just use the feature branches directly from hosted Git, e.g. `"package-under-development": "github:myorg/myrepo#my-feature-branch"`. There are still some rough edges around that workflow (which the team intends to address someday, but probably not in the immediate future), so I can understand not wanting to use it.

You _can_ use prerelease identifiers with a Makefile or small Node script to track these separate releases, in combination with using the `npm version <version>` directly. That, in combination with [`npm dist-tag`](https://docs.npmjs.com/cli/dist-tag), should make it possible to do things like `npm install my-package@next-foo-test` or `npm install my-package@next-bar-test`, which is similar to how the [npm release train](https://github.com/npm/npm/wiki/CLI-Team-Process#the-npm-release-train) works (note that this way we can support both `npm@lts-next` and `npm@next`, although we don't try to support feature prereleases, only separate release lines).

That said, were I in your shoes, I'd probably just use GitHub-based dependencies to test out new features, and publish to your registry of choice once a version is ready for final release. Mostly because it's much simpler.

jbcpollak:
We were using GitHub dependencies, but we are introducing build/compilation
steps (like Typescript) and I'd like to avoid checking the dist/build
directory into GiHub, although I know a lot of projects do just that, it
seems wrong to me.

I'll have to read up on the prerelease id handling you be mentioned.

PR #12324, opened 2016-04-12
doc: Describe how `npm run` sets `NODE` and `PATH` in more detail

addaleax:
Note that `npm run` changes `PATH` to include the current node interpreter’s directory.

Ref: #12318

PR #12968, opened 2016-06-05
run-script: avoid $PATH pollution if possible

### LINKED ISSUE #12318: npm run-script pollutes $PATH
@iarna @segrey the fix from #11175 caused a regression which causes a different executable to be run from inside an npm script than from outside. See #5876 for a description of the original issue. This causes problems when trying to run scripts that depend on the PATH variable, like when using rvm or virtualenv.

segrey:
Let's avoid prepending current node directory to the PATH if current node can be found in the PATH.
It will probably fix #12318

### DIFF
--- a/doc/cli/npm-run-script.md
+++ b/doc/cli/npm-run-script.md
@@ -41,8 +41,8 @@ you should write:
 instead of `"scripts": {"test": "node_modules/.bin/tap test/\*.js"}` to run your tests.
 
 `npm run` sets the `NODE` environment variable to the `node` executable with
-which `npm` is executed, and adds the directory within which it resides to the
-`PATH`, too.
+which `npm` is executed. Also, the directory within which it resides is added to the
+`PATH`, if the `node` executable is not in the `PATH`.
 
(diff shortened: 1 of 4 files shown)


segrey:
Seems the failed checks in this PR are not caused by these changes, because https://ci.appveyor.com/project/npm/npm/ and https://travis-ci.org/npm/npm have the same failed checks.
@othiym23 any ETA for review? Thanks.

iarna:
I'm tentatively ok with this. I would like to see some new documentation describing the behavior here. (We really should have insisted on documentation when this code path was first changed too.)

segrey:
Done, documentation provided by https://github.com/npm/npm/pull/12324 is extended with new behavior description. (thanks @addaleax for that PR!)

iarna:
Landed in 3.10.6!

segrey:
Thanks!

jrgm:
Will this be back-ported to 2.x, where #12318 broke builds on the LTS branch?

PR #12973, opened 2016-06-06
addRemoteTarball: reuse cached packages

misterbyrne:
When adding a package via a URL pointing to a tarball, we can try to use a
cached version by attempting to extract the version from the URL and performing
a cache lookup using the name and version. If the cached package was resolved
using the exact same URL then we can reuse it. This makes it safe for packages
added from other sources, or with unusual URL formats from which we cannot
extract the version.

When the cache is primed, this results in a noticeable speed improvement for shrinkwrapped installs (along with the obvious reduction in network traffic).

Happy to add test(s) if this looks good.

### DIFF
--- a/lib/cache/add-remote-tarball.js
+++ b/lib/cache/add-remote-tarball.js
@@ -12,8 +12,33 @@ var addLocalTarball = require('./add-local-tarball.js')
 var cacheFile = require('npm-cache-filename')
 var rimraf = require('rimraf')
 var pulseTillDone = require('../utils/pulse-till-done.js')
+var cachedPackageRoot = require('./cached-package-root.js')
+var readPackageJson = require('read-package-json')
 
 module.exports = addRemoteTarball
+
+function extractVersionFromUrl (url) {
+  var match = url.match(/(\d+\.\d+\.\d+(-.+)?)\.tgz(\?.*)?$/)
+  return match && match[1]
+}
+
+function findInCache (u, pkgData, shasum, cb) {
+  if (!pkgData.name) return cb()
+  var ver = extractVersionFromUrl(u)
+  if (!ver) return cb()
+
+  var pkgroot = cachedPackageRoot({ name: pkgData.name, version: ver })
+  var pkgjson = path.join(pkgroot, 'package', 'package.json')
+  readPackageJson(pkgjson, function (err, pkgData) {
+    if (err && err.code === 'ENOENT') return cb()
+    if (err) return cb(err)
+    if (pkgData._resolved === u) {
+      var pkgtgz = path.join(pkgroot, 'package.tgz')
+      return cb(null, pkgtgz, pkgData)
+    }
+    return cb()
+  })
+}
 
 function addRemoteTarball (u, pkgData, shasum, auth, cb_) {
   assert(typeof u === 'string', 'must have module URL')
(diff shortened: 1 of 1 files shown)


othiym23:
Including the version number in the tarball URL is a convention that's not formally specified anywhere in particular, and shouldn't be considered a normative requirement. I don't think @npm/registry-team has any plans to change that convention, but I'm still uncomfortable making it a load-bearing part of the cache resolution process. I'm weakly -1 on this change as-is as a result. I'd rather keep moving towards getting the cache content-addressable and including shasums in shrinkwrap files. 

That said, it's possible I'm letting the perfect be the enemy of the good. I'd like to get feedback from a few more people before deciding definitively what to do with this. @iarna, @zkat – thoughts?

misterbyrne:
Thanks. I understand that there's no URL spec - that's why it's a "best effort" kind of thing, failing if `_resolved` is not an exact match. I get your point about moving towards content-addressable cache/shrinkwraps, but I'm aiming for a stopgap which has a marked effect on large shrinkwrapped projects.

If taking an educated guess on the version number is out, then what about looping through all cached versions of that package, reading each version's package metadata and picking the one with the matching `_resolved` URL (if any)? I guess that this could end up being a performance regression in some edge cases (small package, extracts quickly, bunch of cached versions - although this could be addressed by having a maximum number of cached versions that would be considered).

iarna:
I would very much like to have something like this, as not having `addRemoteTarball` go via the cache is really, really unfortunate for shrinkwrap users. (If you delete the `resolved` field from your shrinkwrap then requests go via the cache, but it's also not as tightly locked down.)

Ideally, what I'd like to see is the fetch code here folded into [`lib/cache/caching-client.js`](https://github.com/npm/npm/blob/master/lib/cache/caching-client.js), which is what we use to cache metadata (the results of getting `https://registry.npmjs.com/packagename`). Something like that for tarballs would be fantastic.

For any solution we do need to still do a freshness check, for a few reasons:

1. While versions can't change in the npm registry, this isn't true for third party registries.
2. It's easy to get tarballs in your cache that don't match what's in the registry (`npm install -g .` for ex), but your check here `https://github.com/npm/npm/pull/12973/files` mostly resolves that concern.

Doing a freshness check means storing the etag from the previous request and passing that and the last modified time of the local tarball through, and then looking for 304s. If you look at the existing caching-client code you can see how it approaches this for metadata.

While this wouldn't speed things up as much as your existing patch does, it would still provide a substantial boost, given that 304s would save download time. Plus it would be able to take advantage of `cache-min` (and users who wanted the behavior you implemented here could set their `cache-min` to a high value).

othiym23:
@misterbyrne Are you still interested in working on this? Is there any other information we can provide to unblock you / make it easier for you to get this in shape to merge?

misterbyrne:
Hey - sorry I took so long to pick this up again.

My motivation for this isn't just to reduce bandwidth but to also improve install time.

In my experience the real slowdown is in validating the downloaded tarball, which happens outside the caching-client. After we have the data we call add-local-tarball, which needs to validate the package content by extracting the whole thing, and loading the package metadata from package.json, after which the extracted files are deleted.

Avoiding this cost seems like a bigger win to me (of course that is said with the privilege of a fast internet connection).

I think something which would address both performance and bandwidth would be to store the package sha in the shrinkwrap file and to introduce a means of addressing the package cache via sha. I've done up a PR to illustrate this: https://github.com/npm/npm/pull/13398

misterbyrne:
Closing in favour of the panacea of storing the SHA in the shrinkwrap.

Issue #9186, opened 2015-08-05
question: is the ability to run a dependency's scripts using `npm run-script <pkg> <script>` removed on purpose?


joakimbeng:
**Background:**

I was really glad when I found out that npm has (or had) the ability to run a dependency's scripts by running: `npm run-script <pkg> <script>`. Which is also stated in the [npm documentation](https://docs.npmjs.com/misc/scripts#description) as:

> Additionally, arbitrary scripts can be executed by running npm run-script <pkg> <stage>

When I tried: `npm init -y && npm i -S tap && npm run-script tap test`, it didn't work and I got this output:

```bash
npm ERR! Darwin 14.0.0
npm ERR! argv "node" "/usr/local/bin/npm" "run-script" "tap" "test"
npm ERR! node v0.12.7
npm ERR! npm  v3.2.1

npm ERR! missing script: tap
```

I then began tracing back when the feature was removed and have concluded that it has been so mistakenly.

**Related commits:**

* Taking a look at a commit from [run-script v.1.0](https://github.com/npm/npm/commit/03f782fe2978ddaae80fa5e094e9b319d2192066) one can see that the feature is there back then.
* When the [ability to pass arguments to scripts was added](https://github.com/npm/npm/commit/df4b0e7fc1abd9a54f98db75ec9e4d03d37d125b) the above mentioned feature was removed, even though the `runScripts.completion` still works for dependencies' scripts. Which feels to me like a mistake.

**Closure:**

Am I right that this was mistakenly removed and you would accept a PR for this? Or am I missing something and this feature should actually be removed?

boennemann:
npm scripts changed w/2.0.0.
You might want to try `npm explore tap -- npm test` now. You might have to install its devDependencies first though.

othiym23:
This is behavior that was changed in `npm@2` so that users could pass arguments to run-scripts, akthough I don't believe it was ever the intent that users be able to run arbitrary bins exported from dependencies (like @boennemann pointed out, that's what `npm explore <depname> -- <binname> <arguments>` is for). I'm pretty sure the current documentation is reasonably unambiguous on this, but if it's still confusing, we'd be happy to take a patch to clarify it. Aside from that, I don't think it would make sense to make more big changes to `npm run`, but do check out the discussion on [`npm-exec`](https://github.com/npm/npm/issues/6053), which sounds like exactly what you're looking for, with the teeny tiny caveat that it doesn't exist yet. ;)

joakimbeng:
[...]

PR #12983, opened 2016-06-07
doc: add workaround for running scripts of dependencies

Tapppi:
Add npm explore workaround for running scripts from a dependency.
Related to the removal of the earlier possibility of running these scripts with `npm run <pkg> <script>` (removed from this doc in https://github.com/npm/npm/commit/70a3ae4d4ec76b3ec51f00bf5261f1147829f9fe).

Workaround mentioned in https://github.com/npm/npm/issues/9186.
Ref https://github.com/npm/docs/issues/704.

### DIFF
--- a/doc/misc/npm-scripts.md
+++ b/doc/misc/npm-scripts.md
@@ -36,7 +36,8 @@ following scripts:
 Additionally, arbitrary scripts can be executed by running `npm
 run-script <stage>`. *Pre* and *post* commands with matching
 names will be run for those as well (e.g. `premyscript`, `myscript`,
-`postmyscript`).
+`postmyscript`). Scripts from dependencies can be run with `npm explore
+<pkg> -- npm run <stage>`.
 
 ## COMMON USES
 


Tapppi:
Maybe mention the workaround in cli/run-script.md as well?

ScottFreeCode:
https://docs.npmjs.com/misc/scripts currently says:
> Additionally, arbitrary scripts can be executed by running `npm run-script <pkg> <stage>`.

As far as I can tell this is inaccurate in the current version of npm as it always takes the first argument to be a script name. I was thinking of submitting an issue about this, not knowing whether the documentation is outdated or the functionality is bugged. That said, this PR suggests that it's the documentation that's wrong; if so, that page should be updated too.

Tapppi:
@ScottFreeCode oops, I messed up the reference formatting in the pull request, that was already addressed in https://github.com/npm/npm/commit/70a3ae4d4ec76b3ec51f00bf5261f1147829f9fe. This is just to add a workaround. The updated version has simply not been deployed yet. The issue I created about this in npm/docs is the other ref (https://github.com/npm/docs/issues/704).

ScottFreeCode:
Ah, cool; wonder why that didn't come up in my search... Oh well, nothing beats looking into a fix only to discover it's already done and on its way!

iarna:
LGTM! Scheduling this for this week.

iarna:
Landed in 3.10.0!

Tapppi:
Thanks!

Issue #12994, opened 2016-06-07
npm install --time-machine (ignore packages released after a given date)


jedwards1211:
Recently I've run into an issue where a new version of some transitive dependency of `babel-eslint` fails to parse Flow parameterized types. Downgrading `babel-eslint` didn't work because the newer versions of the transitive deps are semver-compatible with the older version of `babel-eslint`.

Since a fresh `npm install` worked fine several months ago, if there was a time machine-like feature whereby it would install exactly the same packages as it would have on a specific date several months ago, I should be able to get a working project setup again, because I would get all the older versions of all the transitive dependencies.

I can't think of any simpler way to get a working project setup, because I've lost a working copy with all the old transitive deps that worked.

DaveEmmerson:
@jedwards1211 This would be great... but given it's almost a year later, how did you solve your issue? I'm facing a similar problem and my first thought is to run the dependency list through npm to get versions and release dates (if that's possible) and run them through a script to filter out the newer ones.

jedwards1211:
I did solve it using this fork... You could use it temporarily to install what you need, then shrinkwrap and switch back to official npm.  Let me know if you need help installing/running this fork.

DaveEmmerson:
I don't see the fork in your list of repositories. I can see the other issue referenced and the changes on there, but the fork doesn't show. Any clue? I could try to pull npm from the date of your commits and just apply the changes, but if the fork is still around somewhere that would be easier :)

jedwards1211:
@DaveEmmerson ah sorry, here it is: https://github.com/jedwards1211/npm/tree/time-machine

DaveEmmerson:
Thanks. I managed to sort it, although I ended up using a script to
reinstall all the deps at the versions listed, just by removing the ^.

Hopefully the link helps someone else looking. Sorry to trouble you :)

Issue #7160, opened 2015-01-17
Feature like git-bisect to find broken upstream deps


nolanlawson:
In PouchDB, we frequently have our builds broken due to an upstream dependency introducing a bug in a patch/minor version (https://github.com/pouchdb/pouchdb/issues/2235).

Recently it's gotten so bad that we're actually starting to consider npm shrinkwrapping or committing the `node_modules` ([discussion on Twitter](https://twitter.com/daleharvey/status/550996955384786944)). We've also started locking down several deps instead of using the caret- or tilde-versioning. It's unfortunate, though, because this robs the larger community of a valuable "canary in the coal mine."

For projects that have a lot of dependencies and a lot of tests (and are perhaps better-tested than their own dependencies), it would be really nice if we had an easier way to diagnose breakages. I'm picturing some kind of feature like `git-bisect`, where npm would version-lock all dependencies and then unlock one dependency at a time, run a test, and then tell us which dependency broke the build.

It gets complicated when dependencies have had a lot of releases since the one declared in the `package.json`, but again this is where it would be nice if npm could automate that for us.

fschwiet:
A downside of the git-bisect type approach is that you might fix a lot of issues that don't correspond to a real deployment target.

It sounds your tests could have detected the problems.  Is it possible to have PouchDB customers run the tests to check their environment?  At some point PouchDB is included in a project that is deployed, checking in the node_modules folder.  Maybe the tests can be ran at that point.

othiym23:
I think this is the kind of truly useful edge case solution that would work best as an external module, perhaps called `npm-bisect`. It's too niche (and, realistically, too complex to support within the existing codebase) to put into the core CLI, but there are definitely times in the past when it would have saved my bacon. So, I'm rejecting this as a feature request for npm itself, but if you need help getting it built for real, I'd be happy to pitch in.

nolanlawson:
Cool cool, that sounds like a good way to proceed! :)

thadk:
:+1:

mroderick:
I think loose versioning of dependencies is the biggest design flaw of npm. shrinkwrap is a bandaid at best, not a real fix for the problem.

I would love to have such a tool at my disposal, as I have wasted many hours on several occasions, trying to figure out which dependency of which dependency of which dependency that got updated a broke something for everyone.

If anyone knows npm well enough to build such a tool, please start a Kickstarter, and I'll happily donate and market the campaign until I am blue in the face.

PR #12995, opened 2016-06-08
WIP: add --time-machine option (only install deps released on or before a given date)

jedwards1211:
See https://github.com/npm/npm/issues/12994

I'm absolutely sure this will need tests and more work, I'm putting it up now to get feedback.

With these changes I was able to get back a set of working deps and subdeps in my project and create the necessary `npm-shrinkwrap.json` for them.  Yay!

Right now I just filter the versions that `fetchPackageMetadata` and `addNamedRange` take into consideration.  Are there other parts of the code looking at package versions where I need to filter out the ones after the time machine date?

### DIFF
--- a/doc/cli/npm-install.md
+++ b/doc/cli/npm-install.md
@@ -271,6 +271,12 @@ global `node_modules` folder. Only your direct dependencies will show in
 `node_modules` and everything they depend on will be flattened in their
 `node_modules` folders. This obviously will eliminate some deduping.
 
+The `--ignore-versions-after=date` will cause npm to ignore versions of any package
+released after the given date.  This is a useful workaround for bugs in recent releases
+of subdeps that you don't have controlled by a shrinkwrap.  `date` may be anything
+that the `Date` constructor would parse correctly; if it is invalid, this option will
+have no effect.
+
 The `--ignore-scripts` argument will cause npm to not execute any 
 scripts defined in the package.json. See `npm-scripts(7)`.
 
--- a/doc/cli/npm-install.md
+++ b/doc/cli/npm-install.md
@@ -271,7 +271,7 @@ global `node_modules` folder. Only your direct dependencies will show in
 `node_modules` and everything they depend on will be flattened in their
 `node_modules` folders. This obviously will eliminate some deduping.
 
-The `--ignore-versions-after=date` will cause npm to ignore versions of any package
+The `--ignore-versions-after=date` argument will cause npm to ignore versions of any package
 released after the given date.  This is a useful workaround for bugs in recent releases
 of subdeps that you don't have controlled by a shrinkwrap.  `date` may be anything
 that the `Date` constructor would parse correctly; if it is invalid, this option will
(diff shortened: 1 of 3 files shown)


SimenB:
Oh, yes please! Awesome feature.

Would have saved me so much time at one point. Going between a machine with working output, and my own which did not, and comparing output of `npm ls` was annoying. At least we had a working version, so in the end we just generated a shrinkwrap file on his machine, but if we did not have a machine with a working setup, this would have saved us.

jedwards1211:
@SimenB exactly, I had lost a *working* copy, and I thought, wait a minute, if I did a fresh `npm install` a week or two ago, I would get a working project, so why can't I do that now?

jedwards1211:
I'm guessing I probably shouldn't be handling the `--time-machine` option directly in `fetchPackageMetadata` and `addNamedRange` unless we can verify that it works properly with any `npm` command; to make it work only on `npm install`, I could have those functions use a differently-named field in `npm.config` that only `install.js` copies from `--time-machine`.

Would you guys prefer `--ignore-versions-after` for the argument name?

jedwards1211:
@SimenB I noticed in the roadmap that the npm team wants to achieve consistent installs, i.e. `npm install` always does the same thing if the deps haven't changed.  I'm guessing a solution to this problem that doesn't require time machine is possible then.

For instance: perhaps when installing a version of a package, `npm` should ignore subdep versions that were released after that version of the primary package was released (unless the subdep versions are in the shrinkwrap).  This would fail to get minor bugfixes on transitive deps, but if those are needed, one could then install later versions of those transitive deps and create a shrinkwrap.

SimenB:
I don't know if I'd like that as default behavior, I'd expect a fresh `npm install` to install newest semver compatible of all packages. If I want by some date, this option seems fair.

`ignore-versions-after` is probably a better name, more self-explanatory.

jedwards1211:
@SimenB maybe a `--reproducible` option for the per-subdep date restrictions I described above?

zetlen:
This would combine wonderfully with a bisecting utility.

jedwards1211:
@zetlen look who it is, I met you at Volusion!
What do you mean by bisecting?

SimenB:
Probably something like #7160

zetlen:
Oh hey @jedwards1211. Yeah, I mean a utility like `git bisect`, that starts from a broken state, zooms back to a known-good state, and binary searches the intermediate states for the breaking change. This feature wouldn't give you every possible intermediate state, but it would let you determine a smaller time window, and then we could do the expensive work of making and searching a complete set of tree states by sorting nodes on last release date. I expect this probably runs pretty slowly, but at least it runs by itself!

jedwards1211:
Ah, very cool, that would be neat.  At least that process could be parallelized, even though it would be pretty resource intensive.

jedwards1211:
@SimenB I agree that `--ignore-versions-after` would be a good name for the option

jedwards1211:
Hmmm, on my machine this seems to think no version is found unless you specify a specific version.

SimenB:
Any news here 😄

jedwards1211:
@SimenB I think they probably won't touch it again unless I implement tests and get CI to pass.  Which sadly I don't have time for right now, especially given how complicated testing it would be...

iarna:

PR #13089, opened 2016-06-17
tar: Restore our version of applyIgnores

iarna:
This reverts some of #11995.  Without this, bundled dependencies don't
include transitive dependencies.


othiym23:
Landed as 1b6ceca32fc81ca7cc7ac2eb7d11f687e6f87f26 and shipped as part of `npm@3.10.1`. Thanks for the quick turnaround time on this, Rebecca!

PR #13098, opened 2016-06-18
Release 2.15.8

othiym23:
Hotfix update because it looks like we might have a (good enough) fix for #5082.


othiym23:
Gahhhhh....

PR #13116, opened 2016-06-20
Create issue template

emzinnia:
I think that including an issue template would be helpful.

I based this off mostly from what I saw on Angular/Typescript’s issue
templates, as well as our own contributing guidelines.

Only question is: should users be encouraged to submit their debug logs
in gists?

Well, there might also be questions about formatting/etc. 
The debug log generally includes a lot of the info we want like node version, platform, etc.

### DIFF
--- a/issue_template.md
+++ b/issue_template.md
@@ -1,0 +1,34 @@
+<!-- Thank you for contributing to npm! Please review this checklist
+      before contributing to npm.
+      Please check if there's a solution in the troubleshooting wiki
+        https://github.com/npm/npm/wiki/Troubleshooting
+      Also ensure that your new issue conforms to contribution guidelines.
+      Participation in this open source project is subject to the npm Code of Conduct.
+        https://www.npmjs.com/policies/conduct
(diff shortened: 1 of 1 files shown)


EmmaRamirez:
# Meeting Notes
🕔 Tues Jun 21 @ 5:00pm

**Main goal of this PR**: to be better able to identify what an issue is meant for: bugs, support, or feature requests. Issues should contain a clear statement of the problem/request, environment in which it occurred, and steps to reproducing the problem. Bonus points for suggesting fixes.

**How to Accomplish This**
- [ ] We should ask the user if npm crashes, produces an incorrect install, or other.
- [ ] We should ask if the user has an idea on how to make npm better.
- [ ] Find the balance between asking the least we can, and extracting the most information possible. (subjective, I know)
- [ ] Instructions should be clear, with links to further documentation.
- [ ] We should ensure issues filed are in line with the Code of Conduct.
- [ ] We should get information about the user's development setup/environment.

I'll be editing the PR further to make sure it accomplishes these goals.

PR #13117, opened 2016-06-20
Disable progress bar for `npm init` & `npm adduser`

iarna:
Displaying the progress bar during interactive commands results in the prompt be stomped.


othiym23:
Reviewed in-air. :sheep::rocket::fire:

iarna:
This landed in 3.10.2

Issue #13113, opened 2016-06-20
Docs: link to `npm-tag` from cli docs on tag is 404


SimenB:
https://docs.npmjs.com/cli/tag#see-also

Clicking the one called `tag` leads to https://docs.npmjs.com/undefined/tag

PR #13189, opened 2016-06-27
Issue #13113: Docs: link to `npm-tag` from cli docs on tag is 404

macdonst:
User @SimenB found a broken link. I think that the `npm-tag(3)` should actually be `npm-dist-tag(1)` in the see also section to link to the non-deprecated command.


zkat:
LGTM

zkat:
This was included in `npm@2.15.9` and `npm@3.10.4`. Thanks! 🎉

PR #13199, opened 2016-06-28
deps: upgrade node-gyp to v3.4.0 (master)

rvagg:
Headline items:
- Fix for Visual Studio 2015 __pfnDliNotifyHook2 failures https://github.com/nodejs/node-gyp/pull/952
- Fix for AIX using fs.statSync https://github.com/nodejs/node-gyp/pull/955
- More verbose msbuild.exe missing error output https://github.com/nodejs/node-gyp/pull/930/files
- Added --silent for --loglevel=silent https://github.com/nodejs/node-gyp/pull/937
- Enable V8 deprecation warnings for native addons https://github.com/nodejs/node-gyp/pull/920

Full changelog at https://github.com/nodejs/node-gyp/blob/master/CHANGELOG.md

---

I suspect you're already getting bug reports on the first item, addons just won't compile with the latest MSVS 2015.

PR #13200, opened 2016-06-28
deps: upgrade node-gyp to v3.4.0 (lts)

rvagg:
Same as https://github.com/npm/npm/pull/13199 but for 2.x


zkat:
(note: we might need to reinstall this cause there's missing deps, I think, but I'll still attribute it to you, Rod)

zkat:
This was merged into LTS and will be released into our next LTS this week! Thanks,

rvagg:
@zkat for future reference what should I be doing here? A complete `rm -rf node_modules` and reinstall? I just did `npm i node-gyp@latest -S` and I believe it was with npm@3. Perhaps an `npm dedupe` would have done the trick?

zkat:
I'm not actually sure what the difference was? Our `.npmrc` should be taking care of those installed. Doing `rm -rf` on `node_modules/`, installing the way you did, and doing a `git add -A node_modules package.json` should generally be enough to get the commit right.

Also, this was included in `npm@2.15.9`. Thanks! I'm closing this for now, but feel free to keep talking in here.

PR #13201, opened 2016-06-28
fix: nyc@7 places files in node_modules/.cache/* that should be excluded

bcoe:
## Background

nyc@7 is a major overhaul, as we move towards merging the nyc and Istanbul projects once and for all:

https://github.com/istanbuljs/nyc/pull/286

tldr; we've completely swapped out the instrumentation internals, so we might bump into a few more hiccups -- please open issues against nyc, and I'll get them fixed ASAP.
## The Problem

nyc caches instrumented source files in `node_modules/.cache`, the cached versions of modules that represent exceptions for the `config-meta.js` test cause problems, e.g., `config.js`.

I'm not sure why this didn't crop up in pre-7.x versions, we've been using this caching approach for quite some time (it might actually be that the instrumentation was silently failing).
## The Solution

Let's just ignore source files that fall inside of `node_modules/.cache`.

### DIFF
--- a/test/tap/config-meta.js
+++ b/test/tap/config-meta.js
@@ -24,6 +24,7 @@ var exceptions = [
   path.resolve(nm, 'npm-registry-client', 'lib', 'publish.js'),
   path.resolve(nm, 'npm-registry-client', 'lib', 'request.js')
 ]
+var cacheFolder = path.resolve(nm, './.cache')
 
 test('get files', function (t) {
   walk(nm)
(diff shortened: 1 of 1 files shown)


zkat:
I think @othiym23 said something about needing to review this/talk about it? I forget because it was on-air.

iarna:
@bcoe I don't see any errors without this patch with files in `node_modules/.cache` including js files. Can you share the test failure you're seeing?

iarna:
(More meta, but I would prefer this be fixed in the test rather than having something nyc specific in npm itself.)

othiym23:
According to @bcoe `nyc@7` followed a different approach that didn't use this directory / create this cruft. At the same time, it _still_ brings greatly improved handling of ES2015 features. Way to go, Ben! Closing.

PR #13209, opened 2016-06-28
doc: tweak issue template

othiym23:
- move important bits to top
- condense supporting info down to a smaller section
- more structured request for proxy-related info

**r**: @EmmaRamirez, @KenanY, @npm/support-team, @npm/registry-team, @zkat @iarna

### DIFF
--- a/.github/issue_template.md
+++ b/.github/issue_template.md
@@ -1,3 +1,40 @@
+#### I'm opening this issue because:
+
+  - [ ] npm is crashing.
+  - [ ] npm is producing an incorrect install.
+  - [ ] npm is doing something I don't understand.
+  - [ ] Other (_see below for feature requests_):
+
+#### What's going wrong?
+
+#### How can the CLI team reproduce the problem?
+
+<!--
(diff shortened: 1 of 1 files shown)


zkat:
LGTM. This is a great next attempt! 🎉

EmmaRamirez:
First, I think this is definitely an improvement--I think checkboxes are a lot nicer than making someone manually answer. It also gives us an at-a-glance if the user filed the issue according to the template.

Question: should there be any ramifications for deleting the template in an issue? It, at least to me, comes off like the person a) was intimidated by the template (which would be our fault, and definitely something we're addressing) or b) didn't really want to put in the effort to get help from their problem. Maybe that's too harsh though? But we also have the issue, like we've discussed in meetings, that the npm CLI team doesn't really have the personnel bandwidth to reply to most support issues, especially the undecipherable ones.

mmalecki:
Looking good, the questions are pretty exhaustive with regards to network-related issues :+1: :sheep:

othiym23:
Let's see how this iteration goes!

PR #13212, opened 2016-06-29
Resolve local args relative to cwd

iarna:
When you run `npm install ../relative/path/to/module` it should be resolving that relative to your current directory.

This fixes a bug in npm@3 where it would resolve it relative to your package root, so if you were in a subdirectory it would mysteriously fail.


zkat:
LGTM

zkat:
This was included in `npm@3.10.4`. Danke~

PR #13213, opened 2016-06-29
When installing a module from the command line, use shrinkwrap ver by default

iarna:
If you run `npm install modulename` it should, if a `npm-shrinkwrap.json` is
present, use the verison found there.  If not, use the version found in your
`package.json` and failing that use `latest`.

Without this patch the first check was being bypassed because version resolution
was being done prior to loading the shrinkwrap, and so checks to match the shrinkwrap
version couldn't succeed. This was a regression in a earlier version of npm@3.

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -591,18 +591,19 @@ Installer.prototype.readLocalPackageData = function (cb) {
         return cb(er)
       }
       if (!currentTree.package) currentTree.package = {}
-      self.loadArgMetadata(iferr(cb, function () {
-        if (currentTree.package._shrinkwrap) return cb()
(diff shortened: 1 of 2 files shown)


zkat:
Is it reasonable to ask for a test that verifies that if you specify an actual version, you *will* hit the registry if the version doesn't match the shrinkwrapped entry? Is that specific thing already verified elsewhere?

iarna:
Sure is. I've updated it as such.

zkat:
Thanks! LGTM!

zkat:
This was included in npm@3.10.4, danke!

PR #13214, opened 2016-06-29
Relative local specifiers in a shrinkwrap should be relative to the shrinkwrap

iarna:
A bug was introduced in 3.9.0 with the npm-package-arg/realize-package-specifier change where they would be relative to the requiring module's source directory (if the requiring module was also a local dependency).

This fixes https://github.com/JamieMason/shrinkpack/issues/45

### DIFF
--- a/lib/install/deps.js
+++ b/lib/install/deps.js
@@ -1,7 +1,6 @@
 'use strict'
 var assert = require('assert')
 var path = require('path')
-var url = require('url')
 var semver = require('semver')
 var asyncMap = require('slide').asyncMap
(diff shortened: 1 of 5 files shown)


zkat:
LGTM

zkat:
@iarna this patch doesn't cleanly apply because of https://github.com/npm/npm/pull/13212 :\ I'm not exactly sure which `where` to keep, here.

zkat:
I think I got it.

iarna:
Landed in 3.10.4!

PR #13222, opened 2016-06-30
test: tweaking git-races test to maybe work on appveyor

zkat:
This really shouldn't be going to the network like this
but I would need to take more time to carefully reproduce
this with local repositories.

It's really icky that we're relying on an external git repo
owned by someone else to pull this off.

The actual fix in this patch is just to bypass ssh key errors
that happen in AppVeyor. I'm not sure why Travis doesn't have
this issue, but that's probably because Travis does its own
ssh key business?

**r**: @othiym23

### DIFF
--- a/test/tap/git-races.js
+++ b/test/tap/git-races.js
@@ -44,7 +44,7 @@ bar#4.0.0 shouldn't have gotten its own copy if buzz, and if it did, it shouldn'
 */
 
 ;['bar', 'foo', 'buzz'].forEach(function (name) {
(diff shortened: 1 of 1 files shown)


palmerj3:
@zkat do you guys have an id_rsa file on appveyor? It seems like this is the problem from the logs..

othiym23:
Upon team discussion, we decided that this probably isn't going to fix our AppVeyor issues, and so closing. Thanks anyway, @zkat!

PR #13244, opened 2016-07-01
fix env - https://github.com/npm/npm/pull/10073

graingert:


iarna:
PRing someone else's branch is a little weird. This isn't intended for consumption as is. At the moment it's just a little experimental restructuring.

Possibly these should be functions and not computed statically.
It also would require tests.
And finally, a bit more justification for actually making changes to the code. Yes, I wrote it, but I also didn't PR it. =p

PR #13257, opened 2016-07-02
Fix error EXDEV: cross-device link not permitted (utils/rename.js)

rachael:
Fixes issue #9863

Copy and delete when `rename` fails.
A common problem when a node module is mounted on a different mount point, as in common Docker build caching mechanisms.

---

Edit:
In reference to PR #10805 -- the copy/remove action in this PR acts as a fallback that only occurs when `rename` fails, so only causes slowdown when absolutely necessary to fix the problem.

### DIFF
--- a/lib/utils/rename.js
+++ b/lib/utils/rename.js
@@ -1,16 +1,86 @@
 'use strict'
 var fs = require('graceful-fs')
+var log = require('npmlog')
+var path = require('path')
 var SaveStack = require('./save-stack.js')
 
 module.exports = rename
 
 function rename (from, to, cb) {
   var saved = new SaveStack(rename)
-  fs.rename(from, to, function (er) {
-    if (er) {
-      return cb(saved.completeWith(er))
-    } else {
-      return cb()
+  fs.rename(from, to, function (renameErr) {
+    if (renameErr) {
+      try {
+        copyDir(from, to)
+        rmdir(from)
+      } catch (copyErr) {
+        log.silly('renameFail', 'Could not copy directory %s => %s: %s',
+          from, to, copyErr)
+        try {
+          rmdir(to)
+        } catch (cleanupErr) {
+          log.silly('cleanupFailedCopy', 'Could not cleanup partial copy %s: %s',
+            to, cleanupErr)
+        }
+        return cb(saved.completeWith(renameErr))
+      }
     }
+    return cb()
   })
+}
+
+function mkdir (dir) {
+  // making directory without exception if exists
+  try {
+    fs.mkdirSync(dir, '0755')
+  } catch (e) {
+    if (e.code !== 'EEXIST') {
+      throw e
+    }
+  }
+}
+
+function rmdir (dir) {
+  if (fs.existsSync(dir)) {
+    var list = fs.readdirSync(dir)
+    for (var i = 0; i < list.length; i++) {
+      var filename = path.join(dir, list[i])
+      var stat = fs.statSync(filename)
+
+      if (filename === '.' || filename === '..') {
+        // pass these files
+      } else if (stat.isDirectory()) {
+        // rmdir recursively
+        rmdir(filename)
+      } else {
+        // rm fiilename
+        fs.unlinkSync(filename)
+      }
+    }
+    fs.rmdirSync(dir)
+  } else {
+    log.silly('removeDir', 'Does not exist: %s', dir)
+  }
+}
+
+function copyDir (src, dest) {
+  mkdir(dest)
+  var files = fs.readdirSync(src)
+  for (var i = 0; i < files.length; i++) {
+    var current = fs.lstatSync(path.join(src, files[i]))
+    if (current.isDirectory()) {
+      copyDir(path.join(src, files[i]), path.join(dest, files[i]))
+    } else if (current.isSymbolicLink()) {
+      var symlink = fs.readlinkSync(path.join(src, files[i]))
+      fs.symlinkSync(symlink, path.join(dest, files[i]))
+    } else {
+      copy(path.join(src, files[i]), path.join(dest, files[i]))
+    }
+  }
+}
+
+function copy (src, dest) {
+  var oldFile = fs.createReadStream(src)
+  var newFile = fs.createWriteStream(dest)
+  oldFile.pipe(newFile)
 }


akatrevorjay:
This resorts to copying the directory, then deleting it, avoiding cross-device mountpath problems with atomic rename, but only if the atomic rename itself fails. It's standard UNIX semantics.

rachael:
In reference to PR #10805 -- the copy/remove action in this PR acts as a fallback that only occurs when rename() fails, so only causes slowdown when absolutely necessary to fix the problem.

akatrevorjay:
It's important to note that this is the same method utilized by text editors around the world. It is not at all slower unless the rename *fails* and it tries to copy it.

The method npm is using to install packages is already *NOT* fully atomic, so there's no point to consider that as a reason to deny this PR.

This is also not docker specific, but applies anywhere you are using overlay or bind mounts to manage your node_modules from a central place/cache.

Boom.

frank-dspeed:
Ehm this is useless work should get denied !!!!!! look fs.move defaults to use on unix rename and failback to move only in case of cross device

frank-dspeed:
Let me Explain Something more about why this is bad expect one build step has the files already because npm install then the secund layer where i do upgrade or else will have a secund new copy so layers will grow more then 2 x even on snap shooting expect npm upgrade inside container and container tagging

akatrevorjay:
@frank-dspeed Really sorry, but can you rephrase that a bit? It's a bit hard for me to understand the phrasing.

akatrevorjay:
@frank-dspeed Also, it's not useless work at all. I think you may misunderstand the point of this. I'd be delighted to talk about it though if you don't mind!

vincentwoo:
Would love to see this PR merged. Docker users around the world will finally be able to delete the following hack from their `Dockerfile`s:

```
# TODO: remove when https://github.com/npm/npm/issues/9863 is fixed
RUN cd $(npm root -g)/npm \
 && npm install fs-extra \
 && sed -i -e s/graceful-fs/fs-extra/ -e s/fs\.rename/fs.move/ ./lib/utils/rename.js
```

iarna:
I want this to be a module that implements the semantics in https://github.com/npm/npm/issues/9863#issuecomment-152342020
http://aufs.sourceforge.net/aufs.html
> To rename(2) directory may return EXDEV even if both of src and tgt are on the same aufs. When the rename-src dir exists on multiple branches and the lower dir has child(ren), aufs has to copyup all his children. It can be recursive copyup. Current aufs does not support such huge copyup operation at one time in kernel space, instead produces a warning and returns EXDEV. Generally, mv(1) detects this error and tries mkdir(2) and rename(2) or copy/unlink recursively. So the result is harmless. If your application which issues rename(2) for a directory does not support EXDEV, it will not work on aufs. Also this specification is applied to the case when the src directroy exists on the lower readonly branch and it has child(ren).

This new module needs to be used everywhere in `/lib/install/action/*.js` that currently uses rename.

rachael:
@iarna I agree regarding the importance of keeping an atomic operation atomic. However, wouldn't adding in calls to `mkdir` or `rmdir` also break atomicity? `copy/unlink` is also widely accepted as a fallback in unix, text editors, etc when an atomic `rename` fails. In the quote that you posted:

> Generally, mv(1) detects this error and tries mkdir(2) and rename(2) or copy/unlink recursively.

it sounds like the two fallback approaches are being referred to interchangeably. I like the `copy/unlink` approach because it is simple. Are there specific reasons why you think using that approach over `mkdir/rename` warrants closing or revising the PR? I'd love to hear them :)

I think this is a valuable fix and I'd like to get this merged in so we can all stop using hacks and workarounds! 🎉

frank-dspeed:
you should read about the current state of fs.move it already uses in current versions rename and if that fails it moves so it should be working like expected

rachael:
@frank-dspeed are you referring to fs-extra's [`move`](https://www.npmjs.com/package/fs-extra#movesrc-dest-options-callback) function? A potential revision to this PR could be to replace either the `copy/unlink` or the entire `rename` operation with `move`.

Here's the code if anyone's interested: [`node-fs-extra/lib/move/index.js`](https://github.com/jprichardson/node-fs-extra/blob/master/lib/move/index.js)

frank-dspeed:
What i mean is Unix and Windows and as Far as i know all other Operating systems do on the filesystem level on a move action internal a rename (Changing index of data without touching data body) and if that fails they do all the code you writed here in background but you need to expect that npm if you do the code like in this pr is not useable any more with docker.

It will Technical work without questions but the resulting images in many situations will have a bigger size then expected because you create maybe many many versions of the npm trees in sub layers.

for example when i would know this pr gets applyed to npm i would rewrite all my dockerfiles to do actions in one layer as i would do it without this pr to dont get the exdev warning :)

frank-dspeed:
Easy example: 
1. docker build dockerfile (includes node npm) and installs some packages global
2. docker run result then install some npm packages global and in some app directorys. After that Commit
3. docker run the result from 1-2 and now we upgrade all global packages and install some extra.
- The Fail happend now we will have a copy of old global packes in one layer complet and thx to this mv copy action in this pr we will have copyed all old files one layer up as we upgraded in action 2 and expected would be to have only updated files in action 3s layer but we will get a complet copy of the package directorys
so with every npm install upgrade i will end up with more copys.
 now we should think also about! top level packages!!! when i install a new package that depends on some stuff a other package has already NPM will do restructure the dependencys and guess what? yes it will create copys :)

So to compress this infos replacing the fs.rename inside npm with fs.move does also all the technical stuff this PR does without affecting the Concepts of Layers. And in benchmarking it adds near no overhead on diffrent filesystems i use it in Production for more then 1 and a Half Year as you know from diffrent issus posted here.

oh and by the way i am in the lucky situation that when i say production we talk about many automated builds with realy freaky situations on 1000 of hosts with diffrent docker containers managed via nodejs that are running many 100th of nodejs microservices using totall 8 diffrent filesystems for diffrent needs of clustering security and performance needs.

frank-dspeed:
oh and i would suggest you to easy do the most common use case where AUFS is not envolved even :)

DOCKERFILE:
FROM node:6.2.1
ADD ./my_cool_app /app
WORKDIR /app
RUN npm install

docker build -t my_cool_app DOCKERFILE
docker run my_cool_app -v /home/frank/my_cool_module:/app/node_modules/my_cool_module

When my_cool_module will have subfolder node_modules and i will issu npm install or upgrade on /app inside the container.

rachael:
Alternative PR #13299 which utilizes `fs-extra` as discussed has been opened.

akatrevorjay:
@frank-dspeed That scenario you mentioned is against best practices in a few ways. If you make your container immutable (as it should be), then as any CoW filesystem, you have two copies in the end, the old and the new, not 2+, unless I'm misunderstanding what you mean by that?

akatrevorjay:
What this fixes is when you are using standardized, reproducible builds of containers (ie not docker commit), and you maintain a node_modules cache across builds to speed up those nasty build times. This is a pretty standard setup, and is the primary goal of this to fix.

frank-dspeed:
@rachael when it happens in one layer this fix would work right! all is cleaned up!
but the standart situation that can come is that i prepare some packages via docker file and maybe i use diffrent lines so diffrent layers as you see in the build process of the Dockerfile and then when i update that already installed packages (reorder dependencys because new info in the npm packages that get installed on execution of the docker container) it will create on that execution in that current layer a totall new copy of each module that gets touched by this function

iarna:
@rachel I prefer the rename-recursively approach over copy-recursively and unlink because the former should be substantially faster. Also I like following `mv(1)`'s lead as far as behavior is concerned.

iarna:
None of this really solves the atomicity problem– they would require anything accessing the source or destination to take a lock, and personally I'm inclined to give it a pass for the time being. The `move` action is done serially anyway so it won't step on itself even without atomicity guarantees

iarna:
I had originally thought that `lib/utils/rename` was used outside of the installer, but no, I was wrong, so changing it's behavior as you've done is totally fine!

akatrevorjay:
@iarna Thanks! That's what we saw when we looked as well. Glad we can be of service!

akatrevorjay:
@frank-dspeed Those need to be one command in the Dockerfile, or else you're going to run into situations like you mentioned. That's why it's seen as not the best solution, similar to leaving package lists in your layers then removing them in another. You should make your RUNs rather atomic themselves to avoid such situations. :)

rachael:
@iarna awesome! Glad to hear it! Replacing the copy with a recursive rename sounds like a good TODO for a future speedup but this copy/unlink solution does work well to fix the problem. Happy we could help :)

kvervo:
The EXDEV is a real problem currently for us. Would live to see this PR merged :)

AkihiroSuda:
Doesn't it need to preserve the original permission and the timestamp?

zkat:
Hey @rachael! Thanks for this PR. Sorry it's been a while -- could you include a test case with this before you merge it? From what @iarna was saying, it looks like this change is good to go otherwise?

chadnickbok:
I'm not sure how a good test case around this functionality can be written, since right now many of the already existing npm tests rely on `rename.js`, but without explicitly running this on a filesystem that doesn't support the `rename(2)` syscall, its not clear how this functionality can be tested.

rachael:
Hi @zkat! Anything you'd like a test case for specifically? Current tests that test the installer are all passing, at least proving there's no regression (hopefully). There wasn't a unit test for the original `rename()`, just the integration ones.

It would be cool to get this merged. Looks like the spawning issue is still active now and then.

zkat:
Yeah you'd basically have to mock up `rename(2)` in order to test it. We've been trying to get more tests going, specially around basic functionality. But if @othiym23 thinks it's fine to merge this without, then I'm fine with it too 👍

chadnickbok:
@zkat This PR has some serious shortcomings, and I don't think its ready to be merged. As a few people have mentioned in this thread;

 - This approach isn't async, which the rest of npm is. Instead, once `rename` fails this approach uses synchronous filesystem calls
- This approach doesn't address syncing file permissions
- The `copy` function in this approach _is_ asynchronous, but isn't checked. This means that when `rmdir` is called, the files might not have finished being copied when the original files are deleted.

I've addressed all of these issues in an alternate PR based on ncp here: https://github.com/npm/npm/pull/13901

frank-dspeed:
@chadnickbok your right but maybe watch into most current fs.move it uses rename under the hood if possible :) thats why my PR simply was to use that as all the logic is in fs.move core functionality already

also all the gracefull-fs stuff should be integrated already

chadnickbok:
@frank-dspeed I looked into fs.extra, and while it handles the EXDEV issue by falling back onto `ncp`, under the hood `ncp` doesn't _recursively_ use `rename`.

Its a small tweak, but should provide a significant speedup when the reason for the EXDEV error is that the filesystem is using aufs under the hood.

iarna:
I'm closing this in preference to #13901 as I gather @rachael prefers that one over this one. Yell at me if I'm wrong! =)

iarna:
And #13901 has now been superseded by #15901.

Issue #13256, opened 2016-07-02
npm@3.10.4 installation: node-gyp@3.4.0 "src" directory is missing


zaenk:
#### I'm opening this issue because:

  - [x] npm is producing an incorrect install.

#### What's going wrong?

Directory `%appdata%/npm/node_modules/npm/node_modules/node-gyp/src` is missing after updating to npm 3.10.4 via `npm i npm@3.10.4 -g`. 

This folder is also missing from the pre-release ZIP archive.

#### How can the CLI team reproduce the problem?

- run `npm i npm@3.10.4 -g`
  - make sure that this installation is picked up as primary (eg. delete `npm.cmd` from `%ProgramFiles%/nodejs`)
- set up `node-gyp`
  - install [Visual C++ 2015 Build Tools](http://landinghub.visualstudio.com/visual-cpp-build-tools)
  - install [python 2.7.12](https://www.python.org/downloads/release/python-2712/)
  - configure defaults for `node-gyp`: https://github.com/nodejs/node-gyp#installation
- test `node-gyp`
  - run `npm i bufferutil`, [sample output](https://gist.github.com/zaenk/0f97b9402c41db6330d6d9e224d1f6af)
  - [npm-debug.log gist](https://gist.github.com/zaenk/11f36fba96ef6af975c7fabac4ba668e)

### supporting information:

 - `npm -v`: `6.2.2`
 - `node -v`:  `3.10.4`
 - `npm config get registry`: `https://registry.npmjs.org/`
 - OS: Windows 10

rvagg:
yikes! looks like this was my fault some how, very strange. Thanks for picking this up so quickly @zaenk.

othiym23:
Closing as duplicate of #13260, which will be addressed by a testing release cut most likely today by @iarna. Thanks for noticing this and letting us know so quickly!

PR #13260, opened 2016-07-03
deps: fix node-gyp@3.4.0

zkat:
I do not know what happened, but whatever it is, it went poorly.

This is a reinstall of node-gyp@3.4.0 which seems to be significantly
different from the one that was submitted in a PR.

Fixes https://github.com/npm/npm/issues/13256


iarna:
Landed in 3.10.5

PR #13299, opened 2016-07-06
Fix error EXDEV: cross-device link not permitted (alternative approach - replace utils/rename.js -> utils/move.js)

rachael:
Alternative to PR #13257 that replaces a simply `copy/remove` on fail with the [more robust error handling](https://github.com/jprichardson/node-fs-extra/blob/master/lib/move/index.js) of [fs-extra](https://www.npmjs.com/package/fs-extra#movesrc-dest-options-callback).

`utils/rename.js` has been renamed to `utils/move.js` and now wraps fs-extra's `move` instead of graceful-fs's `rename`.

Fixes issue #9863 where `rename` was causing EXDEV cross-device link errors (fairly common because of the popularity of tools like Docker). fs-extra's `move` can handle moving files and directories cross-device, specifically trapping for EXDEV errors and utilizing a recursive async `copy/remove` (`ncp` and `rimraf`) to move directories only after all other attempts to move fail.

I did have some uncertainty when making the changes because of an existing file [`lib/install/action/move.js`](../blob/master/lib/install/action/move.js) that I did not rename. `install/action/move.js` requires `utils/move.js` as `mv`. Renaming one of the `move.js` files is a possibility; however, I like the parallelism between `install/action/move.js` and `install/action/remove.js`, as well as the wrapping of `move` with `move.js` which matches the old behavior of wrapping `rename` with `rename.js`.


frank-dspeed:
i would say you should simply rename the file to move and rename the other file to mv and rm rest of this pr is much better then the other because this dont got this try catch panic but if you only whant to handle exdef its realy enought to rename the fs.rename to fs.move and fs to gracefull fs and your done i dont get any exdev errors with that in any docker situation

iarna:
I'm not super comfortable with `fs-extra` implementation of move as it's handling of EXDEV is focused on cross device issues (which to be fair, is what the error is _supposed_ to be about), but the scenario we're fixing that's not the case. For our scenario its going to fall back to copying, which I would prefer not to do unless necessary for speed reasons.

frank-dspeed:
i am out of discussion and whant to give every one the tip that there are alternates to npm that are not maintained like this :) cheers

iarna:
I'm going to close this as we'd prefer to take a version of #13257 that uses renames instead of copies.

Issue #13290, opened 2016-07-06
npm@3.10 progress bar shown inside editor during `npm config edit`


janslow:
#### I'm opening this issue because:

  - [ ] npm is crashing.
  - [ ] npm is producing an incorrect install.
  - [ ] npm is doing something I don't understand.
  - [x] Other (_see below for feature requests_):

#### What's going wrong?

When using `npm config edit` on npm@3.10.3, the progress bar overwrites the selected line in the editor, making it quite hard to edit the config.

![image](https://cloud.githubusercontent.com/assets/1253367/16615095/32beb954-436c-11e6-91ad-0eff9d5f4034.png)

![image](https://cloud.githubusercontent.com/assets/1253367/16615111/48049eaa-436c-11e6-9023-1766d0558d77.png)

#### How can the CLI team reproduce the problem?

1. Run `npm config edit`
2. Press down key (e.g. to navigate to a config line)
3. Every selected line is replaced with a progress bar, so you can't see what you are typing.

### supporting information:

 - `npm -v` prints: `3.10.3` 
 - `node -v` prints: `v6.2.2`
 - `npm config get registry` prints: `https://registry.npmjs.org/`
 - Windows, OS X, or Linux?: Reproduced on OS X and Linux (using both fish and bash shells).

Also,
- `npm config get editor` prints: `vi`
- `vi --version` prints: 
-- `VIM - Vi IMproved 7.4 (2013 Aug 10, compiled Apr  8 2016 14:34:51)` (on Linux)
-- `VIM - Vi IMproved 7.3 (2010 Aug 15, compiled Apr  5 2016 14:53:37)` (on OS X)
I've also set the `editor` config option to `nano`, and it's the same.

PR #13305, opened 2016-07-06
Disable progress bar for `npm edit` and `npm config edit`

watilde:
Fixes https://github.com/npm/npm/issues/13290 and the similar issue. In v3.10.x, the progress bar is shown inside editor during `npm config edit` and `npm edit`. This kind of issues are fixed by just disabling it: https://github.com/npm/npm/pull/13117, so I added a line into the files.

Thanks!

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -106,6 +106,7 @@ function edit (cb) {
         data,
         function (er) {
           if (er) return cb(er)
+          log.disableProgress()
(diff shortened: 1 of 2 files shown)


iarna:
Thanks for this! We probably want to enable it again after the editor completes as a rebuild can happen then. I'll slip that in though. =)

iarna:
Landed in 3.10.6!

Issue #13303, opened 2016-07-06
shrinkwrap including dev deps deps


iarna:
#### I'm opening this issue because:

  - [ ] npm is crashing.
  - [x] npm is producing an incorrect install.
  - [ ] npm is doing something I don't understand.
  - [ ] Other (_see below for feature requests_):

#### What's going wrong?

`npm shrinkwrap` is including dependencies of dev dependencies even when the shrinkwrap file didn't have any dependencies.

#### How can the CLI team reproduce the problem?

Save any module that has dependencies as a dev dep. Run `npm shrinkwrap` see how the transitive deps are included

SimenB:
This is the same as #13227?

PR #13308, opened 2016-07-07
Fix bug where transitive deps of dev deps would be incorrectly included

iarna:
Fixes: #13303

### DIFF
--- a/lib/install/is-dev.js
+++ b/lib/install/is-dev.js
@@ -1,11 +1,21 @@
 'use strict'
 var moduleName = require('../utils/module-name.js')
 
+function isThisDev (node, name) {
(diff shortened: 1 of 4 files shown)


othiym23:
I like the extraction of the tests into functions, and the way that shrinkwrapping works now is closer to being extracted as freestanding logic, which I like. The distinction between the intended uses of `lib/install/is-dev.js` is fuzzy enough that some short comments describing the intended use of each would make it easier for people reading the file to get oriented. The test is simple and straightforward. Overall, looks good! :sheep::rocket::fire:

iarna:
Landed in 3.10.6!

PR #13316, opened 2016-07-07
Run cache/add-local's install prune and pack in a child process

Volune:
This PR make `add-local` tasks "install, prune, pack" to be run in a child process.

This PR should probably not be merged as it is now, a few questions remain.

The changes are heavily inspired by #8725. It should be a good step to build modules from git/local without polluting node_modules in global and/or production mode (see #3055).

I've tested it (inspired by #7040) with the following files:
- `./package.json`
  
  ```
  {
    "name": "npm-issue-local-dependency-prepublish",
    "version": "0.0.0",
    "dependencies": {
      "local-pkg": "file:local-pkg"
    }
  }
  ```
- `./local-pkg/package.json`
  
  ```
  {
    "name": "local-pkg",
    "version": "1.0.0",
    "scripts": {
      "prepublish": "npm config get cache-lock-retries > npmconfig && tap --version > tapversion"
    },
    "devDependencies": {
      "tap": "^0.4.13"
    }
  }
  ```
- `./local-pkg/.npmrc`
  
  ```
  cache-lock-retries = 9
  ```

I've tested with `../npm/bin/npm-cli.js i` and `../npm/bin/npm-cli.js i --production` and `../npm/bin/npm-cli.js -g i`. Also a test with `"git+file:///f/Dev/some-module#master"` (on Windows)

The benefits of this implementation: don't have unwanted arguments when running install/prune/pack (in fact it'll be easy to chose what arguments and env variables to transmit), also take into consideration any .npmrc in the built module

Remaining questions:
- What arguments should be forwarded to the child process?
- What environment variables should be forwarded to the child process?
- How to configure if install or prune should be executed?
  It may make sense not to execute them for a local directory, and to be able to configure that in the .npmrc
- Add some test / doc in the PR?

Hope this will help, waiting for feedbacks.

### DIFF
--- a/bin/npm-cli.js
+++ b/bin/npm-cli.js
@@ -23,13 +23,9 @@
 
   var path = require('path')
   var npm = require('../lib/npm.js')
-  var npmconf = require('../lib/config/core.js')
   var errorHandler = require('../lib/utils/error-handler.js')
 
-  var configDefs = npmconf.defs
-  var shorthands = configDefs.shorthands
-  var types = configDefs.types
-  var nopt = require('nopt')
+  var buildConf = require('../lib/utils/build-conf.js')
 
   // if npm is called as "npmg" or "npm_g", then
   // run in global mode.
@@ -39,7 +35,7 @@
 
   log.verbose('cli', process.argv)
 
-  var conf = nopt(types, shorthands)
+  var conf = buildConf()
   npm.argv = conf.argv.remain
   if (npm.deref(npm.argv[0])) npm.command = npm.argv.shift()
   else conf.usage = true
--- a/lib/cache/add-local-fork.js
+++ b/lib/cache/add-local-fork.js
@@ -30,8 +30,15 @@ process.once('message', function onMessage (message) {
   }
 
   function cb (er) {
+    var error = undefined
+    if (er instanceof Error && er.stack) {
+      error = er.stack + '\n From fork child, ' +
+        String(er.message).replace(/\n/g, '\n ')
+    } else if (er) {
+      error = String(er)
+    }
     process.send({
-      error: er
+      error: error
     })
     setTimeout(function () {
       process.exit(0)
--- a/lib/cache/add-local-fork.js
+++ b/lib/cache/add-local-fork.js
@@ -1,0 +1,40 @@
+var log = require('npmlog')
+var iferr = require('iferr')
+
+process.once('message', function onMessage (message) {
+  var tgz = message.tgz
+  var p = message.p
+
+  var npm = require('../npm.js')
+  var buildConf = require('../utils/build-conf.js')
+  npm.load(buildConf(), iferr(cb, thenInstall))
+
+  function thenInstall () {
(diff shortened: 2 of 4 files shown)


othiym23:
Thanks for the work you've put into this so far! There are a number of problems with the approach of packing in a child process, including:

- There's a _lot_ of configuration that the child process would need in order to correctly create the cacheable package tarball. Just blanking out the child environment is too broad.
- npm is already using more RAM and CPU than a lot of users are comfortable with; spawning a new child process for each hosted Git dependency is only going to exacerbate this tendency (which will disproportionately affect the kinds of front-end projects that rely on lots of Git dependencies that are already having a tough time with npm).
- This adds a fair amount of complexity and new edge cases (working with child processes is always tricky, as we've learned with package scripts and the npm lifecycle), which seems like a bad idea to take on given that we have a better solution in mind for the long term.

What's the better solution? The CLI team believe that the long-term supportability of the CLI requires that npm change how it uses its configuration. Instead of having a quasi-global shared state object that all of the commands and their support modules can access directly, each command should have any necessary configuration passed in at invocation time via a configuration object. I hope it's clear how that would be useful in this case – the caching and installation code could be called with special options to pack Git dependencies, allowing the whole thing to be done in-process. It also gets us one step closer to having a less confusing and more supportable programmatic API within the CLI.

The most sensible approach to breaking down what we call "the god object" within the team would be to replace uses of `npm.conf` with a configuration object on a per-command basis. For the purposes of this change, that would mean fixing up the caching and install code to set up all their configuration at the top level, and call functions that actually do the caching and installation that take a config object.

I would totally understand if you felt like this was too much work to take on, especially after you put so much work into the current implementation. However, this _is_ work that needs to be done eventually, and the CLI team will get to it if you or community contributors don't first. As such, we have an incentive to support that work as it's done.

After all that, I hope it makes sense that I'm closing this pull request. It's an interesting approach, and I do appreciate the time and effort you put into it. Thanks for your help!

Volune:
No problem, this PR helped me find a few test issues, so it's not all lost.

Also I've realized that the currently implementation, when installing dependencies from local directory or git, ignores any `.npmrc` existing in the dependency package. It doesn't seem coherent with published packages. Should that be considered as a bug?

If I understand the better solution:
- First change the code to not require the `npm` object. I say `npm` and not `npm.config`, because some properties on `npm` depends on `npm.config`; that may be a problem if we only transfer `npm.config`.
  There is no feature change, so this can probably be done progressively on different parts of the code.
- Then recreate/clone/fork/... the `npm` object at the appropriate places in the code.
- Finally to fix the "git dependency" issue, run install inside `add-local`

@othiym23 For the first point, I tried an implementation on the `lib/cache/*.js` files by giving the `npm` object as first argument. I'm not sure this is a clean implementation, and I would like to know if someone on CLI team has an idea of what would be a good architecture/implementation regarding this point.
My try is available on [this branch](//github.com/Volune/npm/tree/temp-npm-as-arg) (and surprisingly it passes the tests on travis-ci).

PR #13336, opened 2016-07-11
Support for shell completion in GitHub shell

OwnageIsMagic:
GitHub Bash for Windows doesn't sets MSYSTEM var


iarna:
As implied by the tags, this would need tests to land. As it's been a year, I'm going to close it, but if you (or someone) gets motivated I'd happily accept it with tests.

OwnageIsMagic:
@iarna
Any idea what shold I test? 

I just changed to test is shell is `bash` instead of checking is it mingw build (it also could be `ash` or any other mingw shell)

PR #13338, opened 2016-07-11
Fix map-to-registry test when having custom npm config

Volune:
Discovered while working on #13316

When running `npm run tap` something dumps all the npm config in the environment variables, messing with the `map-to-registry.js` test

To reproduce, have a `registry` set in `~/.npmrc`, then run `npm run tap -- test/tap/map-to-registry.js`

This PR fixes the test and provides an utility file in case other tests have the same issue later.

### DIFF
--- a/test/fixtures/clean-process-env.js
+++ b/test/fixtures/clean-process-env.js
@@ -1,0 +1,8 @@
+// `npm run tap` dumps a ton of npm_config_* variables in the environment.
+Object.keys(process.env)
+  .filter(function (key) {
+    return /^npm_config_/.test(key)
+  })
+  .forEach(function (key) {
+    delete process.env[key]
+  })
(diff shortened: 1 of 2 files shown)


othiym23:
In general, tests that don't set themselves up such that extraneous environmental config doesn't interfere with the tests are considered buggy. It makes the tests a little brittle in some ways, but it also has helped us catch a number of instances where we've introduced regressions around changes to configuration or the environment. As such, this is kind of a blunt-force solution to that problem, and as such, we're not going to take this change. Thanks for your time!

Volune:
@othiym23 Would force the values that otherwise conflict with environmental config be a better fix?

```
var creds = {
  '//registry.npmjs.org/:username': 'u',
  '//registry.npmjs.org/:_password': new Buffer('p').toString('base64'),
  '//registry.npmjs.org/:email': 'e',
  cache: common.npm_config_cache,

  // ensure not to conflict with environment configuration
  'registry':  'https://registry.npmjs.org/',
  '//registry.npmjs.org/:__authToken': '',
  '//registry.npmjs.org/:always-auth': 'false'
}
```

If so I'll update the PR accordingly

PR #13339, opened 2016-07-11
Make fake package `@test/whoops` packagable in test `gently-rm-overeager`

Volune:
Discovered while working on #13316

In the test `gently-rm-overeager.js`, the package `@test/whoops` is expected to fail in postinstall when installed as a dependency.
Currently it also fails when installed for development, packaging or publishing. (That's why the test fails on my other PR)

This PR changes the test so that the package `@test/whoops` is packageable/publishable, but still fails when installed as a dependency.

By the way, when I was trying to figure out how this test works, I had cases where the `node_modules` or `node_modules/.bin` where not removed (rollback) after the installation of `@test/whoops` failed. Don't know if it was me messing with the code or an actual issue.

### DIFF
--- a/test/tap/gently-rm-overeager.js
+++ b/test/tap/gently-rm-overeager.js
@@ -15,9 +15,18 @@ var fixture = {
   name: '@test/whoops',
   version: '1.0.0',
   scripts: {
-    postinstall: 'echo \'nope\' && exit 1'
+    postinstall: 'echo \'nope\' && node failinstall.js'
   }
 }
+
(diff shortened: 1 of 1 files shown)


iarna:
Having this fail for development, packaging or publishing seems like the test is working as intended. Why would having the `postinstall` fail be a problem?

iarna:
As this has gone a year without comment, I'm going to close it. Please feel free to open a new PR if you have an answer to my question. Thanks!

PR #13344, opened 2016-07-12
Do not overwrite TMPDIR

### LINKED ISSUE #13325: Npm overrides TMPDIR for root users
- [ ] npm is crashing.
#### What's going wrong?

npm assumes that if `unsafe-perm` is false (the default for root users) that we want to use the current working directory as `TMPDIR`.

https://github.com/npm/npm/blob/6da8d1d47eb3f933fffd47952079dd3d94e9fe25/lib/utils/lifecycle.js#L80

This leads to some surprising and annoying behavior when running npm scripts as root on docker.

This was originally reported here: https://github.com/npm/npm/issues/4531 but closed.

It seems that the original "fix" was presumed that you could not write to `/tmp` if `unsafe-perm` is unset. Given that this is the default for the `root` user, this doesn't make much sense to me.

Perhaps npm should check to see if it can write to the directory before silently overriding an environment variable?

Create a `package.json` with: 

``` json
...
"scripts": {
    "test": "echo $TMPDIR"
  },
...
```

Run `sudo npm test`
### supporting information:
- `npm -v` prints: 3.8.9
- `node -v` prints: v6.2.0
  
  ```
  # npm config get registry
  npm info using npm@3.8.9
  npm info using node@v6.2.0
  https://registry.npmjs.org/
  npm info ok
  ```
- Windows, OS X, or Linux?: Linux and OS X
- Network issues:
- Container:
  - [x] I develop / deploy using Docker.

aaronjensen:
Fixes #13325 

Maybe a check for 'nobody' would be more appropriate, I don't know, but this existing behavior is surprising and causes mysterious/confusing problems.

### DIFF
--- a/lib/utils/lifecycle.js
+++ b/lib/utils/lifecycle.js
@@ -2,6 +2,7 @@ exports = module.exports = lifecycle
 exports.cmd = cmd
 exports.makeEnv = makeEnv
 exports._incorrectWorkingDirectory = _incorrectWorkingDirectory
+exports._getTemporaryDirectory = _getTemporaryDirectory
 
 var log = require('npmlog')
 var spawn = require('./spawn')
@@ -75,16 +76,36 @@ function lifecycle (pkg, stage, wd, unsafe, failOk, cb) {
     env.npm_node_execpath = env.NODE = env.NODE || process.execPath
     env.npm_execpath = require.main.filename
 
-    // 'nobody' typically doesn't have permission to write to /tmp
-    // even if it's never used, sh freaks out.
-    if (!npm.config.get('unsafe-perm')) env.TMPDIR = wd
+    _getTemporaryDirectory(wd, process.env, function (er, tmpdir) {
+      if (er) return cb(er)
 
-    lifecycle_(pkg, stage, wd, env, unsafe, failOk, cb)
+      env.TMPDIR = tmpdir
+      lifecycle_(pkg, stage, wd, env, unsafe, failOk, cb)
+    })
   })
 }
 
 function _incorrectWorkingDirectory (wd, pkg) {
   return wd.lastIndexOf(pkg.name) !== wd.length - pkg.name.length
+}
+
+function _getTemporaryDirectory (wd, env, cb) {
+  if (_isWritable(env.TMPDIR)) return cb(null, env.TMPDIR)
+
+  // 'nobody' typically doesn't have permission to write to /tmp
+  // even if it's never used, sh freaks out.
+  if (_isWritable(wd)) return cb(null, wd)
+
+  cb(new Error('Cannot find writable TMPDIR, please set TMPDIR to something writable'))
+}
+
+function _isWritable (dir) {
+  try {
+    fs.accessSync(dir, fs.W_OK)
+    return true
+  } catch (e) {
+    return false
+  }
 }
 
 function lifecycle_ (pkg, stage, wd, env, unsafe, failOk, cb) {
--- a/test/tap/lifecycle.js
+++ b/test/tap/lifecycle.js
@@ -42,3 +42,30 @@ test('lifecycle : rejects wd for ', function (t) {
     t.end()
   })
 })
+
+test('lifecycle : accepts TMPDIR that is writable', function (t) {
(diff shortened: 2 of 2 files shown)


othiym23:
As the comment indicates, just straight-up deleting those lines is going to cause problems for users where bash can't write to `$TMPDIR` as `nobody`. Having some kind of tests as to whether the CLI can write to the current directory and / or `$TMPDIR` and choosing one or the other (and maybe bailing out if there's _nowwhere_ we can write to) seems much less likely to cause unexpected regressions. Also, this needs a test to be merged. Thanks for the work you've done thus far!

othiym23:
See also d7dcacea6f17437fd88321987ca6aa8fe3c017d8, when this behavior was added: `/bin/sh` on Solaris / SmartOS systems, which is "pure" Bourne shell, freaks out when it can't write to TMPDIR. Cross-platform compatibility! 💫

aaronjensen:
@othiym23 Thank you for the feedback. Please see the updated commit and let me know if there's anything I can do to improve it.

lib/utils/lifecycle.js
```diff
@@ -75,18 +76,38 @@ function lifecycle (pkg, stage, wd, unsafe, failOk, cb) {
...
+  cb(new Error('Cannot find writable TMPDIR, please set TMPDIR to something writable'))
+}
+
+function _isWritable (dir) {
+  try {
+    fs.accessSync(dir, fs.W_OK)
+    return true
+  } catch (e) {
+    return false
+  }
```
aaronjensen: I'm not a fan of using exceptions in this way, but it felt like an awkward place to have to use callbacks. If you'd prefer that let me know and I can switch it out.

test/tap/lifecycle.js
```diff
@@ -42,3 +42,30 @@ test('lifecycle : rejects wd for ', function (t) {
     t.end()
   })
 })
+
+test('lifecycle : accepts TMPDIR that is writable', function (t) {
```
aaronjensen: These tests assume TMPDIR is set properly and writeable, hopefully that's not a problem anywhere.

test/tap/lifecycle.js
```diff
@@ -42,3 +42,30 @@ test('lifecycle : rejects wd for ', function (t) {
...
+    t.equal(tmpdir, wd)
+    t.end()
+  })
+})
+
+test('lifecycle : errors if neither TMPDIR or wd are writable', function (t) {
+  var wd = '/probably/not/writable/either'
+
+  lifecycle._getTemporaryDirectory(wd, {TMPDIR: '/probably/not/writable'}, function (er, tmpdir) {
+    t.type(er, Error)
```
aaronjensen: I'm new to tap and `t.error` didn't seem to do what I expected. Is this the proper way to test this?

iarna: Yeah `t.ifError` (and `t.error`) go against the `tap` grain and assert there WASN'T an error.

I would just do: `t.ok(er, 'assertion description')` to assert that an error _was_ produced.

lib/utils/lifecycle.js
iarna: In a test, this is fine (I do it myself).

aaronjensen:
Thanks @iarna. I updated the test to use `t.ok`. Is there anything else you'd like me to update on this or is it good to go?

iarna (approved):
LGTM

iarna:
This does not pass on 0.10 because it uses `fs.accessSync` which is not available on 0.10. We've gotta support 0.10 a little longer so I'm going to defer merging it.

iarna (changes requested):
There's an issue with this: It checks to see if `TMPDIR` is writable _by the current user_. But the error condition we're trying to guard against is `TMPDIR` not being writable by the uid & gid passed in.

I'm not sure there's a meaningful way to check for that, short of spawning a sub process to check for us.

iarna:
As I ultimately found a problem with this that I don't know how to resolve, and there's been no reply in over six months, I'm going to close this. Please feel free to open a new PR if you'd like to continue this discussion. Thanks!

PR #13349, opened 2016-07-12
doc: explain difference between `version` and `preversion` scripts

christophehurpeau:


iarna:
This is merged and will be in the next release (scheduled for later today)

PR #13359, opened 2016-07-12
addRemoteTarball: extract package metadata without writing to disk

misterbyrne:
One of the slowest parts of installing from a shrinkwrap is that `addLocalTarball` must validate the contents before it's added to the package cache ([here](https://github.com/npm/npm/blob/master/lib/cache/add-local-tarball.js#L105-L141)).

How much validation do we need? If it is enough simply to extract the `package.json` file and check the name, then we could stop extracting after `package.json` is extracted - since the tarballs are optimised to include this file first. We can also extract directly to memory to avoid the disk I/O.

I tried implementing this to see what the effect would be:

Timing on `master` for an application with a lot of dependencies on an SSD:

`node ~/src/npm/bin/npm-cli.js install --ignore-scripts  89.43s user 32.43s system 133% cpu 1:30.97 total`

timing on this branch:

`node ~/src/npm/bin/npm-cli.js install --ignore-scripts  63.44s user 19.56s system 128% cpu 1:04.34 total`

While a 30% speed-up is significant, there are _implicit_ validations that occur with the current method.
1. The whole file is extracted so we can be sure it's a valid tarball.
2. Loading the package metadata is more than just loading the JSON data (the README is inlined, etc.)

### DIFF
--- a/lib/cache/add-local-tarball.js
+++ b/lib/cache/add-local-tarball.js
@@ -14,8 +14,6 @@ var chownr = require('chownr')
 var inflight = require('inflight')
 var once = require('once')
 var writeStreamAtomic = require('fs-write-stream-atomic')
-var tempFilename = require('../utils/temp-filename.js')
-var rimraf = require('rimraf')
 var packageId = require('../utils/package-id.js')
 
 module.exports = addLocalTarball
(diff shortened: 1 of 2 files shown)


iarna:
This is a good insight. Currently we do often end up extracting tarballs or parts of them during an install and eliminating that has always been at the top of my list for low-hanging-performance-fruit. But I'd not considered the interactions between this code and shrinkwraps.

I implemented largely the same code over in `fetch-package-metadata` for locating `npm-shrinkwrap.json` files as `untarStream`. (It's a little more complicated because package files don't have to be gzipped– raw tars are ok).

The real sticker is all the weird behavior `read-package-json` does. It does more than including random other files, it also does things like drive the `node-gyp` interactions and many other defaults in the `package.json`. It also does some of its own normalization and then calls `normalize-package-data` to do more.

Because for remote tarballs like those used with shrinkwraps we _only_ have the package file to go on (no registry info), what we read is what will be passed through to the rest of the installer. So we _have_ to get that data through the same code paths as `read-package-json` currently provides.

This is why I didn't end up implementing streaming `package.json` reads for npm@3. I actually did in early versions of it, but ran into the "would have to rewrite `read-package-json` & `normalize-package-data`" wall and decided to punt that down the road. (At the time, they were still pretty much black boxes to me too.)

So rewriting `read-package-json` and `normalize-package-data` to be more useful to us _is_ on the table, but even then I'm pretty sure that won't help us here because `read-package-data` fills in defaults based on what files exist _on disk_. Hmmm (so thinking out loud here), we could build up an index of what's in the tar and have this `read-package-json` successor query that. That probably would be faster than writing all that data to disk, but it would still mean running the extraction more than once.

That heart of the issue, the double extract part, I'm pretty sure won't be fixable without rewriting the whole existing `cache` interface to separate its concerns more cleanly.

misterbyrne:
Cool. I kinda figured as much - I guess the problem is we can't just trust anything to go into the cache.

PR #13397, opened 2016-07-15
Tests that use the network...

iarna:
This PR has two things:
1. Patches for four tests to make them no longer use the network
2. Move the remaining 10 tests that require the network into `test/network`

`npm run test` will still run the network tests but this makes it easy to skip them if we have need to with `npm run tap test/tap/*.js`. It also acts as documentation of which tests remain to be converted.


othiym23:
Good idea, and good work! If the tests are green on Travis, LGTM! :sheep::rocket::fire:

iarna:
Will merge when green

iarna:
(I canceled the push CI build since it's identical to the PR build)

iarna:
Merged into `release-next`.

Issue #10300, opened 2015-11-07
Suggestion: store sha256sums in npm shrinkwrap


nakedible-p:
When a user runs npm-shrinkwrap, the idea is to ensure that no package versions may change. However, this still trusts "registry.npmjs.org" to always produce the same package to be true. In case of an exploit, this might assumption might not hold.

By including SHA256 checksums for each package in `npm-shrinkwrap.json` and having npm verify those checksums on `npm install`, the user could feel safe that nobody has substituted any code after running `npm shrinkwrap`.

This is not a full security solution for npm and is not meant to be, but it is meant to help the case where the project being compiled/installed is trusted (coming from a private git repo, or similar) but no outside sources are trusted - and I can't think where it would be harmful.

I am willing to do most of the legwork in implementing this, but for that I need to know that a) this feature is something that would be accepted in npm, b) somebody else besides me wants this to happen.

anko:
I'd want this. Version numbers say nothing about the actual contents—a SHA256 says everything.

gasteve:
Shrinkwrap is practically useless without this...so yes, I would want it.  On the point of security, it would be nice of npm also had some ability to keep a database of code signing keys and verify signatures on the hashes of the downloads or the git commits.  For each module, a set of acceptable / required code signing keys could be specified and used to check signatures.

braydonf:
I think this is a good idea, fills a missing gap to verify that the file downloaded is exactly what is expected.

busches:
@othiym23 @KenanY is this a feature npm would consider adding with a PR?

othiym23:
It's long been a goal of the CLI team to add content hashes to `npm-shrinkwrap.json`. Getting the shasums into the file is relatively simple, because that metadata's available at shrinkwrap time. What's harder is resolving the shasums back to packages at install time. There are brute-force ways to do this that might be relatively simple to code, but I have concerns about how that would affect install performance when used. So, I certainly don't discourage experimentation here, and am happy to review and provide feedback on patches, but be aware that there are dragons here. Thanks for asking!

PR #13398, opened 2016-07-15
Add shasum to shrinkwrap for use during install

misterbyrne:
Optionally including a shasum when adding a package via a URL also allow us to look up packages in the local cache, and validate when adding packages to the cache using only a URL, e.g.

``` bash
npm cache add https://registry.npmjs.org/camelcase/-/camelcase-2.1.1.tgz 7c1d16d679a1bbe59ca02cacecfb011e201f5a1f
```

``` bash
npm cache add https://registry.npmjs.org/camelcase/-/camelcase-2.1.1.tgz 789
pm ERR! Darwin 15.5.0
npm ERR! argv "/Users/adam/.nvm/versions/node/v5.10.1/bin/node" "/Users/adam/src/npm/bin/npm-cli.js" "cache" "add" "https://registry.npmjs.org/camelcase/-/camelcase-2.1.1.tgz" "789"
npm ERR! node v5.10.1
npm ERR! npm  v3.10.6

npm ERR! shasum check failed for /var/folders/g7/8qlj4bk90lg637c449yjg6mr0000gq/T/npm-19243-30cdcc1b/registry.npmjs.org/camelcase/-/camelcase-2.1.1.tgz
npm ERR! Expected: 789
npm ERR! Actual:   7c1d16d679a1bbe59ca02cacecfb011e201f5a1f
npm ERR! From:     https://registry.npmjs.org/camelcase/-/camelcase-2.1.1.tgz
npm ERR! 

npm ERR!     .../npm-debug.log

```

This speeds up shrinkwrap installs quite a lot when the package is locally cached because the package's contents don't have to be validated if the shasum matches the one that's supplied from the shrinkwrap. It also gives us more confidence that the contents of the packages that we are installing are exactly as we expect.

e.g.

``` bash
time node ~/src/npm/bin/npm-cli.js install --ignore-scripts
node ~/src/npm/bin/npm-cli.js install --ignore-scripts  88.82s user 32.35s system 132% cpu 1:31.50 total

node ~/src/npm/bin/npm-cli.js shrinkwrap --dev
rm -rf node_modules

time node ~/src/npm/bin/npm-cli.js install --ignore-scripts
node ~/src/npm/bin/npm-cli.js install --ignore-scripts  55.17s user 15.99s system 125% cpu 56.762 total
```

This PR is illustrative. There are no tests. The cache lookup is synchronous, and could be improved by keeping a `sha => directory` index in the cache somewhere. I'm sure there are also better ways to pass the shasum around.

### DIFF
--- a/lib/cache.js
+++ b/lib/cache.js
@@ -225,16 +225,22 @@ function clean (args, cb) {
 // npm cache add <tarball>
 // npm cache add <folder>
 cache.add = function (pkg, ver, where, scrub, cb) {
+  addWithShaCheck(pkg, ver, where, scrub, null, cb)
+}
+
+cache.addWithShaCheck = addWithShaCheck
+
+function addWithShaCheck (pkg, ver, where, scrub, shasum, cb) {
   assert(typeof pkg === 'string', 'must include name of package to install')
   assert(typeof cb === 'function', 'must include callback')
 
   if (scrub) {
     return clean([], function (er) {
       if (er) return cb(er)
-      add([pkg, ver], where, cb)
+      add([pkg, ver, shasum], where, cb)
     })
   }
-  return add([pkg, ver], where, cb)
+  add([pkg, ver, shasum], where, cb)
 }
 
 var adding = 0
@@ -250,20 +256,26 @@ function add (args, where, cb) {
   // that just a single argument.
 
   var usage = 'Usage:\n' +
-              '    npm cache add <tarball-url>\n' +
+              '    npm cache add <tarball-url> [shasum]\n' +
               '    npm cache add <pkg>@<ver>\n' +
               '    npm cache add <tarball>\n' +
               '    npm cache add <folder>\n'
-  var spec
-
+  var spec, shasum
   log.silly('cache add', 'args', args)
 
   if (args[1] === undefined) args[1] = null
-
+  if (args.length === 3) {
+    shasum = args[2]
+  } else {
+    if (args[0].indexOf('@') && args[1] !== null) {
+      shasum = args[1]
+      args[1] = null
+    }
+  }
   // at this point the args length must ==2
   if (args[1] !== null) {
     spec = args[0] + '@' + args[1]
-  } else if (args.length === 2) {
+  } else {
     spec = args[0]
   }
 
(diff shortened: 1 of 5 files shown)


othiym23:
See https://github.com/npm/npm/issues/10300#issuecomment-247743485 for what would need to be resolved to move this PR forward / remove `needs-discussion`. Be aware that making `npm-shrinkwrap.json` content-aware is one of the cornerstones for the CLI team's plans for `npm@5`, so this is something that's very close to our hearts; it's just more of an architectural yak shave than this PR currently encompasses.

misterbyrne:
From reading the linked comment, it appears that the only real concern is that of reduced install performance. In this PR, the SHA is not used as the sole package identifier, but is used in addition to the package name. This reduces the number of locations within the cache that need to be probed per package.

The best case scenario here, from the point of view of a performance regression, is that there are no packages by that name in the cache. In order to achieve this we need to perform a `readDirSync` which will result in a `NOENT` being thrown.

The worst case scenario is that there are N non-matching versions of this package in the cache and the following needs to be executed:

1. A single `readDirSync` to enumerate the cached versions of the package in question
2. A `readFileSync`, `JSON.parse`, and simple comparison of the SHA stored in the `package.json` in the cache, for each version

This does have the potential to be slow because of the increased I/O. Albeit, a `getdents` type of call and reading and parsing a `package.json` isn't going to be a big deal in the grander scheme of an npm install.

Of course this could be reduced to something a lot simpler if there was a SHA -> package index somewhere. This could be achieved via symlinks or an in-memory index, persisted as a JSON file, for example.

I'd be happy to have a go at implementing the above if there was a chance that this could progress things. What's the consensus on symlinks in the multi-OS world of npm? Would there be a preference for an index file?

misterbyrne:
How would you feel about me splitting this work into two distinct PRs?

One would include the less contentious changes to include the SHA in the shrinkwrap and validate it during install, and the other would include more contentious SHA addressable cache part.

misterbyrne:
The following PR is a subset of this one, without involving a SHA addressable cache. It will however allow us to ensure that the installed dependencies are exactly those that were used to generate the shrinkwrap.

https://github.com/npm/npm/pull/14173

othiym23:
> the only real concern is that of reduced install performance

This is an important and concern. Aside from producing correct and consistent installs, this is the most significant concern around the npm installer. Performance is a feature, and it's a feature that npm CLI users have serious concerns around. I feel strongly enough about this to say that a patch such as this one won't land until there's a fast way to look up package metadata and tarballs by content address.

> What's the consensus on symlinks in the multi-OS world of npm?

That they're a very bad idea, given the many different operating environments under which npm is run. This is one of the reasons why the CLI team doesn't adopt some of the strategies used by other npm package installers like `ied` or `pnpm`.

Issue #13327, opened 2016-07-08
Maximum call stack size exceeded when running npm shrinkwrap on 3.10.6


JEBoothjr:
#### I'm opening this issue because:

  - [x] npm is crashing.
  - [ ] npm is producing an incorrect install.
  - [ ] npm is doing something I don't understand.
  - [ ] Other (_see below for feature requests_):

#### What's going wrong?
In npm 3.10.6, I'm attempting to run npm shrinkwrap and getting a "Maximum call stack size exceeded" error. I don't get this on 3.10.5

#### How can the CLI team reproduce the problem?
I deleted the node_modules folder.
Ran npm cache clean.
Ran npm install
Ran npm shrinkwrap
npm-debug.log: [https://gist.github.com/JEBoothjr/8d724fe37047de5f0f9765190f933c73](https://gist.github.com/JEBoothjr/8d724fe37047de5f0f9765190f933c73)

### supporting information:

 - `npm -v` prints: 3.10.6
 - `node -v` prints: v0.10.36
 - `npm config get registry` prints: https://registry.npmjs.org/
 - Windows, OS X, or Linux?: OS X 10.10.5
 - Network issues:
   - Geographic location where npm was run: 
   - [ ] I use a proxy to connect to the npm registry.
   - [ ] I use a proxy to connect to the web.
   - [ ] I use a proxy when downloading Git repos.
   - [ ] I access the npm registry via a VPN
   - [ ] I don't use a proxy, but have limited or unreliable internet access.
 - Container:
   - [ ] I develop using Vagrant on Windows.
   - [ ] I develop using Vagrant on OS X or Linux.
   - [ ] I develop / deploy using Docker.
   - [ ] I deploy to a PaaS (Triton, Heroku).

SimenB:
Same for me, using `node@6.3.0`

Stack trace:

```
35 verbose stack RangeError: Maximum call stack size exceeded
35 verbose stack     at module.exports (/Users/simbekkh/.nvm/versions/node/v6.3.0/lib/node_modules/npm/node_modules/aproba/index.js:18:42)
35 verbose stack     at moduleName (/Users/simbekkh/.nvm/versions/node/v6.3.0/lib/node_modules/npm/lib/utils/module-name.js:25:3)
35 verbose stack     at exports.isOnlyDev (/Users/simbekkh/.nvm/versions/node/v6.3.0/lib/node_modules/npm/lib/install/is-dev.js:39:45)
[...]

PR #13399, opened 2016-07-15
shrinkwrap: Improve detection of pkgs req'd by top level dev deps

iarna:
Specifically, we now don't blow the stack on dev deps that have cycles.

Fixes: #13327


othiym23:
How do you feel about the idea of some unit tests demonstrating the expected range of inputs for `isDevDep` and `isProdDep`?

othiym23:
The changes look good (if only because they simplify things a bit), but I'm wondering if there's a way we can fence in the behavior here a little more tightly with tests and make this code generally more resistant to regressions and bugs in the future.

iarna:
`is(Dev|Prod)Dep` are pretty minimal functions. Surely I can write tests for them, but I'm a little dubious that it's the best use of our time. They're pretty uninteresting. (My experience is telling me: That's not where we'll have bugs. I would be very surprised if we ever saw a bug in either of those functions.) I guess what I'm saying is: If you feel that's necessary I'll do it, but if I were making the decision I'd move on to other, IMHO more valuable, work.

Putting aside the refactoring, there are only two substantive changes here:

https://github.com/npm/npm/pull/13399/files#diff-14efa09b371358c4c6934dc02a7884efR158

Whence I added `seen` to break cycles.

And 

https://github.com/npm/npm/pull/13399/files#diff-97a53acbc070b7edcd97498046a5eee3R53

Where I simplified the test as we don't need to identify transitive dev deps, just direct ones.

If your goal is just to have more testing around this corner I'd prioritize unit testing of `isOnlyDev` well ahead of them. It's not exported, but it is a moderately complex recursive function, of sufficient obtuseness that I wrote substantial comments for it.

iarna:

Issue #13088, opened 2016-06-17
npm 3 dependencies of bundled dependencies missing from package


stevengill:
So my package has some `bundleddependencies`. When I run `npm pack`, those `bundleddependencies` are included in the package but the `dependencies` of those `bundleddependenceis` are missing. This is because `npm@3` installs the dependencies as flat. So the deps of my bundleddependenceis are siblings instead of nested.

Two short term solutions for this problem

1) I use npm@2 and run `npm install` before I run `npm pack` to ensure the deps are nested.
2) I include the dependenceis of my `bundledDependencies` as `bundledDependenceis` in my `package.json`

If I'm npm installing the package, not having the dependencies of of `bundledDependencies` isn't a big deal as they will get fetched and installed. But I thought the point of `bundledDependenceis` were that they would be included in the package and wouldn't require extra fetching.

In Cordova, we `npm cache add` some modules instead of `npm install` them. This means, we are relying on `bundledDependencies` to include all necessary dependencies since we don't run `npm install`. In the next major version of Cordova, we are switching to using `npm install` instead of `npm cache add`. 

Thanks! You all are doing an amazing job :)

othiym23:
So, you'll want to read [this chunk of the release notes](https://github.com/npm/npm/blob/b38573fae383aaaeaec0548c2ae616778a3b5dab/CHANGELOG.md#erk) of the new `npm@next` to understand what happened here (you might want to [read this (for context)](https://github.com/npm/npm/blob/b38573fae383aaaeaec0548c2ae616778a3b5dab/CHANGELOG.md#whoa) too). This was an entirely unintentional regression, and should be fixed now. Thanks so much for bringing this to our attention, Steve!

PR #13438, opened 2016-07-20
scoped bundledDeps fixes

### LINKED ISSUE #8614: npm pack doesn't bundle scoped bundledDependencies
test package.json
```json 
{
  "name": "npm-bundle-test",
  "version": "1.0.0",
  "dependencies": {
    "@myscope/debug": "*"
  },
  "bundledDependencies": [
    "@myscope/debug"
  ]
}
```

`npm install` creates `node_modules/@myscope/debug/`, but `npm pack` ignore all scoped bundled dependencies 

```sh
# tar -tvf npm-bundle-test-1.0.0.tgz`
-rw-r--r-- 501/20          152 2015-06-19 16:09 package/package.json
```

It probably does not work because scoped modules are in @*scope* subdirectory.

zkat:
@forivall kindly set up a pr over at https://github.com/npm/fstream-npm/pull/22 to fix #8614. This branch adds a basic bundledDeps test to verify the changes, and will eventually include the dep upgrade commit.

### DIFF
--- a/lib/utils/tar.js
+++ b/lib/utils/tar.js
@@ -117,7 +117,17 @@ BundledPacker.prototype.applyIgnores = function (entry, partial, entryObj) {
   // if they're not already present at a higher level.
   if (this.bundleMagic) {
     // bubbling up.  stop here and allow anything the bundled pkg allows
-    if (entry.indexOf('/') !== -1) return true
+    if (entry.charAt(0) === '@') {
+      var firstSlash = entry.indexOf('/')
+      // continue to list the packages in this scope
+      if (firstSlash === -1) return true
+
+      // bubbling up.  stop here and allow anything the bundled pkg allows
+      if (entry.indexOf('/', firstSlash + 1) !== -1) return true
+    // bubbling up.  stop here and allow anything the bundled pkg allows
+    } else if (entry.indexOf('/') !== -1) {
+      return true
+    }
 
     // never include the .bin.  It's typically full of platform-specific
     // stuff like symlinks and .cmd files anyway.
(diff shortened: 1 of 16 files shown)


forivall:
Hmm, it looks like there was bundledDependency checking being done in tar as well: Once I removed it, it works!

I just did a quick visual inspection, so I don't know if anything was different between tar.js's `BundledPacker.prototype.applyIgnores` and fstream-npm's `Packer.prototype.applyIgnores`.

https://github.com/forivall/npm/commit/27fdf27674aaecd3a5923b94f3706cafffc0f882

zkat:
Hey @forivall -- I know this looks terrible (and it is), but... that's actually kinda how it has to be. And it keeps biting us when we forget. That whole stack is getting rewritten eventually, so the only thing you can really do is duplicate whatever code you need to dupe across both implementations :( Otherwise, things start going awry. It's basically needed for transitive dependencies (see #13088 and #13089)

forivall:
ah, yeah, i diffed it

https://gist.github.com/forivall/58d56561293b6740241343063d8657cc

Well, that explains why transitive dependencies weren't being included when I was using fstream-npm directly for a bundler, and I had to resort to this hack: https://gist.github.com/forivall/80b4284065a354f72f1766504dc7aa5d

So, I guess my change in fstream-npm needs to be duplicated in tar.js's `BundledPacker.prototype.applyIgnores` 😿

forivall:
https://github.com/npm/npm/compare/npm:zkat/scoped-bundled-fix...forivall:forivall/scoped-bundled-fix

iarna:
Essentially `fstream-npm` holds the npm@2 logic, and `lib/utils/tar.js` holds the npm@3 logic.

Yeah, this sucks– it was necessitated by the npm@3 bundle dependency logic needing access to npm@3 internals and `fstream`'s extension mechanism not allowing finer grained insertion of them.

That last patch looks good to me.

zkat:
This is in our next release branch and will be included in the next release :)

PR #13483, opened 2016-07-26
test: fix async version in update-symlink.js

MylesBorins:
A hardcoded value included in `test/update-symlink.js` is no
longer accurate and is causing test failures. This commit
updates the value in order to see a green test suite.

It seems a bit fragile that an update to a module could break
the test suite, is there something I am missing? If there is
a way to modify this to be less fragile please let me know.


iarna:
I've updated this test a bit more comprehensively over in 4bff74a, which will be landing today.

PR #13493, opened 2016-07-26
Document that per-user config file isn't hardcoded

jasonkarns:
Document that the per-user config file respects the `$NPM_CONFIG_USERCONFIG` env var, defaulting to ~/.npmrc


zkat:
Thanks, @jasonkarns! This was included as part of `npm@3.10.8` and `npm@2.15.11`. 🎉

PR #13519, opened 2016-07-28
Don't crash when node.requiredBy is missing.

creationix:
This fix works for us in that we're unblocked and can use the newer npm on our old codebase with some probably malformed packages.

``` sh
$ ls -lh node_modules
...
lrwxrwxrwx  1 creationix admins   19 Jul 28 20:09 thrift -> thrift-performance/
drwxrwsr-x  5 creationix admins 4.0K Jul 28 20:31 thrift-performance
...
```

node_modules/thrift-performance/package.json:

``` json
{
  "name": "thrift-performance",
  "description": "node-thrift",
  "version": "0.6.0dev5",
  "author": "Wade Simmons <wade@wades.im>",
  "directories" : { "lib" : "./lib/thrift" },
  "main": "./lib/thrift",
  "engines": { "node": ">= 0.2.4" },
  "dependencies": {
    "node-int64": "0.2.x"
  }
}
```

There may be a better way to handle this edge case, but this patch at least prevents the crash and reports something.

### DIFF
--- a/lib/install/is-dev.js
+++ b/lib/install/is-dev.js
@@ -20,7 +20,7 @@ function andIsDev (name) {
 }
 
 exports.isDev = function (node) {
-  return node.requiredBy.some(andIsDev(moduleName(node)))
+  return node.requiredBy && node.requiredBy.some(andIsDev(moduleName(node)))
 }
 
 function andIsOnlyDev (name) {
--- a/lib/install/mutate-into-logical-tree.js
+++ b/lib/install/mutate-into-logical-tree.js
@@ -38,7 +38,7 @@ var mutateIntoLogicalTree = module.exports = function (tree) {
 
   Object.keys(flat).sort().forEach(function (flatname) {
     var node = flat[flatname]
-    if (!node.requiredBy.length) return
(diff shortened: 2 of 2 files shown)


creationix:
Hmm, I'm finding more places where this missing `node.requiredBy` property is causing crashes.  It might be best to figure out the root cause of it being missing.    Let me know if I can help you track that down.

iarna:
Yeah, it's _really_ not supposed to be possible for it to be missing.

iarna:
I ... don't get a crash with the example you laid out up there. What command are you running?

I _do_ get empty results from `npm outdated`–I would have sworn we'd fixed this or had a branch for fixing it, but apparently not.

creationix:
I wish I could reduce the test case.  I do know that the npm that comes with node 0.10.x doesn't have this problem and the one that comes with node 6.x does.  Sorry if that doesn't help much, there have probably been a *ton* of npm changes in the intervening years.

Is there anything you would like me to test or look for in our code base?  The symlink was my initial guess, but it might be something different.

zkat:
This has been merged into `release-next` and will be included in `npm@4`

Issue #12483, opened 2016-04-26
package.json/script: path/to/a/script.js


sarbbottam:
`npm scripts` are great, I personally prefer it over any other workflow tools. However, its bloating the `package.json` with [excessive usage](https://github.com/ReactiveX/rxjs/blob/a3ec89605a24a6f54e577d21773dad11f22fdb14/package.json#L14-L96)

[p-s](https://www.npmjs.com/package/p-s) has a workaround to mitigate it.

Please find the excerpt from [p-s](https://www.npmjs.com/package/p-s)'s [readme]

> npm scripts have a ton of advantages ([learn more](https://medium.freecodecamp.com/why-i-left-gulp-and-grunt-for-npm-scripts-3d6853dd22b8#.9qghcfdr9)), it can grow into an
> [unmaintainable mess](https://github.com/ReactiveX/rxjs/blob/a3ec89605a24a6f54e577d21773dad11f22fdb14/package.json#L14-L96) in your `package.json` file. Part of the problem is we're configuring scripts in `json` which has fundamental issues (like no comments).

---

Wondering if a `file path` can be passed instead of an `object like` in `package.json`

```
...
"scripts": "path/to/scripts.js",
...
```

[run-script](https://github.com/npm/npm/blob/master/lib/run-script.js) on reading the `package.json/scripts`

```
if scripts is a string and path to a file
   read the file 
   business as usual, 
else 
   business as usual
```

sarbbottam:
If this is accepted, I am happy to work on a PR, with a little guidance.

gustavnikolaj:
This issue seems to be conflicting with the answer that @othiym23 gave [here](https://github.com/npm/npm/issues/11418#issuecomment-180640530) in issue #11418

While explaining why the proposal weren't good enough, an argument were made against changing the scripts property.

> As @iarna wrote, the technical problem with using arrays and nested object literals to decompose package scripts is that old versions of npm aren't forward-compatible with either, meaning this is a breaking change to the ecosystem. 

The proposal to change it to a reference to a file would have the exact same problem. :-)

Issue #11418, opened 2016-02-04
npm as a first-class task runner


boneskull:
*If this is a dupe, I apologize.  I'm always leery of adding to the pile of issues here.  However, I wanted to begin discussion around this (besides, it's more fun to think about than some `npm install` edge case).*

Many users are abandoning Gulp and Grunt for the npm `scripts` property.  I don't claim to know *all* of the reasons why this is happening, but I can gather a few things:

- Certain lifecycle scripts are not runnable w/o the development dependencies (ostensibly Grunt/Gulp) installed.  This means a user often has to put things in `scripts` *anyway*.  :+1: for keeping your scripts all in one place.
- Task runner plugins introduce bottlenecks to upgrades.  Introducing a dependency on Gulp or Grunt means you will also have dependencies on a variety of plugins.  If you wish to upgrade a package (say, `uglifyjs`), you may have to wait on the maintainer of `gulp-uglifyjs` or `grunt-uglifyjs` to ensure compatibility and update *that* package as well.
- npm is always present.  You do not need to add it as a dependency or development dependency.

On the main, npm works pretty well for this.  However, there are limitations to npm as a task runner (see #6943, #11291 and probably others).  For example, here's a (real) sample `scripts` property I'm working with:

```json
{
  "scripts": {
    "build": "parallel build:node build:browser",
    "build:node": "rimraf lib && babel src --out-dir lib",
    "build:browser": "rimraf dist && mkdirp dist && parallel build:browser:main build:browser:min",
    "build:browser:main": "browserify --transform [ exposify --expose [ --angular angular ] ] --plugin [ bannerify --file .banner.ejs --pkg package.json ] src --outfile dist/angular-loggly-mixin.js",
    "build:browser:min": "browserify --transform [ exposify --expose [ --angular angular ] ] --plugin [ bannerify --file .banner.ejs --pkg package.json ] --global-transform uglifyify src --outfile dist/angular-loggly-mixin.min.js",
    "bump:major": "npm version major -m 'Release v%s'",
    "bump:minor": "npm version minor -m 'Release v%s'",
    "bump:patch": "npm version patch -m 'Release v%s'",
    "check": "parallel nsp deps",
    "deps": "parallel deps:missing deps:extra",
    "deps:extra": "dependency-check package.json --no-default-entries --entry src/index.js --extra --no-dev",
    "deps:missing": "dependency-check package.json --no-default-entries --entry src/index.js --missing",
    "deps:update": "updtr",
[...]

Issue #11284, opened 2016-01-26
Proposal: { "scripts": "moduleA" }


donaldpipowitch:
It is quite common in the Node community to use `$ npm run` for build scripts and other scripts like tests. See [here](http://substack.net/task_automation_with_npm_run) for an older example and [here](https://medium.com/@housecor/why-i-left-gulp-and-grunt-for-npm-scripts-3d6853dd22b8) for a newer example. In my opinion this is generally a good thing as you have fewer dependencies on plugin authors for task runners like Grunt or Gulp as they aren't often _that_ necessary and add complexity. There _are_ some pitfalls you still need to keep in mind like _"Don't use comments in JSON files."_ (surprisingly unexpected for many people) or _"Don't use `$ rm -rf`, but the `rimraf` module for cross-platform compability."_ What is _really_ great about `$ npm run` is that you can use common commands like `$ npm run start` (or `$ npm start`) across projects without learning new commands like `$ grunt` or `$ gulp` and without installing additional modules like `grunt-cli` or `gulp-cli`.

But `$ npm run` can also be very limiting. Look at [this `package.json`](https://github.com/AngularClass/angular2-webpack-starter/blob/master/package.json#L8). It is not uncommon to have so many commands/scripts. We certainly have them. But we also have many projects which all use the same scripts and configurations. So how do we share and version the `scripts` field in our `package.json`? It is not really possible. If we just change one tool, we need to update all `package.json`s manually. We have more constraints. We have many people who are happy command line users and use our tools, but aren't JavaScript developers. We want to offer them a nice experience on the command line without the need to look in our JavaScript source code. Grunt allows us to list all commands with descriptions via `$ grunt -h`. Can we do the same with `$ npm run -h`? Sadly, no.

So what do we do? We don't use Grunt (at least not anymore), we don't use Gulp, we don't use `$ npm run`. We have a small module which acts as a nice little facade to other Node modules we need for our tooling like Babel, JSCS, Karma and so on. This facade is just a small commander.js application which offers a nice and coherent command line interface for a multitude of different Node modules - each offering other interfaces. Our developers learn our interface with generic command like `start`, `test`, `serve` and so on and we as JavaScript developers swap out the underlying modules as needed. (Like switching for JSHint to ESLint or something like, but keep a single `lint` command.)
[...]

Issue #6943, opened 2014-12-15
npm scripts properties can be read as arrays


keithamus:
Sometimes when writing npm scripts, they can start to get a little bit unwieldy if you try to string multiple commands together. Take for example this: 

```json
"scripts": {
  "build": "npm run build:scripts -s && npm run build:styles -s && npm run build:markup -s",
  "build:scripts": "browserify -d assets/scripts/main.js -p [minifyify --compressPath . --map main.js.map --output dist/main.js.map] | hashmark -n dist/main.js -s -l 8 -m assets.json 'dist/{name}{hash}{ext}'"
}
```

It'd be real swell if npm could read a scripts value as an array, and just run `.join('')` before executing, which in my humble opinion would improve the readability of scripts immensely:

```json
"scripts": {
  "build": [
    "npm run build:scripts -s",
    "&& npm run build:styles -s",
    "&& npm run build:markup -s"
  ],
  "build:scripts": [
    "browserify -d assets/scripts/main.js -p ",
    "[minifyify --compressPath . --map main.js.map --output dist/main.js.map]",
    " | hashmark -n dist/main.js -s -l 8 -m assets.json 'dist/{name}{hash}{ext}'"
  ]
}
```

j-san:
Maybe a solution for "&&" operator in windows shell ?
:)

smikes:
@j-san That would be interesting - it could be an `npm` module that uses forward polish (instead of infix) notation

```
npm install and
...
"scripts": { "build": "and \"npm run build:scripts -s\" \"npm run build:styles -s\" " }
...
> npm run build
```

sindresorhus:
:+1:

Qix-:
Why introduce another parser do to weird, non-standard command linkage using `and` instead of using JSON how people would expect it? And why not make it an optional thing, similar to how `repository` works? If it's a string, it's one command. If it's an array, it's multiple commands that are run in succession so long as they exit with `0`.

I was actually kind of surprised NPM doesn't do this by default; as discussed in sindresorhus/xo-init#6, `&&` has its downfalls. Further, being dogmatic about it, while realistically `&&` is pretty reliable, it shouldn't be expected to work on every platform, always.

As stated in the aforementioned issue, manipulating the `package.json` data automatically now becomes a hassle because we have to *guess* (never good) the format of the command string; does it contain a `;` at the end? Does it already have some weird postfix operator that will cause unintended side-effects with `&&`?

[...]

PR #13578, opened 2016-08-04
run-script: __catchall handler

gustavnikolaj:
People want to use `npm run-script` as a simple task runner. Most of the time, a dedicated task runner is overkill, or simply just not nice to work with. Tools like Grunt, Gulp, and Broccoli all define their own structures and require you to install plugins to solve even the most trivial tasks like running eslint. GNU Make is currently one of the best alternatives, as it integrates nicely with CLI tools, which most of the tools we integrate with exports anyway.

The downside to GNU Make is that it needs to be installed before you can run the targets and it is not as portable as a pure javascript solution.
## The proposal

Introduce a new special cased script in the scripts block called `__catchall`, which will be called when the script that was asked for did not exist.

``` json
{
  "name": "some-package",
  "version": "1.0.0",
  "scripts": {
    "__catchall": "node scripts.js"
  }
}
```

``` js
console.log('process.argv', process.argv);
```

```
$ npm run build

> some-package@1.0.0 build /.../some-package
> node scripts.js

process.argv [ 'node', '/.../some-package/scripts.js', 'build' ]
```

This will allow people to implement any script resolution algorithm that they desire, while keeping the impact on npm itself minimal.

It is called slightly different than other scripts, as the name of the script is to the list of arguments (notice the third item in the argv array). This will, among other things, allow you to use e.g. Gulp through `__catchall` instead of requiring a globally installed version:

``` json
{
  "name": "some-package",
  "version": "1.0.0",
  "scripts": { "__catchall": "gulp" },
  "devDependencies": { "gulp": "*" }
}
```

```
$ npm run build

> some-package@1.0.0 build /.../some-package
> gulp build
...
```

Most existing task runners will seamlessly integrate this way, without any changes to them, including:
- GNU Make
- [p-s](https://www.npmjs.com/package/p-s) (which at the moment recommends using `npm start` to achieve something similar)
- [scripty](https://www.npmjs.com/package/scripty) (this proposal solves the
  problem of maintaining the scripts object manually for scripty)
## A basic catchall script

The following is one of the simpler ways of implementing a catchall handler, which will allow you to author your scripts as node code.

``` js
// scripts.js
const script = process.argv[2]
const scripts = {
  build: function () {
    // do the build
    console.log('Done!')
  }
};

if (scripts[script]) {
  scripts[script]()
} else {
  console.error('no such script')
  process.exit(1)
}
```

```
$ npm run build

> some-package@1.0.0 build /.../some-package
> node scripts.js

Done!
$ npm run foobar

> some-package@1.0.0 foobar /.../some-package
> node scripts.js

no such script

npm ERR! Linux 4.4.0-31-generic
npm ERR! argv "node" "/.../v0.10.38/bin/npm" "run" "foobar"
npm ERR! node v0.10.38
npm ERR! npm  v3.10.6
npm ERR! script "foobar" errored when passed to catchall script "node scripts.js"
...
```
## Notes on backwards compatibility

To mitigate backwards compatibility concerns not all scripts will be passed on to the catchall handler. Specifically the relevant to core npm scripts, mentioned in the [scripts documentation](https://docs.npmjs.com/misc/scripts).

This will make sure that people will not start relying on the catchall functionality to run code that is required when installing, or other cases that could potentially be a problem for people with older npm clients.

The use case for this feature is mostly development centric. The scripts I have which it would make sense to migrate to a solution based on this will typically handle stuff like building, testing and preparation for publishing.

It is still possible to run the blacklisted script names through the catchall handler, you will just have to explicitly list them in the scripts section, as you would have done without this feature.

``` json
{
  "name": "some-package",
  "version": "1.0.0",
  "scripts": {
    "__catchall": "make",
    "prepublish": "make prepublish"
  }
}
```
## pre and post hooks

When you choose to shell out to a more powerful tool for your task running needs you are likely to handle complex logic and dependencies of your script within that new tool, thus making the pre and post hooks less relevant for scripts ran outside of npm.

When npm shells out to the catchall script, it will thus not call the pre and post versions of that script, unless they are explicitly defined in the scripts object.

``` json
{
  "scripts": { "__catchall": "echo" }
}
```

```
$ npm run build
build
```

``` json
{
  "scripts": {
    "prebuild": "echo prebuild!",
    "__catchall": "echo"
  }
}
```

```
$ npm run build
prebuild!
build
```
## Why not just build a separate task runner?

This solution is as minimal as possible, and basically just allows you to hook in after npm gave up on finding the script to run. So, why is this better than a standalone task runner? What does this proposal provide that wouldn't be handled just as good by a standalone utility?
1. The task runner would need to be installed as a dependency. That is not necessary with this solution.
2. The task runner needs to be installed globally or the user needs to add `node_modules/.bin` to their `$PATH` for it to be available, where `npm run-script` is available already.
3. This solution gets access to the npm environment variables which is also available when you run commands from `npm run-script` or run `npm start`. This means that a task runner based on this proposal could use the [config property](https://docs.npmjs.com/files/package.json#config) in `package.json` instead of introducing another rcfile or dotfile.
## Why use a special cased property in the scripts object?

If the magic property `__catchall` is not desirable, a new package.json property could be introduced, with a name like `scriptsCatchAll` or `scriptResolver`, but otherwise working in the same way. This could make it less obscure and feel less magic.

The argument to keep it as a property on the script object is that it opens a door for people to use it on older versions of npm. If the script name is passed as the first argument (as discussed in the section above) it would allow npm versions 2 and later to use it like so:

```
$ npm run __catchall -- build # npm without __catchall support
$ npm run build # npm with __catchall support
```

It's not beautiful, but it will at least make it usable everywhere.
### Naming

An alternate name for `__catchall` could be `*`. I'm honestly not sure about it, but I'm leaning towards `*` now. I think that naming was the least interesting part of this proposal, so I decided to submit it as is.
### Script Completion

Currently, if you try to set up `npm run-script` shell completion with this you will get `__catchall` listed as a potential candidate, but not any of the scripts it implement.

I'm not sure how best to support this, if at all, so for now, it is left unanswered. My best idea on the spot is to make a convention that the tool can expose a special command that `npm run-script` will call, which could be used to return the list of available targets.
### Documentation

I did not write any documentation to accompany this pull request, as I would like the proposal validated first.
## Similar Requests:
- https://github.com/npm/npm/issues/12483 (allow script to be a reference to a file)
- https://github.com/npm/npm/issues/11418 (nested objects and arrays @boneskull)
- https://github.com/npm/npm/issues/11284 (allow script to be a reference to a module/file)
- https://github.com/npm/npm/issues/6943 (script entries as arrays that will be auto-concatted)

Thanks to @alexjeffburke and @papandreou for your help with this proposal.

### DIFF
--- a/lib/run-script.js
+++ b/lib/run-script.js
@@ -62,20 +62,21 @@ function runScript (args, cb) {
   })
 }
 
+var cmdList = [
+  'publish',
+  'install',
+  'uninstall',
+  'test',
+  'stop',
+  'start',
+  'restart',
+  'version'
+].reduce(function (l, p) {
+  return l.concat(['pre' + p, p, 'post' + p])
+}, [])
+
 function list (cb) {
   var json = path.join(npm.localPrefix, 'package.json')
-  var cmdList = [
-    'publish',
-    'install',
-    'uninstall',
-    'test',
-    'stop',
-    'start',
-    'restart',
-    'version'
-  ].reduce(function (l, p) {
-    return l.concat(['pre' + p, p, 'post' + p])
-  }, [])
   return readJson(json, function (er, d) {
     if (er && er.code !== 'ENOENT' && er.code !== 'ENOTDIR') return cb(er)
     if (er) d = {}
@@ -127,6 +128,7 @@ function run (pkg, wd, cmd, args, cb) {
   if (!pkg.scripts) pkg.scripts = {}
 
   var cmds
+  var catchAll = false
   if (cmd === 'restart' && !pkg.scripts.restart) {
     cmds = [
       'prestop', 'stop', 'poststop',
@@ -147,6 +149,8 @@ function run (pkg, wd, cmd, args, cb) {
         }
       } else if (npm.config.get('if-present')) {
         return cb(null)
+      } else if (pkg.scripts['__catchall'] && cmdList.indexOf(cmd) === -1) {
+        catchAll = true
       } else {
         return cb(new Error('missing script: ' + cmd))
       }
@@ -155,7 +159,18 @@ function run (pkg, wd, cmd, args, cb) {
   }
 
   if (!cmd.match(/^(pre|post)/)) {
-    cmds = ['pre' + cmd].concat(cmds).concat('post' + cmd)
+    if (catchAll) {
+      // only run pre|post hooks for __catchall when they are explicitly defined
+      // in the scripts section
+      if (pkg.scripts['pre' + cmd]) {
+        cmds = ['pre' + cmd].concat(cmds)
+      }
+      if (pkg.scripts['post' + cmd]) {
+        cmds = cmds.concat('post' + cmd)
+      }
+    } else {
+      cmds = ['pre' + cmd].concat(cmds).concat('post' + cmd)
+    }
   }
 
   log.verbose('run-script', cmds)
@@ -163,11 +178,19 @@ function run (pkg, wd, cmd, args, cb) {
     // pass cli arguments after -- to script.
     if (pkg.scripts[c] && c === cmd) {
       pkg.scripts[c] = pkg.scripts[c] + joinArgs(args)
+    } else if (catchAll && c === cmd) {
+      // pass script name as well as arguments when defering to __catchall
+      pkg.scripts[c] = pkg.scripts['__catchall'] + joinArgs([c].concat(args))
     }
 
     // when running scripts explicitly, assume that they're trusted.
     return [lifecycle, pkg, c, wd, true]
-  }), cb)
+  }), function (err) {
+    if (err && catchAll && err.stage === cmd) {
+      return cb(new Error('script "' + err.stage + '" errored when passed to catchall script "' + pkg.scripts['__catchall'] + '"'))
+    }
+    return cb.apply(null, arguments)
+  })
 }
 
 // join arguments after '--' and pass them to script,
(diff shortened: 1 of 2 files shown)


iarna:
As I didn't note previously: This is purely a product management discussion and thus requires that the CLI team put our heads together and come to a consensus on the feature.

This makes me uncomfortable because it complicates our story around how scripts work. It also further takes them from being primarily package lifecycle scripts and more in to a task-runner mode.

Further it means if you typo a script name, instead of getting an error from npm you'll get an error from an unrelated tool.

Which gets into my other issue with this. It's only used when `npm run` is used, which means `npm run install` might do something even though the module doesn't have an install lifecycle (and won't run one when installing the module). I'm not comfortable losing the symmetry between the behavior when `npm` runs a script for its own reasons and a user runs the same script.

alexjeffburke:
Just a thought, but given the list of top level commands is stable, would it address the concern about for example `npm run install` if it and it's peers were excluded from the set of allowed deferrals?

iarna:
We are not going to have `__catchall` scripts because we don't want npm to be a general task runner. I would encourage you to distribute your own cli to be called directly or via `npx`. Sorry that I'm not more helpful here, we do appreciate these kinds of PRs even when we don't accept them. If you've got further ideas around this sort of thing, I encourage you to join [package.community](http://package.community), where many package manager maintainers hang out.

PR #13579, opened 2016-08-04
Do not include optionalDependencies in shrinkwrap

palmerj3:
Shrinkwrap and optionalDependencies are not interoperable..

The purpose of a shrinkwrap is to get as close to hermetic builds as possible but optionalDependencies do not allow this to happen since, theoretically, different dependencies can be installed on different platforms.

My suggestion is to remove optionalDependencies from generated shrinkwrap files in order to achieve truly cross-platform hermetic builds.


palmerj3:
@iarna does the needs-discussion label imply we'll be discussing here or ... ?

iarna:
Eliminating optional dependencies from the `npm-shrinkwrap.json` file will not result in their not being installed. When npm finds items in your `package.json` that aren’t in your `npm-shrinkwrap.json` it will still try to install them. Removing optional dependencies from `npm-shrinkwrap.json` thus does not allow you to get exactly the same build across platforms. (This behavior is required by existing packages, most notably [hapi](https://github.com/hapijs/hapi/blob/master/npm-shrinkwrap.json).)

This does not improve the situation around identical cross platform builds and so we’re not going to take this change. We have made some other changes related to optional dependencies that you may find interesting, however:

[#10073](https://github.com/npm/npm/pull/10073) annotates the `npm-shrinkwrap.json` with the optional status of a package.

[#13692](https://github.com/npm/npm/pull/13692) ensures that optional dependencies found in shrinwkraps that are incompatible with your architecture or operating system will not be attempted. Previously their installation would be attempted resulting in confusing warning messages (and wasted time waiting on failed compiles).

You can also use `--no-optional` to skip the installation of any optional dependencies. This would ensure that none of your environments get dependencies that might not be installable everywhere.

Thank you for your patience in waiting on us to get back to you on this and for putting this together in the first place.

palmerj3:
If NPM is installing things that aren't defined in npm-shrinkwrap.json then I would consider that a bug... that really defeats the whole purpose of npm-shrinkwrap.. Ok thank you for the feedback. I will submit a more impactful PR soon.

PR #13580, opened 2016-08-04
config: pretend login is a toplevel command

zkat:
So @substack brought this up over at #13581 and @ashleygwilliams mentioned it to me, so I just tossed a quick PR together.

### DIFF
--- a/lib/config/cmd-list.js
+++ b/lib/config/cmd-list.js
@@ -66,6 +66,7 @@ var cmdList = [
   'stars',
   'tag',
   'adduser',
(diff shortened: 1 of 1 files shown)


lib/config/cmd-list.js
```diff
@@ -66,6 +66,7 @@ var cmdList = [
   'stars',
   'tag',
   'adduser',
+  'login',
```
iarna: maybe a comment noting that it's an alias for adduser?

iarna:
keen <3

iarna:

PR #13598, opened 2016-08-07
Explain how to set node env in bin scripts

### LINKED ISSUE #12438: Document that bin scripts should start with #!/usr/bin/env node so npm install will create proper node_modules/.bin scripts
If the .js script referenced in package.json > [bin](https://docs.npmjs.com/files/package.json#bin) does not start with  `#!/usr/bin/env node`, `npm install` generates a script under node_modules/.bin which starts directly the .js script without node executable.
This does not work at least on Windows.

This seems to be handled by [cmd-shim](https://github.com/ForbesLindesay/cmd-shim)

mxstbr:
Closes #12438

_I'm totally not here [for the socks](https://twitter.com/erikawh0/status/762183602519977985)…_

### DIFF
--- a/doc/files/package.json.md
+++ b/doc/files/package.json.md
@@ -247,6 +247,10 @@ would be the same as this:
     , "version": "1.2.5"
     , "bin" : { "my-program" : "./path/to/program" } }
 
+Please make sure that your file(s) referenced in `bin` starts with
+`#!/usr/bin/env node`, otherwise the scripts are started without the node
+executable!
+
 ## man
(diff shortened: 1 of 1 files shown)


zkat:
Yay! Thanks for the patch, @mxstbr! This was included as part of both `npm@3.10.8` and `npm@2.15.11`!

If you wanna get them socks, feel free to shoot an email over to `support@npmjs.org` and link them to this issue. They'll hook you up. :)

PR #13613, opened 2016-08-09
Reduce 'skipping action' log level from 'warn' to 'verbose'.

### LINKED ISSUE #9999: [npm3] Install with a symlinked module results in lots of warnings (for no reason)
During a `npm install` all modules in `node_modules` will be analyzed to build up the ideal dependency tree ... as far as I understand ...

But this doesn't make any sense for modules that are symlinked. This is also realized by `npm` and results in a bunch of warn messages: 

```
npm WARN skippingAction Module is inside a symlinked module: not running add ...
```

So why analyzing this modules anyway? Wouldn't it make sense to filter these out in an earlier install step. Because besides being unnecessary it also takes a lot of time, depending on the size of the module of course.

timoxley:
> npm WARN skippingAction Module is inside a symlinked module: not running …

This message is incredibly verbose and not very helpful, as far as I can tell. If a user has symlinked packages into `node_modules` they've very likely done this on purpose. A user certainly does not need a warning for every single sub-dependency of the symlinked dependency i.e. hundreds of lines of non-actionable output.

Closes #9999.

image

### DIFF
--- a/lib/install/filter-invalid-actions.js
+++ b/lib/install/filter-invalid-actions.js
@@ -24,7 +24,7 @@ module.exports = function (top, differences, next) {
     if (pkg.isInLink || (pkg.parent && (pkg.parent.target || pkg.parent.isLink))) {
       // we want to skip warning if this is a child of another module that we're removing
       if (!pkg.parent.removing) {
-        log.warn('skippingAction', 'Module is inside a symlinked module: not running ' +
+        log.verbose('skippingAction', 'Module is inside a symlinked module: not running ' +
           cmd + ' ' + packageId(pkg) + ' ' + path.relative(top, pkg.path))
       }
     } else {


timoxley:
A single test [when update is called linked packages should be excluded](https://github.com/npm/npm/blob/6da8d1d47eb3f933fffd47952079dd3d94e9fe25/test/tap/update-symlink.js#L64) from `test/tap/update-symlink.js` fails on my local machine, but also fails on the master so I am to conclude it was not my change that lead to this breakage.

Actually it appears to be because it's checking for `async@1.5.2` after an `npm update` but `async@2.0.0` is now released so this test breaks (and is not a very good test).

It seems the test suite couldn't have been passing (using public registry data anyway) since `async@2.0.0` was released on `2016-07-13`?

iarna:
Yeah, I have a fix for that test issue (and many other "why are we hitting the network in a test?" issues) in the upcoming release (to be merged to master later today).

iarna:
I'd be uncomfortable about taking this (It's not supposed to be warning about more than the top level symlink) except that the plan is to make symlinked modules entirely opaque in which case it won't be looking inside them _at all_. So I'll take this as a stop-gap.

iarna:

Issue #11398, opened 2016-02-03
patr dependency fails on 3.6.0


winek:
```js
{ 'diagoa-core': '0.0.0',
  npm: '3.6.0',
  ares: '1.10.1-DEV',
  http_parser: '2.6.0',
  icu: '56.1',
  modules: '47',
  node: '5.5.0',
  openssl: '1.0.2e',
  uv: '1.8.0',
  v8: '4.6.85.31',
  zlib: '1.2.8' }
```

`npm install compose` fails on patr dependency:

![image](https://cloud.githubusercontent.com/assets/2938672/12782943/b6d07a64-ca7e-11e5-9841-5541c696f11d.png)

```
0 info it worked if it ends with ok
1 verbose cli [ '/usr/bin/node', '/usr/bin/npm', 'install', '--save', 'compose' ]
2 info using npm@3.6.0
3 info using node@v5.5.0
4 silly loadCurrentTree Starting
5 silly install loadCurrentTree
6 silly install readLocalPackageData
7 silly fetchPackageMetaData compose
8 silly fetchNamedPackageData compose
9 silly mapToRegistry name compose
10 silly mapToRegistry using default registry
11 silly mapToRegistry registry https://registry.npmjs.org/
12 silly mapToRegistry uri https://registry.npmjs.org/compose
13 verbose request uri https://registry.npmjs.org/compose
14 verbose request no auth needed
15 info attempt registry request try #1 at 14:00:05
16 verbose request id db394a2c7acc890f
17 verbose etag "8FO5F8FYAPVQ3ONV2LO5EASCI"
18 http request GET https://registry.npmjs.org/compose
19 http 304 https://registry.npmjs.org/compose
20 verbose headers { date: 'Wed, 03 Feb 2016 13:00:05 GMT',
20 verbose headers   via: '1.1 varnish',
20 verbose headers   'cache-control': 'max-age=300',
20 verbose headers   etag: '"8FO5F8FYAPVQ3ONV2LO5EASCI"',
20 verbose headers   age: '0',
20 verbose headers   connection: 'keep-alive',
20 verbose headers   'x-served-by': 'cache-ams4142-AMS',
20 verbose headers   'x-cache': 'MISS',
20 verbose headers   'x-cache-hits': '0',
20 verbose headers   'x-timer': 'S1454504405.745262,VS0,VE336',
20 verbose headers   vary: 'Accept' }
21 silly get cb [ 304,
[...]

PR #13635, opened 2016-08-11
install: fix hiccup that caused a conflict

zkat:
Basically, there was some duck typing in `resolveWithNewModule`
in situations where the function got a `package.json` from `npa`,
rather than the internal request object.

The issue reared its head because of a package that had `type`
in its `package.json`, which confused the duck typing.

Instead, the special case was inlined into (afaict) the only
place where it was actually triggered from.

Fixes: https://github.com/npm/npm/issues/11398
Credit: @zkat

### DIFF
--- a/lib/install/deps.js
+++ b/lib/install/deps.js
@@ -452,7 +452,9 @@ function addDependency (name, versionSpec, tree, log, done) {
         }
       }))
     } else {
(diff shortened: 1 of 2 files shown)


iarna:
LGTM! 🐑 💣 💥

iarna:

PR #13648, opened 2016-08-12
Added sanitation of the maintainers field.

larsw:
Some (at least Artifactory) doesn't sanitize non-conforming package.json content, and serves it back verbatim. There exists many? packages where the maintainers property is a string instead of an array. This pollutes the NPM cache (when using Artifactory) and breaks the search.js/stripData() function where it tries to do a map over the maintainers property (or an empty array).

### DIFF
--- a/lib/search.js
+++ b/lib/search.js
@@ -98,13 +98,11 @@ function filter (data, args, notArgs) {
 }
 
 function stripData (data) {
-  
   // Sanitize the maintainers field to ensure that NPM repositories (Artifactory) that
   // does not sanitize this won't break the search functionality.
-  if (typeof data.maintainers === "string") {
-    data.maintainers = [data.maintainers];
+  if (typeof data.maintainers === 'string') {
+    data.maintainers = [data.maintainers]
   }
-  
   return {
     name: data.name,
     description: npm.config.get('description') ? data.description : '',
--- a/lib/search.js
+++ b/lib/search.js
@@ -98,6 +98,13 @@ function filter (data, args, notArgs) {
 }
 
 function stripData (data) {
+  
+  // Sanitize the maintainers field to ensure that NPM repositories (Artifactory) that
+  // does not sanitize this won't break the search functionality.
+  if (typeof data.maintainers === "string") {
+    data.maintainers = [data.maintainers];
+  }
+  
   return {
     name: data.name,
     description: npm.config.get('description') ? data.description : '',


lib/search.js
```diff
@@ -98,6 +98,11 @@ function filter (data, args, notArgs) {
 }
 
 function stripData (data) {
+  // Sanitize the maintainers field to ensure that NPM repositories (Artifactory) that
+  // does not sanitize this won't break the search functionality.
+  if (typeof data.maintainers === 'string') {
```
iarna: Instead of mutating this maybe just put the array version in a variable and use that below?

lib/search.js
```diff
@@ -98,6 +98,11 @@ function filter (data, args, notArgs) {
...
+    data.maintainers = [data.maintainers]
+  }
   return {
     name: data.name,
     description: npm.config.get('description') ? data.description : '',
```
iarna: ah but either way that breaks down at usage, since `m.name` isn't gonna work if its a string.

iarna:
Ok, so in principle I think this is a good idea. As a rule, npm can't trust any of its data sources to give it clean data. But just sticking the string in an array fixes the crash while not really fixing the problem.

What we probably want to do here is something closer to what [normalize-package-data](https://github.com/npm/normalize-package-data) does, which is pass the maintainer through the following two functions:

```
function unParsePerson (person) {
  if (typeof person === "string") return person
  var name = person.name || ""
  var u = person.url || person.web
  var url = u ? (" ("+u+")") : ""
  var e = person.email || person.mail
  var email = e ? (" <"+e+">") : ""
  return name+email+url
}

function parsePerson (person) {
  if (typeof person !== "string") return person
  var name = person.match(/^([^\(<]+)/)
  var url = person.match(/\(([^\)]+)\)/)
  var email = person.match(/<([^>]+)>/)
  var obj = {}
  if (name && name[0].trim()) obj.name = name[0].trim()
  if (email) obj.email = email[1];
  if (url) obj.url = url[1];
  return obj
}
```

Essentially it ends up calling: ```
maintainers = maintainers.map(function (person) { parsePerson(unparsePerson(person)) })
```

othiym23:
is there a way to actually run `normalize-package-data` on the stream of search results? That seems like the most consistent (and potentially simplest) way to do this, if the `normalize-package-data` API supports it. I'd really like to avoid duplicating this logic, if possible.

iarna:
Since we requested changes and it's been almost a year, I'm going to close this issue. If inspiration strikes again and you're interested in continuing this work, please open a new PR. Thanks!

PR #13655, opened 2016-08-12
Update .npmrc documentation

mdjasper:
I found that information about commenting lines in a `.npmrc` file was lacking from the official documentation. I've updated the doc to reflect currently supported ways to commenting lines in the npm config file as found through reading other sources and through experimentation with different ways of adding comments.

### DIFF
--- a/doc/files/npmrc.md
+++ b/doc/files/npmrc.md
@@ -41,6 +41,15 @@ sensitive credentials, they must be readable and writable _only_ by your user
 account (i.e. must have a mode of `0600`), otherwise they _will be ignored by
 npm!_
 
+#### Comments
+
+Because lines in an npm config file follow the `key = value` pattern, lines may be effectivly commmented-out by placing any character at the beginning of the line (creating a `key` which is not recoginized, or used). Common comment characters are `;`, `#`, or `//`.
+
+For example:
+
+    ; Set a new registry for a scoped package
+    @myscope:registry=https://mycustomregistry.example.org
+
 ### Per-project config file
 
 When working locally in a project, a `.npmrc` file in the root of the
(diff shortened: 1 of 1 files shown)


iarna:
`.npmrc` files are parsed with https://github.com/npm/ini so `;` and `#` do explicitly mark [start-of-line comments](https://github.com/npm/ini/blob/master/ini.js#L76).

doc/files/npmrc.md
```diff
@@ -41,6 +41,15 @@ sensitive credentials, they must be readable and writable _only_ by your user
 account (i.e. must have a mode of `0600`), otherwise they _will be ignored by
 npm!_
 
+#### Comments
+
+Because lines in an npm config file follow the `key = value` pattern, lines may be effectivly commmented-out by placing any character at the beginning of the line (creating a `key` which is not recoginized, or used). Common comment characters are `;`, `#`, or `//`.
```
iarna: `;` and `#` represent actual comments which won't end up in the key-space for loaded config, by contrast `//` does not and in fact can be a part of valid keys. For instance, your authentication key is stored as:

```
//registry.npmjs.org/:_authToken=AUTH-TOKEN-HERE
```

mdjasper: Great feedback. I reached the conclusion in the PR through experimentation. I will update the doc PR to reflect the ini config specified comments

mdjasper:
Addressed and corrected an issue pointed out by @iarna on the initial PR on what constitutes valid comments in the `.npmrc` file

legodude17:
Is this still in progress?

mdjasper:
@legodude17 I believe it it complete, but happy for more review if anyone thinks that is necessary

legodude17:
@mdjasper, then it LGTM.

Of course, I don't actually have any authority, but I think this is a good idea.

zkat:
Thanks for the patch, @mdjasper! This was included as part of `npm@3.10.8`, so I'm gonna close the PR now. 🎉

PR #13692, opened 2016-08-17
Don't try to install known-uninstallable opt deps that are in shrinkwraps

iarna:
_The commits here will be rolled up a bit before merge._

Ordinarily, if an optional dep fails the arch/cpu checks then we never try to extract it or install it.

But if an optional dep is in an `npm-shrinkwrap.json` file then it would get injected into the tree without checking instability–this fixes that.

In a perfect world we would do this step in `inflate-shrinkwrap`. Adding that there will be somewhat involved however as it currently has no facility for skipping a dependency. We could just skip the opt dep that can't work but what we _need_ it to do is ALSO skip any of its transitive deps as well. Worse, it might be a non-optional transitive dep of an optional dep that failed and we have to go back and remove the entire tree associated with the optional dep. Fortunately this is already implemented elsewhere…

So instead I've implemented this by doing the check prior to running any actions. The action running stage already has facilities to deal with failing optional deps and by hooking this as a prerequisite to ANY actions, it means it will happen before the very first one, which is `fetch`.  This means performance-wise it should be roughly equivalent to doing this in `inflate-shrinkwrap` and it saves us from duplicating the optional-dep-failing code.


lib/install/actions.js
```diff
@@ -46,7 +48,24 @@ Object.keys(actions).forEach(function (actionName) {
...
+      return isInstallable(pkg.package, iferr(andDone(next), andMarkInstallable(thenRunAction)))
+    }
+    function andMarkInstallable (cb) {
+      return function () {
+        pkg.knownInstallable = true
+        cb()
+      }
+    }
+    function thenRunAction () {
+      pkg.knownInstallable = true
```
zkat: Why do you need `andMarkInstallable` at all here? It looks like both code paths hit `thenRunAction` invariably. Alternately, why is this in `thenRunAction`? In the path that skips `andMarkInstallable`, `pkg.knownInstallable` is already true.

lib/utils/error-message.js
```diff
@@ -150,17 +150,21 @@ function errorMessage (er) {
...
     case 'EBADPLATFORM':
+      var validOs = er.os.join ? er.os.join(',') : er.os
+      var validArch = er.cpu.join ? er.cpu.join(',') : er.cpu
+      var expected = {os: validOs, arch: validArch}
+      var actual = {os: process.platform, arch: process.arch}
       short.push([
         'notsup',
         [
-          'Not compatible with your operating system or architecture: ' + er.pkgid
+          util.format('Unsupported platform for %s: wanted %j (current: %j)', er.pkgid, expected, actual)
```
zkat: This `pkgid` includes the package name and version, right?

test/tap/full-warning-messages.js
```diff
@@ -94,8 +94,8 @@ test('tree-style', function (t) {
     t.notMatch(stdout, /modB/, 'modB not installed')
     var stderrlines = stderr.trim().split(/\n/)
     t.is(stderrlines.length, 2, 'two lines of warnings')
-    t.match(stderr, /Skipping failed optional dependency/, 'expected optional failure warning')
-    t.match(stderr, /Not compatible with your operating system or architecture/, 'reason for optional failure')
+    t.match(stderr, /SKIPPING OPTIONAL DEPENDENCY/, 'expected optional failure warning')
+    t.match(stderr, /Unsupported platform/, 'reason for optional failure')
```
zkat: Should this test that we're actually reporting platforms + pkgid?

zkat:
LGTM! Couple of nits you're free to ignore, otherwise 🐑 🎉

lib/utils/error-message.js
iarna: right

lib/install/actions.js
iarna: This is a good question…

iarna: The answer is: This wasn't supposed to still be in `thenRunAction`

iarna:
Landed in `release-next`!

PR #13705, opened 2016-08-19
Update README.md

emigenix:
Added README info about non-support of Cygwin.


othiym23:
The team has no interest in encouraging users to install Cygwin under an unsupported platform, so we won't be landing this patch. Thanks for your time, though!

PR #13708, opened 2016-08-19
tar: ignore *.orig files

boneskull:
This PR ignores causes `*.orig` files to explicitly be ignored upon `npm publish`.  Docs & tests updated accordingly.  

Original issue: mochajs/mocha#2446.

In Git, when performing a merge, the original file(s) with conflict markers can be saved as `file.orig`.  Git's default behavior is to create them ([reference](https://git-scm.com/docs/git-config)).

Then, unless:
- a `.npmignore` contains `*.orig`, or
- an npm user is careful to remove these files before publishing, or
- a Git user disables this behavior via `git config mergetool.keepBackup false` (and thus the files are never created),

any of these untracked files present in a working copy will be published upon `npm publish`.  

Since it's often desirable to published untracked files, there's no issue with `npm publish`'s behavior _per se_--but I cannot think of a use case for publishing an `.orig` file.

### DIFF
--- a/doc/files/package.json.md
+++ b/doc/files/package.json.md
@@ -203,6 +203,7 @@ Conversely, some files are always ignored:
 * `.npmrc`
 * `node_modules`
 * `config.gypi`
+* `*.orig`
 
 ## main
 
--- a/lib/utils/tar.js
+++ b/lib/utils/tar.js
@@ -102,7 +102,8 @@ BundledPacker.prototype.applyIgnores = function (entry, partial, entryObj) {
       entry === '.npmrc' ||
       entry.match(/^\..*\.swp$/) ||
       entry === '.DS_Store' ||
-      entry.match(/^\._/)
+      entry.match(/^\._/) ||
+      entry.match(/^.*\.orig$/)
     ) {
     return false
   }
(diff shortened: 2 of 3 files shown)


boneskull:
appveyor failures (anyone surprised?) seem unrelated

legodude17:
This makes sense to me. Ready to merge?

boneskull:
I'm good here.

zkat:
This LGTM. I'll merge it in alongside https://github.com/npm/fstream-npm/pull/23 which will take care of the other side of this (I know, it's weird, but we need it in both places, and all this code is going away someday not-too-far in the future so it's not worth fixing).

Thanks! 🎉 

Also, @legodude17: would you mind withholding your approval of PRs like that? Some folks might be confused about who the maintainers are if they don't have context, and assume that you were rubber-stamping something for us.

zkat:
Alright, this was included, along with https://github.com/npm/fstream-npm/pull/23, as part of `npm@3.10.8`, so you shouldn't need to `.npmignore` these files. Here's hoping we don't completely ruin someone's day with this by accident :fingers_crossed:

PR #13717, opened 2016-08-21
doc: npm link will link bin of project.

legodude17:
This documents the fact that `npm link` will link the files specified in the `bin` field of `package.json` to `{prefix}/bin/{name}`.


legodude17:
Should I be worried about the appveyor build failing? This did not change any code, only the docs.

zkat:
@legodude17 Yeah, we occasionally poke at trying to get AppVeyor tests to be more reliable, but it's a bit of a time sink so it's taking a while. You're safe to ignore that.

And thanks for this patch! It was merged and included as part of `npm@3.10.8` and `npm@2.15.11`. 🎉

legodude17:
Thank you @zkat.

PR #13735, opened 2016-08-23
test: update test-node to work on windows

gibfahn:
See https://github.com/nodejs/node/pull/7867#discussion_r72723803

The test-node script needs to be updated to handle windows. It's probably worth waiting for https://github.com/nodejs/node/pull/7867 to land (or at least be accepted) before going ahead with this, just in case any more changes need to be made.

cc: @iarna

### DIFF
--- a/package.json
+++ b/package.json
@@ -213,12 +213,10 @@
     "dumpconf": "env | grep npm | sort | uniq",
     "prepublish": "node bin/npm-cli.js prune --prefix=. --no-global && rimraf test/*/*/node_modules && make doc-clean && make -j4 doc",
     "preversion": "bash scripts/update-authors.sh && git add AUTHORS && git commit -m \"update AUTHORS\" || true",
-    "tap": "tap --timeout 300",
     "tap-cover": "tap --nyc-arg='--cache' --coverage --timeout 600",
-    "test": "standard && npm run test-tap",
+    "test": "standard && npm run tap",
     "test-coverage": "npm run tap-cover -- \"test/tap/*.js\" \"test/network/*.js\" \"test/broken-under-*/*.js\"",
-    "test-tap": "npm run tap -- \"test/tap/*.js\" \"test/network/*.js\" \"test/broken-under-*/*.js\"",
-    "test-node": "tap --timeout 240 \"test/tap/*.js\" \"test/network/*.js\" \"test/broken-under-nyc*/*.js\""
+    "tap": "tap --timeout 300 \"test/tap/*.js\" \"test/network/*.js\" \"test/broken-under-*/*.js\""
   },
   "license": "Artistic-2.0"
 }
--- a/lib/doctor/check-files-permission.js
+++ b/lib/doctor/check-files-permission.js
@@ -34,7 +34,7 @@ function checkFilesPermission (root, mask, cb) {
       function checkFile (f, next) {
         var file = path.join(root, f)
         tracker.silly('checkFilePermissions', f)
-        fs.stat(file, function (e, stat) {
+        fs.lstat(file, function (e, stat) {
           tracker.completeWork(1)
           if (e) return next(e)
           if (!stat.isFile()) return next()


package.json
```diff
@@ -207,7 +207,7 @@
     "test": "standard && npm run test-tap",
     "test-coverage": "npm run tap-cover -- \"test/tap/*.js\" \"test/network/*.js\"",
     "test-tap": "npm run tap -- \"test/tap/*.js\"",
-    "test-node": "\"$NODE\" \"node_modules/.bin/tap\" --timeout 240 \"test/tap/*.js\" \"test/network/*.js\""
+    "test-node": "tap --timeout 240 \"test/tap/*.js\" \"test/network/*.js\""
```
iarna: I wonder if we could do: 
```
"%NODE%" "node_modules/tap/bin/run.js"
```
on Windows? Will `cmd` do env var substitution for the command part?

If that will work I'd be inclined to have a `test-node-win` target with the above.

iarna (commented):
In order to use this we'll have to go back to path mangling in Node.js's `tools/test-npm.sh` (plus, I guess, some new `tools/test-npm.cmd`?) to ensure the version of node we're testing is used to execute tap.

I'm not really particular one way or another as to what solution we take so I'm happy to merge this (or whatever other solution) when you're ready. Please feel free to ping me on IRC or Twitter when that time comes so I can be more responsive than if you ping me here. =)

gibfahn:
We currently do the path mangling in `test-npm.sh` already, and yes, we'd have to have a windows equivalent (I used powershell so we can have the output teed to a file).

I'll try your `test-node-win` target suggestion, no idea if it will work...

As for landing this, I want to do some more testing of https://github.com/nodejs/node/pull/7867 in Node first, so this should probably wait until that gets merged.

Thanks!

iarna:
@gibfahn sounds good, I'll keep an eye on it.

gibfahn:
@iarna So I've now got to the point where `make test-npm` and `vcbuild test-npm` are working (there are still failures, but I'm not sure they're due to the test suite). See the [unix build job](https://ci.nodejs.org/view/All/job/gibfahn-test-npm/) and [windows build job](https://ci.nodejs.org/view/All/job/gibfahn-test-npm-win/). 

I'm currently using the change that's in this PR, I haven't tried using `"%NODE%" "node_modules/tap/bin/run.js"`, but it should work, npm scripts can definitely pick up windows environment variables. Of course that would break people trying to run scripts using Git Bash or Cygwin. 

I assume the reason you want to use `%NODE%` instead of just `tap` is to make sure you're using the right version of Node? That shouldn't be an issue, as the test-npm scripts in Node now both prepend the `%NODE%`/`$NODE` path to the `PATH`, so `where node`/`which node` should always give the right node.

All that said, the Node PR is now nearly ready, so I'd like to get this change merged in if it's okay.

iarna:
@gifahn Your one failing test on unix is due to, I think, a difference between how Linux and OSX handle or report signals. Currently we disable one test if in Windows or if the TRAVIS env var is true. I suspect that the latter is due to a difference in Linux and not specifically in TRAVIS, so I intend to update it be more general.

gibfahn:
@iarna So it looks like some of the linux distros are having more issues with the tests:

https://ci.nodejs.org/view/All/job/gibfahn-test-npm/lastBuild/testReport/

gibfahn:
@iarna the original change in this PR has already gone in, which is great.

I don't think the `test-node` script in the [package.json](https://github.com/npm/npm/blob/892f531130d38dfd40000794dc9087b483bb1ee9/package.json#L221) is required any more (looking at `latest`). As far as I can tell:

`test-node`:
`"tap --timeout 240 \"test/tap/*.js\" \"test/network/*.js\" \"test/broken-under-nyc*/*.js\""`

`test-tap`:
`tap --timeout 300 \"test/tap/*.js\" \"test/network/*.js\" \"test/broken-under-*/*.js\""`

So I updated this to remove the test-node line entirely. I also cleaned up the npm run-scripts, I'm not sure if this is an okay change or not, let me know.

iarna:
Since this has distilled down to basically nothing, I'm going to close it.

We keep the `tap` test that has no target so that we can manually run targets from the command line with the full npm run-script/testing environment, eg `npm run tap test/tap/name-of-test.js`

node's build scripts still call `test-node`. I'm not going to remove that from npm's package.json till it's updated on the node side, but yeah, it does seem unnecessary now.

Issue #13247, opened 2016-07-01
npm search failing with "npm ERR! Unexpected token g"


geuis:
#### I'm opening this issue because:
  - [x] npm is crashing.
  - [ ] npm is producing an incorrect install.
  - [ ] npm is doing something I don't understand.
  - [ ] Other (_see below for feature requests_):

#### What's going wrong?
When I search for something, I get results for the first search. Subsequent searches, even for the same term, produce invalid output.

I have uninstalled and reinstalled the latest npm@3.10.3 and have the same issue that was occuring with 3.9 (previously installed).

#### How can the CLI team reproduce the problem?
npm-debug.log gist here: https://gist.github.com/geuis/198fd71410a40bfeaf0a875f42b276ef

npm cache clean
npm search <term>
    // get correct results
npm search <term or another term>
    // get error and invalid results

// now do the whole thing again
npm cache clean
npm search <term again>
    // get correct results
npm search <term or another term>
    // get error and invalid results

### supporting information:

 - `npm -v` prints: 3.10.3
 - `node -v` prints: v4.3.1
 - `npm config get registry` prints: http://registry.npmjs.org/
 - Windows, OS X, or Linux?: OS X
 - Network issues:
   - Geographic location where npm was run: San Francisco
   - [ ] I use a proxy to connect to the npm registry.
   - [ ] I use a proxy to connect to the web.
   - [ ] I use a proxy when downloading Git repos.
   - [ ] I access the npm registry via a VPN
   - [ ] I don't use a proxy, but have limited or unreliable internet access.
 - Container:
   - [ ] I develop using Vagrant on Windows.
   - [ ] I develop using Vagrant on OS X or Linux.
   - [ ] I develop / deploy using Docker.
   - [ ] I deploy to a PaaS (Triton, Heroku).

zkat:
This should be fixed by https://github.com/npm/npm/pull/13746 once that's done.

Issue #12619, opened 2016-05-07
npm search crashes with GC error when building local cache


iankronquist:
OS X 10.11, npm 3.8.6, node 6.0.0 all installed using homebrew. I managed to get this to happen twice in a row before succeeding.

```
npm search rtfm
npm WARN Building the local index for the first time, please be patient
▐ ╢░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░╟

<--- Last few GCs --->

  131112 ms: Mark-sweep 807.6 (1148.1) -> 782.3 (1130.1) MB, 640.0 / 0 ms [allocation failure] [GC in old space requested].
  131777 ms: Mark-sweep 782.3 (1130.1) -> 782.1 (1131.1) MB, 665.3 / 0 ms [allocation failure] [GC in old space requested].
  132454 ms: Mark-sweep 782.1 (1131.1) -> 782.1 (1125.1) MB, 676.7 / 0 ms [last resort gc].
  133124 ms: Mark-sweep 782.1 (1125.1) -> 782.1 (1119.1) MB, 669.4 / 0 ms [last resort gc].


<--- JS stacktrace --->

==== JS stack trace =========================================

Security context: 0x3c83930c9fa9 <JS Object>
    1: fromString(aka fromString) [buffer.js:234] [pc=0x12dd12bb19ab] (this=0x3c8393004189 <undefined>,string=0x1e951b6647c1 <Very long string[166224244]>,encoding=0x3c83930d79a9 <String[4]: utf8>)
    2: from [buffer.js:131] [pc=0x12dd12b35c36] (this=0xb3a135de721 <JS Function Buffer (SharedFunctionInfo 0x2376c91f5a61)>,value=0x1e951b6647c1 <Very long string[166224244]>,encodingOrOffset=0x3c83...

FATAL ERROR: Abort trap: 6

```

jhaungs:
Same here on OSX 10.11, Node 6.1.0 and npm 3.8.9

iankronquist:
Yeah, look like the same issue.

Tref:
I am getting the same issue via node 6.1.0 and npm 3.8.9

```
0 info it worked if it ends with ok
1 verbose cli [ '/usr/local/Cellar/node/6.1.0/bin/node',
[...]

Issue #12587, opened 2016-05-05
Callback called more than once.


afzalhussein:
[npm-debug.zip](https://github.com/npm/npm/files/249860/npm-debug.zip)

guiwp:
```
npm ERR! Windows_NT 10.0.10586
npm ERR! argv "C:\\Program Files\\nodejs\\node.exe" "C:\\Program Files\\nodejs\\node_modules\\npm\\bin\\npm-cli.js" "search" "beautify"
npm ERR! node v6.2.0
npm ERR! npm  v3.8.9

npm ERR! Callback called more than once.
npm ERR!
npm ERR! If you need help, you may report this error at:
npm ERR!     <https://github.com/npm/npm/issues>

npm ERR! Please include the following file with any support request:
npm ERR!     E:\Wiki\npm-debug.log
```

[npm-debug.zip](https://github.com/npm/npm/files/282200/npm-debug.zip)

zkat:
Since it's such a big rewrite, https://github.com/npm/npm/pull/13746 will almost certainly fix this.

Issue #12439, opened 2016-04-22
npm ERR! Cannot read property '_updated' of undefined


qkboo:
I just search by npm what express version are available
But npm error "npm ERR! Cannot read property '_updated' of undefined"

below contents are *npm-debug.log*
--
```
0 info it worked if it ends with ok
1 verbose cli [ '/home/qkboo/.nvm/versions/node/v5.10.1/bin/node',
1 verbose cli   '/home/qkboo/.nvm/versions/node/v5.10.1/bin/npm',
1 verbose cli   'search',
1 verbose cli   'express' ]
2 info using npm@3.8.3
3 info using node@v5.10.1
4 silly mapToRegistry name -/all
5 silly mapToRegistry using default registry
6 silly mapToRegistry registry https://registry.npmjs.org/
7 silly mapToRegistry data Result {
7 silly mapToRegistry   raw: '-/all',
7 silly mapToRegistry   scope: null,
7 silly mapToRegistry   name: null,
7 silly mapToRegistry   rawSpec: '-/all',
7 silly mapToRegistry   spec: '/home/qkboo/express_v4/uploader_multer/-/all',
7 silly mapToRegistry   type: 'local' }
8 silly mapToRegistry uri https://registry.npmjs.org/-/all
9 info updateIndex /home/qkboo/.npm/registry.npmjs.org/-/all/.cache.json
10 verbose correctMkdir /home/qkboo/.npm correctMkdir not in flight; initializing
11 verbose stack TypeError: Cannot read property '_updated' of undefined
11 verbose stack     at /home/qkboo/.nvm/versions/node/v5.10.1/lib/node_modules/npm/lib/cache/update-ind
ex.js:61:26
11 verbose stack     at /home/qkboo/.nvm/versions/node/v5.10.1/lib/node_modules/npm/node_modules/gracefu
l-fs/polyfills.js:210:7
11 verbose stack     at FSReqWrap.oncomplete (fs.js:82:15)
12 verbose cwd /home/qkboo/express_v4/uploader_multer
13 error Linux 4.4.4-std-3
14 error argv "/home/qkboo/.nvm/versions/node/v5.10.1/bin/node" "/home/qkboo/.nvm/versions/node/v5.10.1/
bin/npm" "search" "express"
15 error node v5.10.1
16 error npm  v3.8.3
17 error Cannot read property '_updated' of undefined
18 error If you need help, you may report this error at:
18 error     <https://github.com/npm/npm/issues>
[...]

PR #13746, opened 2016-08-23
Streaming CLI Search

zkat:
This is a WIP `npm search` rewrite branch, based off the work done by @aredridel over at #10615.

The entire search process is now streamed, end-to-end. This branch brings in a couple of stream support libraries to help things along, too.

It should be able to fix pretty much all the heap-blowup crashers in npm.

Fixes: #12619
Fixes: #13247
Fixes: #12587
Fixes: #12439
Fixes: #6254
Fixes: #7090

streamed-search-pretty

It's not great to still be doing a linear search every time, but it's a significant step for search to be at least usable (without crashing or waiting forever for any results).
### TODO
- [x] Stream results from `request`
- [x] Stream results from cache
- [x] Stream output into cache file
- [x] Find issues related to crashing search, close duplicates, and reference master issue here.
- [x] Fix `cb not called` corner case when cache is an empty object (probably when request is empty, too -- check and test both)
- [x] Figure out the `cb not called` thing with `request`'s `res` grrr
- [x] Write unit tests
  - [x] `createEntryStream`
  - [x] `createCacheEntryStream`
  - [x] `createEntryUpdateStream`
  - [x] `createCacheWriteStream`
  - [x] `updateIndex`
- [x] Prettify output
- [x] Do the thing from https://github.com/npm/npm/issues/6254
- [x] Support `--json` for streamed JSON output
- [x] Add message for when there's no search matches
- [X] Rewrite history to not be a crapmess
- [x] check `npm.config.get('unicode')`
- [x] Comb through code and make sure none of the object streams have absurd `highWaterMark` settings.
- [x] Merge and release https://github.com/npm/npm-registry-client/pull/131
- [x] Remove hacked-in copy of `npm-registry-client`
- [x] see if new deps can be minimized (do we need all of `mississippi`? really?) -- decision: keep mississippi cause at least 5 of its utillities are getting used by this code, and it's not worth the extra hassle
- [x] Rewrite integration test for `npm search`
- [x] Document `--tab-separated`
- [x] Refactor all the various stream bits
- [x] Replace `--tab-separated` with `--parseable`
- [x] Make sure searching all fields is well-tested


legodude17:
This seems really cool. How can I help?

zkat:
@legodude17 Hey! Thanks for the offer -- for now, I think the best thing to do is to just hang in there while I get this PR finished/ready. Since the bulk of the implementation is done, if you find any *glaring* issues with the patches themselves, you can try commenting on them. For the most part, though, a lot about this PR will change before I pass it on to the rest of my team for review.

legodude17:
Okay. Keep up the good work @zkat!

timoxley:
This is awesome. Is the gif in the first post up-to-date?

zkat:
@timoxley yup! The gif is how it looks now.

martinheidegger:
@zkat @legodude17 wouldn't it be good if JSONStream was updated to Streams3 ? https://github.com/dominictarr/JSONStream/pull/61#issuecomment-114876997

zkat:
@martinheidegger Please note that we have to support down to `node@0.10` right now, which rules out relying on `streams3`. Beyond that, there's nothing specific that I'd like our of `streams3` enough to nudge an external project into upgrading.

node_modules/sorted-union-stream/index.js
```diff
@@ -33,8 +33,9 @@ var union = function (streamA, streamB, toKey) {
         var keyB = toKey(dataB)
 
         if (keyA === keyB) {
```
iarna: Can you elaborate on what's going on here? And what this being fixed looks like?

lib/search.js
```diff
@@ -46,6 +34,9 @@ function search (args, silent, staleness, cb) {
 
   var searchopts = npm.config.get('searchopts')
   var searchexclude = npm.config.get('searchexclude')
+  if (npm.config.get('searchsort', 'cli')) {
+    log.warn('search', 'Search result sorting not currently supported')
+  }
```
legodude17: @zkat, do you think that search sorting will ever be supported?

zkat: @legodude17 not for this iteration, and since this PR is a stopgap to get search working *at all*, it's unlikely that we'll have more featureful searching until we get better backend support for fast searching.

lib/search.js
```diff
@@ -70,24 +57,50 @@ function search (args, silent, staleness, cb) {
...
-    log.clearProgress()
-    output(prettify(data, args))
-    cb(null, data)
+  updateIndex(staleness, function (er, stream) {
+    if (er) return cb(er)
+    log.silly('search', 'searching packages')
+    var matchCount = 0
+    var filterStream = ms.through.obj(function (pkg, enc, cb) {
+      log.gauge.pulse('search')
+      log.gauge.show({section: '🤔 search', logline: 'scanning ' + pkg.name})
```
iarna: heehee, a+ emoji use buuut... you need to only include it if `npm.config.get('unicode')` is true.

zeg-io:
Sadly the GIF doesn't work in Safari for some reason :( 
(Works on Chrome though!)

legodude17:
About the last task (That you don't need all of mississippi), based on my initial reading you use almost all of it except concat-stream, from, and to. You could possibly just install those modules (through, stream-finished, pipeline (or whatever they are called)).

lib/search.js
legodude17: One option for searching that could work is to stream, but then stream the results to a full object or something, which could then be sorted and displayed. This will decrease the amount of memory required currently, but could still cause a heap blowup if the search term has a lot of results and the computer has a very small amount of RAM.

lib/search.js
```diff
@@ -6,38 +6,26 @@ var columnify = require('columnify')
...
+var log = require('npmlog')
+var jsonstream = require('JSONStream')
+var ms = require('mississippi')
 
 search.usage = usage(
   'search',
   'npm search [--long] [search terms ...]'
 )
 
+search.usage = 'npm search [--long] [search terms ...]' +
```
othiym23: This replaces the setting of `search.usage` immediately above it. That seems unnecessary?

lib/search.js
```diff
@@ -66,35 +57,59 @@ function search (args, silent, staleness, cb) {
...
   })
 
-  getFilteredData(staleness, opts, searchexclude, function (er, data) {
-    // now data is the list of data that we want to show.
-    // prettify and print it, and then provide the raw
-    // data to the cb.
-    if (er || silent) return cb(er, data)
-    output(prettify(data, args))
-    cb(null, data)
+  updateIndex(staleness, function (er, stream) {
```
othiym23: I think this callback should be named and hoisted, given its length, complexity, and the fact there are further callbacks nested in it.

othiym23 (commented):
This is looking good! I think I'm going to do another pass through this at the individual commit level, because there are too many files in the PR to use the global view (maybe after you do an initial squash?). I also think I'll probably want to pull the branch down and play with it / try to break it some before it gets squashed and merged. It might be worth putting all of the stuff that affects both result and cache streaming into a single file, pulled out of `lib/cache/update-index.js` and `lib/search.js`, because right now it's a little hard to correlate some of the stream hijinx between the two. Aside from that, the architectural basis looks sound.

I do have some concerns about the amount of dependencies we're adding with this change, though. Are we using enough of `mississippi` to justify using the whole package, or might we save some space by only including the bits that we're using?

Again, this is looking good, and I appreciate all of the care and time you've put into this. This is a huge improvement over what we have now.

doc/cli/npm-search.md
```diff
@@ -17,6 +17,20 @@ expression characters must be escaped or quoted in most shells.)
...
 ## CONFIGURATION
 
+### json
+
+* Default: false
+* Type: Boolean
+
+Output search results as a JSON array.
+
+### tab-separated
```
zkat: I remember we had an over-the-air conversation about `--parseable` vs `--tab-separated` but I don't remember if there was a particular reason for it. So then: would you like me to use `--parseable` instead for this? It's no skin off my back. I have no strong opinion either way, and I understand your point. (and, fwiw, being specific about what the "parseable" format is seems a little nicer, but it's not the convention we already have)

zkat:
@othiym23 I've addressed all the specific comments now, I think, and now to respond to the initial review comment:

> It might be worth putting all of the stuff that affects both result and cache streaming into a single file, pulled out of `lib/cache/update-index.js` and `lib/search.js`.

I'm not entirely sure how to parse this one: the divide between those two files is relatively clear to me. The API of `update-index.js` is pretty much to give you a stream, by whatever means it feels appropriate, of all package entries. As its own internal side-effect, it does caching, but the cache is not something `search.js` is really aware of at all. Did I miss something here about what affects both of them, that you're referring to? I don't mind moving around whatever, I'm just not sure which things. :)

> I do have some concerns about the amount of dependencies we're adding with this change, though. Are we using enough of mississippi to justify using the whole package, or might we save some space by only including the bits that we're using?

I was a bit concerned, too, but the only significant dependency we don't use is `concat-stream` (376K uncompressed, 60K tarballed up). The others are tiny (`flush-write-stream`: 28K uncompressed; `stream-each`: 32K uncompressed), and I can totally see us using all 3 eventually. All other utilities are used either by the code, or by the tests. I started to install all the mississippi deps individually at one point, and it just made things a lot noisier, and I realized I was dragging a significant % of the library into a bunch of files. I think it's nice to have it bundled like this. I can go ahead and do that anyway if you're uncomfortable with the 60K+ that the unused deps will add, though.

othiym23:
> I'm not entirely sure how to parse this one: the divide between those two files is relatively clear to me.

Both are working with streams to a single representation on disk. There isn't really a datatype for that representation, so there's not any kind of encapsulated interface, which makes me concerned that a. it will be tough for newbies (as it was for me, a lil bit) to correlate the two different halves of updating that cache and scanning it, and b. that it's easier for the two chunks of code to drift if we're not careful (especially via contributors making patches or changes – see a.). From looking at the code, it _seems_ like all the streaming code can be pulled into a single module with a defined interface, but it could be that it's tricky in practice. What do you think?

>  I can go ahead and do that anyway if you're uncomfortable with the 60K+ that the unused deps will add, though.

The npm package tarball is now over 3MiB, which feels large to me. However, the point about noise is a good one, and there is a good chance we'll end up using those things later. Consider that case made. I just want to encourage a healthy sense of parsimony on the part of the team, because I think we haven't really been paying attention to how large the CLI is getting, and it has a significant effect on everyone downstream from us.

Still waiting on the squash for a complete, final review!

doc/cli/npm-search.md
othiym23: Yeah, I think I would. It's fine if it stays tab-separated, but I don't think we need a separate format for tab-separated results just for this subcommand. Just out of curiosity, what happens if there are tabs embedded in the description?

test/tap/update-index.js
```diff
@@ -1,200 +1,209 @@
...
+  var cacheContents = {
+    '_updated': now,
+    bar: { name: 'bar', version: '1.0.0' },
+    cool: { name: 'cool', version: '1.0.0' },
+    foo: { name: 'foo', version: '2.0.0' },
+    other: { name: 'other', version: '1.0.0' }
+  }
+  var fixture = new Tacks(File(cacheContents))
+  fixture.create(cachePath)
+  // TODO - figure out how to get the time stuff in here
```
othiym23: Is this a matter of it being tricky to get the configuration here? Or having control over what times things happen at, and avoiding race conditions? This particular TODO has a bit of a smell for me.

test/tap/search.js
```diff
@@ -1,84 +1,185 @@
-var fs = require('graceful-fs')
 var path = require('path')
```
othiym23: Do any of the tests in here match against the description field? Or username fields? These are both things that I've found surprisingly useful in the past, when search worked. I'm pretty sure they work in practice (testing right now), but it would be good to have them included in the tests.

othiym23 (commented):
Thanks for the rebase! I think I've checked everything I wanted here, and replied to your responses from the last go-round. This is looking good! I wouldn't consider any of the items I've raised in this round blockers, but several of them are changes I think we'd be happy we made later. I'm still trying to exercise edge cases a little more, and will let you know if I find anything egregious (although the tests, which are great, do a pretty good job of nailing down the edges). ガンバレよー！

lib/cache/update-index.js
```diff
@@ -10,184 +10,293 @@ var log = require('npmlog')
...
+var sortedUnionStream = require('sorted-union-stream')
 var once = require('once')
 
-/* /-/all is special.
- * It uses timestamp-based caching and partial updates,
- * because it is a monster.
- */
-function updateIndex (staleness, args, notArgs, filter, cb) {
-  assert(typeof filter === 'function', 'must pass filter callback to updateIndex')
+// Updates the search index (aka the entire metadata db in a flat file)
```
zkat: This is a good idea!

zkat: done!

lib/cache/update-index.js
```diff
@@ -10,184 +10,293 @@ var log = require('npmlog')
...
 }
 
-function updateIndex_ (all, params, latest, filter, args, notArgs, cacheBase, cb) {
-  cb = once(cb)
-  log.silly('update-index', 'fetching', all)
-  npm.registry.request(all, params, pulseTillDone('updateIndex', function (er, res) {
-    if (er) return cb(er)
+// Merges `a` and `b` into one stream, dropping duplicates in favor of entries
+// in `b`. Both input streams should already be individually sorted, and the
+// returned output stream will have semantic resembling the merge step of a
```
zkat: This is done.

lib/search.js
```diff
@@ -70,24 +54,69 @@ function search (args, silent, staleness, cb) {
...
+    if (er) return cb(er)
+    searchInStream(allEntriesStream, filter, args, opts, searchexclude, silent, cb)
+  })
+}
+
+function searchInStream (stream, filter, args, opts, exclude, silent, cb) {
+  log.silly('search', 'searching packages')
+
+  var filterStream = createFilterStream(filter, opts, exclude, silent)
+  var outputStream = npm.config.get('json')
```
zkat: This was done as part of the refactor. All these options are passed in the `opts` object to the formatter stream.

test/tap/update-index.js
zkat: This comment predates a lot of stuff I ended up doing to make doubleplus sure these timestamps were good in the tests. I've removed it.

lib/search.js
```diff
@@ -69,7 +69,9 @@ function search (args, silent, staleness, cb) {
     log.gauge.pulse('search')
     log.gauge.show({section: searchSection, logline: 'scanning ' + pkg.name})
     // Simply 'true' if the package matches search parameters.
-    var match = packageFilter(pkg, opts, searchexclude)
+    var match = packageFilter(pkg, opts, searchexclude, {
```
othiym23: Changing this to include the description flag is in keeping with the spirit of keeping the config out of the refactored search, but the cascading effects of this are pretty big. It's also weird that there are two options arguments here now.

lib/search/package-filter.js
```diff
@@ -2,14 +2,14 @@
 
 module.exports = filter
 
-function filter (data, args, notArgs) {
+function filter (data, args, notArgs, opts) {
```
othiym23: Picking up from up above, this is a hella confusing arguments list: should it be `included`, `excluded`, and `opts`?

lib/search.js
othiym23: Looking at this in the context of `filter()`, fixing this could be as simple as changing `opts` to `included`, and `searchexclude` to `excluded` to match.

lib/search/all-package-metadata.js
```diff
@@ -18,6 +18,13 @@ var once = require('once')
 // Returns a sorted stream of all package metadata. Internally, takes care of
 // maintaining its metadata cache and making partial or full remote requests,
 // according to staleness, validity, etc.
+//
+// The local cache must hold certain invariants:
```
othiym23: Outstanding. :tada:

lib/search/all-package-metadata.js
```diff
@@ -124,11 +124,11 @@ function createEntryUpdateStream (all, auth, staleness, latest, cb) {
   var partialUpdate = false
   if (latest && (Date.now() - latest < (staleness * 1000))) {
     // Skip the request altogether if our `latest` isn't stale.
-    log.info('all-package-metadata', 'Local data up to date, skipping update')
+    log.verbose('all-package-metadata', 'Local data up to date, skipping update')
     return cb(null)
   } else if (latest === 0) {
     log.warn('', 'Building the local index for the first time, please be patient')
-    log.silly('all-package-metadata', 'No cached data: requesting full metadata db')
+    log.verbose('all-package-metadata', 'No cached data: requesting full metadata db')
```
othiym23: Good eye, making this match with the above.

doc/misc/npm-config.md
```diff
@@ -655,7 +655,7 @@ process is not aborted.
 * Type: Boolean
 
 Output parseable results from commands that write to
-standard output.
+standard output. For `npm search`, this will be tab-separated table format.
```
othiym23: Good start. This is yet more evidence that [`doc/misc/config.md` must be destroyed](https://github.com/npm/npm/issues/7185), because now I really want this to include what the parseable form means for other commands (which is totally out of scope for this PR – DON'T DO THIS).

lib/search.js
```diff
@@ -202,7 +202,7 @@ function prettify (data, num, args) {
 
   if (npm.config.get('tab-separated')) {
     return columns.map(function (col) {
-      return dat[col] && ('' + dat[col]).replace('\t', ' ')
+      return dat[col] && ('' + dat[col]).replace(/\t/g, ' ')
```
othiym23: _sings the "this is why I like tests" song_

lib/search/format-package-stream.js
```diff
@@ -0,0 +1,188 @@
+'use strict'
+
+var ms = require('mississippi')
+var jsonstream = require('JSONStream')
+var columnify = require('columnify')
+
+module.exports = function (opts) {
```
othiym23: Functions feel less lonely with names (also, the CLI's code base has a house style of using

```js
module.exports = hoistedFunctionName
function hoistedFunctionName () {
//...
```
)

lib/search/all-package-metadata.js
```diff
@@ -0,0 +1,302 @@
...
+var jsonstream = require('JSONStream')
+var writeStreamAtomic = require('fs-write-stream-atomic')
+var ms = require('mississippi')
+var sortedUnionStream = require('sorted-union-stream')
+var once = require('once')
+
+// Returns a sorted stream of all package metadata. Internally, takes care of
+// maintaining its metadata cache and making partial or full remote requests,
+// according to staleness, validity, etc.
+module.exports = function (staleness) {
```
othiym23: Same comment as below giving this function a name. Otherwise, _so beautiful_ 😭.

lib/search/stream-filter.js
```diff
@@ -0,0 +1,10 @@
+var ms = require('mississippi')
+
+module.exports = function (filter) {
```
othiym23: Does this need to be a standalone file, given that it's only used in one place and is a pretty simple direct invocation?

othiym23 (approved):
Found a bunch of small things that you can tweak or not, to your taste. I don't think this will need another round of review, although it would probably be good to do another pass at squashing commits (all of the post-refactor test changes can probably be squashed into the refactor changesets, for example).

Your refactor is pretty much exactly what I had in mind. Thank you for bearing with me on that! I think the result is a lot clearer than what we had before, and I think should make things simpler when we decide to make further changes to search.

The documentation for https://docs.npmjs.com/misc/config#searchopts and https://docs.npmjs.com/misc/config#searchexclude are pretty terrible, in that I had to read the code to learn exactly what they did. It might be good to update `doc/misc/config.md` to include a bit more detail about how the two are meant to be used (including mentioning that they're treated like JS regexps, which is a pretty cool feature that I bet _almost nobody_ who hasn't been using npm since 2011 or earlier knows about).

Congratulations! This is really good work!

othiym23:
We _do_ have to get the tests passing before `npm@4` is released, though. I know that @iarna was looking at the Travis failures – maybe the two of you could pair and see if you can get everything ironed out?

lib/search/stream-filter.js
zkat: This is such a baseline tool that I wanted to hide the implementation but it wasn't worth adding `stream-filter` as a dependency. I guess I can just define it as a utility in `search.js`

PR #13858, opened 2016-09-04
Do not needlessly require bash

AMDmi3:
This script does not need any bash features and works fine with any POSIX-compliant shell. Fix for systems which do not have /bin/bash (e.g. FreeBSD).

### DIFF
--- a/node_modules/which/changelog.sh
+++ b/node_modules/which/changelog.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 (
   echo '# Changes'
   echo ''


KenanY:
@AMDmi3 Why not send this [upstream](https://github.com/npm/node-which)?

AMDmi3:
I wasn't aware that's separate project. [Done](https://github.com/npm/node-which/pull/38).

legodude17:
I believe this can be closed, right?

zkat:
closing. the new version of `which` will be included once released as part of our usual dep update process. Thanks for paying attention to this!

PR #13860, opened 2016-09-04
shrinkwrap: save-dev updates shrinkwrap file

szimek:
It fixes the issue where running `npm install --save-dev` would not update the shrinkwrap file. However, it looks like the assumption was to not update it, if there are no dev dependencies already in this file and I haven't changed this behavior (I also added a test for this case). I think that a better idea would be to store info if the shrinkwrap file was created with `--dev` option inside this this file and decide based on that instead.

I'm not sure about the test. I couldn't find any test that would simply check if `install --save` updates the shrinkwrap file, only the [one](https://github.com/npm/npm/blob/master/test/tap/shrinkwrap-save-with-existing-dev-deps.js) that says `install --save leaves dev deps alone`, but seemed to also test if the shrinkwrap file was updated. The new test is based on this one.

Fixes: #11735

### DIFF
--- a/doc/cli/npm-shrinkwrap.md
+++ b/doc/cli/npm-shrinkwrap.md
@@ -137,11 +137,11 @@ To add or update a dependency in a shrinkwrapped package:
 
 1. Run `npm install` in the package root to install the current
    versions of all dependencies.
-2. Add or update dependencies. `npm install --save` each new or updated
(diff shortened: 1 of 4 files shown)


iarna:
This looks fantastic, thank you!

zkat:
Woo! This was merged and included with `npm@3.10.8`. Thanks!

PR #13894, opened 2016-09-08
deps: remove non-deps from bundleDependencies

zkat:
This was spitting out a bunch of warnings when I did `npm install`


zkat:
oops, this is a thing for upgrades from `npm@2` lol. Closing.

PR #13896, opened 2016-09-08
Release: 3.10.8

zkat:


addaleax:
> ` 	write-file-atomic@1.2.0`
> `* Preserve chmod and chown from the overwritten file`

That update is responsible for https://github.com/npm/npm/issues/13918 …  as far as I can tell, that means that npm@3.10.8 won’t perform global installs correctly on OS X with all current Node versions.

It’s due to a bug in Node core, and a fix for that is available, but maybe you’d like to patch up `graceful-fs` or something like that so that older Node.js versions will work fine, too? (I’d put a PR together myself but I’m packing for Amsterdam right now… :smile:)

PR #13901, opened 2016-09-08
Move when rename fails (fix EXDEV docker issues)

chadnickbok:
Fixes issue: https://github.com/npm/npm/issues/9863

Currently, the npm 'rename' function is a thin wrapper around the `rename(2)` syscall. However, rename isn't guaranteed to be implemented by all filesystems, and can cause issues when running inside docker containers.

This patch changes the behavior of the 'rename' function to behave more like the traditional unix `mv` command, falling back to recursively copying and deleting files if renaming them isn't possible.

I based my work on https://github.com/npm/npm/pull/13257 and I've tried to take into account the feedback on this PR:
- Use a recursive strategy, trying to make use of 'rename' for as much of the work as possible (making it slower across-devices, but faster on AUFS for Docker).
- Preserve most important file attributes when copying.
- Testing - I don't think there are any valuable test cases that can be written for this functionality, as the only way to actually test it would be to setup a copy across filesystems, which is very system-dependent.

### DIFF
--- a/lib/utils/rename.js
+++ b/lib/utils/rename.js
@@ -6,85 +6,177 @@ var SaveStack = require('./save-stack.js')
 
 module.exports = rename
 
-/**
- * Rename or move a single directory.
- */
-function rename (src, dest, cb) {
+function rename (from, to, cb) {
   var saved = new SaveStack(rename)
-  try {
-    renameOrCopy(src, dest)
-    try {
-      removeFile(src)
-    } catch (_) {
-      // XXX: Cleanup failures don't throw exception
-    }
-  } catch (renameErr) {
-    return cb(saved.completeWith(renameErr))
-  }
-  return cb()
-}
 
-function removeFile (filename) {
-  var exists = false
-  try {
-    fs.accessSync(filename)
-    exists = true
-  } catch (_) {}
+  var basePath = process.cwd()
+  var currentPath = path.resolve(basePath, from)
+  var targetPath = path.resolve(basePath, to)
 
-  if (exists) {
-    rimraf.sync(filename)
-  }
-}
+  var hadErr = false
 
-function copyDir (src, dest) {
-  var filenames = fs.readdirSync(src)
-  fs.mkdirSync(dest)
+  var started = 1 // Unlike ncp we always start with 1 (the top-level rename)
+  var finished = 0
+  var running = 0
+  var limit = 512
 
-  for (var i = 0; i < filenames.length; i++) {
-    renameOrCopy(path.join(src, filenames[i]), path.join(dest, filenames[i]))
-  }
-}
-
-function copyFile (src, dest) {
-  var oldFile = fs.createReadStream(src)
-  var newFile = fs.createWriteStream(dest)
-  oldFile.pipe(newFile)
-}
-
-function copyLink (src, dest) {
-  var symlink = fs.readlinkSync(src)
-  fs.symlinkSync(symlink, dest)
-}
-
-function syncPermissions (srcStats, dest) {
-  var destFD = fs.openSync(dest, 'r+')
-
-  fs.fchownSync(destFD, srcStats.uid, srcStats.gid)
-  fs.fchmodSync(destFD, srcStats.mode)
-  fs.futimesSync(destFD, srcStats.atime, srcStats.mtime)
-
-  fs.closeSync(destFD)
-}
-
-function renameOrCopy (src, dest) {
-  try {
-    fs.renameSync(src, dest)
-  } catch (renameErr) {
-    var srcStats
-    try {
-      removeFile(dest)
-      srcStats = fs.lstatSync(src)
-    } catch (_) {
-      throw renameErr
+  // First, try a 'standard' rename on the top-level directory
+  fs.rename(currentPath, targetPath, function (renameErr) {
+    if (!renameErr) {
+      return cb(null)
     }
 
-    if (srcStats.isDirectory()) {
-      copyDir(src, dest)
-    } else if (srcStats.isSymbolicLink()) {
-      copyLink(src, dest)
-    } else {
-      copyFile(src, dest)
+    if (renameErr.code !== 'EXDEV' && renameErr.code !== 'EPERM') {
+      return cb(saved.completeWith(renameErr))
     }
-    syncPermissions(srcStats, dest)
+
+    getStats(currentPath)
+  })
+
+  function startRenameOrCopy (source) {
+    if (hadErr) return
+
+    var target = source.replace(currentPath, targetPath)
+    started++
+
+    // Always try to rename first
+    fs.rename(source, target, function (err) {
+      if (!err) {
+        return doneOne(true)
+      }
+
+      return getStats(source)
+    })
+  }
+
+  function getStats (source) {
+    if (running >= limit) {
+      return setImmediate(function () {
+        getStats(source)
+      })
+    }
+    running++
+
+    fs.lstat(source, function (err, stats) {
+      if (err) return onError(err)
+      // We need to get the mode from the stats object and preserve it.
+      var item = {
+        name: source,
+        mode: stats.mode,
+        mtime: stats.mtime, // modified time
+        atime: stats.atime, // access time
+        stats: stats // temporary
+      }
+
+      if (stats.isDirectory()) {
+        return onDir(item)
+      } else if (stats.isFile()) {
+        return onFile(item)
+      } else if (stats.isSymbolicLink()) {
+        // Symlinks don't really need to know about the mode.
+        return onLink(source)
+      }
+    })
+  }
+
+  function onFile (file) {
+    var target = file.name.replace(currentPath, targetPath)
+    copyFile(file, target)
+  }
+
+  function copyFile (file, target) {
+    var readStream = fs.createReadStream(file.name)
+    var writeStream = fs.createWriteStream(target, { mode: file.mode })
+
+    readStream.on('error', onError)
+    writeStream.on('error', onError)
+
+    writeStream.on('open', function () {
+      readStream.pipe(writeStream)
+    })
+
+    writeStream.once('finish', function () {
+      setFilePermissions(file, target)
+    })
+  }
+
+  function setFilePermissions (file, target) {
+    fs.chmod(target, file.mode, function (err) {
+      if (err) return onError(err)
+
+      fs.open(path, 'r+', function (err, fd) {
+        if (err) return onError(err)
+
+        fs.futimes(fd, file.atime, file.mtime, function (futimesErr) {
+          fs.close(fd, function (closeErr) {
+            if (futimesErr) return onError(futimesErr)
+            if (closeErr) return onError(closeErr)
+            return doneOne()
+          })
+        })
+      })
+    })
+  }
+
+  function onDir (dir) {
+    var target = dir.name.replace(currentPath, targetPath)
+    return mkDir(dir, target)
+  }
+
+  function mkDir (dir, target) {
+    fs.mkdir(target, dir.mode, function (err) {
+      if (err) return onError(err)
+      // despite setting mode in fs.mkdir, doesn't seem to work
+      // so we set it here.
+      fs.chmod(target, dir.mode, function (err) {
+        if (err) return onError(err)
+        copyDir(dir.name)
+      })
+    })
+  }
+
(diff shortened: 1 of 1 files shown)


lib/utils/rename.js
```diff
@@ -1,16 +1,90 @@
...
+  var exists = false
+  try {
+    fs.accessSync(filename)
+    exists = true
+  } catch (_) {}
+
+  if (exists) {
+    rimraf.sync(filename)
+  }
+}
```
legodude17: Any reason this file is sync? It should be async to conform to the rest of npm.

lib/utils/rename.js
```diff
@@ -1,16 +1,90 @@
...
+
+function copyDir (src, dest) {
+  var filenames = fs.readdirSync(src)
+  fs.mkdirSync(dest)
+
+  for (var i = 0; i < filenames.length; i++) {
+    renameOrCopy(path.join(src, filenames[i]), path.join(dest, filenames[i]))
+  }
+}
```
legodude17: As well as this one.

lib/utils/rename.js
```diff
@@ -1,16 +1,90 @@
...
+
+function copyFile (src, dest) {
+  var oldFile = fs.createReadStream(src)
+  var newFile = fs.createWriteStream(dest)
+  oldFile.pipe(newFile)
+}
```
legodude17: Shouldn't this take a call back?

lib/utils/rename.js
```diff
@@ -1,16 +1,90 @@
...
+
+function copyLink (src, dest) {
+  var symlink = fs.readlinkSync(src)
+  fs.symlinkSync(symlink, dest)
+}
```
legodude17: This should be sync.

lib/utils/rename.js
```diff
@@ -1,16 +1,90 @@
...
+
+function syncPermissions (srcStats, dest) {
+  var destFD = fs.openSync(dest, 'r+')
+
+  fs.fchownSync(destFD, srcStats.uid, srcStats.gid)
+  fs.fchmodSync(destFD, srcStats.mode)
+  fs.futimesSync(destFD, srcStats.atime, srcStats.mtime)
+
+  fs.closeSync(destFD)
+}
```
legodude17: And this.

lib/utils/rename.js
```diff
@@ -1,16 +1,90 @@
...
+    } else if (srcStats.isSymbolicLink()) {
+      copyLink(src, dest)
     } else {
-      return cb()
+      copyFile(src, dest)
     }
-  })
+    syncPermissions(srcStats, dest)
+  }
 }
```
legodude17: And this.

legodude17:
A bunch of line comments, mostly on the use of sync. This seems like a really good idea.

lib/utils/rename.js
chadnickbok: Yes, looks like I didn't read the documentation closely enough!

chadnickbok:
Thanks for the comments!

Answer to all of them right now is - because the original `rename` was an atomic call, I can't drop-in replace it with an async operation. Basically right now anywhere that uses `rename` should expect it to be async, but it isn't. This patch fixes the behavior of `rename` to actually work on filesystems that don't support the `rename(2)` syscall, but it doesn't address all the places that would need to change to support an async call.

My proposal is to first fix the behavior of rename to be _correct_; then we can review and address making it async.

chadnickbok:
Great comments, this should absolutely be async. I'll move over to async operations for what I can

lib/utils/rename.js
```diff
@@ -1,16 +1,182 @@
 'use strict'
 var fs = require('graceful-fs')
+var path = require('path')
+var rimraf = require('rimraf')
```
legodude17: Shouldn't you use `lib/utils/gentlyRm.js` instead?

chadnickbok: Yes, great suggestion. I probably should have familiarized myself with the rest of the guild directory before jumping in with a rewrite!

What do you think of the rest of this code?

lib/utils/rename.js
```diff
@@ -1,16 +1,182 @@
...
 var SaveStack = require('./save-stack.js')
 
 module.exports = rename
 
 function rename (from, to, cb) {
   var saved = new SaveStack(rename)
```
legodude17: What does this do?

chadnickbok: Its from the original `rename` - I believe it makes sure that the stack trace includes the call to this function (although tbh I'm not entirely sure)

lib/utils/rename.js
chadnickbok: Unfortunately not.

gentlyRm takes paths relative to the npm prefix: https://github.com/npm/npm/blob/master/lib/utils/gently-rm.js#L54

Whereas `rename` takes paths relative to `process.pwd` (this is implicit in the call to `rename(2)`)

Changing over to using `gentlyRm` breaks tests; I'd need to refactor all callers of `rename` to make this change, and its not something I think should be in scope for this PR

chadnickbok: Great suggestion! I didn't realize what gentlyRm was doing, but it seems reasonable :)

lib/utils/rename.js
legodude17: Yeah I saw that now.

rachael:
I like this PR as a replacement for my PR #13257. @chadnickbok is more knowledgeable than me about the necessities of interacting with filesystems and the addition of async makes it much better than my band-aid fix.

@zkat if my PR was good to merge without any additional tests would that mean this one is good to go?

lib/utils/rename.js
iarna: Yup, without this errors in rename have no stacktrace at all, because they lead back to the callback being called from Node.js internals.

lib/utils/rename.js
```diff
@@ -1,16 +1,182 @@
...
+      if (!err) {
+        return doneOne(true)
+      }
+
+      return getStats(source)
+    })
+  }
+
+  function getStats (source) {
+    if (running >= limit) {
```
iarna: So... if it hits its concurrency limit it... hard loops using setImmediate until some of the outstanding `lstat`s complete?

I'm not comfortable with this implementation. This should either be using a tool to limit concurrency (like [call-limit](https://www.npmjs.com/package/call-limit)) or doing so structurally.

I'll come back to that second bit below…

lib/utils/rename.js
```diff
@@ -1,16 +1,182 @@
...
+  }
+
+  function copyFile (file, target) {
+    var readStream = fs.createReadStream(file.name)
+    var writeStream = fs.createWriteStream(target, { mode: file.mode })
+
+    readStream.on('error', onError)
+    writeStream.on('error', onError)
+
+    writeStream.on('open', function () {
```
iarna: Why wait for `open` to start piping?

lib/utils/rename.js
```diff
@@ -1,16 +1,182 @@
...
+      return cb(saved.completeWith(renameErr))
+    }
+
+    getStats(currentPath)
   })
+
+  function startRenameOrCopy (source) {
+    if (hadErr) return
+
+    var target = source.replace(currentPath, targetPath)
```
iarna: Using text replacement here feels brittle to me…

lib/utils/rename.js
```diff
@@ -1,16 +1,182 @@
...
+    fs.rename(source, target, function (err) {
+      if (!err) {
+        return doneOne(true)
+      }
+
+      return getStats(source)
+    })
+  }
+
+  function getStats (source) {
```
iarna: This function name seems a bit too centered on the mechanics and not the logic of what it's doing. Which is… looking at what kind of thing the path is and selecting the appropriate path for renameOrCopy?

lib/utils/rename.js
```diff
@@ -1,16 +1,182 @@
...
+        if (err) return onError(err)
+        copyDir(dir.name)
+      })
+    })
+  }
+
+  function copyDir (dir) {
+    fs.readdir(dir, function (err, items) {
+      if (err) return onError(err)
+      items.forEach(function (item) {
```
iarna: So... this is where you end up having to have your unusual concurrency controls.

I would use something like [`asyncMap`](https://www.npmjs.com/package/slide#asyncmap) or [`chain`](https://www.npmjs.com/package/slide#chain) to iterate over the directory entries.

That doesn't automatically save you from the giant async-bomb of recursion going on here but I'll talk more about that in my general review comment.

iarna (changes requested):
Mechanically I'm happy with what this code is doing.

I'm not comfortable with the way it's recursing. Limiting that recursion with the `setImmediate` loop is a symptom of that, but not in and of itself the problem.

I would prefer that this use a work queue and handle one file at a time. This theoretically is slower, but in practice this shouldn't be happening too often and copying is already plenty slow.

That... might be a little vague? Here's some psuedocode of how I'd approach it: (Uhm, sorry in advance if I've swung way too far over into "too specific" here. =D)

``` js
function safeRename (from, to, cb):
  workQueue = [[tryRename, from, to]]
  _safeRename()
  function _safeRename:
    if (!workQueue.length)
    [action, from, to] = workQueue.shift()
    action(from, to, _safeRename)
  function tryRename (from, to, cb):
    fs.rename(from, to, function (err) {
      if (err):
        if (err is EXDEV or other docker error):
          workQueue.push([copy, from, to])
        else:
          return cb(err)
      cb()
    })
  function copy (from, to, cb):
    fs.stat(from, to, function (err, stat) {
      if (err) return cb(err)
      if (isFile) workQueue.push([copyFile, from, to])
      else if (isSymlink) workQueue.push([copySymlink, from to])
      else if (isDir) workQueue.push([copyDir, from, to])
      else return cb(someError)
      cb()
    })
  function copyDir(from, to, cb):
    fs.readdir(from, function (err, items) {
      if (err) return cb(err)
      items.forEach(function (item) {
        workQueue.push([tryRename, path.join(from, item), path.join(to, item)])
      })
      cb()
    })

  copyFile & copySymlink would work as they do now, except with callbacks for when they're done.
```

iarna:
I've discussed this further with @othiym23 & @zkat and we agree that this will need tests. I'd like to see two sorts of things:

1. A test that shows that exercises the copying code being introduced here. To write this you'd need to use `requireInject` to replace `graceful-fs`'s `rename` method with something that failed with EXDEV for certain directories. I'd like to see this test exercise copyFile, copyDir & copySymlink.

2. A functional test that, without this patch, fails consistently in a docker environment would be super useful. Bonus points for including a Dockerfile that runs said test. I expect this to be doable in light of the Dockerfile in #9863. The only changeup from what it's doing is that the test needs to not install modules off the network.

chadnickbok:
Okay, great comments I'll try to implement the suggestions over the next day or two.

One thing I do want to mention is that I based these changes on the widely-deploy `ncp` utility - https://github.com/AvianFlu/ncp/blob/master/lib/ncp.js
That's where the non-standard concurrency controls come in.

Before I go any further, I want to be _really_ clear that the proposed work-queue based approach will be slow in the case where all files need to be copied, and that we're _explicitly_ signing up for that slowness. In the case of docker and aufs, this shouldn't actually be terribly slow, since we'll be making use of `rename` for most of that (and making the kernel do the hard work), but for filesystems that require a full copy its going to be significantly slower.

iarna:
It is, but I'm also not aware of any scenarios where this would trigger besides Docker and aufs. At least, thus far, you all are the only one's making noise about it.

And if need be we can rejigger the work queue to handle _n_ items in it concurrently, when the time comes.

iarna:
Incidentally: I traced the ncp code back to a commit from 2011(!) which really does a lot to explain why it's so weird. =D

lib/utils/rename.js
legodude17: Yeah, shouldn't pipe automatically wait for both streams to be ready?

justincormack:
It will be difficult to do a functional test that fails in a Docker environment as that Docker environment needs to be using `aufs` or `overlay` which not all are.

You can perhaps make a test in an environment with some bind mounted files, not sure.

iarna:
@justincormack Personally, I'd make a functional test that injects an `fs` replacement that emits `EXDEV` when trying to rename from `.staging` to an ordinary `node_modules` location.

iarna:
This is superseded by #15901.

PR #13904, opened 2016-09-08
Update example to show feature branch reference

stevenkissack:
When manually adding a dependency reference to a github branch with the pattern feature/[branchName](same for bugfix, etc) npm fails with a forward slash, you need a backslash to escape this and everything works fine. Updated the documentation to reflect this.


legodude17:
I would suggest changing the text before the thing you changed as well.

iarna:

Issue #13496, opened 2016-07-27
npm3 shrinkwrap not support latest semver well.


xuxin:
#### I'm opening this issue because:

  - [ ] npm is crashing.
  - [x] npm is producing an incorrect install.
  - [ ] npm is doing something I don't understand.
  - [ ] Other (_see below for feature requests_):

#### What's going wrong?
shrinkwrap with devendencies semver latest goes wrong.

1、my package.json is like this:
```javascript
    {
         "dependencies": {
            "A": "^0.0.1"}}
``` 
the A package's dependencies is : 
```javascript
    {  
           "name": "A",  
           "version": "0.1.0",  
           "dependencies": { "B": "latest" } }
```    
 `npm install` build my node_module:
```
  └─┬   A@0.0.1
     ├── B@0.0.1
```
`npm shrinkwrap` get my shrinkwrap.json:
```javascript
{
  "dependencies": {
    "A": {
      "version": "0.0.1",
      "from": "A@>=0.0.1 <0.1.0",
      "resolved": "**** A-0.0.1.tgz"
    },
    "B":  {
      "version": "0.0.1",
      "from": "B@latest",
      "resolved": "**** A-0.0.1.tgz"
    }
    }
}
```
now ,everything is right.

2、 delete node_module dir and run `npm install` from npm-shrinkwrap.json again, my node_module become:
```
    ├─┬ A@0.0.1
    │  └──  B@0.0.1
    ├── B@0.0.1
```
and when I run `npm shrinkwrap`:
```javascript
{
  "dependencies": {
    "A": {
      "version": "0.0.1",
      "from": "A@>=0.0.1 <0.1.0",
      "resolved": "**** A-0.0.1.tgz",
      "dependencies": {
           "B":  {
          "version": "0.0.1",
          "from": "B@latest",
          "resolved": "**** A-0.0.1.tgz"}
        }
    }
    }
}
```
now the npm-shrinkwarp.json was changed.

3、remove node_module , and `npm install` the module goes **wrong**
```
    ├─┬ A@0.0.1
    │  └──  B@0.0.1
    ├── B@0.0.1  invalid
```
`npm ls` show: 
```
npm ERR! invalid: B@0.0.1 /node_modules/A/node_modules/B
```
`npm shrinkwrap` show like this:
```
npm ERR! Darwin 15.6.0
[...]

PR #13941, opened 2016-09-12
Fix tagged dependency matching with shrinkwraps

iarna:
When you depend on a tag, eg `foo@latest`, your `npm-shrinkwrap.json` will record this in its `_from` field. Similarly when you depend on a git dependency with something like `github:foo/bar` it saves this to your `_from`.

When you install under the above circumstances the installed module will be installed from the associated tarball URL for tagged dependencies or git URL with commit hash for git dependencies.

As a result the `_requested` section of the `package.json` will be based on the tarball URL or githash and cann't be matched up to the `package.json`.

It was _supposed_ to fallback to comparing the `package.json` `_from` field with but it was doing this by comparing `rawSpec` to `_from`, which _can't_ work consistently. 

Fixes: #13496
Fixes: #11736

### DIFF
--- a/lib/install/deps.js
+++ b/lib/install/deps.js
@@ -64,11 +64,26 @@ function doesChildVersionMatch (child, requested, requestor) {
   if (requested.spec === '*') return true
 
   var childReq = child.package._requested
+  if (!childReq) childReq = npa(moduleName(child) + '@' + child.package._from)
   if (childReq) {
     if (childReq.rawSpec === requested.rawSpec) return true
     if (childReq.type === requested.type && childReq.spec === requested.spec) return true
   }
-  if (!registryTypes[requested.type]) return requested.rawSpec === child.package._from
+  // If _requested didn't exist OR if it didn't match then we'll try using
+  // _from. We pass it through npa to normalize the specifier.
+  // This can happen when installing from an `npm-shrinkwrap.json` where `_requested` will
+  // be the tarball URL from `resolved` and thus can't match what's in the `package.json`.
+  // In those cases _from, will be preserved and we can compare that to ensure that they
+  // really came from the same sources.
+  // You'll see this scenario happen with at least tags and git dependencies.
+  if (!registryTypes[requested.type]) {
+    if (child.package._from) {
+      var fromReq = npa(child.package._from)
+      if (fromReq.rawSpec === requested.rawSpec) return true
+      if (fromReq.type === requested.type && fromReq.spec === requested.spec) return true
+    }
+    return false
+  }
   return semver.satisfies(child.package.version, requested.spec)
 }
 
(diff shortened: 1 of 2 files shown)


othiym23:
This looks like a simple, straightforward bug fix to me, and the test makes sense in light of #13946 – maybe drop in a URL to that in a comment in the test itself so we know why it's set up the way it is?

My only other comment is that this change could end up having an effect on Git dependencies locked via shrinkwrap (see #11736 for an example). Probably for the better, but worth reviewing before merging.

Aside from that one nit and one observation, this looks good to me! :sheep::rocket::fire:

iarna:
I've updated this with a slightly different patch in order to fix both git AND tagged dependencies. I also improved the test to be a little more specific and self-descriptive around what it expects.

lib/install/deps.js
```diff
@@ -64,11 +64,25 @@ function doesChildVersionMatch (child, requested, requestor) {
...
   var childReq = child.package._requested
+  if (!childReq) childReq = npa(moduleName(child) + '@' + child.package._from)
   if (childReq) {
     if (childReq.rawSpec === requested.rawSpec) return true
     if (childReq.type === requested.type && childReq.spec === requested.spec) return true
   }
-  if (!registryTypes[requested.type]) return requested.rawSpec === child.package._from
+  // If _requested didn't exist OR if it didn't match then we'll try using
+  // _from. We pass it through npa to normalize the specifier.
+  // This can happen when isntalling from an `npm-shrinkwrap.json` where `_requested` will
```
othiym23: s/isntalling/installing/

test/tap/tagged-version-matching.js
```diff
@@ -0,0 +1,162 @@
...
+test('tagged-version-matching', function (t) {
+  common.npm(['ls', '--json'], conf, function (err, code, stdout, stderr) {
+    if (err) throw err
+    t.is(code, 0, 'command ran ok')
+    if (stderr.trim()) t.comment(stderr.trim())
+    var result = JSON.parse(stdout.trim())
+    var problems = result.problems || []
+    // Original PR: https://github.com/npm/npm/pull/13941
+    // Original issue: https://github.com/npm/npm/issues/13496
+    // Original issue: https://github.com/npm/npm/issues/11736
```
othiym23: ty!

lib/install/deps.js
```diff
@@ -64,11 +64,25 @@ function doesChildVersionMatch (child, requested, requestor) {
...
+  // be the tarball URL from `resolved` and thus can't match what's in the `package.json`.
+  // In those cases _from, will be preserved and we can compare that to ensure that they
+  // really came from the same sources.
```
othiym23: Since I know you made some changes to this to make sure Git dependencies work properly here, maybe mention Git in this comment?

othiym23 (approved):
Two nits, otherwise still LGTM! :sheep::rocket::fire:

lib/install/deps.js
iarna: This is currently a description of the higher level thing that was wrong with both tagged and git dependencies. (It was a singular root cause.)

Issue #13252, opened 2016-07-02
Shrinkwrap lifecycle scripts have wrong cwd


STRML:
#### I'm opening this issue because:

Shrinkwrap lifecycle scripts behave oddly.

#### What's going wrong?

For most lifecycle scripts, the cwd is the root of the project. For example,
a scripts like this resolve fine:

```json
"dev": "./run-dev.sh",
"preCommit": "scripts/precommit.sh"
```

However, `preshrinkwrap` and `postshrinkwrap` resolve relative to `./node_modules`, making the below an error:

```json
"preshrinkwrap": "scripts/pre-shrinkwrap.sh",
"postshrinkwrap": "scripts/post-shrinkwrap.sh",
```

```bash
$ npm shrinkwrap

> x-x @1.1.0 preshrinkwrap /Users/x/x/x/node_modules
> scripts/pre-shrinkwrap.sh

sh: scripts/pre-shrinkwrap.sh: No such file or directory
```

You must instead prepend `../`:

```json
"preshrinkwrap": "../scripts/pre-shrinkwrap.sh",
"postshrinkwrap": "../scripts/post-shrinkwrap.sh",
```

This also means that any scripts these lifecycles call have a CWD in `node_modules` instead of the root of the project, which can cause breakage.

#### How can the CLI team reproduce the problem?

Add any preshrinkwrap/postshrinkwrap lifecycle scripts.

### Supporting information:

 - `npm -v` prints: 3.10.3
 - `node -v` prints: v6.2.1
 - Windows, OS X, or Linux?: OS X

iarna:
I've tried to reproduce this and it's not happening for me. Can you put together a git repo that displays this behavior?

Here's what I tried:

**package.json**:
```
{
  "name": "13253",
  "version": "1.0.0",
  "scripts": {
    "preshrinkwrap": "scripts/shrinkwrap.sh pre",
    "shrinkwrap": "scripts/shrinkwrap.sh -",
    "postshrinkwrap": "scripts/shrinkwrap.sh post"
  }
}
```

**scripts/shrinkwrap.sh** (mode: 0755):
```
#!/bin/bash
echo Shrinkwrap $1 $(pwd)
```

Result of run:
```
$ npm --version
3.10.3
$ npm shrinkwrap

> 13253@1.0.0 preshrinkwrap /Users/rebecca/code/npmtest/13253
> scripts/shrinkwrap.sh pre

Shrinkwrap pre /Users/rebecca/code/npmtest/13253

[...]

PR #13942, opened 2016-09-13
Fix CWD while running shrinkwrap lifecycle scripts

iarna:
Previously if you ran a shrinkwrap from another lifecycle script AND `node_modules` existed (and if you're running `npm shrinkwrap` it probably should) then `npm` would run the shrinkwrap lifecycle from the `node_modules` folder instead of the package folder.

Fixes: #13252

### DIFF
--- a/lib/shrinkwrap.js
+++ b/lib/shrinkwrap.js
@@ -35,12 +35,11 @@ function shrinkwrap (args, silent, cb) {
     log.warn('shrinkwrap', "doesn't take positional args")
   }
 
-  var dir = path.resolve(npm.dir, '..')
   var packagePath = path.join(npm.localPrefix, 'package.json')
   var dev = !!npm.config.get('dev') || /^dev(elopment)?$/.test(npm.config.get('also'))
 
   readPackageJson(packagePath, iferr(cb, function (pkg) {
-    createShrinkwrap(dir, pkg, dev, silent, cb)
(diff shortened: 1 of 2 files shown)


othiym23:
LGTM! Had to track down how we came to rely on `dir` in the first place, but once I did the changes seem simple and unlikely to cause problems. :sheep::rocket::fire:

evocateur:
Hah, I knew I shoulda opened my pull request last night: https://github.com/npm/npm/compare/master...evocateur:shrinkwrap-lifecycle-path

Identical fix, with some more paranoid (probably redundant) tests. Also like the simplification of `dir` to `npm.localPrefix`, though I omitted that to keep the PR scope narrow.

For those looking for a workaround in the meantime, preface your shrinkwrap lifecycle executions with `cd $(npm prefix) && $(npm bin)/<executable> ...`. (All other lifecycle scripts work fine without this workaround)

For example,
```json
"scripts": {
  "preshrinkwrap": "./scripts/localThing.sh",
  "postshrinkwrap": "devDependencyBin --with-args"
}
```
becomes
```json
"scripts": {
  "preshrinkwrap": "cd $(npm prefix) && ./scripts/localThing.sh",
  "postshrinkwrap": "cd $(npm prefix) && $(npm bin)/devDependencyBin --with-args"
}
```

Thanks @iarna for the quick turnaround!

iarna:
This has been merged and will be included in the next npm release.

Issue #13963, opened 2016-09-15
preuninstall script stops working in npm 3.0


tarmolov:
#### I'm opening this issue because:

  - [ ] npm is crashing.
  - [x] npm is producing an incorrect install.
  - [ ] npm is doing something I don't understand.
  - [ ] Other (_see below for feature requests_):

#### How can the CLI team reproduce the problem?

Do some preparation work.
```
$ git clone https://github.com/tarmolov/git-hooks-tests
$ cd git-hooks-tests
$ npm install
```

In npm 2.x everything works fine
```
$ node --version
v4.3.2
$ npm --version
2.14.12
$ npm install git-hooks@1.1.1

> git-hooks@1.1.0 preuninstall /Users/tarmolov/git-hooks-test/node_modules/git-hooks
> git-hooks --uninstall


> git-hooks@1.1.1 postinstall /Users/tarmolov/git-hooks-test/node_modules/git-hooks
> git-hooks --install

git-hooks@1.1.1 node_modules/git-hooks
```

But in npm 3.x — doesn't.
```
$ node --version
v6.5.0
$ npm --version
3.10.3
$ npm install git-hooks@1.1.1

> git-hooks@1.1.1 postinstall /Users/tarmolov/git-hooks-test/node_modules/git-hooks
> git-hooks --install

git-hooks-test@0.0.1 /Users/tarmolov/git-hooks-test
└── git-hooks@1.1.1  invalid
```

Only `install` section was performed. Also it shows `invalid` label right from the package name.

### supporting information:

 - `npm -v` prints:
3.10.3
 - `node -v` prints:
v6.5.0
 - `npm config get registry` prints: https://registry.npmjs.org/
 - Windows, OS X, or Linux?:
osx, version 10.11.6
 - Network issues:
   - Geographic location where npm was run:
   - [ ] I use a proxy to connect to the npm registry.
   - [ ] I use a proxy to connect to the web.
   - [ ] I use a proxy when downloading Git repos.
   - [ ] I access the npm registry via a VPN
   - [ ] I don't use a proxy, but have limited or unreliable internet access.
 - Container:
   - [ ] I develop using Vagrant on Windows.
   - [ ] I develop using Vagrant on OS X or Linux.
   - [ ] I develop / deploy using Docker.
   - [ ] I deploy to a PaaS (Triton, Heroku).

iarna:
To be specific, it doesn't run the uninstall lifecycles when _updating_ a module.

PR #13964, opened 2016-09-15
Make sure remove steps are run when updating

iarna:
This ultimately means running the `unbuild` phase. 

Fixes: #13963


othiym23:
I really need to figure out why `test/tap/whoami.js` is failing on Travis, but aside from that, this LGTM. :sheep::rocket::fire:

PR #13989, opened 2016-09-17
git: Use shallow clones making some installs faster

odinho:
I noticed that some packages took a lot of time installing.  Some Git repositories have a lot of cruft in them. `npm` sometimes installs from Git repositories, I'd guess this happens a bit more often behind closed doors in companies. Though I guess it's somewhat frowned upon for other reasons thas just speed. Anyway, this change will help me a lot.

By doing a shallow clone, `npm` will get the code that was requested, but without any extra history that probably won't be used anyway.  In some cases this will be a lot faster.
## Example

I did some tests on a big-ish repository and created a hopefully reproducible example:

```
nodetest$ cat package.json
{ "name": "clonetest", "dependencies": {
   "videojs-contrib-hls": "hola/videojs-contrib-hls#v2.2.0-12" } }

nodetest (before)$ time npm install
real    6m9.858s
user    0m25.391s
sys     0m21.009s

nodetest (after)$ time npm install
real    2m24.738s
user    0m14.777s
sys     0m18.674s
```

YMMV based on bandwidth, computer speed etc.  There's also a nice side effect:

```
(before)$ du -hs ~/.npm/_git-remotes/
184M    /home/odin/.npm/_git-remotes/

(after)$ du -hs ~/.npm/_git-remotes/
40M     /home/odin/.npm/_git-remotes/
```
## Information about the pull request (commit body)

> git: Use shallow clones making some installs faster
> 
> This change makes npm require Git 1.9, released February 2014.
> 
> Some git repositories are huge, npm only needs the head of each ref it
> wants to install.  To make sure it can still potentially reuse the
> repository for different branches/tags and also update itself, it will
> always 'git fetch origin <the_ref>' which makes sure that single ref is
> available and the newest version.
> 
> If the ref is a SHA, however, npm will fall back to fetching the entire
> repository again, because we can't know where said SHA is.  Some
> upstream servers allow you to fetch a SHA, but we can't rely on it.

### DIFF
--- a/doc/cli/npm-install.md
+++ b/doc/cli/npm-install.md
@@ -179,6 +179,9 @@ after packing it up into a tarball (b).
     If the repository makes use of submodules, those submodules will
     be cloned as well.
 
+    If you set the "git-shallow-clone" config, the clone will be shallow with
+    depth 1, unless `<commit-ish>` is a SHA.
+
     The following git environment variables are recognized by npm and will be added
     to the environment when running git:
 
--- a/doc/misc/npm-config.md
+++ b/doc/misc/npm-config.md
@@ -364,6 +364,14 @@ The command to use for git commands.  If git is installed on the
 computer, but is not in the `PATH`, then set this to the full path to
 the git binary.
 
+### git-shallow-clone
+
+* Default: `false`
+* Type: Boolean
+
+Use shallow clone when fetching git repositories.  This will only fetch the
+newest commit, making clones faster.  Only supported with Git 1.9+.
+
 ### git-tag-version
 
 * Default: `true`
--- a/lib/cache/add-remote-git.js
+++ b/lib/cache/add-remote-git.js
@@ -201,8 +201,11 @@ function mirrorRemote (from, cloneURL, treeish, cachedRemote, silent, cb) {
       '--mirror',
       cloneURL, cachedRemote
     ]
+    if (npm.config.get('git-shallow-clone')) {
+      args.splice(1, 0, '--depth=1')
+    }
     git.whichAndExec(
-      ['clone', '--template=' + templates, '--mirror', cloneURL, cachedRemote],
+      args,
       { cwd: cachedRemote, env: gitEnv() },
       function (er, stdout, stderr) {
         if (er) {
@@ -216,7 +219,7 @@ function mirrorRemote (from, cloneURL, treeish, cachedRemote, silent, cb) {
           return cb(er)
         }
         log.verbose('mirrorRemote', from, 'git clone ' + cloneURL, stdout.trim())
-        setPermissions(from, cloneURL, treeish, cachedRemote, cb)
+        updateRemote(from, cloneURL, treeish, cachedRemote, cb)
       }
     )
   })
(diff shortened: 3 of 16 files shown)


odinho:
Hm, so I see that Ubuntu 12.04 Precise LTS is supported all the way until May 2017.  That's the last one which has old Git without the required 1.9.  I thought almost 3 years old would be enough, but it seems like it's very sadly not :(

Might be some things to check on Windows too.

How have you handled similar problems earlier?  Do you do version checks and branching, or always go for lowest common denominator?

odinho:
I added a new commit which introduces a configuration variable `git-shallow-clone` that is disabled by default.

This way it should be safe for Ubuntu 12.04.  It'd be possible to flip the switch and remove the config in a few versions.

I didn't add this as a new commit, as your contributor guidelines seems to ask for squashed commits.

iarna:
A couple things:

* The current process let's us fetch down a repo once (for your cache) even when you're installing from multiple branches.
* For Github, shallow clones can be FAR more expensive: https://github.com/CocoaPods/CocoaPods/issues/4989#issuecomment-193772935
* Bumping the required version of `git` may potentially break existing users. My bar for this would be that all supported LTS RedHat, Ubuntu & Debian releases need to ship at least this version.

Don't get me wrong, the benefits of doing a shallow clone here may out weigh those concerns.

TBH, personally I would be more inclined to just directly fetch tarballs for the commitish when the source is a service (like Github) known to provide that feature. (This'd require a bit more discussion, I'm not sure how comfortable @othiym23 is around this idea. If we were to do it, it'd be something along the lines of adding a `tarball()` method to [hosted-git-info](https://github.com/npm/hosted-git-info) and using that if it was available to route to `lib/cache/add-remote-tarball.js`.)

odinho:
Hi iarna!

> - The current process let's us fetch down a repo once (for your cache) even when you're installing from multiple branches.

Indeed, I've made that also work in the patch.  It will `fetch --depth=1` those other branches too.  It still goes a lot faster, and Git will reuse objects, so that if the difference between `branch1` and `branch2` is a file, it will be smart about not taking stuff it already has.

> - For Github, shallow clones can be FAR more expensive: CocoaPods/CocoaPods#4989 (comment)

That's a great discussion and reference. Thank you. There are some big differences to this patch and the CocoaPods though, first and most importantly of all, in this patch it always uses `fetch --depth=1`, so we don't get into the badness of CocoaPods.  If we ever get into the situation that we need to do a full fetch (that is, when users have specified a `SHA` instead of a ref) it will actually do so by doing `--depth=32_MAX_INT` that git has a special case for.  This will make the clone a full clone.

However, the point is a really good one and it's another reason to having an option and keeping that off by default.

> - Bumping the required version of git may potentially break existing users. My bar for this would be that all supported LTS RedHat, Ubuntu & Debian releases need to ship at least this version.

Yep, I assumed that, I was using Ubuntu LTS as a proxy since last time I checked it was slowest at updating, might not be like that this time around though. :)  Anyway, I already made the change dependent on a config - if this was at all to be considered, I'm now thoroughly convinced it should be this way.

Shallow clones *do* make a lot of sense, but if and only if:
- You have a big repository
- You reference using tags and/or branches, not SHAs

I've been running this myself on our projects and it has been a huge boost in performance.  I expect most uses will not be as severe though.

--

Fetching using the GH tarballs is also another fix that will solve the slowness. Though in some cases the clone will still be faster.

This change fixes the problem, but requires user configuration, which I think is the right move, especially after taking GitHubs issues into consideration (though this will be nicer since it doesn't *not* use `--depth=1` also on subsequent fetches).

Anyway, thanks for your time and having a look at this issue :)

othiym23:
After reviewing the back-and-forth here, the CLI team isn't going to land this change, mostly for the reasons that @iarna enumerates. Perhaps the biggest flag for me personally is using anything that might make using npm on Windows more difficult; I think the common distros of Git for Windows are all post-1.9 now, but not all of them are, and people get very confused by the dependency on Git on Windows as it is. Also, we do have a number of enterprise users who are using distributions that are far older than anybody, possibly including them, are comfortable with. The current Git-handling code is a result of a lot of trying something, and then fixing it when somebody inevitably ends up hitting an edge case that blocks them.

The performance improvement you demonstrate in your benchmark is significant but (I think) atypical; most JavaScript Git dependencies are much smaller, and really, it's the repeated cloning and repacking of Git dependencies that makes them so much slower than registry-hosted dependencies. The team intends to overhaul how we deal with Git dependencies overall as part of the cache rewrite we're doing as part of `npm@5`; my hope is that will vastly diminish the amount of superfluous cloning and packing we do with Git dependencies (although they'll probably always be slower to install than registry dependencies, given that registry dependencies come down prepared for use in a way that Git deps just can't).

In any case, thank you for your time and your discussion!

odinho:
OK, it was a good try :+1:  Good luck with the improvements.

We were running this patched for a while, but I have now transferred all the big repos to `@scoped` NPM packages instead, since I anticipated this not going in. ;)

PR #13995, opened 2016-09-18
Fix afterUser() function to get root npm dir

alissonlinneker:
The directory in afterUser() can not be obtained by conf.get('prefix'), as it must necessarily be the root directory npm, or it will not proceed with loadExtras, line 241 `// Without prefix, nothing will ever work`

My problem was that conf.get('prefix') was getting a bug directory and I could not run any command npm! See my problem: 

http://stackoverflow.com/questions/39540427/npm-commands-not-work

### DIFF
--- a/lib/config/core.js
+++ b/lib/config/core.js
@@ -151,14 +151,14 @@ function load_ (builtin, rc, cli, cb) {
     // Eg, `npm config get globalconfig --prefix ~/local` should
     // return `~/local/etc/npmrc`
     // annoying humans and their expectations!
-    var rootdir = path.resolve(__dirname, '../..', '')
-    if (rootdir) {
-      var etc = path.resolve(root, 'etc')
+    conf.set('prefix', path.resolve(__dirname, '../..', ''))
+    if (conf.get('prefix')) {
+      var etc = path.resolve(conf.get('prefix'), 'etc')
       mkdirp(etc, function () {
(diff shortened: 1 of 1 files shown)


alissonlinneker:
Consider the fix:

```
function afterUser () {
    // globalconfig and globalignorefile defaults
    // need to respond to the 'prefix' setting up to this point.
    // Eg, `npm config get globalconfig --prefix ~/local` should
    // return `~/local/etc/npmrc`
    // annoying humans and their expectations!
	conf.set('prefix', path.resolve(__dirname, '../..', ''))
    if (conf.get('prefix')) {
      var etc = path.resolve(conf.get('prefix'), 'etc')
      mkdirp(etc, function () {
        defaults.globalconfig = path.resolve(etc, 'npmrc')
        defaults.globalignorefile = path.resolve(etc, 'npmignore')
        afterUserContinuation()
      })
    } else {
      afterUserContinuation()
    }
  }
```

iarna (commented):
This change just overrides whatever prefix was already set. If there is indeed a bug in how prefix is being set, this is just going to paper over it.

Further, looking at your stack overflow issue I _suspect_ this is a configuration issue. I encourage you to contact support@npmjs.com if you'd like help figuring out how to correct it without patching the source.

Thank you for taking the time to put this together. I'm going to close this as we won't be merging it.

PR #13999, opened 2016-09-19
Fix installation error.

### LINKED ISSUE #13997: npm install ends with "Cannot read property '<dependency>' of null"

  - [x] npm is crashing.

#### What's going wrong?

The `npm install` process ends with an error: `Cannot read property '<dependency>' of null` (where `<dependency>` is a dependency from my `package.json`).

Relevant log:

```console
1087 verbose afterAdd /Users/ionicabizau/.npm/cli-size/1.0.0/package/package.json not in flight; writing
1088 verbose correctMkdir /Users/ionicabizau/.npm correctMkdir not in flight; initializing
1089 verbose afterAdd /Users/ionicabizau/.npm/cli-size/1.0.0/package/package.json written
1090 verbose stack TypeError: Cannot read property 'diable' of null
1090 verbose stack     at /Users/ionicabizau/Documents/npm/lib/install/deps.js:604:21
1090 verbose stack     at Array.some (native)
1090 verbose stack     at /Users/ionicabizau/Documents/npm/lib/install/deps.js:603:43
1090 verbose stack     at Array.some (native)
1090 verbose stack     at exports.earliestInstallable (/Users/ionicabizau/Documents/npm/lib/install/deps.js:600:68)
1090 verbose stack     at exports.earliestInstallable (/Users/ionicabizau/Documents/npm/lib/install/deps.js:625:11)
1090 verbose stack     at /Users/ionicabizau/Documents/npm/lib/install/deps.js:500:22
1090 verbose stack     at /Users/ionicabizau/Documents/npm/node_modules/iferr/index.js:13:50
1090 verbose stack     at addBundled (/Users/ionicabizau/Documents/npm/lib/fetch-package-metadata.js:254:39)
1090 verbose stack     at /Users/ionicabizau/Documents/npm/lib/install/deps.js:499:7
1091 verbose cwd /Users/ionicabizau/Documents/3abn
1092 error Darwin 15.6.0
1093 error argv "/usr/local/Cellar/node/6.1.0/bin/node" "/Users/ionicabizau/Documents/npm/bin/npm-cli.js" "install"
1094 error node v6.1.0
1095 error npm  v3.10.8
1096 error Cannot read property 'diable' of null
1097 error If you need help, you may report this error at:
1097 error     <https://github.com/npm/npm/issues>
1098 verbose exit [ 1, true ]
```

Use this [`package.json`](https://gist.github.com/IonicaBizau/57b8aea565449f73d253719abcc127a0) and run `npm install`.

### supporting information:

 - `npm -v` prints: `3.8.6`, still a problem using the latest version (cloned from GitHub)
 - `node -v` prints:`v6.1.0`
 - `npm config get registry` prints: `https://registry.npmjs.org/`
 - Windows, OS X, or Linux?: OSX and Linux. I speculate it's happening on Windows too.
 - Network issues: We have [amazing internet in Romania](http://berniespeedtest.com). 🚀 
   - Geographic location where npm was run: Romania, Europe
 - Container:

IonicaBizau:
Fixes #13997. Check if the `bin` object exists to prevent the errors.

### DIFF
--- a/lib/install/deps.js
+++ b/lib/install/deps.js
@@ -597,13 +597,13 @@ var earliestInstallable = exports.earliestInstallable = function (requiredBy, tr
 
   // If any of the children of this tree have conflicting
   // binaries then we need to decline to install this package here.
-  var binaryMatches = typeof pkg.bin === 'object' && tree.children.some(function (child) {
-    if (child.removed) return false
-    if (typeof child.package.bin !== 'object') return false
+  var binaryMatches = pkg.bin && tree.children.some(function (child) {
+    if (child.removed || !child.package.bin) return false
     return Object.keys(child.package.bin).some(function (bin) {
       return pkg.bin[bin]
     })
   })
+
   if (binaryMatches) return null
 
   // if this tree location requested the same module then we KNOW it
--- a/test/tap/install-bin-null.js
+++ b/test/tap/install-bin-null.js
@@ -1,0 +1,91 @@
+var fs = require('graceful-fs')
+var path = require('path')
+
+var mkdirp = require('mkdirp')
+var osenv = require('osenv')
+var rimraf = require('rimraf')
+var test = require('tap').test
+
+var common = require('../common-tap.js')
+
+var pkg = path.join(__dirname, 'install-bin-null')
+
+var EXEC_OPTS = { cwd: pkg }
+
+var parentPkg = {
+  name: 'parent-package',
+  version: '0.0.0',
+  dependencies: {
+    'child-package-a': 'file:./child-package-a',
+    'child-package-b': 'file:./child-package-b'
+  }
+}
+
+var childPkgA = {
+  name: 'child-package-a',
+  version: '0.0.0',
+  bin: 'index.js'
+}
+
+var childPkgB = {
+  name: 'child-package-b',
+  version: '0.0.0',
+  dependencies: {
+    'grandchild-package': 'file:../grandchild-package'
+  }
+}
+
+var grandchildPkg = {
+  name: 'grandchild-package',
+  version: '0.0.0',
+  bin: null
+}
+
+var pkgs = [childPkgA, childPkgB, grandchildPkg]
+
+test('the grandchild has bin:null', function (t) {
+  setup()
+  common.npm(['install'], EXEC_OPTS, function (err, code, stdout, stderr) {
+    t.ifErr(err, 'npm link finished without error')
+    t.equal(code, 0, 'exited ok')
+    t.ok(stdout, 'output indicating success')
+    t.notOk(stderr, 'no output stderr')
+    t.end()
+  })
+})
+
+test('cleanup', function (t) {
+  cleanup()
+  t.end()
+})
+
+function cleanup () {
+  process.chdir(osenv.tmpdir())
+  rimraf.sync(pkg)
+}
+
+function setup () {
+  cleanup()
+  mkdirp.sync(pkg)
+  fs.writeFileSync(
+    path.join(pkg, 'package.json'),
+    JSON.stringify(parentPkg, null, 2)
+  )
+  pkgs.forEach(function (json) {
+    process.chdir(mkPkg(json))
+  })
+  fs.writeFileSync(
+    path.join(pkg, childPkgA.name, 'index.js'),
+    ''
+  )
+}
+
+function mkPkg (json) {
+  var pkgPath = path.resolve(pkg, json.name)
+  mkdirp.sync(pkgPath)
+  fs.writeFileSync(
+    path.join(pkgPath, 'package.json'),
+    JSON.stringify(json, null, 2)
+  )
+  return pkgPath
+}


IonicaBizau:
Not quite sure why the tests are failing on Windows, but I guess it's not because my change.

lib/install/deps.js
```diff
@@ -600,6 +600,7 @@ var earliestInstallable = exports.earliestInstallable = function (requiredBy, tr
   var binaryMatches = typeof pkg.bin === 'object' && tree.children.some(function (child) {
     if (child.removed) return false
     if (typeof child.package.bin !== 'object') return false
+    if (!pkg.bin || !child.package.bin) return false
```
watilde: Thanks for the good catch/patch! I think the actual point of the bug is the conditional expression `typeof variable === 'object'` and `typeof variable !== 'object'` . It's sometimes annoy since the variable could be `null` at the case and `typeof null` is `object`. So checking `null` would be better imo, like the following:
```diff
-- var binaryMatches = typeof pkg.bin === 'object' && tree.children.some(function (child) {
++ var binaryMatches = pkg.bin !== null && typeof pkg.bin === 'object' && tree.children.some(function (child) {
     if (child.removed) return false
++   if (child.package.bin === null) return false
     if (typeof child.package.bin !== 'object') return false
       return Object.keys(child.package.bin).some(function (bin) {
         return pkg.bin[bin]
     })
   })
```
Cheers!

iarna: The only values that are supposed to come out of `read-package-json`/`normalize-package-json` are an array or undefined/null. (If you had a string in your `package.json` that gets converted to an object before this code can see it.)

Truthyness is sufficient for `pkg.bin` but `child.package.bin` needs truthyness (or not null) checking AND object-ness-checking as it's gonna call `Object.keys`.

My inclination is to have the guard be the same for both fields.

iarna:
Thank you for putting this together. After reviewing the code in `read-package-json` and `normalize-package-json` I can confirm that a guard here is appropriate. The one thing that we'll need before I can land this is a test.

lib/install/deps.js
IonicaBizau: @iarna Makes sense. Will definitely fix that. I was actually thinking the same, but I thought there are other reasons to check against the type. 👍

IonicaBizau:
> The one thing that we'll need before I can land this is a test.

@iarna Can one of you help me writing a test for this? Should we directly call the `findRequirement` method with data that was making the things crash before?

Where would be the appropriate place to put the test in?

Thanks! 🍀

IonicaBizau:
I added a test which is failing without this change and is passing with the patch.

But I feel I didn't do it the right way. Can you check that and help me improve it if it's the case? 😊

I also simplified the fix by simply checking `pkg.bin && ...` since `pkg.bin` can be `null` which is an object. ✨

lib/install/deps.js
IonicaBizau: In my latest commit I changed `typeof pkg.bin === 'object' && ...` into `pkg.bin && ...`. 👍 

@iarna @watilde Can you review it and help me to improve the tests? 💭

watilde: LGTM for the last change! Great :smiley_cat:

In this case, the package that has `bin: null` is [cli-size](https://github.com/IonicaBizau/cli-size/blob/master/package.json#L38) which is used by `cli-update`, so I've written the test case to reproduce the error:
https://gist.github.com/watilde/2e2533299bf2942bd719893062017307

Here is a small example to reproduce the original issue that I do some spelunking:
```json
{
  "name": "test",
  "version": "0.0.0",
  "dependencies": {
    "diable": "*",
    "cli-update": "*",
  }
}
```

Thanks!

watilde:
In this case, the package that has `bin: null` is [cli-size](https://github.com/IonicaBizau/cli-size/blob/master/package.json#L38) which is used by `cli-update`, so I've written the test case to reproduce the error:
https://gist.github.com/watilde/2e2533299bf2942bd719893062017307

Here is a small example `packagejson` to reproduce the original issue that I do some spelunking:
```json
{
  "name": "test",
  "version": "0.0.0",
  "dependencies": {
    "diable": "*",
    "cli-update": "*",
  }
}
```

Cheers!

IonicaBizau:
@watilde Thanks a lot for that! :cake: I just added it. :rocket:

IonicaBizau:
I'm not sure how the process works––I guess this is ready to be merged, or should I do anything else? :thought_balloon:

iarna:
Regarding process: This is the week of our next release, so I'm reviewing outstanding PRs. =)

iarna (approved):
This test seems good to me. You asked for feedback:

These days I typically use the `tacks` module to handle creating and destroying my fixture directories.

And to make using it easier, I further use `scripts/maketest` to generate the scaffolding for my test.

IonicaBizau:
Sounds amazing! ✨

iarna:

Issue #5999, opened 2014-08-20
exitcode != 0 on outdated dependencies (npm outdated)


seriousManual:
this a proposal for a non zero exit value for `npm outdated`.

if dependencies are found that don't match the specification from the package.json, either if they are wrong (red) or outdated (yellow) the command should return a value that is not zero.

this feature could be utilized to invalidate a build if dependencies are not up to date any more.

by default the return value should be zero, via a cli parameter the level should be configurable. e.g. return non zero if there are errors (red) or non zero if there are warnings.

if this proposal finds acceptance i'll be happy to devise a pull request.

othiym23:
This has been moved to the [npm roadmap](https://github.com/npm/npm/wiki/Roadmap), which we're using instead of the confusing `next-*` labels now.

othiym23:
As @isaacs said, this is a breaking change. In addition, this isn’t a change the CLI team is going to have the resources to make itself. As such, I’m going to close this feature request, but the CLI team will be happy to ship a community-contributed patch with this behavior in the next major version of npm after the patch is landed. Thanks for your time, and patience!

watilde:
@othiym23 Can we keep opening this issue like the others breaking-change issues that are added `4.x` milestone?
+ https://github.com/npm/npm/issues?q=is%3Aopen+is%3Aissue+milestone%3A4.x

I'm curious very much about adding this feature into npm officially because I've created [a module](https://github.com/watilde/freshness) to do this before so I'd like to try building this patch if it's ok :)

iancrowther:
if you append ``` || true``` to the command then your outdated wont fail :-)

I just wanted to use this as a status

viskin:
@iancrowther `|| true` will not work on Windows.

iancrowther:
@viskin fair point that.. hmm

I'd be happy to pair up and have a stab at patching this properly.

smyrman:
> if you append || true to the command then your outdated wont fail :-)

That doesn't help to use it as a status as it would then _never_ return non-zero. It _should_ return non-zero when there is something to do...

PR #14013, opened 2016-09-19
outdated: exitcode != 0 on outdated dependencies

watilde:
Fixes: #5999.

This update is a breaking change requested a long time ago. When outdated dependencies exist in a project, the command displays a helpful error message and then finish the action with `exit 1`. To ignore the error and keep the previous behavior, it's possible via passing `--silent` option.

### DIFF
--- a/lib/outdated.js
+++ b/lib/outdated.js
@@ -111,7 +111,12 @@ function outdated (args, silent, cb) {
         }
         output(table(outTable, tableOpts))
       }
-      cb(null, list.map(function (item) { return [item[0].parent.path].concat(item.slice(1, 7)) }))
+      var msg = [
+        'The outdated dependencies were found. Please make sure',
+        'the versions in node_module, or look over package.json',
+        'to use the most suitable semver range.'
+      ].join('\n')
+      cb(msg, list.map(function (item) { return [item[0].parent.path].concat(item.slice(1, 7)) }))
     })
   }))
 }
--- a/test/tap/outdated-color.js
+++ b/test/tap/outdated-color.js
@@ -56,8 +56,8 @@ test('does not use ansi styling', function (t) {
       ],
     EXEC_OPTS,
     function (err, code, stdout) {
-      t.ifError(err)
-      t.notOk(code, 'npm outdated exited with code 0')
+      t.ifError(err, 'npm outdated ran with a error message')
+      t.is(code, 1, 'npm outdated exited with code 1')
       t.ok(stdout, stdout.length)
       t.ok(!hasControlCodes(stdout))
       s.close()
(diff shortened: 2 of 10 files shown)


iarna (approved):
For the record: this all looks good. We'll be including it in npm@4.

lib/outdated.js
```diff
@@ -111,7 +111,12 @@ function outdated (args, silent, cb) {
         }
         output(table(outTable, tableOpts))
       }
-      cb(null, list.map(function (item) { return [item[0].parent.path].concat(item.slice(1, 7)) }))
+      var msg = [
+        'The outdated dependencies were found. Please make sure',
+        'the versions in node_module, or look over package.json',
+        'to use the most suitable semver range.'
+      ].join('\n')
+      cb(msg, list.map(function (item) { return [item[0].parent.path].concat(item.slice(1, 7)) }))
```
mhart: Quick question – should the first arg here be an Error instead of a String? (not sure about npm's style preference – but most Node.js projects typically use Errors so that stack information is retained, among other things)

iarna: This isn't an internal error, this is a callback to provide user visible output.

Errors from commands (like this one) are processed via `lib/utils/error-handler.js`, which, if the error argument is a string, prints it to the user using `log.error` and exits with a non-zero exit code.

By contrast, if we called back with an error object (which we often do), then its output message is determined via `lib/utils/error-message.js` and it's logged along with a bunch of other diagnostic information.

This is actually the only way currently to trigger a non-zero exit code without printing all sorts of diagnostic information on folks screens.

If I were to design this, I'd have an error class to use for this kind of scenario and always pass through objects. But IMO that kind of rewrite (while positive) is out of scope for this PR and so I wouldn't ask someone contributing a change like this one to make that kind of rewrite.

zkat:
This was merged into `release-next`!

We had some discussion at the last minute about the UX for this and I ended up making some quick patches on top of what you did in a separate PR, btw. If we'd had more time pre-npm@4 we would've done it in review though. ^_^;;

This is really great, though, @watilde! Thank you as always for the excellent work <3

fearphage:
FYI, I don't see tests for `--silent` nor can I get it to work as expected.

```
➜  unity git:(starting-point) ✗ node --version
v7.7.1
➜  unity git:(starting-point) ✗ npm --version
4.1.2
➜  unity git:(starting-point) ✗ npm outdated > /dev/null; echo $?
1
➜  unity git:(starting-point) ✗ npm outdated --silent > /dev/null; echo $?
1
➜  unity git:(starting-point) ✗ npm --slient outdated > /dev/null; echo $?       
1
➜  unity git:(starting-point) ✗ npm --slient outdated --silent > /dev/null; echo $?
1
```

Am I missing something?

PR #14032, opened 2016-09-21
Fixes #13973 #14030

### LINKED ISSUE #13973: "npm show <package>@<unpublished-version>" prints "undefined"

  - [ ] npm is crashing.

Sorry if this has already been reported. I spent several minutes trying to find a filed ticket, and couldn't. 

#### What's going wrong?

We automatically publish some internal npm packages via a CI build. That build checks if the package has previously been published via `npm show package@<expected version>`. In npm2, that returned an empty string, in npm3 it returns the string "undefined":

Example:

```
$ npm show react-server@12345
undefined
$ npm show --json react-server@12345
undefined
```

I guess my expectation was that the output would be the same between npm2 and npm3, but it's possible that's just silly of me. :) I also couldn't find documentation of the intended behavior, so I wasn't sure whether it's really a bug.

Example: `npm show react-server@12345`

### supporting information:

 - `npm -v` prints: 3.10.3
 - `node -v` prints: v6.5.0
 - `npm config get registry` prints: has same behavior on both our internal registry and the public one
 - Windows, OS X, or Linux?: Linux, OSX
 - Network issues:
   - Geographic location where npm was run: Oregon
 - Container:

### LINKED ISSUE #14030: npm view --json <package>@<semver range> returns only one result

  - [ ] npm is crashing.

#### What's going wrong?

```
$ npm -v
3.10.3
$ npm view react-server@^0.4.6 version
react-server@0.4.6 '0.4.6'
react-server@0.4.7 '0.4.7'
$ npm view --json react-server@^0.4.6 version
"0.4.6"     # this result seems unexpected
```

I think this is similar to #13973.

npm2 returns the same thing both with and without `--json` (which is also not good):

```
$ npm -v
2.14.12
$ npm view --json react-server@^0.4.6 version
react-server@0.4.6 "0.4.6"
react-server@0.4.7 "0.4.7"
```

I'm a little unsure as to which output is to be expected. Docs for `npm view` seem to indicate you should get more than one row though:

>        If a version range is provided, then data will be printed for every matching version of the package.  This will show which version of jsdom was required by each matching version of yui3:
>
>         npm view yui3@'>0.5.4' dependencies.jsdom

See above.

### supporting information:

 - `npm -v` prints:
 - `node -v` prints:
 - Windows, OS X, or Linux?:
 - Network issues:
 - Container:

roblg:
This makes `npm show` for an unpublished version print empty string, regardless of whether `--json` is passed. As a (intentional) side effect of making it return empty string for `--json`, it fixes #14030.

Empty string for `--json` might be the wrong thing, but it's what npm2 did, and it seemed weird at first blush for `--json` to print `undefined`, since I don't think that's valid JSON. 

Approach: wrap some stuff towards the end of `printData` in a `--json` check, since it doesn't seem like it should apply if we're not printing JSON. The check for `if (!msg)` was what was ultimately causing the `undefined` to be printed, and it seems like we should be going into that branch (line 283 in the patch) if we have no `msgJson` data to begin with.

LMK if there's anything I can clean up here (or if I'm way off base). Thanks!
# 

Commits
13973: npm show w/ undefined version prints 'undefined'
14030: npm show --json w/ semver range prints single result

### DIFF
--- a/lib/view.js
+++ b/lib/view.js
@@ -273,15 +273,17 @@ function printData (data, name, cb) {
     })
   })
 
-  if (msgJson.length && Object.keys(msgJson[0]).length === 1) {
-    var k = Object.keys(msgJson[0])[0]
-    msgJson = msgJson.map(function (m) { return m[k] })
-  }
+  if (npm.config.get('json')) {
+    if (msgJson.length && Object.keys(msgJson[0]).length === 1) {
+      var k = Object.keys(msgJson[0])[0]
+      msgJson = msgJson.map(function (m) { return m[k] })
+    }
 
-  if (!msg) {
-    msg = JSON.stringify(msgJson[0], null, 2) + '\n'
-  } else if (msgJson.length > 1) {
-    msg = JSON.stringify(msgJson, null, 2) + '\n'
+    if (msgJson.length === 1) {
+      msg = JSON.stringify(msgJson[0], null, 2) + '\n'
+    } else if (msgJson.length > 1) {
+      msg = JSON.stringify(msgJson, null, 2) + '\n'
+    }
   }
 
   // preserve output symmetry by adding a whitespace-only line at the end if
--- a/test/tap/view.js
+++ b/test/tap/view.js
@@ -277,6 +277,27 @@ test('npm view <package name>@<semver range> versions', function (t) {
   })
 })
 
+test('npm view <package name>@<semver range> version --json', function (t) {
+  mr({ port: common.port, plugin: plugin }, function (er, s) {
+    common.npm([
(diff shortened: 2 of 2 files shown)


roblg:
Are the appveyor checks expected to pass? It looks like they've been failing a lot (https://ci.appveyor.com/project/npm/npm/history)

iarna:
Our AppVeyor config currently has some bugs around git paths that stop it from completing successfully (even though we can complete testing on Windows locally).

iarna (approved):
This looks great, thank you!

iarna:

PR #14045, opened 2016-09-22
test: handle the order of the cli args for windows

watilde:
Hello!

I've fixed the ci error on Appveyor. Some of the test code are depending on the order of the command line arguments in the git command, and it doesn't work on Windows since prefixes are added in `lib/utils/git.js`, like the following:

```
$ git -c core.longpaths=true clone
```

For the test cases related to bitbucket, github and gitlab, I used the same solution with [github-shortcut-package.js](https://github.com/npm/npm/blob/8fa75cd0313e3cea8459f89b79208461e54b033b/test/tap/github-shortcut-package.js#L34-L49) which doesn't depend on the order of the command line arguments. For the other case, I've added an index number into it: [git-race.js](https://github.com/watilde/npm/blob/e58b4b308f38acbbe3ed1f8e6f73e01a47308888/test/tap/git-races.js#L63-L84).

### DIFF
--- a/test/tap/bitbucket-https-url-with-creds-package.js
+++ b/test/tap/bitbucket-https-url-with-creds-package.js
@@ -34,10 +34,10 @@ test('bitbucket-https-url-with-creds-package', function (t) {
     'child_process': {
       'execFile': function (cmd, args, options, cb) {
         process.nextTick(function () {
-          if (args[0] !== 'clone') return cb(null, '', '')
+          if (args.indexOf('clone') === -1) return cb(null, '', '')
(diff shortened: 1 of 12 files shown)


watilde:
Oh I didn't check on Node.js v0.10.x in my laptop, I will check it again later.

watilde:
Well, the errors on Node.js v0.10/0.12 on Windows seem to be from the tacks mainly. I've created an another PR for it.

iarna (approved):
Looks good to me!

iarna:

PR #14071, opened 2016-09-25
Add .npmignore to always ignored files

fson:
Many packages unintentionally publish `.npmignore` to npm. For example, looking at the dependencies of `npm` package alone, almost half of them include `.npmignore`:

```
npm init -f
npm install npm --save
npm list | wc -l
     267
find node_modules -name ".npmignore" | wc -l
      98
```

However, by definition, `.npmignore` is unnecessary once the package has been published to npm. Removing it from published packages would speed up npm installs everywhere. But asking all package authors to ignore it manually is almost an impossible task.

npm already ignores certain files (e.g. `.npmrc`) by default. Let's add `.npmignore` to this list of ignores, and make lots of future releases smaller.

### DIFF
--- a/doc/files/package.json.md
+++ b/doc/files/package.json.md
@@ -200,6 +200,7 @@ Conversely, some files are always ignored:
 * `.DS_Store`
 * `._*`
 * `npm-debug.log`
+* `.npmignore`
 * `.npmrc`
 * `node_modules`
 * `config.gypi`
--- a/doc/misc/npm-developers.md
+++ b/doc/misc/npm-developers.md
@@ -120,6 +120,7 @@ need to add them to `.npmignore` explicitly:
 * `.DS_Store`
 * `.git`
 * `.hg`
+* `.npmignore`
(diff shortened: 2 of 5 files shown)


iarna:
The CLI team has some hesitation about adding any more files to our already overly complicated forced include/exclude lists. We don't have a consensus on if we should accept this change. (I'm pretty +0 personally.) I'll try to put that together and get back to you.

If we were to take this, this change would also need to be made to [fstream-npm](https://github.com/npm/fstream-npm) (which is almost, but not entirely, the same, for reasons). It would also need to have tests.

othiym23:
After further discussion, this is a very small change that's likely to result in a less than marginal improvement to install performance, at the cost of adding to an already somewhat surprising hardcoded ignore rule system, and there are users of npm who like to work from packages pulled down from the registry. All of these are small on their own, but add up to something that's not worth the effort of landing it. That said, thanks very much for your time and contribution!

Issue #14042, opened 2016-09-22
NPM 3.10.8 tries to install optional dep on unsupported platform when using npm-shrinkwrap.json


paglias:
#### I'm opening this issue because:
- [x] npm is crashing.
- [ ] npm is producing an incorrect install.
- [ ] npm is doing something I don't understand.
- [ ] Other (_see below for feature requests_):
#### What's going wrong?

This problem started appearing with NPM 3.10.8, with 3.10.7 it worked well. With 3.10.8 TravisCI tries to install `fsevents` which is an optional dependency of a dev dependency that only works on OS X / mac OS and so it fails.

```
matteo@macbook~/dev/habitrpg/habitrpg$ npm ls fsevents
habitica@3.42.1 /Users/matteo/dev/habitrpg/habitrpg
└─┬ karma@1.3.0
  └─┬ chokidar@1.6.0
    └── fsevents@1.0.14
```

If it can help we're using npm shrinkwrap where the package is listed as `optional: true`

```
"fsevents": {
  "version": "1.0.14",
  "from": "fsevents@>=1.0.0 <2.0.0",
  "resolved": "https://registry.npmjs.org/fsevents/-/fsevents-1.0.14.tgz",
  "optional": true,
  "dependencies": {
    ...
  }
},
```

An example of a failing build is https://travis-ci.org/HabitRPG/habitrpg/builds/161951399, I don't have access to npm-debug.log

```
npm ERR! Linux 3.13.0-92-generic
npm ERR! argv "/home/travis/.nvm/versions/node/v4.3.1/bin/node" "/home/travis/.nvm/versions/node/v4.3.1/bin/npm" "install"
npm ERR! node v4.3.1
npm ERR! npm  v3.10.8
npm ERR! code EBADPLATFORM
npm ERR! notsup Unsupported platform for fsevents@1.0.14: wanted {"os":"darwin","arch":"any"} (current: {"os":"linux","arch":"x64"})
npm ERR! notsup Valid OS:    darwin
npm ERR! notsup Valid Arch:  any
npm ERR! notsup Actual OS:   linux
npm ERR! notsup Actual Arch: x64
npm ERR! Please include the following file with any support request:
npm ERR!     /home/travis/build/HabitRPG/habitrpg/npm-debug.log

```

paglias:
Just verified that this is happening on our production server that runs Linux.

paglias:
[...]

PR #14117, opened 2016-09-29
shrinkwrap: fix optional deps check

watilde:
Fixes https://github.com/npm/npm/issues/14042. From the last release, `npm-shrinkwrap.json` has been updated to tell us the deps is optional or not, but instal-check in `install/action.js` still has been checking the deps could be installable or not even the deps has `optional: true`.

I've updated `inflate-shrinkwrap.js` to pass the property of the optional to `action.js` and  `isInstallable` should be skipped if the deps is optional. I will add a test and write more detail into the commit message later. Thanks!

### DIFF
--- a/lib/install/deps.js
+++ b/lib/install/deps.js
@@ -311,14 +311,16 @@ function andForEachChild (load, next) {
   }
 }
 
-function isDepOptional (tree, name) {
+function isDepOptional (tree, name, pkg) {
+  if (pkg.package && pkg.package._optional) return true
   if (!tree.package.optionalDependencies) return false
   if (tree.package.optionalDependencies[name] != null) return true
   return false
 }
 
 var failedDependency = exports.failedDependency = function (tree, name_pkg) {
-  var name, pkg
+  var name
+  var pkg = {}
   if (typeof name_pkg === 'string') {
     name = name_pkg
   } else {
@@ -327,7 +329,7 @@ var failedDependency = exports.failedDependency = function (tree, name_pkg) {
   }
   tree.children = tree.children.filter(noModuleNameMatches(name))
 
-  if (isDepOptional(tree, name)) {
+  if (isDepOptional(tree, name, pkg)) {
     return false
   }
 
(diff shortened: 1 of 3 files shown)


lib/install/actions.js
```diff
@@ -46,7 +46,8 @@ Object.keys(actions).forEach(function (actionName) {
       if (!pkg.commit) pkg.commit = []
       pkg.commit.push(action.commit)
     }
-    if (pkg.knownInstallable) {
+
+    if (pkg.knownInstallable || pkg.package._optional) {
```
iarna: My reading of this code:

If the shrinkwrap flagged this package as optional then don't bother checking to see if it meets archictecture or OS requirements?

Isn't that the opposite of what we want?

iarna (changes requested):
This is going to need a test, because my reading of the code is that this makes the scenario outlined in the issue worse, not better.

I _believe_ the scenario in the issue is related to rollbacks of transitive deps of optional deps causing an "exit with error" condition when they shouldn't.

lib/install/actions.js
watilde: True. It should be handled through the errorMessage like the general `optionalDependencies` way of `install.js`. I will update them again!

watilde: Ok! I've updated the code and added a test for the case.  Here is the minimum case to reproduce the original issue: https://github.com/watilde/npm-issue-14042

iarna:
This looks much better! I'm still concerned that apparently our optionalDependency checker is failing under some circumstances (that you've isolated). I'd _prefer_ not to have to fall back on assuming the shrinkwrap is correct regarding the optional status of packages.

So this is my plan:

1. I plan on taking this patch, because as long as we have optional in the shrinkwrap I think we should believe it.
2. I'm going to pull out your reproducer and see if I can't figure out what the root cause of the failure is.

I'd like to get #2 done before I land #1 as having #1 in place will make it harder to reproduce.

watilde:
Your plan looks good to me! Let me know if you need more help :)
I guess the optional status of the package should be transferred to our optionalDependency checker as just one optional dependency instead of the optional property eventually.

iarna:
This has been merged into `release-next`

Issue #14009, opened 2016-09-19
npm version fails if using shrinkwrap and command isn't executed at same level


AlexChesters:
#### I'm opening this issue because:

  - [ ] npm is crashing.
  - [ ] npm is producing an incorrect install.
  - [X] npm is doing something I don't understand.
  - [ ] Other (_see below for feature requests_):

#### What's going wrong?
`npm version` fails if you are:
1. Using an `npm-shrinkwrap.json`
2. Running `npm version` from somewhere other than the same level as `npm-shrinkwrap.json`

#### How can the CLI team reproduce the problem?
1. `npm init -y`
2. `npm install express --save` (or any node module)
3. `npm shrinkwrap`
4. `mkdir src`
5. `cd src`
6. `npm version minor`

Gist of `npm-debug.log` file can be found [here](https://gist.github.com/AlexChesters/30de975e94950d33125b53cd92051df1)

### supporting information:

 - `npm -v` prints: `3.10.3`
 - `node -v` prints: `v6.5.0`
 - `npm config get registry` prints: `https://registry.npmjs.org/`
 - Windows, OS X, or Linux?: OS X
 - Network issues:
   - Geographic location where npm was run: UK
   - [ ] I use a proxy to connect to the npm registry.
   - [ ] I use a proxy to connect to the web.
   - [ ] I use a proxy when downloading Git repos.
   - [ ] I access the npm registry via a VPN
   - [ ] I don't use a proxy, but have limited or unreliable internet access.
 - Container:
   - [ ] I develop using Vagrant on Windows.
   - [ ] I develop using Vagrant on OS X or Linux.
   - [ ] I develop / deploy using Docker.
   - [ ] I deploy to a PaaS (Triton, Heroku).

lholmquist:
@AlexChesters i've tried to reproduce your issue with the steps provided, but everything works fine.

npm version : 3.10.8
node version: 6.5.0
macOS 10.12

lholmquist:
@AlexChesters did you already run `git init` or it is under gits control before you did those steps?

when i walked through the steps above, my directory was not under git control

lholmquist:
@AlexChesters @othiym23 ok,  i was able to reproduce,  it fails when the project is also a git project.  There error is trying to do a `git add` on the shirnkwrap.json file, but it doesn't see it since it is in another directory.  

I can probably send a PR for this

lholmquist:
[...]

PR #14143, opened 2016-09-30
The npm shrinkwrap file needed to have is full path specified.

lholmquist:
fixes issue #14009.  

The problem was that the `git add` of the npm-shrinkwrap.json was failing, since it was looking for it in the wrong directory.  Adding in the actual path to the shrinkwrap file similar to https://github.com/npm/npm/blob/master/lib/version.js#L145 fixes the issue

### DIFF
--- a/lib/version.js
+++ b/lib/version.js
@@ -263,7 +263,7 @@ function _commit (version, localData, cb) {
   chain(
     [
       git.chainableExec([ 'add', packagePath ], options),
-      localData.hasShrinkwrap && git.chainableExec([ 'add', 'npm-shrinkwrap.json' ], options),
+      localData.hasShrinkwrap && git.chainableExec([ 'add', path.join(npm.localPrefix, 'npm-shrinkwrap.json') ], options),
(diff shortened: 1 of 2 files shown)


iarna (changes requested):
Thank you! This is very similar to the bug we fixed in the version lifecycle a little bit ago. Unfortunately it seems our test suite doesn't often test running lifecycles outside the project root.

While getting more general tests across our various lifecycles is outside of scope, I'd like to see a test for this particular bug as part of this PR.

lholmquist:
@iarna sure thing

lholmquist:
@iarna added a test for this.  it was basically just a c/p of another test.  :)  I also squashed the commits into 1, if they need to be 2, let me know

lholmquist:
@iarna not really sure why that appveyor check is failing though

iarna:
Yeah, the AppVeyor definitely isn't your fault =) 

And 1 commit is perfect, thanks!

iarna:

Issue #14166, opened 2016-10-03
NPM install throws exception in some symlink scenarios


dzearing:
#### I'm opening this issue because:
- [X] npm is crashing.
- [ ] npm is producing an incorrect install.
- [ ] npm is doing something I don't understand.
- [ ] Other (_see below for feature requests_):
#### What's going wrong?

Running the normal `npm install --save <module>` in a folder where modules are symlinked (by an internal tool) causes install to throw unhandled exceptions in 2 places, therefore preventing installs until the node_modules folder is removed.

The 2 places in the code are:

lib/install.js, where `child.requiredBy` is null, you can see below that .length reference would nullref:

``` js
Installer.prototype.normalizeTree = function (log, cb) {
  validate('OF', arguments)
  log.silly('install', 'normalizeTree')
  recalculateMetadata(this.currentTree, log, iferr(cb, function (tree) {
    tree.children.forEach(function (child) {
      if (child.requiredBy.length === 0) {
        child.existing = true
      }
    })
    cb(null, tree)
  }))
}
```

lib/install/copy-tree.js where `tree[key]` is null:

``` js
function copyModuleList (tree, key, cache) {
  var newList = []
  tree[key].forEach(function (child) {
    newList.push(copyTree(child, cache))
  })
  tree[key] = newList
}
```

If the conditions are adjusted to check for null first before proceding, npm install works just fine. Recommend that null checks are added, but ideally it would also be nice to know why these scenarios would be hit in the first place.

Pull request here: https://github.com/npm/npm/pull/14167
#### How can the CLI team reproduce the problem?

Create a condition where requiredBy for an installed child package is not populated with an array. I couldn't find in the code where this is possible and don't quite understand what conditions could cause this. Perhaps someone more versed in how this gets populated might be able to deduce the conditions where requiredBy is not populated. I can offer a PR to treat the undefined array as an empty array rather than nullrefing and can unblock our scenario.
### supporting information:
- `npm -v` prints:
  v3.10.8
- `node -v` prints:
  v6.6.0
- `npm config get registry` prints:
  https://registry.npmjs.org/
- Windows, OS X, or Linux?:
  OS X
- Network issues:
  - Geographic location where npm was run:
  - [ ] I use a proxy to connect to the npm registry.
  - [ ] I use a proxy to connect to the web.
  - [ ] I use a proxy when downloading Git repos.
  - [ ] I access the npm registry via a VPN
  - [ ] I don't use a proxy, but have limited or unreliable internet access.
[...]

PR #14167, opened 2016-10-03
Null check to prevent npm install from throwing in some cases.

dzearing:
This adds a simple null check to cases that are hit by `npm install` where a child package doesn't have requiredBy array populated.

Addresses issue: https://github.com/npm/npm/issues/14166


iarna (changes requested):
I'd like to see a test for this, particularly because `child.requiredBy` _should_ always be set and a test would allow us to track down the root cause and not just the proximate cause.

dzearing:
@iarna Sure, but i could definitely use help here. Do you have any guesses about what code may be short circuited here? I can debug in but just need some pointers on where to look.

Thanks!

dzearing:
I'm closing because I don't want to leave this open while I am unsure of the root cause causing the null value. I will dig in when i can.

PR #14173, opened 2016-10-04
Add shasum to shrinkwrap for use during install

misterbyrne:
Extend the cache add command to optionally accept a SHA to validate
against when adding a package by URL.

NB: This PR does not implement any code which attempts to perform any cache lookups using the package's shasum. The only changes is to shrinkwrap generation and inflation.

This means that an error will occur when running an npm install if the
downloaded package's checksum is different to that in the shrinkwrap

e.g.

image

TODO:
- [ ] tests

### DIFF
--- a/lib/cache.js
+++ b/lib/cache.js
@@ -84,9 +84,9 @@ var cachedPackageRoot = require('./cache/cached-package-root.js')
 var mapToRegistry = require('./utils/map-to-registry.js')
 var output = require('./utils/output.js')
 
-cache.usage = 'npm cache add <tarball file>' +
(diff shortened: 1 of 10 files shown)


othiym23:
It turns out that adding shasums to `npm-shrinkwrap.json` is one of the most significant changes the CLI team intends to support for `npm@5`, as part of our initiative to improve shrinkwrap support. The team may or may not pick up this PR as a base. This is very much top of mind for the team, as we specify what goes in `npm-shrinkwrap.json` and how we want everything to _actually work_, so we're not ignoring this patch, we just need to come to some decisions about what we want to happen (and also, this may end up turning into a breaking change).

PR #14230, opened 2016-10-10
npm: Detect unsupported Node.js versions and warn

iarna:
Also error on really old versions where we know we can't work.


othiym23:
With a test to verify that this doesn't cause false positives, this is good to go. Thanks for putting this together!

iarna:
This has been merged into `release-next`

Issue #7168, opened 2015-01-19
Support for no-proxy in .npmrc file


jfbibeau:
It looks like NPM will correctly honor the ENV variables for http_proxy, https_proxy and recently also for no_proxy.

However .npmrc only seems to support proxy and https-proxy. This leads to some funny behavior...

I am trying to configure a no_proxy and http_proxy setting for NPM. So I thought I would configure `proxy` in .npmrc and `no_proxy` in my ENV variable (since npmrc doesn't support it). But .npmrc's `proxy` overrides the environment variable's `no_proxy` :(

So the only way for me to get no_proxy to work with NPM is to set it as an environment variable, which will enable it system wide (and I'm trying to avoid that).

Thanks!

othiym23:
The current proxy settings are a pretty delicate balance, because there are many conflicting use cases around them (in particular, figuring out the precedence of the configuration file vs environment variables is pretty much a nightmare). The basic rule, for now, is that if you want to use `no_proxy`, you should configure both `http_proxy` and `no_proxy` via environment variables. The reason for this is an implementation detail (the proxy handling is actually done by `request`). A patch that adds support for `no-proxy` to `.npmrc` would be welcome, as would be a patch that tweaks the documentation to make the current setup easier to use.

ehsalazar:
We haven’t heard back and we’re trying to clean up some old issues.  If this is still a problem, can you please reply and let us know?  We’ll be happy to reopen if necessary.

brunowego:
+1

wildpixeldesign:
+1

goodloot:
+1, please, reopen.

othiym23:
Reopening, as this is a feature request. It would be nice to get this as a patch, because this is such a low priority for the CLI team, it's unlikely the team will get to it in the next year.

marcellodesales:
Would love this one as well...

karthikeayan:
+1

rogerwelin:
+1

Mischi:
:+1:

Mischi:
:+1:

BrianLenzo:
+1 
Setting proxy in env vars messes other applications up. Having it all in .npmrc would help my company use NPM behind a very secure proxy.

marcellodesales:
@BrianLenzo Just stepped on this again today :( Please guys, add this support... We have NPMO installation internally and we are building applications behind the HTTP_PROXY as well...

dg1sek:
+1

[...]

PR #14248, opened 2016-10-11
config: Set process.env.no_proxy in adaptConfig

legodude17:
Solution for #7168.


othiym23:
Here's what needs to be added before this will be ready to merge:

- [ ] The configuration variable name should follow the convention of the CLI's other configuration, and use a hyphen instead of an underscore (`no-proxy`).
- [ ] It also will need to be added to `lib/config/defaults.js` (the tests won't pass otherwise – this is part of how the project ensures that all of the CLI's configuration is documented), and have a sensible type and default set. See the rest of that file to get the gist of how that works. It's a little tricky, because `nopt` will typically try to convert a `no-<name>` config setting into a boolean set to a falsy value, so this also probably needs 
    - [ ] a few additional test cases in the current config tests to ensure that nothing untoward happens as a result of this change.
- [ ] A few general tests that the behavior makes it to the child environment of run-scripts as `no_proxy` is probably a good enough test, because the CLI doesn't really have control over `request`.
- [ ] Also, `doc/misc/npm-config.md` will need an update to include information about the new configuration, and the existing documentation doesn't mention `no_proxy` / `NO_PROXY`'s functionality, even as it currently stands, so that could probably use some cleanup overall, and cross referencing between the sections for `proxy`, `https-proxy`, and `no-proxy`.

Let us know if you have any questions about any of the above, and thanks for the contribution!

legodude17:
Thanks for the quick feedback. I was wondering what the default should be in `lib/config/defaults.js`. I think it should be `null`, but I was wondering what you thought.

legodude17:
Where should the tests go in the folder structure? I am still a little unclear about where the tests go and what organization system they have.

legodude17:
Any updates? I just need to find a bit of time to write some tests. The docs are done, can I get feedback on them?

iarna:
All tests go in `test/tap/`

legodude17:
Should I make some new files or piggyback on some existing ones?

jd-carroll:
What's the status of getting this merged?  Just setup an internal registry and this is suddenly going to become a problem for 200+ developers internally.

legodude17:
@jd-carroll I need to find some time to write tests, and then someone from `npm` needs to resolve all of the merge errors.

jd-carroll:
Would it make sense to close this PR and open a new one just to avoid all of the merges?

legodude17:
That was an idea I had. Is there anyone who wants to continue this?

iarna:
As suggested I'm going to close this but if it's something you're still interested please feel free to open a new PR based on `latest`.

PR #14251, opened 2016-10-12
Fix node download links

ArtskydJ:
The old link (http://nodejs.org/download/) just doesn't work anymore.

Changed the links to point to https://nodejs.org/en/download/

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,15 +27,15 @@ terms of use for the default public registry are available at
 
 ## Super Easy Install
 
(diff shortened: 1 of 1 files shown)


ArtskydJ:
@Blaccexican I don't understand what you're saying. I didn't change any of the header levels. Also, the node.js site now redirects to https if you go to a http URL, so it's better to just go straight to the https version.

watilde:
Good catch! LGTM.

PR #14327, opened 2016-10-19
No Partial Shrinkwraps

iarna:


zkat:
This looks great!

I had a thought about the way we're dealing with devDeps -- should we try and make the devDep shrinkwrapping more explicit rather than assume it based on devDeps in the shrinkwrap?

For example, what happens if I do something like this?

```
npm i -D eggplant
npm shrinkwrap --dev
npm rm -D eggplant
npm i -D eggplant
```

As a user, I would expect the above to add `eggplant` back into the shrinkwrap file.

Conceptually, I believe this patch will make it so the new `eggplant` will not be added to the shrinkwrap, because the shrinkwrap was no longer in a "look for devDeps" mode.

But when I was testing it, I found what I believe is a different bug: After `npm rm -D eggplant` above, `eggplant` won't be removed from the shrinkwrap -- the `rm` basically becomes a noop. This seems to happen only when it's the last package left.

iarna:
A is confusing and B is definitely a bug, but also everything you described there is already existing behavior so I'm calling them out-of-scope for this PR.

zkat:
Fair enough. I'm alright calling those things out of scope. We'll have a good enough chance to overhaul shrinkwrap with `@5` so idk if I care to nitpick about weirdness like this for the sake of `@4`.

Additionally, this seems to have a test failure with one of the shrinkwrap tests that should probs be addressed.

'sides that, LGTM <3

iarna:
It passes for me 😭

iarna:
def need to dig into it

iarna:
oh, no I was running from the wrong folder. Oh... haha, that's the test for the thing we're breaking. I'M GLAD IT BROKE.

iarna:
Welp, that test (with a small tweak) will continue to be relevant until npm@5, at which point it will need to just be removed.

iarna:
merged into `release-next`

Issue #6274, opened 2014-09-22
`npm tag` → `npm dist-tag`


othiym23:
- [x] add `npm dist-tag` as an alias for `npm tag` (`npm@2.x`)
- [x] deprecate `npm tag`, make it an alias for `npm dist-tag` (`npm@3.x`)
- [x] remove `npm tag` as an alias for `npm dist-tag` (`npm@4.x`)
- [ ] make `npm tag` an alias for `npm ?keyword? ?add?`‽

See also #6243, which will have the correct name from the start.

othiym23:
This indeed seems like a good thing to include as a (small) breaking change in `npm@4`! So labeled and milestoned!

iarna:
Taking this off the v4 milestone as the work for v4 for it has been merged.

PR #14328, opened 2016-10-19
Remove deprecated tag command

iarna:
Folks should be using the `dist-tag` command now.

Fixes: #6274


zkat:
This has a test failure with semver tests. Otherwise LGTM once that's fixed.

iarna:
Ah, that test needed to away (all it did was run the obsolete command).

PR #14334, opened 2016-10-19
lifecycle: tweak a few things with new warning

zkat:
This is a relatively minor nudge to @addaleax's #13409 PR. It slightly rewords the error messages for the warning, and also adds an additional test that checks that both types of warnings are properly spit out.


iarna (approved):
👍

addaleax (approved):
fwiw, this lgtm :)

PR #14356, opened 2016-10-20
error-handler: stop logging the error code

zkat:


lib/outdated.js
```diff
@@ -111,12 +111,8 @@ function outdated (args, silent, cb) {
         }
         output(table(outTable, tableOpts))
       }
-      var msg = [
-        'The outdated dependencies were found. Please make sure',
-        'the versions in node_module, or look over package.json',
-        'to use the most suitable semver range.'
-      ].join('\n')
-      cb(msg, list.map(function (item) { return [item[0].parent.path].concat(item.slice(1, 7)) }))
+      process.exitCode = list.length
```
iarna: I don't this should vary based on how many deps are outdated.

iarna (commented):
one nit

iarna (approved):
Looks good 🐑 🚀 🎆

PR #14365, opened 2016-10-21
Fixed crash during 'npm i' or 'npm u' on win32 when repo URL is too l…

ockhalaw:
Fixed crash during 'npm i' or 'npm u' on win32 when repo URL is too long.

### DIFF
--- a/lib/cache/add-remote-git.js
+++ b/lib/cache/add-remote-git.js
@@ -117,7 +117,16 @@ function tryClone (from, combinedURL, silent, cb) {
   var treeish = normalized.branch
 
   // ensure that similarly-named remotes don't collide
-  var cachedRemote = uniqueFilename(remotes, combinedURL.replace(/[^a-zA-Z0-9]+/g, '-'), cloneURL)
+  //
+  // on win32 we should limit dirname due to 260-character path length
+  var dirPrefix = combinedURL.replace(/[^a-zA-Z0-9]+/g, '-')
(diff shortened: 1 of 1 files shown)


legodude17:
I don't use windows, but I think that this will help a lot. Can you review this @zkat?

othiym23:
It's not clear, from the change, exactly what the problem is – the repo name can be long, but except in very unusual cases, it shouldn't exceed `MAXLEN` for Windows paths when using the default cache path. In other words, this PR would be a lot easier to evaluate if it was associated with a reproducible bug report.

In addition, a platform-specific truncation like the one used in the code seems, at best, ungainly – however this gets solved, it should be done in a way that's applicable to both Windows and the other platforms supported by the CLI. I would be fine, if this is indeed required, with just using a truncated version of the hash everywhere, except for the fact that it means that it becomes impossible for debuggers to know what repo is in which folder without some fairly nitpicky examination using the Git CLI.

In any case, this is going to need tests to ensure that it doesn't cause regressions elsewhere in the caching and installation paths, and may or may not end up being a breaking change. Thanks for your time, and thanks for your patience in dealing with these additional concerns!

iarna:
As this PR hasn't seen activity in a year and is based on code that doesn't exist any more, I'm going to close it. If this is still an issue, we would welcome a similar patch with accompanying tests to `pacote` or `cacache` (the two modules responsible for this sort of work now).

Issue #14372, opened 2016-10-21
npm@4 warning a lot about node binary path


dpogue:
#### I'm opening this issue because:
* npm is doing something I don't understand.

#### What's going wrong?

After updating to npm@4, running any script from package.json is printing a warning:

```
npm WARN lifecycle The node binary used for scripts is /usr/local/bin/node but npm is using /usr/local/Cellar/node/6.9.1/bin/node itself. Use the `--scripts-prepend-node-path` option to include the path for the node binary npm was executed with.
```

I understand this is related to a change in npm@4 to warn if the node binary used by npm and the scripts are different, but it probably shouldn't warn in the case of a symlink.

#### How can the CLI team reproduce the problem?

This seems to happen consistently every time a script is run, when node is installed via homebrew.

In my case, `/usr/local/bin/node` is a symlink to `/usr/local/Cellar/node/6.9.1/bin/node`

### supporting information:

 - `npm -v` prints: 4.0.0
 - `node -v` prints: v6.9.1 (installed via homebrew)
 - Windows, OS X/macOS, or Linux?: OSX 10.11.6

zkat:
Fixed by #14374.

diegodesouza:
I'm still experiencing this issue, what should I do in order to fix it? 
`npm -v => 4.1.1` 
`node -v => v7.4.0`
Mac OS `10.12.2`

addaleax:
@diegodesouza Hey, what do `which npm` and `which node` print for you when run from the command line? Also, what is the exact warning message that `npm` prints?

diegodesouza:
```
npm WARN lifecycle The node binary used for scripts is /usr/local/opt/node but npm is using /usr/local/Cellar/node/7.4.0/bin/node itself. Use the `--scripts-prepend-node-path` option to include the path for the node binary npm was executed with.
```

```
➜  ~ which npm
/usr/local/bin/npm
➜  ~ which node
/usr/local/bin/node
```

Here's the output @addaleax.

addaleax:
Oh, yea, maybe I should have been asking for `realpath $(which node)` and `realpath $(which npm)`.

In any case, the situation is about this: You have (at least) two node binaries installed on your machine, and running `npm` from the command line doesn’t seem to use the one that would come first in the `PATH` environment variable (that you can view using `echo $PATH`).

[...]

PR #14374, opened 2016-10-21
lifecycle: don’t warn about PATH for symlinked node

addaleax:
Apply a `fs.realpath()` check to the found `node` executable to
mirror the fact that `process.execPath` is always a real path, too.

Fixes: https://github.com/npm/npm/issues/14372

/cc @zkat

### DIFF
--- a/test/tap/lifecycle-path.js
+++ b/test/tap/lifecycle-path.js
@@ -80,9 +80,12 @@ test('make sure the path is correct, with directory of current node and warn-onl
   }, t)
 })
 
-test('make sure there is no warning with a symlinked node and warn-only detection', function (t) {
+test('make sure there is no warning with a symlinked node and warn-only detection', {
+  skip: isWindows && 'symlinks are weird on windows'
+}, function (t) {
   checkPath({
-    withDirOfCurrentNode: 'extra-node',
+    withDirOfCurrentNode: false,
+    extraNode: true,
     prependNodePathSetting: 'warn-only',
     symlinkNodeInsteadOfCopying: true
   }, t)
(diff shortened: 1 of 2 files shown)


addaleax:
Updated the PR a fixed patch, this should do better in CI

test/tap/lifecycle-path.js
```diff
@@ -32,46 +32,88 @@ test('setup', function (t) {
...
+    withDirOfCurrentNode: true,
+    prependNodePathSetting: 'warn-only'
+  }, t)
+})
+
+test('make sure there is no warning with a symlinked node and warn-only detection', function (t) {
+  checkPath({
+    withDirOfCurrentNode: 'extra-node',
+    prependNodePathSetting: 'warn-only',
+    symlinkNodeInsteadOfCopying: true
```
zkat: so it turns out this isn't going to work consistently across Windows platforms, depending on system configurations and stuff. Could we skip this entirely on Windows? (we have an `isWindows` utility for this)

test/tap/lifecycle-path.js
```diff
@@ -32,46 +32,88 @@ test('setup', function (t) {
...
-  checkPath(true, 'warn-only', t)
+  checkPath({
+    withDirOfCurrentNode: true,
+    prependNodePathSetting: 'warn-only'
+  }, t)
+})
+
+test('make sure there is no warning with a symlinked node and warn-only detection', function (t) {
+  checkPath({
+    withDirOfCurrentNode: 'extra-node',
```
zkat: I originally made this a `boolean+` for the sake of not adding more arguments, but since you're switching to a config object, do you mind making `withDirOfCurrentNode` a plain bool, and adding an extra config for the `extra-node` bit?

test/tap/lifecycle-path.js
zkat: (the skip is probably better off inside `checkPath` itself)

test/tap/lifecycle-path.js
addaleax: @zkat done!

zkat (approved):
LGTM! 🐑 🎉

PR #14375, opened 2016-10-21
shrinkwrap: Make including dev deps the default

iarna:
This is important because our plan for v5 is to NOT install dev deps if you have a shrinkwrap that doesn't contain them. So we ought to start getting folks making those shrinkwraps.

### DIFF
--- a/test/tap/shrinkwrap-default-dev.js
+++ b/test/tap/shrinkwrap-default-dev.js
@@ -1,0 +1,126 @@
+'use strict'
+var path = require('path')
+var test = require('tap').test
+var fs = require('fs')
+var Tacks = require('tacks')
+var File = Tacks.File
+var Dir = Tacks.Dir
+var extend = Object.assign || require('util')._extend
+var common = require('../common-tap.js')
+
+var basedir = path.join(__dirname, path.basename(__filename, '.js'))
+var testdir = path.join(basedir, 'testdir')
+var cachedir = path.join(basedir, 'cache')
+var globaldir = path.join(basedir, 'global')
+var tmpdir = path.join(basedir, 'tmp')
+
+var conf = {
+  cwd: testdir,
+  env: extend(extend({}, process.env), {
+    npm_config_cache: cachedir,
+    npm_config_tmp: tmpdir,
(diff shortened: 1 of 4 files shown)


test/tap/shrinkwrap-default-dev.js
```diff
@@ -0,0 +1,128 @@
...
+
+test('shrinkwrap-only-prod', function (t) {
+  fs.unlinkSync(shrinkwrapPath)
+  common.npm(['shrinkwrap', '--only=prod'], conf, function (err, code, stdout, stderr) {
+    if (err) throw err
+    t.is(code, 0, 'command ran ok')
+    t.comment(stdout.trim())
+    t.comment(stderr.trim())
+    var swrap = JSON.parse(fs.readFileSync(shrinkwrapPath))
+    t.isDeeply(swrap, shrinkwrapWithoutDev, 'Shrinkwrap included dev deps by default')
```
zkat: Can we update the message here?

test/tap/shrinkwrap-default-dev.js
```diff
@@ -0,0 +1,128 @@
...
+
+test('shrinkwrap-production', function (t) {
+  fs.unlinkSync(shrinkwrapPath)
+  common.npm(['shrinkwrap', '--production'], conf, function (err, code, stdout, stderr) {
+    if (err) throw err
+    t.is(code, 0, 'command ran ok')
+    t.comment(stdout.trim())
+    t.comment(stderr.trim())
+    var swrap = JSON.parse(fs.readFileSync(shrinkwrapPath))
+    t.isDeeply(swrap, shrinkwrapWithoutDev, 'Shrinkwrap included dev deps by default')
```
zkat: also here

test/tap/shrinkwrap-default-dev.js
iarna: oh oops

zkat (approved):
LGTM!

jasonrhodes:
Quick question: `npm install --production` still ignores devDependencies, even if they are in the shrinkwrap ... yes?

PR #14324, opened 2016-10-18
Fixed installing bundled dependencies of scoped packages

NatalieWolfe:
Added extra steps to resolve staging directory which are required for handling the bundled dependencies of scoped packages. I have confirmed this resolves the issue reported in #14311 and I believe it would fix #13528 as well.

Fixed #14311 
Fixed #13528

Issue #14311, opened 2016-10-18
npm@3.5.0 to current fails to install scoped package with bundled dependencies


NatalieWolfe:
#### I'm opening this issue because:
- [x] npm is crashing.
#### What's going wrong?

When attempting to install a scoped package with a bundled dependency npm errors. See below for the output on the command line from attempting to install such a package, and here is the debug log: [download](https://github.com/npm/npm/files/534949/npm-debug.log.txt) or [gist](https://gist.github.com/NatalieWolfe/450bef9603cbfb88511d7f1477227bb7).

I have traced the issue down to commit aed71fb9. In particular I believe it is [this line of the diff](https://github.com/npm/npm/commit/aed71fb#diff-6effc31b3479dd85e58db2f9c50dfe2fR58) that is causing the issue. It assumes that the staging directory is in the parent of the build path, however in the case of a scoped module the staging directory should be in the grandparent of the build path. The issue could also stem from [this line instead](https://github.com/npm/npm/commit/aed71fb#diff-6effc31b3479dd85e58db2f9c50dfe2fR30).
#### How can the CLI team reproduce the problem?

Using `npm` version `>=3.5.0` try `npm install @n.wolfe/scoped-test@1.1.0`.

Alternately:
1. Create a scoped package (e.g. `@foo/bar`).
2. Add a dependency to its `package.json` (e.g. `npm install --save express`).
3. Make that dependency bundled.
4. Publish your package (`npm publish --access public`)
5. Attempt to install your package using `npm >=3.5.0` (`npm install @foo/bar`).
### supporting information:

``` sh
$ npm --version
3.10.8
$ node --version
v6.8.1
$ npm config get registry
https://registry.npmjs.org/
$ ls
$ npm install @n.wolfe/scoped-test@1.0.0 # <-- Dependency is not bundled.
/Users/nwolfe/work/me/test
└─┬ @n.wolfe/scoped-test@1.0.0
  └─┬ concat-stream@1.5.2
    ├── inherits@2.0.3
    ├─┬ readable-stream@2.0.6
    │ ├── core-util-is@1.0.2
    │ ├── isarray@1.0.0
    │ ├── process-nextick-args@1.0.7
    │ ├── string_decoder@0.10.31
    │ └── util-deprecate@1.0.2
    └── typedarray@0.0.6

npm WARN enoent ENOENT: no such file or directory, open '/Users/nwolfe/work/me/test/package.json'
npm WARN test No description
npm WARN test No repository field.
[...]

Issue #13528, opened 2016-07-30
ENOENT errors, `.staging`?


tejasmanohar:
```
npm WARN enoent ENOENT: no such file or directory, open '/root/dev/src/github.com/segment-sources/zendesk/node_modules/esrecurse/package.json'
npm WARN enoent ENOENT: no such file or directory, open '/root/dev/src/github.com/segment-sources/zendesk/node_modules/fs-plus/package.json'
npm WARN enoent ENOENT: no such file or directory, open '/root/dev/src/github.com/segment-sources/zendesk/node_modules/grpc/package.json'
npm WARN enoent ENOENT: no such file or directory, open '/root/dev/src/github.com/segment-sources/zendesk/node_modules/doctrine/package.json'
npm WARN @segment/eslint-config@3.2.1 requires a peer of eslint-plugin-react@^4.0.0 but none was installed.
npm ERR! Linux 4.4.15-moby
npm ERR! argv "/usr/local/bin/node" "/usr/local/bin/npm" "i"
npm ERR! node v6.3.1
npm ERR! npm  v3.10.3
npm ERR! path /root/dev/src/github.com/segment-sources/zendesk/node_modules/.staging/mkdirp-b0f0282a
npm ERR! code ENOENT
npm ERR! errno -2
npm ERR! syscall rename

npm ERR! enoent ENOENT: no such file or directory, rename '/root/dev/src/github.com/segment-sources/zendesk/node_modules/.staging/mkdirp-b0f0282a' -> '/root/dev/src/github.com/segment-sources/zendesk/node_modules/@segment/source/node_modules/grpc/node_modules/node-pre-gyp/node_modules/mkdirp'
npm ERR! enoent ENOENT: no such file or directory, rename '/root/dev/src/github.com/segment-sources/zendesk/node_modules/.staging/mkdirp-b0f0282a' -> '/root/dev/src/github.com/segment-sources/zendesk/node_modules/@segment/source/node_modules/grpc/node_modules/node-pre-gyp/node_modules/mkdirp'
npm ERR! enoent This is most likely not a problem with npm itself
npm ERR! enoent and is related to npm not being able to find a file.
npm ERR! enoent
```

full output- https://gist.github.com/tejasmanohar/ed2497ed7d35d37d9b485134d2a3f1a3
npm debug log- https://gist.github.com/tejasmanohar/93f63476e7bed1bba92c81dc083d2bec

npm -v -> 3.10.3
node -v -> v6.3.1
[...]

PR #14403, opened 2016-10-24
Rewrite how arguments are passed to actions

iarna:
The arguments passed to actions had been... a bit awkward and we often found ourselves intuiting where the staging directory was, and doing that wrong (see #14311, #13528, #14324). This rejiggers it to pass through the staging folder and only figure out the module specific staging path if it actually needs it.

Also included is a test from #14324 and @NatalieWolfe for one of those cases where the staging path was being miscomputed.

### DIFF
--- a/lib/install/action/fetch.js
+++ b/lib/install/action/fetch.js
@@ -3,7 +3,7 @@
 // var packageId = require('../../utils/package-id.js')
 // var moduleName = require('../../utils/module-name.js')
 
-module.exports = function (top, buildpath, pkg, log, next) {
+module.exports = function (staging, pkg, log, next) {
   next()
 /*
 // FIXME: Unnecessary as long as we have to have the tarball to resolve all deps, which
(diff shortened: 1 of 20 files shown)


lib/install/action/fetch.js
```diff
@@ -24,6 +24,6 @@ module.exports = function (top, buildpath, pkg, log, next) {
       name = pkg.package._requested.raw
   }
   log.silly('fetch', packageId(pkg))
-  cache.add(name, version, top, false, next)
+  cache.add(name, version, pkg.parent.path, false, next)
```
zkat: this commit implies this is a refactor -- why is `pkg.parent.path` expected to always be the same as `top`? Was this a bug before?

iarna: This is actually in commented out code. (`action/fetch.js` is at this time, and until some registry 3 features, a no-op). It does strictly speaking fix a bug in that comment out code…

and yes, keeping this commented out action is weird, although it should be handy soon, hopefully!

othiym23:
With a rebase against `release-next` and a green Travis build, this LGTM. :sheep::rocket::fire:

iarna:
Landed in 4.0.3

PR #14461, opened 2016-10-28
Improve output of `npm ping`

legodude17:
This makes it much clearer when `npm ping` succeeds or fails.

### DIFF
--- a/test/tap/ping.js
+++ b/test/tap/ping.js
@@ -47,7 +47,7 @@ test('npm ping', function (t) {
       t.ifError(err, 'no error output')
       t.notOk(code, 'exited OK')
 
-      t.same(JSON.parse(stdout), pingResponse)
+      t.same(stdout, 'Ping success: ' + JSON.stringify(pingResponse) + '\n')
       t.end()
     })
   })
--- a/doc/cli/npm-ping.md
+++ b/doc/cli/npm-ping.md
@@ -8,6 +8,14 @@ npm-ping(1) -- Ping npm registry
 ## DESCRIPTION
 
 Ping the configured or given npm registry and verify authentication.
+If it works it will output something like:
+```
+Ping success: {*Details about registry*}
+```
+otherwise you will get:
+```
+Ping error: {*Detail about error}
+```
 
 ## SEE ALSO
 
--- a/lib/ping.js
+++ b/lib/ping.js
@@ -15,7 +15,13 @@ function ping (args, silent, cb) {
   var auth = npm.config.getCredentialsByURI(registry)
 
   npm.registry.ping(registry, {auth: auth}, function (er, pong) {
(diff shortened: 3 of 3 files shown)


test/tap/ping.js
```diff
@@ -47,7 +47,7 @@ test('npm ping', function (t) {
       t.ifError(err, 'no error output')
       t.notOk(code, 'exited OK')
 
-      t.same(JSON.parse(stdout), pingResponse)
+      t.same(stdout, 'Ping succeeded: ' + JSON.stringify(pingResponse))
```
watilde: This message should be `t.same(stdout, 'Ping success: ' + JSON.stringify(pingResponse))`

legodude17: Ok all done

iarna (changes requested):
It would be helpful if you could squash your commits and use our commit message style. (That is: `<thing changed>: <less than 60 char desc>`, so in this case `ping: something something`)

I'd like to see `doc/cli/npm-ping.md` describe its output. It currently doesn't, which was an oversight from when we originally added this feature.

As a rule, we generally consider what's printed to STDOUT to be a part of our contract, and thus changing that constituting a breaking change. In this case though, particularly seeing as we hadn't documented it previously, I think I'm ok with changing the output.

legodude17:
I am sorry, but I can't figure out how to rebase. I am not sure what is going on. I ran `git rebase -i HEAD~7`, but then it took me to `vi` and I couldn't get out. ☹️

legodude17:
What? I thought I rebased and squashed, but now it says I added two commits. I am so confused. What should I do now @iarna?

KenanY:
@legodude17 Maybe try (off the top of my head):

1. Set your `git` editor to your preferred one: `git config --global core.editor "vim"` (replace `"vim"`)
2. Do `git rebase -i HEAD~9`. Your editor should open a file that looks something like this:

```
pick 7ae4be0
pick 2790146
pick 47b23e9
pick 437a6f6
pick 48ebcfe
pick 20639b1
pick c49565d
pick 21c66f7
pick ca63015
```

3\. For every line but the first, change "pick" to "squash", like so:

```
pick 7ae4be0
squash 2790146
squash 47b23e9
squash 437a6f6
squash 48ebcfe
squash 20639b1
squash c49565d
squash 21c66f7
squash ca63015
```

4\. Save this file. You can then do `git commit --amend` if you would like to change the commit message.

5\. Verify it worked with `git log`
6\. Force-push your changes: `git push -f origin patch-2`

legodude17:
Ok @iarna. Did both of those things. Re review?

iarna (approved):
Looking good, thank you!

iarna:
This has been merged into bde151902 in release-next and will be in the next release of npm.

PR #14462, opened 2016-10-29
shrinkwrap: fix scoped dependency resolution from from

zkat:
346bba1 broke the way shrinkwrap `from` got parsed.
Because of the way the `lookup` function here is treating
it as a version, the specified was being resolved as a
local package in the current directory.

This brings back the URL check to protect against this case,
which was only happening for scoped packages.

...this is probably not the right solution just yet (might need to look further into `isRegistrySpecifier` here and how it's getting used), and this needs test, so it's WIP.

### DIFF
--- a/lib/install/realize-shrinkwrap-specifier.js
+++ b/lib/install/realize-shrinkwrap-specifier.js
@@ -1,6 +1,7 @@
 'use strict'
 var realizePackageSpecifier = require('realize-package-specifier')
 var isRegistrySpecifier = require('./is-registry-specifier.js')
+var url = require('url')
 
 module.exports = function (name, sw, where, cb) {
   function lookup (ver, cb) {
@@ -8,7 +9,7 @@ module.exports = function (name, sw, where, cb) {
   }
   if (sw.resolved) {
     return lookup(sw.resolved, cb)
-  } else if (sw.from) {
+  } else if (sw.from && url.parse(sw.from).protocol) {
     return lookup(sw.from, function (err, spec) {
       if (err || isRegistrySpecifier(spec)) {
         return thenUseVersion()
--- a/test/tap/shrinkwrap-no-resolved.js
+++ b/test/tap/shrinkwrap-no-resolved.js
@@ -1,0 +1,59 @@
+'use strict'
+var test = require('tap').test
+var Tacks = require('tacks')
+var File = Tacks.File
+var Dir = Tacks.Dir
+var path = require('path')
+var common = require('../common-tap.js')
+
+var testdir = path.resolve(__dirname, path.basename(__filename, '.js'))
+
+var fixture = new Tacks(Dir({
+  'package.json': File({
+    name: 'x',
+    version: '1.0.0',
+    dependencies: {
+      '@blarg/foo': 'file://./bfoo',
+      'foo': 'file://./foo'
+    }
+  }),
+  'npm-shrinkwrap.json': File({
+    dependencies: {
+      '@blarg/foo': {
+        version: '1.0.0',
+        from: '@blarg/foo@1.0.0'
+      },
+      'foo': {
+        version: '1.0.0',
+        from: 'foo'
+      }
+    }
+  })
+}))
+
+function setup () {
+  fixture.create(testdir)
+}
+
+function cleanup () {
+  fixture.remove(testdir)
+}
+
+test('setup', function (t) {
+  cleanup()
+  setup()
+  t.end()
+})
+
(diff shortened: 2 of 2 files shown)


test/tap/shrinkwrap-no-resolved.js
```diff
@@ -0,0 +1,76 @@
...
+        version: '1.0.0',
+        from: '@blarg/foo@1.0.0'
+      },
+      'foo': {
+        version: '1.0.0',
+        from: 'foo'
+      }
+    }
+  }),
+  // 'foo': Dir({
```
othiym23: baleet plz

test/tap/shrinkwrap-no-resolved.js
```diff
@@ -0,0 +1,76 @@
+'use strict'
+var test = require('tap').test
+var Tacks = require('tacks')
+var File = Tacks.File
+var Dir = Tacks.Dir
+var fs = require('fs')
+var path = require('path')
+var common = require('../common-tap.js')
+
+// NOTE: This test will only remain relavent until npm@5 when
```
othiym23: s/relavent/relevent/

test/tap/shrinkwrap-no-resolved.js
```diff
@@ -0,0 +1,76 @@
...
+// npm-shrinkwrap.json will override EVERYTHING in you package.json.
```
othiym23: s/you/your/

othiym23 (changes requested):
Some semi-nitpicky changes, but the change & test are pretty simple. Lemme know when to look again!

othiym23:
LGTM. Merge at will. :sheep::rocket::fire:

othiym23 (approved):

test/tap/shrinkwrap-no-resolved.js
```diff
@@ -0,0 +1,59 @@
...
+}
+
+test('setup', function (t) {
+  cleanup()
+  setup()
+  t.end()
+})
+
+test('missing resolved field', function (t) {
+  common.npm(['install'], {registry: 'nope', cwd: testdir}, function (err, code, out, stderr) {
```
iarna: The options object is passed to `child_process.spawn`. This should be:

`common.npm(['install', '--registry=nope'],…`

test/tap/shrinkwrap-no-resolved.js
```diff
@@ -0,0 +1,59 @@
...
+    t.is(err, null, 'No fatal errors running npm')
```
iarna: That's not what `err` means. `err` will ONLY be set if there's an OS error when forking. Check `code` for `npm`'s error status.

I strongly the use of `scripts/maketest` to generate the boilerplate, as it does all this stuff for you.

test/tap/shrinkwrap-no-resolved.js
```diff
@@ -0,0 +1,59 @@
...
+  }),
+  'npm-shrinkwrap.json': File({
+    dependencies: {
+      '@blarg/foo': {
+        version: '1.0.0',
+        from: '@blarg/foo@1.0.0'
+      },
+      'foo': {
+        version: '1.0.0',
+        from: 'foo'
```
iarna: So if you take out the `@blarg/foo` you'll see that `npm` tries to install this from the network even though it's a local file. Similarly if you have a `github` shortcut, eg `foo/bar`, here, it'll also use the network inappropriately.

iarna: I think testing this will require splitting this out into two separate test cases to show that they both fail in the way we expect.

lib/install/realize-shrinkwrap-specifier.js
```diff
@@ -1,14 +1,15 @@
...
+var url = require('url')
 
 module.exports = function (name, sw, where, cb) {
   function lookup (ver, cb) {
     realizePackageSpecifier(name + '@' + ver, where, cb)
   }
   if (sw.resolved) {
     return lookup(sw.resolved, cb)
-  } else if (sw.from) {
+  } else if (sw.from && url.parse(sw.from).protocol) {
```
iarna: Maybe pass this to `npa` and if it's not a version or range then do this code path?

iarna: On second thought, that's basically what the code in here does already. I think we need to dig further into the root cause here, which I believe would be that `isRegistrySpecifier` is doing the wrong thing somehow? Which implies that in turn `realizePackageSpecifier` is doing the wrong thing?

PR #14501, opened 2016-11-02
errors: clarifying the ENOSELF error message a bit

zkat:
This is a tricky error message to word.

It's also an error that normally bites new users.

Perhaps it's worth trying to have a clearer message here.

### DIFF
--- a/doc/cli/npm-install.md
+++ b/doc/cli/npm-install.md
@@ -343,6 +343,10 @@ folder structures that npm creates.
 
 ### Limitations of npm's Install Algorithm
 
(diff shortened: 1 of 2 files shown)


othiym23:
This is an improvement, in that it provides a clue as to why npm is grumpy, but it still doesn't give new users enough information to understand why what they're trying to do isn't allowed. Perhaps pointing to the docs (and making sure the docs clearly explain why this is a bad idea)?

legodude17:
I really like this, it will help a lot with people filing issues about that error.

othiym23:
:sheep::rocket::fire:

Issue #14493, opened 2016-11-02
prepublishOnly environment variable not include node_modules/.bin 


leecade:
#### I'm opening this issue because:

  - [ ] npm is crashing.
  - [ ] npm is producing an incorrect install.
  - [ ] npm is doing something I don't understand.
  - [x] Other (_see below for feature requests_):

#### What's going wrong?

prepublishOnly environment variable not include `node_modules/.bin`

#### How can the CLI team reproduce the problem?

```sh
$ npm init -y
$ mkdir node_modules/.bin
$ echo "#\!/usr/bin/env node\nconsole.log('output')" > node_modules/.bin/test-build
```

`package.json`

```json
{
  "name": "test",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1",
    "build": "test-build",
    "prepublishOnly": "npm run build"
  },
  "keywords": [],
  "author": "",
  "license": "ISC"
}
```

Run `npm run build` got the right outout:

```
❯ npm run build

> test@1.0.0 build /Users/yuji/tmp/test
> test-build

output
```

but, run `npm publish` not found the `test-build` command

```
❯ npm publish

> test@1.0.0 prepublishOnly /Users/yuji/.npm/test/1.0.0/package
> npm run build


> test@1.0.0 build /Users/yuji/.npm/test/1.0.0/package
> test-build

sh: test-build: command not found

npm ERR! Darwin 16.1.0
npm ERR! argv "/usr/local/Cellar/node/6.8.1/bin/node" "/usr/local/bin/npm" "run" "build"
npm ERR! node v6.8.1
npm ERR! npm  v4.0.1
npm ERR! file sh
npm ERR! code ELIFECYCLE
npm ERR! errno ENOENT
npm ERR! syscall spawn
npm ERR! test@1.0.0 build: `test-build`
npm ERR! spawn ENOENT
npm ERR!
npm ERR! Failed at the test@1.0.0 build script 'test-build'.
npm ERR! Make sure you have the latest version of node.js and npm installed.
npm ERR! If you do, this is most likely a problem with the test package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     test-build
npm ERR! You can get information on how to open an issue for this project with:
npm ERR!     npm bugs test
npm ERR! Or if that isn't available, you can get their info via:
npm ERR!     npm owner ls test
[...]

PR #14502, opened 2016-11-02
simplify publish lifecycle invocation

othiym23:
Previously (in 0ab0c11) support was added for running the publish and postpublish lifecycle scripts. Some work was done to ensure that this was run in the correct path both when publishing from the package's working directory, as well as when publishing from a package tarball file or URL. The latter case is interesting, because it would only ever work for lifecycle scripts that didn't rely on bins exported by (dev)dependencies, due to being run directly from the cached package.json. In the process, there was a lot of inline use of ternary operators that made it very hard to understand what was going on.

There are two ways to fix this:

1. An expensive way, which is to explode the package once it's been cached, install it, and then run the lifecycle scripts from that  directory.
2. A cheap way, which is to assume that whatever scripts needed to be run were already run when packing the tarball or remote package, and just don't run the lifecycle scripts at all unless you're publishing from the working directory.

This patch opts for approach #2. Publishing directly from tarballs is an option for power users, and having lifecycle events fire in that case is a conflation of the build and upload stages of publish that cuts against what (I believe) most users are trying to do, which is to build an
artifact, certify it, and then only publish it if certification succeeds.

However, this is technically a breaking change, and as such should either ship before `npm@4` becomes `latest`, or be held for `npm@5`. (I'd prefer the former, but leave that for the team to decide.)

**r**: @iarna 
**r**: @zkat

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -190,7 +190,7 @@ function linkBins (pkg, folder, parent, gtop, cb) {
   }
   var binRoot = gtop ? npm.globalBin
                      : path.resolve(parent, '.bin')
-  log.verbose('link bins', [pkg.bin, binRoot, gtop])
+  log.verbose('linkBins', [pkg.bin, binRoot, gtop])
 
   asyncMap(Object.keys(pkg.bin), function (b, cb) {
     linkBin(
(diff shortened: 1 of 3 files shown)


othiym23:
This is intended to fix #14493.

iarna (approved):
LGTM 🚀

PR #14559, opened 2016-11-08
doc: we do NOT check npmrc permissions

iarna:
These docs slipped into the codebase without any code. If this was ever true, it's not now.


iarna:
This is now merged into `release-next` and will be a part of the release this Thursday (2016-11-17).

PR #14569, opened 2016-11-09
Add save options to link command

mrstork:
Working on #1166

Still need to write tests. I tried copying the ideas from `test/tap/install-save-local.js` into `test/tap/link.js` but the dependencies weren't being added to package.json. However, running the same scenario locally worked as expected. Could use some guidance.

### DIFF
--- a/doc/cli/npm-link.md
+++ b/doc/cli/npm-link.md
@@ -4,7 +4,7 @@ npm-link(1) -- Symlink a package folder
 ## SYNOPSIS
 
     npm link (in package dir)
-    npm link [<@scope>/]<pkg>[@<version>]
+    npm link [<@scope>/]<pkg>[@<version>] [--save|--save-dev|--save-optional] [--save-exact]
 
     alias: npm ln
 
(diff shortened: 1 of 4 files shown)


mrstork:
I thought it would be helpful to see the tests I tried to add so I also committed those

iarna:
The CLI team has been thinking a lot lately about how we want symlinks to work, not just for `npm link` but more generally. We're in the process of putting together a set of documents describing the behavior we'd like to see and that will encompass how `npm link` works and if it should include `--save` and if it does, exactly what that would mean. 

As such, I'd like to hold off on this until we have those documents out, so we can discuss as a community what we want the behavior to be in the larger context of `npm` and the Node.js ecosystem. Ideally they should be ready by/before the first week of January and I'll try to make a note here when they are.

iarna:
I'm going to close this as have `file:` saving links to package.json and package-lock.json and we will have a new RFC up soon for new `npm link` behavior. Thanks again for this though!

Issue #6756, opened 2014-11-20
npm doctor outline


smikes:
This should run as a standalone tool as well as a builtin command.  As a standalone tool it would (ideally) be a single script e.g.,

```
curl -L https://npmjs.org/doctor.sh | sh
```

should not require the existence of a working `npm` or any dependencies.
1. run `npm ping`  - determine if registry is reachable
2. `npm ping` result should include version of latest stable `node` and `npm`
3. `doctor` should check if `node` and `npm` are out of date
   - if `node` out of date.. what? 
   - if `npm` out of date, offer to run `npm upgrade` (qv) or just `npm i -g npm`
4. `doctor` should check for the presence of `npm-debug.log` and offer to send it to `npm report` (qv)
5. Always/sometimes run `npm cache clean` ? prompt first?
6. check for auth problems -- if not auth'd, check if there are saved credentials and (?) nuke them?
7. provide links to change password/change email scripts
8. ... other stuff?  

A further note -- First-class windows support implies the existence of

```
curl -L https://npmjs.org/doctor.cmd | cmd
```

_for windows it's not doctor.~~sh~~, but doctor.cmd_

what color should we paint this thing?  /cc @othiym23 @iarna @faiq @kenanY @isaacs

othiym23:
This has been moved to the [npm roadmap](https://github.com/npm/npm/wiki/Roadmap), which we're using instead of the confusing `next-*` labels now.

othiym23:
The CLI team probably isn't going to have the bandwidth to schedule this in the next year, but this is functionality we'd dearly like to have. Is there interest from the community in picking up this work?

watilde:
I still have an interest in this feature! Here is the scetch of the simple `npm doctor`:

1. run `npm ping` - determine if registry is reachable
2. doctor should check if node and npm are out of date
  1. if node is out of date, suggest go nodejs.org
  2. if npm is out of date, suggest run `npm i -g npm`
3. if npm has cache(`npm cache ls`), suggest `npm cache clear`
4. Display tips: https://github.com/npm/npm/wiki/Troubleshooting
5. Suggest run `npm report` which is like [npm-report](https://github.com/watilde/npm-report).

watilde:
Hello! Let me share my progress of `npm doctor`:

![screen shot 2016-11-06 at 20 55 22](https://cloud.githubusercontent.com/assets/1716463/20041090/57fcca2c-a464-11e6-85d6-b992fa43ea3e.png)

[...]

PR #14582, opened 2016-11-10
Add new subcommand `npm doctor`

watilde:
From here https://github.com/npm/npm/issues/6756#issuecomment-251798653

*this screenshot was updated with current progress at 4th December 2016*
screenshot

ToDo:
- [x] Add new subcommand
- [x] Write a test
- [x] Replace `getLatestNodejsVersion` with a fixturable way
- [x] Update docs

### DIFF
--- a/doc/cli/npm-doctor.md
+++ b/doc/cli/npm-doctor.md
@@ -1,0 +1,44 @@
+npm-doctor(1) -- Check your environments
+========================================================
+
+## SYNOPSIS
+
+    npm doctor
+
+## DESCRIPTION
+
+npm command is just a single command, but depends on several things outside
+the code base. Broadly speaking, it is the following three types:
+
++ Technology stack: Node.js, git
++ Registry: `registry.npmjs.com`
++ Files: `node_modules` in local/global, cached files in `npm config get cache`
+
+Without all of these working properly, the npm command will not work properly.
+Many issue reports that arrive under us are often attributable to things that
+are outside the code base as described above, and it is necessary to confirm
+that this is correctly set with one command would help you solve your issue.
+Also, in addition to this, there are also very many issue reports due to using
+old versions of npm. Since npm is constantly improving, in every aspect the
+`latest` npm is better than the old version(it should be, and we are trying).
+
+From the above reasons, `npm doctor` will investigate the following items in
+your environment and if there are any other recommended settings, it will
+display the recommended example.
+
+|What to check|What we recommend|
+|---|---|
+|`npm ping`|It must be able to communicate with the registry|
+|`npm -v`|It is preferable that the latest version of LTS is used|
+|`node -v`|It is preferable that the latest version of LTS is used|
+|`npm config get registry`|In order not to consider exceptions, it is better to set default values `registry.npmjs.org`|
+|`which git`|Git has to be installed|
+|`Perms check on cached files`|All cached module files must be readable.|
+|`Perms check on global node_modules`|All global module files must be executable.|
+|`Perms check on local node_modules`|All local module files must be executable.|
+|`Checksum cached files`|All cached files must not be broken|
+
+## SEE ALSO
+
+* npm-bugs(1)
+* npm-help(1)
--- a/lib/doctor.js
+++ b/lib/doctor.js
@@ -4,18 +4,19 @@ var table = require('text-table')
 var color = require('ansicolors')
 var styles = require('ansistyles')
 var semver = require('semver')
-var which = require('which')
 var npm = require('./npm.js')
-var ping = require('./ping')
-var ansiTrim = require('./utils/ansi-trim')
+var log = require('npmlog')
+var ansiTrim = require('./utils/ansi-trim.js')
 var output = require('./utils/output.js')
-var defaultRegistry = require('./config/defaults').defaults.registry
-var checksumCachedFiles = require('./doctor/checksum-cached-files')
-var checkFilesPermission = require('./doctor/check-files-permission')
-var getLatestNodejsVersion = require('./doctor/get-latest-nodejs-version')
+var defaultRegistry = require('./config/defaults.js').defaults.registry
+var checkPing = require('./doctor/check-ping.js')
+var getGitPath = require('./doctor/get-git-path.js')
+var checksumCachedFiles = require('./doctor/checksum-cached-files.js')
+var checkFilesPermission = require('./doctor/check-files-permission.js')
+var getLatestNodejsVersion = require('./doctor/get-latest-nodejs-version.js')
 var getLatestNpmVersion = require('./doctor/get-latest-npm-version')
 var globalNodeModules = path.join(npm.config.globalPrefix, 'lib', 'node_modules')
-var localNodeModules = path.join(npm.config.localPrefix, 'lib', 'node_modules')
+var localNodeModules = path.join(npm.config.localPrefix, 'node_modules')
 
 module.exports = doctor
 
@@ -29,16 +30,17 @@ function doctor (args, silent, cb) {
   }
 
   var actionsToRun = [
-    [ping, {}, true],
+    [checkPing],
     [getLatestNpmVersion],
     [getLatestNodejsVersion, args['node-url']],
-    [which, 'git'],
-    [checkFilesPermission, npm.cache, 4],
-    [checkFilesPermission, globalNodeModules, 1],
-    [checkFilesPermission, localNodeModules, 1],
+    [getGitPath],
+    [checkFilesPermission, npm.cache, 6],
+    [checkFilesPermission, globalNodeModules, 4],
+    [checkFilesPermission, localNodeModules, 6],
     [checksumCachedFiles]
   ]
 
+  log.info('doctor', 'Running checkup')
   chain(actionsToRun, function (stderr, stdout) {
     if (stderr && stderr.message !== 'not found: git') return cb(stderr)
     var outHead = ['Check', 'Value', 'Recommendation']
@@ -75,9 +77,9 @@ function makePretty (p) {
   var nodeLTS = p[2].replace('v', '')
   var whichGit = p[3] || 'not installed'
   var readbleCaches = p[4] ? 'ok' : 'notOk'
-  var executableGlobalModules = p[4] ? 'ok' : 'notOk'
-  var executableLocalModules = p[5] ? 'ok' : 'notOk'
-  var checksumCachedFiles = p[6] ? 'ok' : 'notOk'
+  var executableGlobalModules = p[5] ? 'ok' : 'notOk'
+  var executableLocalModules = p[6] ? 'ok' : 'notOk'
+  var checksumCachedFiles = p[7] ? 'ok' : 'notOk'
   var npmV = npm.version
   var nodeV = process.version.replace('v', '')
   var registry = npm.config.get('registry')
--- a/lib/doctor.js
+++ b/lib/doctor.js
@@ -1,0 +1,107 @@
+var path = require('path')
+var chain = require('slide').chain
+var table = require('text-table')
+var color = require('ansicolors')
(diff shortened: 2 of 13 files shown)


legodude17:
Does this add any new deps?

othiym23:
@watilde I owe you a response! I'll try to get to it today! The summary is this looks fantastic, but we need to talk a lil more about the functionality everybody wants.

legodude17:
Current comments:
+ Any ideas on a more exhaustive test then, "It prints things out with 5 lines"? I was thinking maybe make sure it prints a table and that the columns match.
+ What is the best way to get the latest `node` version? I think that yours is the best way.

watilde:
@legodude17 Yes, that's a good point. Before that, I'm thinking about a way to make the test having a fixture in test.

@othiym23 Sure thing! Let's continue the discussion at https://github.com/npm/npm/issues/6756

legodude17:
@watilde is this still `in-progress`?

watilde:
@legodude17 It's fixed technically for the original plan, but the code potentially will be changed depending on what kind of request comes from the issue thread. Then the status doesn't matter anymore :)

legodude17:
One suggestion from #6756 is to offer to fix it like `Would you like me to run npm cache clean for you?`. I think that would be great, but maybe in a separate PR.

othiym23 (changes requested):
@watilde, my apologies for the lengthy delay. It's been a bruiser of a month.

Here, in no particular order, are my thoughts:

1. I'm not in love with the idea of prescribing that users nuke their caches every time they run `npm doctor`. It makes installs (potentially much) slower until the cache rewarms, and in cases in which the user is on a flaky internet connection, can actually make installs even flakier. If at all possible, I'd like to move to advising people to only `npm cache clean` individual broken dependencies.

    It's probably out of scope for this initial spike, but something that validates that the cache is in good shape (and maybe went ahead and cleaned up any shasum mismatches automatically) would be a super useful alternative.
2. Something that validates that `$HOME/.npm` and all its children is owned and writable by `$USER` would be very helpful for diagnostic purposes.
3. Similarly, something that checks whether `join(npm.config.globalPrefix, 'lib', 'node_modules')` is writable and tells the user whether they need to use `sudo` to run global installs would be very helpful.
4. It would be _very_ helpful to verify that Git is installed and available, especially for Windows users.
5. This is another thing that doesn't have to be added yet, but something that builds a very simple canned native module with `node-gyp` would be extremely helpful for Windows users unsure whether they have their toolchain correctly installed.
6. Finally, and longest-term, it would be _fantastic_ to have some way for npm to verify its own integrity, which probably requires that some kind of install manifest be generated for it. While it would be useful to have this functionality for npm packages in general, it's especially important for npm, because of the Node.js installer's bad habit of just copying the new npm right over the top of the old one, which causes a bunch of the stranger `module not found` bugs that show up on the issue tracker. Being able to catch this automatically would be hugely useful.

This is basically my wishlist, though. What you've got is a fantastic start, and when you get the above comments addressed, I'll ship what you've got so far. Thank you so much for putting this together!

lib/doctor.js
```diff
@@ -0,0 +1,112 @@
...
+  var actionsToRun = [
+    [ping, {}, true],
+    [getLatestNpmVersion],
+    [getLatestNodejsVersion, args.url],
+    [exec, 'npm cache ls']
+  ]
+
+  chain(actionsToRun, function (stderr, stdout) {
+    if (stderr) return cb(stderr)
+    var outHead = ['Question', 'Answer', 'Prescription']
```
legodude17: Good point @othiym23.

doc/cli/npm-doctor.md
```diff
@@ -0,0 +1,17 @@
...
+
+## SYNOPSIS
+
+    npm doctor
+
+## DESCRIPTION
+
+This command helps you when you get struck by environments which
+potentially have a problem. Some of the questions are same with
+we ask in the ISSUE_TEMPLATE.md.
```
legodude17: I think the best way to do that would be to template it, but I am not sure about what anyone's thoughts are about including a template engine with `npm`.

lib/doctor.js
```diff
@@ -0,0 +1,112 @@
...
+      return cb(null, version)
+    } catch (e) {
+      return cb(e)
+    }
+  })
+}
+
+function getLatestNpmVersion (cb) {
+  fetchPackageMetadata('npm', '.', function (er, d) {
+    cb(er, d.version)
```
legodude17: So `cb(er, d['dist-tags'].latest)`?

test/tap/doctor.js
```diff
@@ -0,0 +1,29 @@
...
+    npm.load({}, function (err) {
+      t.ifError(err, 'npm loaded successfully')
+      t.pass('all set up')
+      t.done()
+    })
+  })
+})
+
+test('npm doctor', function (t) {
+  npm.commands.doctor({url: 'http://localhost:1337/'}, true, function (err, list) {
```
watilde: That's right! I've updated.

watilde:
@othiym23 Hey Forrest, thanks for your thoughtful wish list! I'd like to separate examination processes and auto-fix features and let me reply to your every wish-item.

1. Agreed! This request is the rest of the things I'd like to put into this PR.
2. I've added into this PR already.
3. I've added into this PR already.
4. I've added into this PR already.
5. I'd like to have [node-gyp doctor](https://github.com/nodejs/node-gyp/issues/1048) instead of putting into `npm doctor`. Let's take out this from the current initial scope.
6. This request may make this PR turning into a giant case of scope creep like you said. Could we also take out this from the current initial scope?

Additionally, I've added more permission check for `join(npm.config.localPrefix, 'lib', 'node_modules')` since I was missing! As the initial plan to iterate from it, let's start with the following check-list as the smallest subset of information necessary to improve the supportability of npm.
1. npm ping
2. npm -v
3. node -v
4. npm config get registry
5. which git
6. Perms check on cached files
7. Perms check on `join(npm.config.globalPrefix, 'lib', 'node_modules')`
8. Perms check on `join(npm.config.localPrefix, 'lib', 'node_modules')`

Let me ping you when they're ready.

legodude17:
Some random ideas for things to add:
  1. Ensure the proxy is setup correctly (somehow)
  2. Ensure git is present and works

Keep up the great work @watilde!

test/tap/doctor.js
```diff
@@ -0,0 +1,29 @@
...
+    t.ifError(err, 'npm loaded successfully')
+    t.same(list.length, 5, 'list should have 5 prop')
```
watilde: Yeah it should be since `npm doctor` ask the things to the registry through `npm ping` and `fetchPackageMetadata`. I've added `npm-registry-mock` into this test and verifying all of the output!

doc/cli/npm-doctor.md
watilde: That's absolutly good idea, and I'd like to have the feature in `npm report` command which have been proposaling at https://github.com/npm/npm/issues/5214. I'm planning to use the `ISSUE_TEMPLATE.md` as a acutual template, but let's separete the reporting part from this diagnosis part to make the initial scope clear and simple.

watilde:
@othiym23 I've added some check items that users need, and written up the detailed documentation for the `npm doctor` command! Could I possibly request your review again?

(๑˃̵ᴗ˂̵)و let me cc to you as well @zkat

othiym23:
I think this will be ready in time to include in this week's release, which might end up being the last release of the CLI in 2016! I only have a few further comments:

1. Right now, the command seems to hang while it's running (I'm going to guess most of that is due to checking the shasums in the cache). If you add a log statement to each check at loglevel info and enable the progress bar, this will let users know that in fact something is happening.
2. Because validating the hashes can be very time-consuming if people have a lot of packages in their cache, it might make sense to log at level `debug` each shasum you're validating, especially because you're doing so synchronously. Because the completion routine returns the number of files to check, it should be pretty easy to set up a progress group that will track the process of validating the cache. (This is a nice to have, and if it turns out to be tricky, don't worry about it and somebody can add it in later.)
3. In the next day or two, I'll make a pass at expanding the docs you've written thus far, and make sure they'll tie in nicely with the the new `TROUBLESHOOTING.md` that @snopeks and @zkat are in the process of assembling.

This is really great! I can't wait to get it out into people's hands. Thank you so much for your patience and careful work, Daijiro!

lib/doctor.js
```diff
@@ -0,0 +1,107 @@
...
+  })
+}
+
+function makePretty (p) {
+  var ping = p[0] ? 'ok' : 'notOk'
+  var npmLTS = p[1]
+  var nodeLTS = p[2].replace('v', '')
+  var whichGit = p[3] || 'not installed'
+  var readbleCaches = p[4] ? 'ok' : 'notOk'
+  var executableGlobalModules = p[4] ? 'ok' : 'notOk'
```
iarna: Shouldn't these from this line down be 5, 6 & 7?

lib/doctor/checksum-cached-files.js
```diff
@@ -0,0 +1,35 @@
...
+  var valid = true
+  fileCompletion(npm.cache, '.', Infinity, function (e, files) {
+    if (e) return cb(e)
+    files.some(function (f) {
+      if (!valid) return true
+      var file = path.join(npm.cache, f)
+      if (!/.tgz$/.test(file)) return
+      var tgz = fs.readFileSync(file)
+      try {
+        var pkgJSON = fs.readFileSync(path.join(file, '../package/package.json'))
```
iarna: I think what's wanted here is:

`path.join(path.dirname(file), 'package/package.json')`

lib/doctor/checksum-cached-files.js
```diff
@@ -0,0 +1,35 @@
...
+        var shasum = JSON.parse(pkgJSON).dist.shasum
```
iarna: `dist` is added by the registry and won't be present for modules installed from other sources. `_shasum` should be available however.

lib/doctor/check-files-permission.js
```diff
@@ -0,0 +1,38 @@
+var fs = require('fs')
+var path = require('path')
+var fileCompletion = require('../utils/completion/file-completion.js')
+var getUid = require('uid-number')
+var npm = require('../npm.js')
+
+function checkFilesPermission (root, mask, cb) {
+  var accessible = true
+  if (process.platform === 'win32') return cb(null, !accessible)
```
iarna: This should probably be calling back with `true` shouldn't it? We don't want it to mark windows as always failing right?

lib/doctor.js
```diff
@@ -0,0 +1,107 @@
...
+    cb = silent
+    silent = false
+  }
+
+  var actionsToRun = [
+    [ping, {}, true],
+    [getLatestNpmVersion],
+    [getLatestNodejsVersion, args['node-url']],
+    [which, 'git'],
+    [checkFilesPermission, npm.cache, 4],
```
iarna: `filemode & 0400` would be true for files you have read access to. Should this be rewrite instead? (`6`)

lib/doctor.js
```diff
@@ -0,0 +1,107 @@
...
+    [checkFilesPermission, globalNodeModules, 1],
```
iarna: `filemode & 0100` is true if you have execute access to. I'm assuming that this is supposed to be checking for read access? (`4`)

iarna:
I put together a little patch that fixes the bugs that I encountered and reworks things to allow progress bars. You can see that here:

https://github.com/npm/npm/commit/7cc791df91bc66446f410711c9058763df41ca26

watilde:
@iarna Thanks for your comments, and your patch seems so brilliant.. Let me do cherry-pick it into this PR, or you also can handle the commits to be suited for next-release.

iarna:
Thanks again for this! It was merged and released in 4.1.0.

Issue #14427, opened 2016-10-26
error in install Ghost on Node.js 7


jourjine:
just install node.js and have little big problem

```
$ sudo npm install --production
npm ERR! Linux 4.4.19-29.55.amzn1.x86_64
npm ERR! argv "/usr/local/bin/node" "/usr/local/bin/npm" "install" "--production"
npm ERR! node v7.0.0
npm ERR! npm  v4.0.1

npm ERR! Cannot read property 'deep-extend' of undefined
npm ERR! 
npm ERR! If you need help, you may report this error at:
npm ERR!     <https://github.com/npm/npm/issues>
npm ERR! Linux 4.4.19-29.55.amzn1.x86_64
npm ERR! argv "/usr/local/bin/node" "/usr/local/bin/npm" "install" "--production"
npm ERR! node v7.0.0
npm ERR! npm  v4.0.1

npm ERR! Cannot read property 'extsprintf' of undefined
npm ERR! 
npm ERR! If you need help, you may report this error at:
npm ERR!     <https://github.com/npm/npm/issues>

npm ERR! Please include the following file with any support request:
npm ERR!     /var/www/ghost/npm-debug.log

```

zkat:
This is a strange issue and I'm not sure it's a bug with npm itself, but maybe something is up with `npm@4.0.1` -- so could you upload your `npm-debug.log` to a gist or similar so we can take a look? :)

jourjine:
hi
i change node.js to version 4.6.0 but have same error:

```
sudo npm install --production
npm ERR! Linux 4.4.19-29.55.amzn1.x86_64
npm ERR! argv "/usr/local/bin/node" "/usr/local/bin/npm" "install" "--production"
npm ERR! node v4.6.1
npm ERR! npm  v4.0.1

npm ERR! Cannot read property 'deep-extend' of undefined
npm ERR! 
npm ERR! If you need help, you may report this error at:
npm ERR!     <https://github.com/npm/npm/issues>
npm ERR! Linux 4.4.19-29.55.amzn1.x86_64
npm ERR! argv "/usr/local/bin/node" "/usr/local/bin/npm" "install" "--production"
npm ERR! node v4.6.1
npm ERR! npm  v4.0.1

npm ERR! Cannot read property 'extsprintf' of undefined
npm ERR! 
npm ERR! If you need help, you may report this error at:
npm ERR!     <https://github.com/npm/npm/issues>

npm ERR! Please include the following file with any support request:
[...]

PR #14601, opened 2016-11-11
shrinkwrap: guard against hard-to-find bug

zkat:
This is a quick guard that should take care of https://github.com/npm/npm/issues/14427. Unfortunately, I don't know how to reproduce this. There's a mystery place in the current architecture where that property becomes `undefined` even though it should _really_ always be assigned. We can't really merge this until we have a failing test for it :(


zkat:
@iarna found the source of this problem, so I'm closing this PR pending a separate PR

stefanpenner:
@zkat mind sharing the PR that introduced the fix?

PR #14602, opened 2016-11-12
chore: switch to using nyc directly with --cache=true

bcoe:
switch to using nyc directly; makes it slightly easier to use advanced options like `--cache`.

let's see if `--cache=true` speeds things up!


bcoe:
@othiym23 @iarna hooray, caching works like I was hoping:

screen shot 2016-11-12 at 9 52 09 pm

iarna (approved):

iarna:
Thank you @bcoe!!!! It's now merged into `release-next` and will be a part of the release this Thursday (2016-11-17).

PR #15082, opened 2016-12-01
inflate-bundled: Initialize nodes from bundled dependencies

iarna:
This solves the root cause of the errors where an attempt to read `node.package.devDependencies` crashes because `devDependencies` is not defined.

Fixes: #14427

### DIFF
--- a/lib/install/inflate-bundled.js
+++ b/lib/install/inflate-bundled.js
@@ -1,10 +1,12 @@
 'use strict'
 var validate = require('aproba')
 var childPath = require('../utils/child-path.js')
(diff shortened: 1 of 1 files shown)


iarna:
I am having the damnedest time coming up with a repro for this, even believing I know the root cause. That is, I haven't figured out what combination of bundles and shrinkwraps and modules on disk produce this scenario.

zkat (approved):
Considering how difficult this one is to repro, I'd like to have something at least in place for folks dealing with the issue. Let's just avoid closing the connected issue(s) until we have an actual test case for it, though.

PR #15083, opened 2016-12-01
dedupe: Remove unused code

iarna:
Remove useless normalizeTree override. This was needed back before loadCurrentTree used readGlobalPackageData for globals.


zkat (approved):
Talked to Rebecca and the tests already pass. This is ded. LGTM!

Issue #12529, opened 2016-04-29
Implement statistics tracking


iarna:
We don't know how people use npm. We don't know how well it works for them. Not really. Our ways of hear our users, here, twitter, the web, only reach a tiny fraction of our users. We don't know how fast or slow installs typically are, nor how big they are, nor how often folks see errors.

In order to work toward answering these questions so that we can make better decisions about the tool, we are going to start adding some metrics. Initially we will just track, on an opt-in basis, how often installs end in error. This is intentionally small, both so that implementation on our part will be easy, but also so that the privacy implications will be clear.
### The work to be done on this project is:
- [ ] For users who opt in (by adding a configuration option), start recording how many times installs succeed or end in an error in a file in the cache (eg, something like `CACHEDIR/metrics.json`).
  
  When npm completes:
  - If a metrics file exists, increment the successful or failed installs and update the `to` time.
  - If a metrics file does not exist, create a new stats file with a new `metricId` associated with it and fresh `from` and `to` times.

---
- [ ] Add a new API to [`npm-registry-client`](https://github.com/npm/npm-registry-client/issues/138) and the [registry itself](https://github.com/npm/registry/issues/11) for sending stats information to a registry.

---
- [ ] Using the new API, send stats data to the registry.
  #### When starting a new install command:
  - If the metric file exists:
    - Call `sendAnonymousCLIMetrics` to send the metrics to the default registry. If the request fails or does not complete before npm exits the metrics file should be left in place. If the request succeeds then the current metrics file should be removed.

_We may not be able to support 0.8, as it can't `unref` sockets (AFAIK)._

lazywithclass:
I am willing to help with this, do you have some more details / some key points as a result of the CLI team meeting?

iarna:
Sure, I need to finish updating these issues with the latest round of discussions internally, but the place to start is → https://github.com/npm/npm-registry-client/issues/138

When I do that update, I'll tag you over on that issue.

lazywithclass:
Many thanks @iarna

lazywithclass:
@iarna I've just committed (https://github.com/lazywithclass/npm/commit/e58df224b5bfca0b64652ce7aa84c4caf3667a5f) my idea on how to collect the required metrics. It's just a stub, to see if I'm going in the right direction.

Could you please have a look to see if this makes sense at all? :P

iarna:
@lazywithclass Ahh, sorry about not getting back to you there.

I ended up implementing it as seen here:
[...]

PR #15084, opened 2016-12-01
Record and report success metrics

iarna:
See: https://github.com/npm/npm/issues/12529


legodude17:
Nice idea! Is this implemented on the registry yet?

test/tap/anon-cli-metrics.js
```diff
@@ -0,0 +1,148 @@
...
+      delete metrics.from
+      delete metrics.to
+      return JSON.stringify(metrics)
+    })
+    t.done()
+  })
+})
+
+test('record success', function (t) {
+  common.npm(['install', '--no-send-metrics', 'success'], conf, function (err, code, stdout, stderr) {
```
othiym23: I like that the tests don't rely on specific knowledge of what the metrics file looks like. This is a good set of integration tests! But running three installs to check one piece of behavior feels like a lot of overhead for the results. I don't think whacking this (and the next) install to speed up the test is a blocker, but if it could be done without a huge amount of trouble, it would be beneficial.

lib/install.js
```diff
@@ -216,9 +217,46 @@ function Installer (where, dryrun, args) {
...
 
-Installer.prototype.run = function (cb) {
+function updateStatsFile (itWorked) {
+  // If the stats reporter hasn't managed to PUT yet then kill it so that it doesn't
+  // step on our updating the cli-stats json
+  if (npm.statsProcess) npm.statsProcess.kill('SIGKILL')
+  var statsFile = path.join(npm.config.get('cache'), 'anonymous-cli-stats.json')
+  var stats
+  try {
+    stats = JSON.parse(fs.readFileSync(statsFile))
```
othiym23: Is it premature to abstract the metrics out into a defined interface, if our goal is to introduce more supportability metrics over time? In other words, would it make sense to put the metrics object into a module that can be required as needed across the project (or maybe hung off a `process.on` listener, a la numbat)?

I'm fine with shipping this as-is, but it does seem like this is something we're going to want to change pretty soon, and having it behind a defined interface from the get-go therefore makes a certain amount of sense.

othiym23 (commented):
Here's what I've got thus far – nothing in here should be construed as a blocker to merging this, but I am curious about your responses.

lib/utils/error-handler.js
```diff
@@ -17,6 +17,10 @@ var errorMessage = require('./error-message.js')
 process.on('exit', function (code) {
   log.disableProgress()
   if (!npm.config || !npm.config.loaded) return
+
+  // kill any outstanding stats reporter if it hasn't finished yet
+  if (npm.statsProcess) npm.statsProcess.kill('SIGKILL')
```
iarna: `send-metrics` is explicitly written with the intent that killing it should be safe. I consider it more important that we be certain that there aren't any lingering processes (that are maybe updating the stats file) then that `send-metrics` not be interrupted.

lib/utils/send-metrics.js
```diff
@@ -0,0 +1,52 @@
...
+  var path = require('path')
+  var npm = require('../npm.js')
+  var metricsFile = process.argv[2]
+  var metricsRegistry = process.argv[3]
+
+  var cliMetrics = JSON.parse(fs.readFileSync(metricsFile))
+  npm.load({}, function (err) {
+    if (err) return
+    npm.registry.config.retry.retries = 0
+    npm.registry.sendAnonymousCLIMetrics(metricsRegistry, cliMetrics, function (err) {
```
iarna: This is intended to be allowed, per the spec I wrote, without double counting. This is why the metrics file gets a unique ID that's changed only when sending the CLI metrics is successful. 

Without a succesfull send we update the metrics file without changing that unique ID. This way the registry can just count the most recent set of data with that unique ID.

othiym23: Excellent! Consider my comment withdrawn, then.

test/tap/anon-cli-metrics.js
iarna: One set of behavior? I'm not sure I follow? There's no way to tell the installer to do all (or any) of these things all in one go…

othiym23: I hadn't considered the fact that you're testing that the stats are being written correctly. With that in mind, my objection makes much less sense. Never mind!

lib/install.js
iarna: I've just pushed a little refactor that moves all of the existing logic into a single file (without changing any of those interfaces), which makes this a bit clearer.

At the moment the entire interface that npm has is `saveMetrics` with `true` or `false`. And yeah, that's clearly not going to be sufficient going forward.

On the other hand, I think I'm inclined to leave it as is until we add another metric. My inclination is to wait until we have a concrete addition that necessitates the change in interface. Right now it'd be pretty hypothetical. We can make educated guesses, sure, but front loading the work uh... hmm, doesn't right?

othiym23: I think that the latest refactor is more than sufficient for now. Thanks for the clear answers!

othiym23:
After this round of review, this is ready to merge. :sheep::rocket::fire:

iarna:
@legodude17 Nope. You can see the issue for tracking that part over here → https://github.com/npm/registry/issues/11

PR #15086, opened 2016-12-01
deps: Move shrinkwrap extraneous check suppression to is-extraneous

iarna:
If your `npm-shrinkwrap.json` contains modules not found in your `package.json` we don't complain about them at install time.

We WILL complain about them at `npm ls` time.  And we WILL complain if you try to make changes to your shrinkwrap though.

This fixes a bug where we were NOT complaining about them at `npm ls` time.


iarna:
This is now obsolete.

PR #15088, opened 2016-12-01
ls: Don't exclude modules that are both dev & prod with --production

iarna:
Previously if a module was in devDependencies then it would never show up with `npm ls --production` even if it was ALSO in dependencies.


iarna:
merged as 7a7710ba7

PR #15091, opened 2016-12-01
actions: Make sure we don't recurse infinately when failing a module

iarna:
This is just a guard to ensure that requires cycles on failing modules won't result in infinite recursion. I'm not certain that I can produce this outcome in an integration test (though I have some ideas to try), but it at the very least could use a unit test.


iarna:
Landed from #12784

PR #15092, opened 2016-12-01
Run preinstall lifecycle before loading the tree

iarna:
Needs:

* [ ] Tests (add assertions to test/tap/preinstall-early)


iarna:
This was implemented elsewhere.

PR #15103, opened 2016-12-01
Removing unused dependencies.

helio-frota:


iarna:
Thanks for noticing these. It turns out that we still want them there, for a rather obscure reason:

These are included to deduplicate them. While npm itself doesn't use them, a bunch of npm's transitive dependencies do.

And because npm uses bundled dependencies and needs to be installable by npm@1.x, it can't rely on its automatic deduplication.

helio-frota:
ok, also, thanks for the clarification.

Issue #15097, opened 2016-12-01
Cannot install package from tarball out of package.json entry generated by npm


mastertinner:
#### I'm opening this issue because:

  - [ ] npm is crashing.
  - [ ] npm is producing an incorrect install.
  - [x] npm is doing something I don't understand.
  - [ ] Other (_see below for feature requests_):

#### What's going wrong?

Cannot install package from tarball out of `package.json` entry generated by npm

#### How can the CLI team reproduce the problem?

1. Have an npm package as a tarball, e.g. `external-package-1.0.0.tgz` in the root folder of your project
1. Run `npm install --save ./external-package-1.0.0.tgz`
1. The package gets saved as `"external-package": "file://external-package-1.0.0.tgz"` in `package.json` and gets installed correctly
1. Run `npm install`
1. `npm ERR! addLocal Could not install /` gets thrown

I would expect the package to be written into `package.json` as `"external-package": "file:./external-package-1.0.0.tgz"` which would then install correctly when running `npm install`.

### supporting information:

 - `npm -v` prints: 4.0.2
 - `node -v` prints: v7.2.0
 - `npm config get registry` prints: https://registry.npmjs.org/
 - Windows, OS X/macOS, or Linux?: macOS

legodude17:
This does appear to be a bug. Can anyone reproduce this?

zkat:
lol of course. Thanks for submitting this! Turns out I'd already patched this earlier this month, but completely forgot to actually put up the PR after that. Probably because of the election.

https://github.com/npm/npm/pull/15114 should fix this, and I'll try and get it in the next release. There's also an explanation over there of what's going on. :)

If you need this working *now*, downgrading to `node@7` while that patch lands should fix it for you.

mastertinner:
Awesome! Thx, @zkat!

PR #15114, opened 2016-12-02
save: stop using url.format for relative local dep paths

zkat:
`node@7` [introduced an issue](https://github.com/nodejs/node/issues/9521) with file URIs that broke the way npm was expecting them to be parsed. Turns out we don't really need to be doing any parsing in this particular case. I've added tests to verify things, too.

Fixes: #15097


iarna (approved):

iarna:

PR #15157, opened 2016-12-07
Update CLI docs regarding npm cache clear/clean

JonahMoses:
It looks to be that `npm cache clear` is being deprecated in favor of `npm cache clean` so adjusting the docs to represent that (if that is in fact what is happening).

Or if both are going to continue to remain then I could make the docs more verbose to reflect that so that the community knows.

### DIFF
--- a/doc/cli/npm-cache.md
+++ b/doc/cli/npm-cache.md
@@ -14,7 +14,7 @@ npm-cache(1) -- Manipulates packages cache
 
 ## DESCRIPTION
 
-Used to add, list, or clear the npm cache folder.
+Used to add, list, or clean the npm cache folder.
 
 * add:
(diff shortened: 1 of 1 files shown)


watilde:
This is good catch! We have affordance for the subcommands of `npm cache`, and it would be helpful to users if we could mention the aliases in the document like you said.

Aliases:

|subcommand name|aliases|
|---|---|
|add| - |
|ls|sl (doesn't need to be mentioned IMO)|
|clean|clear, rm|

In the code:
+ https://github.com/npm/npm/blob/latest/lib/cache.js#L119-L121

iarna:
Thank you for this! It was merged and released in 4.1.0.

PR #15196, opened 2016-12-11
Allow default tag config for splats

zkat:
When using `*` selectors and there's nothing but pre-release versions available, allow configuring which tag is looked up, instead of hardcoding it to `latest`.


iarna:
Worth noting that this is already what `fetch-package-metadata` was doing, so this just brings them into alignment with each other.

iarna (approved):
LGTM

iarna:

PR #10662, opened 2015-12-02
install: simplify fetchPackageMetadata

othiym23:
Make the logic a little more similar to what `addNamed` does in the cache, but mostly take advantage of `semver.maxSatisfying()`.

Needs tests.

Fixes: #10641 

**r**: @iarna

PR #15205, opened 2016-12-13
fetch-package-metadata,add-named: Unify package version selection

iarna:
Supplants #10662.

Gives us a single source of truth when it comes to selecting which version to use when given a range specifier.


othiym23:
Add or repurpose some unit tests for the extracted function and this will be ready to go. Nice work!

othiym23:
Tests – and everything else – LGTM. :sheep::rocket::fire:

iarna:

PR #15234, opened 2016-12-15
Clarify the behavior of the `files` array

jbcpollak:
The `files` documentation is a bit confusing (and actually incorrect with respect to current `files` vs `.npmignore` behavior, so I hope this clarifies it.

### DIFF
--- a/doc/files/package.json.md
+++ b/doc/files/package.json.md
@@ -170,12 +170,15 @@ npm also sets a top-level "maintainers" field with your npm user info.
 
 The "files" field is an array of files to include in your project.  If
 you name a folder in the array, then it will also include the files
-inside that folder. (Unless they would be ignored by another rule.)
+inside that folder. (Unless they would be ignored by another rule in this section.)
 
-You can also provide a ".npmignore" file in the root of your package or
-in subdirectories, which will keep files from being included, even
-if they would be picked up by the files array.  The `.npmignore` file
-works just like a `.gitignore`.
+You can also provide a `.npmignore` file in the root of your package or
+in subdirectories, which will keep files from being included. The `.npmignore` 
+file works just like a `.gitignore`. If there is a `.gitignore` 
+file, and `.npmignore` is missing, `.gitignore`'s contents will be used instead.
+
+Files included with the "package.json#files" field _cannot_ be excluded through
+`.npmignore` or `.gitignore`.
 
 Certain files are always included, regardless of settings:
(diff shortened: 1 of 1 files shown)


coconutperm (approved):

doc/files/package.json.md
```diff
@@ -170,12 +170,15 @@ npm also sets a top-level "maintainers" field with your npm user info.
 
 The "files" field is an array of files to include in your project.  If
 you name a folder in the array, then it will also include the files
-inside that folder. (Unless they would be ignored by another rule.)
+inside that folder. (Unless they would be ignored by another rule in this section.)
```
legodude17: Move this paragraph down.

doc/files/package.json.md
```diff
@@ -170,12 +170,15 @@ npm also sets a top-level "maintainers" field with your npm user info.
...
-in subdirectories, which will keep files from being included, even
-if they would be picked up by the files array.  The `.npmignore` file
-works just like a `.gitignore`.
+You can also provide a `.npmignore` file in the root of your package or
+in subdirectories, which will keep files from being included. The `.npmignore` 
+file works just like a `.gitignore`. If there is a `.gitignore` 
+file, and `.npmignore` is missing, `.gitignore`'s contents will be used instead.
+
+Files included with the "package.json#files" field _cannot_ be excluded through
+`.npmignore` or `.gitignore`.
```
legodude17: Put the description of files down here.

legodude17 (changes requested):
Nice PR!

(Note that I am not a developer of `npm`, just a community member)

doc/files/package.json.md
jbcpollak: @legodude17 I am ok with the change you suggest, but was trying to minimize the documentation change under the theory that would be more acceptable to the core team. I'm happy to update if that is preferred.

legodude17: I would say it makes more sense to merge the `Files included with the "package.json#files" filed....` and the `The "files" field....` paragraphs.

iarna:
This is definitely an improvement (for future reference, you _can_ override `files` with a `.gitignore` or `.npmignore` in a _subdirectory_, but not at the top level—while this'd be good to document, I don't want to hold up this PR any longer. There's no need for the perfect, we can always do another pass.) 

I've merged  this as a0fdd1571. Thank you for your patience.

PR #15237, opened 2016-12-16
metrics: Don't launch if we're runnning from a sender process

iarna:
This saves our users from having almost launcher processes that are forever launching their own children before going away. As it is, once one of these has started it's very difficult to kill them short of rebooting. =/


iarna:
this landed as 51c393f in v4.1.1

PR #15400, opened 2017-01-05
gitignore: ignore npm-debug.log*

watilde:
Since `npm-debug.log` sometimes has numbers, we can use `*` which is same pattern with the `Node.gitignore`:
+ https://github.com/github/gitignore/blob/master/Node.gitignore#L4

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,5 @@
 *.swp
-npm-debug.log
+npm-debug.log*
 .nyc_output/
 /test/bin
 /test/output.log


iarna:
Wait, how does `npm-debug.log` get numbers after it? I think that only happens if you kill a process such that npm can't clean up after itself before exiting (eg with `kill -9` or by rebooting).

Either way, what I'd like to do is land a much older PR to move the `npm-debug.log` into the cache. I even have some hours to finish it up this week! =)

watilde:
Now I saw the commits! I'll close this as resolved then :)

PR #15235, opened 2016-12-15
doc: rewrite npm doctor docs

othiym23:
Kept most of @watilde's docs (cleaned up a little). Added a bunch more!

PR #15402, opened 2017-01-05
doctor: close the message brackets

watilde:
From https://github.com/npm/npm/pull/15235, what @Charlotteis said


iarna (approved):
Ah, good catch.

Issue #3559, opened 2013-06-14
npm config https proxy value not used for git+https dependencies


explunit:
**Steps to reproduce:**
1. do **not** have https_proxy environment variable set in your host OS
2. Set the https proxy using the `npm config set https-proxy` command
3. `npm install jsdoc@3.2.0-dev` (or any module that has git+https dependencies)

**Expected result:** the proxy configuration in npm config is passed to the spawned git process

**Actual result:** the git process does not use the configured proxy information and an error such as this occurs:

```
npm ERR! git fetch -a origin (https://github.com/jsdoc3/markdown-js.git) 
error: Failed connect to github.com:443; 
Connection timed out while accessing https://github.com/jsdoc3/markdown-js.git/info/refs?service=git-upload-pack
```

**Workaround:** set the https_proxy environment variable in your host OS instead of trying to use the `npm config set https-proxy` command

raptoria:
+1

othiym23:
It would be extremely helpful to have this set up automatically, I agree. The CLI team will get to this as part of our initiative to improve the CLI's integration with Git, but we may not get to it for a little while, so if somebody else wants to fix it first, we'd be happy to offer time and expertise on how all of this is wired together. Thanks for your time and patience!

natejgardner:
Any updates on this? This issue makes npm pretty difficult to use behind a corporate proxy.

muehan:
yes it would be very helpful. proxy makes life hard and missing configuration possabilities doesn't help

PR #15474, opened 2017-01-13
Use https-proxy config for spawned processes

burka:
Prefill HTTPS_PROXY and HTTP_PROXY from https-proxy and proxy if not yet set.
This avoids issues if you are behind a proxy while using 
 - curl (like during webdriver-manager update) 
 - git+https, see #3559

I took a look at all those require('child_process') for execFile/spawn/exec and found that setting process.env['HTTPS_PROXY'] is a easy way to catch all externally launched programs. 
Ìn case HTTPS_PROXY is already set should we issue some kind of warning?

### DIFF
--- a/lib/npm.js
+++ b/lib/npm.js
@@ -295,6 +295,13 @@
           umask: umask
         }
 
+        if (config.get('proxy') && !process.env['HTTP_PROXY']){
+          process.env['HTTP_PROXY'] = config.get('proxy')
+        }
+        if (config.get('https-proxy') && !process.env['HTTPS_PROXY']){
+          process.env['HTTPS_PROXY'] = config.get('https-proxy')
+        }
+
         var gp = Object.getOwnPropertyDescriptor(config, 'globalPrefix')
         Object.defineProperty(npm, 'globalPrefix', gp)
 


lib/npm.js
```diff
@@ -295,6 +295,13 @@
           umask: umask
         }
 
+        if (config.get('proxy') && !process.env['HTTP_PROXY']){
+          process.env['HTTP_PROXY'] = config.get('proxy')
```
iarna: I'm not entirely comfortable with tweaking the env globally like this. Particularly because this kind of thing makes refactoring to put config all in one place much harder. (And that's something we desperately need to do.)

I'd prefer that this either get added separately to `lib/utils/lifecycle.js` and `lib/utils/git.js` or put into a library that both those modules use.

Either way, it will need a test and a documentation update.

Docs would be `doc/misc/npm-scripts.md` and `doc/misc/npm-config.md` (where the proxy config options are discussed).

iarna (changes requested):

lib/npm.js
burka: Thanks for your feedback.

I would prefer only lib/utils/git.js too, but I had the impression that there are some cases where `git` is directly invoked  and it would miss the `curl` use-case.

The git.js only approach would at least make a test possible, whereas it will be almost impossible to test a general inclusion like I proposed. As there is a workaround (setting  HTTP_PROXY/HTTPS_PROXY) and thus I have no urgent need to fix it fast it might take a while.

iarna: There are no other cases where we run child processes, so if you do those two cases then you've covered everything.

Tests should be pretty easy regardless... set proxy config and check `npm run env` for changes made to `lib/utils/lifecycle.js`. Slightly trickier with `git`. My approach would be to provide a stub replacement for `git` that just prints out the value of `HTTP_PROXY`, then you can run something like `npm install git://…` and then look for the `HTTP_PROXY` message in the output.

iarna:
As this still needs tests and it's been over 6 months, I'm going to close this. Thank you for putting it together originally and if you get time to write tests in the future, please feel free to open a new pr with them.

PR #15481, opened 2017-01-14
API-based npm search command

zkat:
Yay!

So this PR makes it so the npm CLI actually hits the shiny new search endpoint! Default `npm search` on the main registry should be blazing fast now.

No. Seriously:

fastaf

The search system still falls back to cache or `/-/all`-based searches for registries that don't support this endpoint, and if you happen to have a cache while running offline, it'll use that cache with classic search.

## To-Do

Note: The PR is basically ready to review of the bulk of the changes, but below are a couple of items that are still pending.

- [x] unit tests for `esearch`
- [x] get rid of `gunzip-maybe` because it adds almost 2MB
- [x] Figure out what to do about `--searchlimit` when using `all-package-search`
- [x] unit tests for `all-package-search`
- [x] Check what negation is gonna be like
- [x] figure out whether/how to add sorting to esearch


bcoe:
@zkat this is amazing; I will read through the pull while OSSing over the next couple days. We should get you added to the npms.io slack too; so we can discuss adding `--searchexclude`.

CC: @satazor, @atduarte, how cool is this?!

lib/search/esearch.js
```diff
@@ -0,0 +1,65 @@
...
+var jsonstream = require('JSONStream')
+var ms = require('mississippi')
+var gunzip = require('gunzip-maybe')
+
+module.exports = esearch
+
+function esearch (opts) {
+  var stream = ms.through.obj()
+
+  mapToRegistry('-/v1/search', npm.config, function (er, uri, auth) {
```
bcoe: this warms my heart 👍

bcoe (approved):
this is looking pretty solid to me; I like the how you fallback to the old search.

One thing comes to mind, because browserify-zlib check in there fixture directory, we end up with some pretty odd artifacts in the bundled dependencies. e.g., this picture of a person:

https://github.com/npm/npm/blob/936896722f6ba18a90120c7fac475ddecfecd9cd/node_modules/gunzip-maybe/node_modules/browserify-zlib/test/fixtures/person.jpg

It might be worth opening an issue, or simply sending a pull-request to browserify-zlib that removes test files from the package.

satazor:
This is definitively cool! One question: does it support qualifiers? e.g.: `npm search author:sindresorhus`

@bcoe I guess you have permissions to invite @zkat to our slack.

zkat:
@satazor all the current search syntax is supported because we pretty much pass arguments directly to the endpoint. You can do `maintainer:zkat` just fine. :)

@bcoe any news on negation support? I think pretty much all the other questions are more things for me and @iarna to talk out. :)

atduarte:
Awesome! :D

bcoe:
:tada: :balloon: this is really exciting.

test/tap/search.js
```diff
@@ -358,6 +184,7 @@ function setup () {
 }
 
 function cleanup () {
+  server.done()
```
iarna: why is this in a different place than `server.close()`?

test/tap/search.esearch.js
```diff
@@ -0,0 +1,192 @@
+var common = require('../common-tap.js')
```
iarna: `search.esearch.js`? Not `search-esearch.js`?

iarna (approved):
This looks great! Can't wait to have it!

I had a couple of nits but nothing blocking.

lib/search/esearch.js
```diff
@@ -0,0 +1,64 @@
...
+  var q = buildQuery(opts)
+  npm.registry.request(uri + '?text=' + encodeURIComponent(q) + '&size=' + opts.limit, params, function (er, res) {
+    if (er) return cb(er)
+    log.silly('esearch', 'request stream opened, code:', res.statusCode)
+    // NOTE - The stream returned by `request` seems to be very persnickety
+    //        and this is almost a magic incantation to get it to work.
+    //        Modify how `res` is used here at your own risk.
+    var entryStream = ms.pipeline.obj(
+      res,
+      ms.through(function (chunk, enc, cb) {
```
zkat: No it's just literally that weird and there's something about using this particular type of passthrough that worked through an issue I was having. I didn't do the due diligence of figuring out what was happening back then (during the original rewrite, which this code is based on). So maybe I'll just have to do it but siiiiiggghhhh

test/tap/search.js
zkat: close shuts down the server. done does missing-request checks while letting us reuse the same server (so, it errors if we don't hit the mocked endpoint at all)

test/tap/search.esearch.js
zkat: That's just me being inconsistent. I've been using the former with pacote and cacache and it just feels nicer for hierarchies of tests in a single flat dir

lib/search/esearch.js
bsc1780: Lol to funny

PR #15492, opened 2017-01-15
ci: master is 7

watilde:
Use the same versions with `.travis.yml`.
+ https://github.com/npm/npm/blob/latest/.travis.yml

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -5,7 +5,7 @@ environment:
     # previous LTS is next most important
     - nodejs_version: "4"
     # then master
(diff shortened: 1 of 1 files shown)


zkat (approved):
Looks fine! Travis hasn't updated its [list of available versions](https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#Available-Versions) in the docs, but 7 does work.

Thanks, @watilde!

iarna:
Merged into `release-next` so this'll be out a week from Thursday. Thanks! =)

PR #15515, opened 2017-01-18
docs: update supported version number of Node.js

watilde:
Since v0.10 is not supported anymore, we can update the important part in readme


iarna (approved):
Good catch, thank you!

PR #15519, opened 2017-01-18
Respect PREFIX and DESTDIR in combination

### LINKED ISSUE #15517: PREFIX and DESTDIR cannot be used in combination

  - [x] npm is producing an incorrect install. (kinda)

#### What's going wrong?

npm currently treats DESTDIR and PREFIX as exclusive to each other, when they should be able to be used in combination. The use-case behind this is to produce, without requiring root permissions, an install, to, for example `/usr/local/`, residing inside some user-writable directory (e.g. `/tmp/`, `$HOME`).

Running from a clean git checkout of v4.1.2 (commit 58aa3b6da4):

```
% make install DESTDIR=/tmp/npm-install PREFIX=/usr/local
...
npm ERR! Linux 4.7.0-1-amd64
npm ERR! argv "/usr/bin/nodejs" "/home/rotty/src/_readonly/npm/cli.js" "install" "-g" "-f"
npm ERR! node v4.6.1
npm ERR! npm  v4.1.2
npm ERR! path /usr/local/lib/node_modules
npm ERR! code EACCES
npm ERR! errno -13
npm ERR! syscall mkdir

npm ERR! Error: EACCES: permission denied, mkdir '/usr/local/lib/node_modules'
npm ERR!     at Error (native)
npm ERR!  { [Error: EACCES: permission denied, mkdir '/usr/local/lib/node_modules']
npm ERR!   errno: -13,
npm ERR!   code: 'EACCES',
npm ERR!   syscall: 'mkdir',
npm ERR!   path: '/usr/local/lib/node_modules' }
npm ERR! 
npm ERR! Please try running this command again as root/Administrator.
```

This should result in an install under `/tmp/npm-install/usr/local`, with the "prefix" inside the installed files pointing to `/usr/local`, instead of ignoring `DESTDIR` and attempting to install to `/usr/local`.

The code causing this issue is in `lib/config/defaults.js:92`:

```js
  if (process.env.PREFIX) {
    globalPrefix = process.env.PREFIX
  } else if (process.platform === 'win32') {
    // c:\node\node.exe --> prefix=c:\node\
    globalPrefix = path.dirname(process.execPath)
  } else {
    // /usr/local/bin/node --> prefix=/usr/local
    globalPrefix = path.dirname(path.dirname(process.execPath))

    // destdir only is respected on Unix
    if (process.env.DESTDIR) {
      globalPrefix = path.join(process.env.DESTDIR, globalPrefix)
    }
  }
```

For more information about the DESTDIR and PREFIX conventions, see <https://www.freebsd.org/doc/en/books/porters-handbook/porting-prefix.html> or <https://www.gnu.org/prep/standards/html_node/DESTDIR.html>.

rotty:
Convention dictates that these two variables should be usable in
combination. Fixes #15517.

### DIFF
--- a/lib/config/defaults.js
+++ b/lib/config/defaults.js
@@ -97,11 +97,11 @@ Object.defineProperty(exports, 'defaults', {get: function () {
   } else {
     // /usr/local/bin/node --> prefix=/usr/local
     globalPrefix = path.dirname(path.dirname(process.execPath))
+  }
 
-    // destdir only is respected on Unix
-    if (process.env.DESTDIR) {
-      globalPrefix = path.join(process.env.DESTDIR, globalPrefix)
-    }
+  // destdir only is respected on Unix
+  if (process.platform !== 'win32' && process.env.DESTDIR) {
+    globalPrefix = path.join(process.env.DESTDIR, globalPrefix)
   }
 
   defaults = {


lib/config/defaults.js
```diff
@@ -97,11 +97,11 @@ Object.defineProperty(exports, 'defaults', {get: function () {
...
     // /usr/local/bin/node --> prefix=/usr/local
     globalPrefix = path.dirname(path.dirname(process.execPath))
+  }
 
-    // destdir only is respected on Unix
-    if (process.env.DESTDIR) {
-      globalPrefix = path.join(process.env.DESTDIR, globalPrefix)
-    }
+  // destdir only is respected on Unix
+  if (process.platform !== 'win32' && process.env.DESTDIR) {
```
iarna: This should use `lib/utils/is-windows.js`

iarna (changes requested):
Thanks for finding this! This needs one very minor tweak and a test.

lib/config/defaults.js
dangonzo: use ruby

iarna:
Looking at this again 7 months later, I see that the win32 test is done directly throughout that file, so I'm fine with not using `is-windows.js` but it would still need a test to land.

As it's been so long with no activity I'm going to close this, but we would be happy to take a new PR in future that includes a test for this functionality.

PR #15520, opened 2017-01-18
doc: add section about scopes only belonging to user/org

hzoo:
Fixes https://github.com/npm/docs/issues/795


iarna (approved):
LGTM!

Thank you!

PR #15545, opened 2017-01-22
doc: update Node.js download link

watilde:
Replace http://nodejs.org/download/ with https://nodejs.org/en/download/.


iarna (approved):
LGTM!

PR #15716, opened 2017-02-09
cleaner errors

iarna:
This takes the instruction to visit the issue tracker out of the error message except in the two scenarios where we know for a fact that it's an npm bug.

This adds logs for `cb() never called`. Previously no log file was produced.

This removes a bunch of bolierplate from the start of error messages, that hopefully will make them easier to parse for novices. All of the boilerplate is now being logged at `verbose`.

For instance, if you run `npm run nope` it will now output (in total):
```
npm ERR! missing script: nope

npm ERR!     /Users/rebecca/.npm/_logs/2017-02-09T01_18_59_931Z-debug.log
```

Now here's a question: Are we making things easier or harder by prefixing every line with `npm ERR!`? Would this be clearer if we didn't use `npmlog` to print out our errors like this?


zkat:
I vote for throwing away the `npm ERR!` bits. This seems like a nice first step towards a larger error-reporting overhaul.

lib/utils/error-handler.js
```diff
@@ -32,10 +31,15 @@ process.on('exit', function (code) {
...
-  else {
+  if (itWorked) {
+    log.info('ok')
+  } else {
     if (!cbCalled) {
       log.error('', 'cb() never called!')
+      console.error('')
+      log.error('', 'This is an error with npm itself. Please report this error at:')
+      log.error('', '    <https://github.com/npm/npm/issues>')
+      writeLogFile()
```
ceejbot: Oh hey, nice catch of an omission.

lib/utils/error-message.js
```diff
@@ -304,14 +304,6 @@ function errorMessage (er) {
...
       short.push(['', er.message || er])
-      detail.push([
-        '',
-        [
-          '',
-          'If you need help, you may report this error at:',
-          '    <https://github.com/npm/npm/issues>'
-        ].join('\n')
-      ])
       break
```
ceejbot: And the key omission. An end of an era, but it's time.

iarna:
Added https://github.com/npm/npm/pull/15716/commits/a592d98407e118ed11c16d718b8e0caa860f6818 which switches output from `log.error` to `console.error`.

iarna:
I'm going to hold off on this one, as there's one aspect of this change that we haven't yet discussed:

By changing error reporting from `log.error` to `console.error` it means that `--loglevel=silent` no longer suppresses the messages. We need to decide if that's a good thing or a bad thing.

legodude17:
I would say that this is a pretty good thing, because you can ignore everything except critical errors. The one problem would be build servers.

iarna:
Yeah, I'm not keen on `--loglevel=silent` actually hiding ordinary output OR error output. But it is a change in behavior (that breaks our tests!), so maybe we'll queue that up for `npm@5`.

lib/utils/error-message.js
legodude17: Maybe without that there won't be so many strangely opened issues.

iarna:
Merged into v4.4.0!

PR #15735, opened 2017-02-11
completion: handle colons (e.g. in script names) correctly

### LINKED ISSUE #11696: npm completion does not work with :
I have the following scripts `copy` `copy:all` `copy:js` `copy:css`

I have added the completion script.

If I type `npm run copy:<tab>` It completes to `npm run copy:copy:`

Also if I run `npm run copy:cs<tab>` it completes to `npm run copy:copy:css`

beyondcompute:
As this problem mentioned a few times on the internet including npm's very own Github issues. I thought I'd propose a fix.

Based on this answer: http://stackoverflow.com/questions/2805412/bash-completion-for-maven-escapes-colon/12495727#12495727

Does not modify `COMP_WORDBREAKS`. Uses `__ltrim_colon_completions`

Fixes #11696

P.S. There's also a possibility to implement a solution that does not rely on `_get_comp_words_by_ref` or `__ltrim_colon_completions`. For example, like this: http://willcode4beer.com/tips.jsp?set=tabMaven

### DIFF
--- a/lib/utils/completion.sh
+++ b/lib/utils/completion.sh
@@ -11,7 +11,7 @@ if type complete &>/dev/null; then
   _npm_completion () {
     local words cword
     if type _get_comp_words_by_ref &>/dev/null; then
-      _get_comp_words_by_ref -n = -n @ -w words -i cword
+      _get_comp_words_by_ref -n = -n @ -n : -w words -i cword
     else
       cword="$COMP_CWORD"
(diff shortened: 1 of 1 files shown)


iarna (approved):
Ordinarily I'd be asking for a test, but we don't actually have any facilities for testing our bash completions. 😭 

Thank you for this!

PR #15757, opened 2017-02-14
Default reporting metrics to default registry

iarna:
Previously we defaulted to `https://registry.npmjs.com`. With this patch, we default to whatever the value of the `registry` config is. (The default value for the `registry` config happens to be `https://registry.npmjs.com` so this means no change for most folks.)


iarna:
Released in v4.4.0

Issue #6725, opened 2014-11-17
warn on including package in both dependencies & devDependencies


smikes:
Spin off of https://github.com/npm/npm/issues/5403 which has a somewhat wider scope

> I discussed this a little with @isaacs; our consensus was that having the same packages in dependencies and devDependencies should be considered an error worth warning on (even in the absence of shrinkwrapping). It seems to me that what should happen is that when shrinkwrapping, npm should look to package.json only to see in which stanza a particular package belongs.

This may be relevant to development of multi-stage-install, so I am pulling it out as a separate issue /cc @iarna 

Feel free to close if this is already addressed in your plans; otherwise it should probably be tagged `feature-request`

goloroden:
Would love to see such a warning, too :-)

iarna:
The place to put this would be `lib/install/validate-tree.js` in `thenCheckTop`. You'd then compare `idealTree.package.dependencies` to `idealTree.package.devDependencies`.

othiym23:
Another thing to keep in mind is that this warning is only relevant for the current top-level `package.json` – warnings shouldn't happen for any transitive dependencies, or people will get _way_ infuriated by being given non-actionable warnings. We'd love to see this added, though, because it's kind of surprising that npm doesn't already do this.

TedYav:
On it 👍  Use NPM a lot and want to start contributing

TedYav:
Okay, I found the function and wrote code to issue warning.
Some questions:

- Should I treat same package but different versions as equivalent?
- I know @iarna suggested putting this in `thenCheckTop` but I'm thinking it might do well in its own function. Right now I factored it out in a function as shown below, but I can also add it to its own function called `thenCheckForDuplicates` and add it to the chain at the top of the file.
- I gave it an arbitrary error code. Is there a central list of error codes or should I just give it a sensible one?

Thanks! Pardon me for asking so many questions :) I'll write tests up and submit a PR when I hear back.

Here's the function I wrote right now that works:

```
function checkForDuplicateDependencies(idealTree){
  for(let pkg in idealTree.package.dependencies){
    if(pkg in idealTree.package.devDependencies){
      var warnObj = new Error('The package ' + pkg + ' is included as both a dev and production dependency.')
      warnObj.code = 'EDUPLICATEDEP'
      idealTree.warnings.push(warnObj)
    }
  }
}
```

ianchanning:
The warning is useful - but is there a guide about what commands you should run when you get this warning?

[...]

PR #15772, opened 2017-02-15
Created Warning Message for Duplicate Deps and Test for Issue: 6725

TedYav:
Original Issue: https://github.com/npm/npm/issues/6725

Decisions I made:

- Chose a pretty vanilla warning message. Can change if suggestions are made
- I opted to have it warn if dependencies are same version or different versions

Instead of including in `thenCheckTop` as @iarna  suggested, I opted to factor the duplicate checking to a separate function. I thought this kept things a little cleaner.

Wrote test as well. I could unit test the function in particular but I think this test verifies that it's properly created the warning.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -451,3 +451,4 @@ Dmitry Litvinchenko <karaliti@gmail.com>
 chocolateboy <chocolate@cpan.org>
 Henry Zhu <hi@henryzoo.com>
 Nate Goldman <ungoldman@gmail.com>
+Ted Yavuzkurt <hello@TedY.io>
--- a/lib/install/validate-tree.js
+++ b/lib/install/validate-tree.js
@@ -26,7 +26,8 @@ module.exports = function (idealTree, log, next) {
       ], done)
     }],
     [thenValidateAllPeerDeps, idealTree],
-    [thenCheckTop, idealTree]
+    [thenCheckTop, idealTree],
+    [thenCheckDuplicateDeps, idealTree]
   ], andFinishTracker(log, next))
 }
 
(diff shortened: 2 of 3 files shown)


TedYav:
Edit: fixed bug with `idealTree.package.dependencies` or `idealTree.package.devDependencies` being undefined for test `normalize-package-explode.js`

legodude17:
That does look like a config error...

iarna:
Those things quoted aren't actually errors. The first is the summary of the results and the second is... well, no problem at all.

But no, appveyor has never been passing. IIRC, the git tests don't pass on it (even though they pass on a real windows machine).

AUTHORS
```diff
@@ -451,3 +451,4 @@ Dmitry Litvinchenko <karaliti@gmail.com>
 chocolateboy <chocolate@cpan.org>
 Henry Zhu <hi@henryzoo.com>
 Nate Goldman <ungoldman@gmail.com>
+Ted Yavuzkurt <hello@TedY.io>
```
iarna: AUTHORS is machine generated (when we do releases). To use a name/email address other than the one on your git commits we add an entry to `.mailcap`.

lib/install/validate-tree.js
```diff
@@ -73,5 +74,31 @@ function thenCheckTop (idealTree, next) {
...
+  next()
+}
+
+// check if dependencies are duplicated between dependencies and
+// dev-dependencies. Issue warning if so.
+function thenCheckDuplicateDeps (idealTree, next) {
+  var shorterDepsList = idealTree.package.devDependencies
+  var longerDepsList = idealTree.package.dependencies
+  // might be undefined -- skip if so
+  if (!(idealTree.package.devDependencies === undefined || idealTree.package.dependencies === undefined)) {
```
iarna: I would prefer to just set defaults for the deplists and skip this check, eg:

`var shorterDepsList = idealTree.package.devDependencies || {}`

lib/install/validate-tree.js
```diff
@@ -73,5 +74,31 @@ function thenCheckTop (idealTree, next) {
...
+    // speed optimization --> iterate through shorter dependencies list
```
iarna: The longest possible list is going to have a couple hundred items here and this code is called once, so I don't think optimization is justified.

iarna (approved):
Thanks for putting this together! I'll be making a few minor tweaks (as discussed in the review and merging it.

AUTHORS
TedYav: Ah, okay thanks :) Wasn't sure!

lib/install/validate-tree.js
TedYav: Good idea—makes it simpler.

TedYav:
Okay, sounds good, will keep that in mind in future. Output I pasted from AppVeyor was just showing that I thought failing tests might be related to CouchDB not being running (thus causing them to time out or something). Didn't dig in too closely.

iarna:
Thank you again for this. It's released in v4.4.0!

PR #15773, opened 2017-02-15
docs(npm-scripts): added link to eggheadio video course

elijahmanor:
I've recently made an [egghead.io](https://egghead.io) course on [How to Use npm scripts as Your Build Tool](https://egghead.io/courses/how-to-use-npm-scripts-as-your-build-tool). The course is 52 minutes long and has 21 focused lessons.

Normally these courses start out free, but after about a week or so they get converted to a paid-only offering.

However, I've chatted with egghead.io and they are willing to keep the course **free** if I'm able to get a link to the course from npm's documentation.

### DIFF
--- a/doc/misc/npm-scripts.md
+++ b/doc/misc/npm-scripts.md
@@ -257,6 +257,11 @@ above.
   there is another option. The only valid use of `install` or `preinstall`
   scripts is for compilation which must be done on the target architecture.
 
+## VIDEO COURSE
+
+To see scripts in action you can watch the free [How to Use npm scripts as Your Build
+Tool](https://egghead.io/courses/how-to-use-npm-scripts-as-your-build-tool) course on [egghead.io](https://egghead.io) by [Elijah Manor](https://twitter.com/elijahmanor).
+
 ## SEE ALSO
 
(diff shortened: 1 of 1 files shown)


isaacs:
Hi, @elijahmanor.

Typically, we don't provide links to other services like this from the official npm documentation, just because it adds quite a bit of maintenance overhead which is outside of our control, and sends a confusing message about the level of the relationship between npm and the third party content provider.  For example, even if these video tutorials are of very high quality, it sets a precedent where we're now vouching for the quality of npm tutorials on egghead.io _in general_, sending the message that we're connected in some official capacity, and so on.  We just don't have bandwidth to manage that for all the services that would like to do something like this.

I'd be happy to have us make some noise about this tutorial series in a less permanent/committal way, though.  It would be a good thing to note in [our weekly newsletter](https://www.npmjs.com/npm-weekly).

elijahmanor:
Hi @isaacs 

I totally understand where you are coming from. I can see how including egghead.io could open the flood gates for other 3rd party requests.

That'd be great if you could include something in your weekly newsletter. Thank you for your time and your efforts.

Issue #11473, opened 2016-02-08
freshness self-check


othiym23:
In https://github.com/npm/npm/issues/10800#issuecomment-181439557, @rosskevin says:

> It's easy to forget to update npm, I wish it had a reminder for me to keep it up to date when it falls behind. Bower does this and I find it quite useful

Homebrew does something like this as well.

Given that it's only going to get more important over time to have an up-to-date npm (at some point, a new Node release will become available that will break pretty much every old release of npm), and given how much of our support traffic comes from users stuck on antique versions of npm, the sooner the CLI team implements something like this, the sooner it will start to become useful.

KenanY:
ref. previous discussion https://github.com/npm/npm/issues/5495

othiym23:
Thanks, Kenan. I'm going to close this issue in favor of #5495, which I (obviously?) forgot was around.

KenanY:
No problem :)

othiym23:
We're going to reopen this issue and close #5495, because the sense of the CLI team is that _notifying_ people that their CLI is out of date is useful, but _auto-updating_ the CLI is more dangerous than the team is comfortable with.

PR #15774, opened 2017-02-15
bin: use update-notifier to tell users about updates

ceejbot:
The cli bin script now uses the [update-notifier](https://www.npmjs.com/package/update-notifier) package to inform users of the CLI that an update is available. The current usage uses the default notification message with the default check interval of one week. The notification display is deferred until after the process exits.


KenanY:
This would fix https://github.com/npm/npm/issues/11473 🎉

iarna (approved):
Bundled deps are good.

iarna:
Merged into v4.4.0!

PR #15777, opened 2017-02-16
config: change apply to call

notarseniy:
Since it doesn't have to apply an array of arguments, we change `apply` to more performant `call`.

### DIFF
--- a/lib/config/core.js
+++ b/lib/config/core.js
@@ -208,7 +208,7 @@ inherits(Conf, CC)
 function Conf (base) {
   if (!(this instanceof Conf)) return new Conf(base)
 
-  CC.apply(this)
+  CC.call(this)
 
   if (base) {
     if (base instanceof Conf) {


notarseniy:
yeah, really minor change! 😅

benchmark:
```javascript
function work(a, b, c) {
  // do something
}

for (var j = 0; j < 4; j++) {
  console.time('apply-ing');
  for (var i = 0; i < 10000000; i++) {
    work.apply(this);
  }
  console.timeEnd('apply-ing');

  console.time('call-ing');
  for (var i = 0; i < 10000000; i++) {
    work.call(this);
  }
  console.timeEnd('call-ing');
}
```
results on my MBPr early 2015:
```javascript
apply-ing: 151.142ms
call-ing: 113.402ms
apply-ing: 141.471ms
call-ing: 115.478ms
apply-ing: 148.534ms
call-ing: 111.212ms
apply-ing: 137.648ms
call-ing: 109.652ms
```

iarna:
I'm not, as a rule, a fan of micro-optimizations like this outside of performance critical code, but in this case this had no impact on code legibility so it's fine. This was released in v4.4.0.

PR #15824, opened 2017-02-22
fix: we were not populating version # in new search results

bcoe:
the new `/-/v1/search` API returns the version number on the key `.version`, we were accidentally not populating this in the UI.

### DIFF
--- a/lib/search/format-package-stream.js
+++ b/lib/search/format-package-stream.js
@@ -150,8 +150,6 @@ function highlightSearchTerms (str, terms) {
 
 function normalizePackage (data, opts) {
   opts = opts || {}
-  var version = Object.keys(data.versions || {})[0] || []
-  if (data.version) version = data.version
   return {
     name: data.name,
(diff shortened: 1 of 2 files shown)


lib/search/format-package-stream.js
```diff
@@ -150,6 +150,8 @@ function highlightSearchTerms (str, terms) {
 
 function normalizePackage (data, opts) {
   opts = opts || {}
+  var version = Object.keys(data.versions || {})[0] || []
+  if (data.version) version = data.version
```
zkat: These two lines are basically unnecessary -- both `esearch` and `all-package-search` return a consistent data format (specified at the top of this file). The patch itself just needs to be `version: data.version`, unless I brain farted hard.

zkat (changes requested):
Thanks for putting this up, Ben! I I pretty much just have one comment to make here :)

zkat (approved):
LGTM :D

PR #15833, opened 2017-02-22
Mention the 24-hour restriction on unpublish

carols10cents:
As detailed in http://blog.npmjs.org/post/141905368000/changes-to-npms-unpublish-policy

Looked for this here and didn't find it, but did find the blog post eventually :)

### DIFF
--- a/doc/cli/npm-unpublish.md
+++ b/doc/cli/npm-unpublish.md
@@ -27,6 +27,10 @@ Even if a package version is unpublished, that specific name and
 version combination can never be reused.  In order to publish the
 package again, a new version number must be used.
 
+Unpublish is only allowed with versions published in the last 24
+hours. If you are trying to unpublish a version published longer
+ago than that, contact support@npmjs.com.
+
 The scope is optional and follows the usual rules for `npm-scope(7)`.
 
 ## SEE ALSO
--- a/doc/cli/npm-unpublish.md
+++ b/doc/cli/npm-unpublish.md
@@ -27,9 +27,10 @@ Even if a package version is unpublished, that specific name and
 version combination can never be reused.  In order to publish the
 package again, a new version number must be used.
 
-Unpublish is only allowed with versions published in the last 24
-hours. If you are trying to unpublish a version published longer
-ago than that, contact support@npmjs.com.
+With the default registry (`registry.npmjs.org`), unpublish is
+only allowed with versions published in the last 24 hours. If you
+are trying to unpublish a version published longer ago than that,
+contact support@npmjs.com.
 
 The scope is optional and follows the usual rules for `npm-scope(7)`.
 


doc/cli/npm-unpublish.md
```diff
@@ -27,6 +27,10 @@ Even if a package version is unpublished, that specific name and
 version combination can never be reused.  In order to publish the
 package again, a new version number must be used.
 
+Unpublish is only allowed with versions published in the last 24
+hours. If you are trying to unpublish a version published longer
+ago than that, contact support@npmjs.com.
```
iarna: The `npm` client doesn't only run against `registry.npmjs.org` and private registries may have different policies when it comes to unpublishing. I think your addition here is good, but it should have the caveat that this only applies if you're using the default registry.

iarna (dismissed):

doc/cli/npm-unpublish.md
bsc1780: Do it i gave permissions.  And help me get this published please . your contribute will be valued but check on this Node.js v7.6.0 Documentation
Index | View on single page | View as JSON

Table of Contents

Child Process
Asynchronous Process Creation
Spawning .bat and .cmd files on Windows
child_process.exec(command[, options][, callback])
child_process.execFile(file[, args][, options][, callback])
child_process.fork(modulePath[, args][, options])
child_process.spawn(command[, args][, options])
options.detached
options.stdio
Synchronous Process Creation
child_process.execFileSync(file[, args][, options])
child_process.execSync(command[, options])
child_process.spawnSync(command[, args][, options])
Class: ChildProcess
Event: 'close'
Event: 'disconnect'
Event: 'error'
Event: 'exit'
Event: 'message'
child.channel
child.connected
child.disconnect()
child.kill([signal])
child.pid
child.send(message[, sendHandle[, options]][, callback])
Example: sending a server object
Example: sending a socket object
child.stderr
child.stdin
child.stdio
child.stdout
maxBuffer and Unicode
Child Process#

Stability: 2 - Stable
The child_process module provides the ability to spawn child processes in a manner that is similar, but not identical, to popen(3). This capability is primarily provided by the child_process.spawn() function:

const spawn = require('child_process').spawn;
const ls = spawn('ls', ['-lh', '/usr']);

ls.stdout.on('data', (data) => {
  console.log(`stdout: ${data}`);
});

ls.stderr.on('data', (data) => {
  console.log(`stderr: ${data}`);
});

ls.on('close', (code) => {
  console.log(`child process exited with code ${code}`);
});
By default, pipes for stdin, stdout and stderr are established between the parent Node.js process and the spawned child. It is possible to stream data through these pipes in a non-blocking way. Note, however, that some programs use line-buffered I/O internally. While that does not affect Node.js, it can mean that data sent to the child process may not be immediately consumed.

The child_process.spawn() method spawns the child process asynchronously, without blocking the Node.js event loop. The child_process.spawnSync() function provides equivalent functionality in a synchronous manner that blocks the event loop until the spawned process either exits or is terminated.

For convenience, the child_process module provides a handful of synchronous and asynchronous alternatives to child_process.spawn() and child_process.spawnSync(). Note that each of these alternatives are implemented on top of child_process.spawn() or child_process.spawnSync().

child_process.exec(): spawns a shell and runs a command within that shell, passing the stdout and stderr to a callback function when complete.
child_process.execFile(): similar to child_process.exec() except that it spawns the command directly without first spawning a shell.
child_process.fork(): spawns a new Node.js process and invokes a specified module with an IPC communication channel established that allows sending messages between parent and child.
child_process.execSync(): a synchronous version of child_process.exec() that will block the Node.js event loop.
child_process.execFileSync(): a synchronous version of child_process.execFile() that will block the Node.js event loop.
For certain use cases, such as automating shell scripts, the synchronous counterparts may be more convenient. In many cases, however, the synchronous methods can have significant impact on performance due to stalling the event loop while spawned processes complete.

Asynchronous Process Creation#
The child_process.spawn(), child_process.fork(), child_process.exec(), and child_process.execFile() methods all follow the idiomatic asynchronous programming pattern typical of other Node.js APIs.

Each of the methods returns a ChildProcess instance. These objects implement the Node.js EventEmitter API, allowing the parent process to register listener functions that are called when certain events occur during the life cycle of the child process.

The child_process.exec() and child_process.execFile() methods additionally allow for an optional

legodude17: Like, prefix it with "With the default registry (`registry.npmjs.org`),".

legodude17:
@bsc1780 please stop leaving unhelpful comments.

carols10cents:
@iarna Done! I clarified with the text that @legodude17 suggested :)

iarna:
@bsc1780 is now blocked. Github won't let me remove the comments however.

legodude17:
@iarna Thank you for that.

zkat (approved):
Looks like this is good now 👍

PR #15888, opened 2017-02-28
ls: stop flattening output, show the actual tree

iarna:
This is the infamous `debfsify` branch, finally coming home to land. It only needed a few tests tweaked that relied on specific `npm ls` output. Previously `npm` would trim and reshape the output tree to try to make it more concise. This could be confusing because it didn't represent what was actually there. Further, when modules were deduped (that is, were a shared transitive dependency) there was no way to see this was happening.

With this patch you now get output like:
```
npm@4.4.0 /Users/rebecca/code/npm-latest
├── abbrev@1.1.0
├── ansi-regex@2.1.1
├── ansicolors@0.3.2
├── ansistyles@0.1.3
├── aproba@1.1.1
├── archy@1.0.0
├── asap@2.0.5
├── chownr@1.0.1
├─┬ cmd-shim@2.0.2
│ ├── graceful-fs@4.1.11 deduped
│ └── mkdirp@0.5.1 deduped
├─┬ columnify@1.5.4
│ ├── strip-ansi@3.0.1 deduped
│ └─┬ wcwidth@1.0.0
│   └─┬ defaults@1.0.3
│     └── clone@1.0.2
```

### DIFF
--- a/lib/ls.js
+++ b/lib/ls.js
@@ -15,6 +15,8 @@ var semver = require('semver')
 var color = require('ansicolors')
 var npa = require('npm-package-arg')
 var iferr = require('iferr')
+var sortedObject = require('sorted-object')
+var extend = Object.assign || require('util')._extend
 var npm = require('./npm.js')
 var mutateIntoLogicalTree = require('./install/mutate-into-logical-tree.js')
 var recalculateMetadata = require('./install/deps.js').recalculateMetadata
@@ -76,8 +78,8 @@ var lsFromTree = ls.fromTree = function (dir, physicalTree, args, silent, cb) {
 
   pruneNestedExtraneous(data)
   filterByEnv(data)
-  var bfs = filterFound(bfsify(data), args)
-  var lite = getLite(bfs)
+  var unlooped = filterFound(unloop(data), args)
+  var lite = getLite(unlooped)
 
   if (silent) return cb(null, data, lite)
 
@@ -86,7 +88,7 @@ var lsFromTree = ls.fromTree = function (dir, physicalTree, args, silent, cb) {
   var out
   if (json) {
     var seen = []
-    var d = long ? bfs : lite
+    var d = long ? unlooped : lite
     // the raw data can be circular
     out = JSON.stringify(d, function (k, o) {
       if (typeof o === 'object') {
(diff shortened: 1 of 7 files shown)


test/tap/peer-deps-toplevel.js
```diff
@@ -59,57 +72,65 @@ var expected = {
...
+    t.done()
+  })
+})
 
-        npm.commands.ls([], true, function (err, _, results) {
-          t.ifError(err, 'listed tree without problems')
+test('installs the peer dependency directory structure', function (t) {
+  common.npm(['install'], conf, function (err, code, stdout, stderr) {
+    if (err) throw err
+    t.is(code, 0, 'install ran ok even w/ missing peeer deps')
```
zkat: `peeeeeeeeeeeeeeeeer deps`?

lib/ls.js
```diff
@@ -444,10 +466,11 @@ function getExtras (data) {
 }
 
 function makeParseable (data, long, dir, depth, parent, d) {
+  if (data._deduped) return []
   depth = depth || 0
   if (depth > npm.config.get('depth')) return [ makeParseable_(data, long, dir, depth, parent, d) ]
   return [ makeParseable_(data, long, dir, depth, parent, d) ]
-  .concat(Object.keys(data.dependencies || {})
+ .concat(Object.keys(data.dependencies || {})
```
zkat: isn't this indent weird?

zkat (approved):
LGTM 🐑

lib/ls.js
legodude17: Yeah, shouldn't it be indented two spaces? Instead of outdented once.

iarna: sure looks like it, I wonder why `standard` isn't yelling…

test/tap/peer-deps-toplevel.js
iarna: yuuuup, we still have tests for them

lib/ls.js
legodude17: Maybe standard doesn't enforce spacing in method chains?

iarna: Maybe, it's awfully picky otherwise =p

legodude17: I read through the rules and there was no mention of method chains...
I don't know `standard` well enough to call whether it does or doesn't.

lib/ls.js
```diff
@@ -354,6 +367,14 @@ function makeArchy_ (data, long, dir, depth, parent, d) {
   }
   if (data.link) out.label += ' -> ' + data.link
 
+  if (data._deduped) {
+    if (npm.color) {
+      out.label += ' ' + color.brightBlack('deduped')
+    } else {
+      out.label += ' deduped'
+    }
```
legodude17: Maybe this should be in parens? Like ` (deduped)`.

legodude17:
Did this get merged?

zkat:
yes

legodude17:
🎆

PR #15900, opened 2017-02-28
RFC: file: specifier changes

iarna:
Easy reading link for the new [specification](https://github.com/npm/npm/blob/link-specifier/doc/spec/file-specifiers.md). This also introduces written specifications for npm features, something we intend to extend out to all of npm's activity.

# The Problem We're Solving

File type dependency specifiers are confusing. How they interact across npm's many varied commands (`npm install --save`, `npm update`, `npm outdated`, `npm shrinkwrap`) has never been defined in one place. Most of the behaviors "happened by default". That is, the minimum implementation was done to make them install ok and the rest was left as a side effect.

Currently `npm update` and `npm outdated` simply don't do anything for local dependencies. If you've updated the source you have to manually reinstall to get a copy of it. How `npm shrinkwrap` saves local dependencies to `npm-shrinkwrap.json` and how it resolves them has varied over time as well.

## Root Causes

When local dependencies were added, we didn't have any process around defining behavior or ensuring that all use cases were specified.

## Proposal

In addition to actually specifying behavior (often simply writing down what
things currently do), we propose in important breaking change:

* `file:`-type specifiers that refer to directories will be soft deprecated, their behavior being identical to the new `link:` specifier and their existence becoming a footnote in the documentation.
* A new specifer type, `link:`, will be introduced for linking local directories.  For the duration of `npm@5, `file:` specifiers that refer to directories will be treated identically to `link:` specifiers.
* `link:` specifiers will result in a symlink or junction made from the specifier path into your `node_modules`.  On Windows try a junction and if that fails, try a symlink.  If both fail, the error from the junction should be used.

This RFC essentially brings [linklocal](https://github.com/timoxley/linklocal)'s bevior into core.

## RISKS

* Diagnostic information changes need to be handle delicately in order to not increase user confusion.
* Some users may be unhappy with the changes to `file:` semantics.


timoxley:
Nice! Why new `link:` vs just using `file:`?

iarna:
The reason for introducing the new specifier are:

1. Have a specifier that immediately tells someone reading a `package.json` what it will do.
2. Not have fights over what a valid `file:` URL looks like.

Originally there was a plan to actually hard deprecate the `file:` specifier and remove them, but we decided that was pointlessly disruptive. Instead `file:` will be an alias for `link:`.

timoxley:
@iarna Sounds reasonable 👍 

Have you considered the impact of having different module resolution paths when using symlinks? 
I currently only use symlinks in development, while in production I let the current `file:` behaviour do its thing. Without the symlinks npm is able to move many shared packages to the top level, decreasing overall install time. This could have significant impact on deploy times if local dependencies share a lot of binary dependencies.

iarna:
@timoxley It's a fair question. As it's specced right now it would do that. Maybe, if the linked module is inside the top level install root we should flatten its deps down, instead of keeping them inside the linked module. The module loader would still be able to pick them up and it'd not increase disk usage.

timoxley:
> if the linked module is inside the top level install root we should flatten its deps down

@iarna that sounds good. Being aware of where symlinks resolve to perhaps could be considered a good general improvement to the dedupe/flattening algorithm.

doc/spec/file-and-link-specifiers.md
```diff
@@ -0,0 +1,159 @@
...
+
+* A valid `link:` specifier is either an absolute path, (eg
+  `/path/to/thing`, `d:\path\to\thing`) or a relative path, eg
+  `../path/to/thing`, `path\to\subdir`).  `link:` specifiers do not
+  distinguish between forward and backslashes.  That is, we consider `link:a\b`
+  and `link:a/b` to be the same value.
+* Attempting to install a specifer that has a windows drive letter will
+  produce an error on non-Windows systems.
+
+* A valid `file:` specifier points is:
```
thefourtheye: Did you mean "A valid `file:` specifier also is" like in the next item? I am finding it difficult to understand this line.

doc/spec/file-and-link-specifiers.md
```diff
@@ -0,0 +1,159 @@
...
+
+On the command line, plain paths are allowed.  These paths can be ambiguous
+as they could be a path, a plain package name or a github shortcut.  This
+ambiguity is resolved by checking to see if either a directory exists that
+contains a `package.json`.  If either is the case then the specifier is a
+link specifier, otherwise it's a registry or github specifier.
+
+Historically, these ambiguous specifiers were also allowed in the
+`package.json`.  Starting in `npm@5` using an ambiguous specifier in your
+shrinkwrap will be depricated and will warn.  In `npm@6` it will be an
```
mantoni: Typo: depricated -> deprecated

mantoni:
This is great! I would like to know if you considered supporting the current `npm ln` use-case? I'm thinking of `link:module-name` to resolve to the "global" link which in turn links to the actual module directory. Contrary to the relative link paths, this wouldn't enforce a specific directory structure. What do you think?

doc/spec/file-and-link-specifiers.md
legodude17: I would say it should be "Also, a valid `file:` specifier is:"

doc/spec/file-and-link-specifiers.md
```diff
@@ -0,0 +1,159 @@
...
+  `../path/to/thing`, `path\to\subdir`).  `link:` specifiers do not
+  distinguish between forward and backslashes.  That is, we consider `link:a\b`
+  and `link:a/b` to be the same value.
+* Attempting to install a specifer that has a windows drive letter will
+  produce an error on non-Windows systems.
+
+* A valid `file:` specifier points is:
+  * a valid package file. That is, a `.tar`, `.tar.gz` or `.tgz` containing
+  `<dir>/package.json`.
+  * OR, a directory that contains a `package.json`
```
legodude17: Should that 'OR' be removed or should all of these start with 'OR'? Consistency.

doc/spec/file-and-link-specifiers.md
```diff
@@ -0,0 +1,159 @@
...
+* And a valid `file:` specifier also is:
```
legodude17: Shouldn't this not be a bullet? Or just remove it.

doc/spec/file-and-link-specifiers.md
```diff
@@ -0,0 +1,159 @@
...
+
+File-type specifiers will necessarily not do anything for `fetch` and
+`extract` phases.
+
+The symlink should be created during the `finalize` phase
+
+The `preinstall` for file-type specifiers MUST be run AFTER the
+`finalize` phase as the symlink may be a relative path reaching outside the
+current project root and a symlink that resolves in `.staging` won't resolve
+in the package's final resting place.
```
legodude17: Would the `preinstall` for a package depended on via `link:` be run from it's original folder, or somewhere else?

doc/spec/file-and-link-specifiers.md
```diff
@@ -0,0 +1,159 @@
...
+
+```
+example-package@1.0.0 /path/to/example-package
+└── a → Copied from: link:../a
+```
+
+```
+example-package@1.0.0 /path/to/example-package
++-- a -> Copied from: link:../a
+```
```
legodude17: Again, is this repetition intentional?

doc/spec/file-and-link-specifiers.md
```diff
@@ -0,0 +1,159 @@
...
+
+```
+example-package@1.0.0 /path/to/example-package
+└── a → link:../a
+```
+
+```
+example-package@1.0.0 /path/to/example-package
++-- a -> link:../a
+```
```
legodude17: Why are there two parts with the same thing? Is that intentional?

legodude17 (commented):
This seems like it would really help a lot of people. 🎆

iarna:
So after some discussion in air between @zkat and I, we're of a consensus that we should go ahead with just `file:` and not add a new specifier. I'll update the RFC as appropriate.

doc/spec/file-and-link-specifiers.md
iarna: It should. It's part of the top level list.

doc/spec/file-and-link-specifiers.md
iarna: There are only two of them and the OR separates the two. So I'm not sure what you mean by "all of these".

doc/spec/file-and-link-specifiers.md
iarna: Basically this is mean trying to not screw people who are using `pwd` (which would be relative to where your package root's node_modules).

Folks who are using `cwd` would always get the same answer (the destination of the link).

doc/spec/file-and-link-specifiers.md
```diff
@@ -0,0 +1,156 @@
...
+#### File type specifiers pointing at tarballs
+
+File-type specifiers pointing at a `.tgz` or `.tar.gz or `.tar` file will
+install it as a package file in the same way we would a remote tarball.  The
+checksum of the package file should be recorded so that we can check for updates.
+
+#### File type specifers pointing at directories
+
+File-type specifiers that point at directories will necessarily not do
+anything for `fetch` and `extract` phases.
```
zkat: How do you feel about pacote doing this automagically on `pacote.extract`?

doc/spec/file-and-link-specifiers.md
```diff
@@ -0,0 +1,156 @@
...
+
+### Dependency Specifiers
+
+* A `file:` specifier is either an absolute path (eg `/path/to/thing`, `d:\path\to\thing`):
+  * An absolute `file:///absolute/path` with any number of leading slashes
+    being treated as a single slash. That is, `file:/foo/bar` and
+    `file:///foo/bar` reference the same package.
+* … or a relative path (eg `../path/to/thing`, `path\to\subdir`).  Leading
+  slashes on a file specifier will be removed, that is 'file://../foo/bar`
+  references the same package as same as `file:../foo/bar`.  The latter is
```
zkat: Can you add a quick clarification that `file://foo/bar` is considered relative, too? At least that's what I assume from reading this. It's a weird, sandwichy corner case, but it's probably worth actually _specifying_:

* `file:///foo` -> Absolute `/foo`
* `file://foo` -> Relative, same as `./foo`
* `file:/foo` -> Absolute `/foo`

doc/spec/file-and-link-specifiers.md
```diff
@@ -0,0 +1,156 @@
...
+  considered canonical.
+* Attempting to install a specifer that has a windows drive letter will
+  produce an error on non-Windows systems.
+* A valid `file:` specifier points is:
+  * a valid package file. That is, a `.tar`, `.tar.gz` or `.tgz` containing
```
zkat: We should make sure the error message from this is VERY CLEAR that we determine tarballness based on file extension. I know we talked about this a lot, but I still want to avoid linuxy/unixy folks pulling their hair out if they, say, try to install a tarball they downloaded from a webservice that does not specify a filename and they end up with an unsuffixed filename that is still technically a tarball.

Or, for example, if someone tries to `npm install ~/.npm/.pacote/content/deadbeef`, which do not have `.tgz` suffixes.

doc/spec/file-and-link-specifiers.md
```diff
@@ -0,0 +1,156 @@
...
+and they'd step on each other any time you ran an install in one or the other.
+
+NOTE: This should not have an effect on shrinkwrapping of other sorts of
+shrinkwrapped packages.
+
+### Installation
+
+#### File type specifiers pointing at tarballs
+
+File-type specifiers pointing at a `.tgz` or `.tar.gz or `.tar` file will
```
zkat: missing backtick (`) here after .tar.gz

doc/spec/file-and-link-specifiers.md
```diff
@@ -0,0 +1,156 @@
...
+You would see:
+
+```
+example-package@1.0.0 /path/to/example-package
+└── a → file:../a
+```
+
+```
+example-package@1.0.0 /path/to/example-package
++-- a -> file:../a
```
legodude17: Still, is this repetition intentional?

doc/spec/file-and-link-specifiers.md
```diff
@@ -0,0 +1,156 @@
...
+package came from, not what version happened to be in that path.
+
+### Outdated
+
+Local specifiers should only show up in `npm outdated` if they're missing
+and when they do, they should be reported as:
+
+```
+Package  Current  Wanted  Latest  Location
+a        MISSING   LOCAL   LOCAL  example-package
```
legodude17: I would suggest that it should also log where it is looking for the package in.

doc/spec/file-and-link-specifiers.md
```diff
@@ -0,0 +1,156 @@
...
+The `preinstall` for file-type specifiers MUST be run AFTER the
+`finalize` phase as the symlink may be a relative path reaching outside the
+current project root and a symlink that resolves in `.staging` won't resolve
+in the package's final resting place.
+
+If the module is inside the package root that we're running the install for then
+dependencies of the linked package will be hoisted to the top level as usual.
+
+If the module is outside the package root then dependencies will be installed inside
+the linked module's `node_modules` folder.
```
zkat: Is this going to be done as a separate npm run within that project folder (possibly a subprocess?) or will we build a single ideal tree with some special semantics, and do it in a single npm run? This seems trickier to me than what this single sentence describes. 🤔

doc/spec/file-and-link-specifiers.md
zkat: oh god `Location` is such a horrible, confusing name when we factor this in. I really bloody wish we had a different name for that column.

zkat: Additionally, I wish we had somewhere to put more information about the package getting installed, such as `from` or something, because the package name by itself doesn't give us much. But that's something for a separate RFC.

zkat (changes requested):
pretty excited about this. Thanks for writing the spec up <3

doc/spec/file-and-link-specifiers.md
iarna: Yup, I'd love to change that too

doc/spec/file-and-link-specifiers.md
iarna: Hrm. I don't think pacote has quite enough information currently to resolve this sort of thing. The specifier and the target destination aren't enough. You also need to know the location of the module that required it, because the specifier is relative to THAT.

doc/spec/file-and-link-specifiers.md
iarna: Nah, single thing, it doesn't even require much in the way of special semantics. It's basically the same logic as how we define "global" mode currently.

doc/spec/file-and-link-specifiers.md
zkat: Can/should we add this to `realize-package-specifier`? We use `where` there already.

doc/spec/file-and-link-specifiers.md
zkat: what repetition? directories and package names can be different: `a -> file:../b` is perfectly valid

doc/spec/file-and-link-specifiers.md
iarna: Number of leading slashes changes nothing:

file:///foo -> /foo
file://foo -> /foo
file:/foo -> /foo
With no leading slashes, relative paths are evident:

file:foo -> ./foo
file:../ -> ../foo

I'm proposing that if you have ../ as your first path element then we ignore any leading slashes:

file:/../foo -> ../foo
file://../foo -> ../foo
file:///../foo -> ../foo

I think this would eliminate whole classes of likely errors, particularly around users who are trying to make these things work like URLs.

doc/spec/file-and-link-specifiers.md
iarna: One of these is with unicode enabled, one without. I wanted to specify both.

doc/spec/file-and-link-specifiers.md
iarna: As zkat says below, changing outdated to be more useful is really a separate RFC.

doc/spec/file-and-link-specifiers.md
iarna: Maybe, I mean, we might actually already be resolving come to think. It just requires some end-to-end care.

doc/spec/file-and-link-specifiers.md
legodude17: Oh. Sorry. I did't notice that.

doc/spec/file-and-link-specifiers.md
legodude17: Ok, cool. Sounds like a really good thing.

legodude17: Wait, I am confused. So `package` is the name, and `Location` is the path on disk? Or is it the path to it through the dependency chain? Or is it the package that depended on this?

zkat: @legodude17 `npm help outdated` will answer your questions here.

legodude17: @zkat thanks for the tip. You are right that `Location` is a very confusing name for that. Maybe `Parent`?

zkat: @legodude17 we'll talk about that once there's an RFC for it. Talking about it in here is bound to get lost in oblivion (and is offtopic for this PR)

legodude17: Sounds like a good idea. Do you have a estimated time for this?

iarna: Nope, not part of the current schedule. I have on my todo-list writing up an up-to-date "state of NPM" blog post to bring everyone up-to-date on our goals and plans and expectations.

mlucool:
Using symlinks with node, requires a flag set for it to work correctly: https://github.com/nodejs/node/pull/8749#pullrequestreview-1388063. Is the intent to change how node handles symlinks also?

One nice feature about how file works today is for peer dependencies. Due to the copy instead of a symlink, peers with global singletons work as expected. When this moves to symlinks only, the resolution will make certain cases impossible to resolve. Example:
* Module A is a react component and has react as a `peerDependency` and `devDependency`
* While working on A, I want to test this locally in Module B , so I use `file:/path/to/a`.
  * Today it will do the right thing and share the global version of react as it is copied. The dependency resolution won't include a second react.
  * If instead we did a symlink, this would not work because React is installed in in A's `node_modules` and would not be the same instance as in as B's `node_modules`. 

Let me know if that example needs more clarification or if I have misunderstood RFC.

zkat:
@mlucool my understanding of this case is that it would continue to work so long as Module A is anywhere inside the directory structure for B when you link it.

That is:

```
module-b
\
 | - module-a
 |   \
 |     node_modules
 \
  node_modules
  \
    react
```

`module-a` will have its `react` hoisted to the top. It's worth being more explicit about how we handle peerDeps in this situation. I hopefully didn't misunderstand the RFC in this case.

@iarna it might be worth specifying the logic for `peerDependencies` and `devDependencies` for these linked submodules in the RFC, cause I don't think I'm that clear on it rn.

mlucool:
@zkat In my case A is outside of B always. We do not have a monorepo. Still [this](https://github.com/npm/npm/blob/43db0527310770647f3e9d0b0d76dce53e767de6/doc/spec/file-and-link-specifiers.md#installation) seems to imply as long as it is note `tar` it'll use symlinks.

My current hack (which works quite well) is to create a symlink like util that does:
1. Uninstall A
2. Install via file:/path/to/a
3. Start a file watcher for all files in A's `package.json:files`
4. Start a file watcher for A's package.json to start from 1 again

Now for any change I make it appears nearly instantly. Clearly this is a hack, but it maintains that I can separate components across repos and not worry about dependency issues.

zkat:
@mlucool if you're not going to use a monorepo structure, you could literally just run `npm pack` and _then_ `npm install file://path/to/module-a/module-a-1.2.3.tgz`, which will have the same behavior as `npm i file://path/to/module-a` currently does. We're not changing local tarballs. It's a single extra step, but also something the CLI was already doing (it had to, in order to install the package)

zkat:
and yes, we will _always_ create symlinks for _directory_ specs. The difference is that we hoist deps normally _iff_ the target package is in a subdir of the current package.

mlucool:
@zkat I could do that to keep my hack working. I am still a little unsure about how this hoisting will play out because you can imagine something like:
```
examples
\
  example1
    \ 
     package.json
components
\
  component1
   \ 
    package.json
```
If I did a `file:` in example1, this would not work as far as I can tell now (assuming this is still a React example). I *believe* this paradigm is relatively common also.

Any comments on the node flag required for this to work?

P.S. I am a BIG fan of better support for symlinks, per my commit to node.

zkat:
@mlucool It's worth noting that it's tremendously unlikely for us at npm to try to change module resolution, including symlink stuff, ourselves. afaik, there's no intent for us to actually go and push for a change on that end in order to land this.

More likely, I think, is discussing the possibility of taking that flag into account when deciding how to build the hoisted tree -- some users will want that symlink preservation, some won't. I'll defer to @iarna about whether she thinks this is something worth exploring, with a note that any current hacks around `directory` dependencies can continue to exist with that `npm pack` thing I mentioned above.

@iarna does it make sense to you to read `process.env.NODE_PRESERVE_SYMLINKS` from the env and hoisting across symlink boundaries if we find that flag?

mlucool:
IMO if something is outside of a module you should never touch its contents.

iarna:
@mlucool As @zkat says, changing the default Node.js module loader behavior isn't on the table, which is why 'preserve-symlinks' landed as an option.

Thank you for bringing this up! What you described isn't a use-case we had discussed previously.

I think having it change install behavior based on `NODE_PRESERVE_SYMLINKS` is actually mandatory. `npm` as part of installation has to model how Node.js would load the package to correctly determine where to put its dependencies. This is a detail we've thus far not had to deal with because `npm install` has treated symlinks as opaque things it can't know details of.

So yeah, if you have `NODE_PRESERVE_SYMLINKS` then npm will install transitive dependencies of linked dependencies in your local project, not in the symlinked module. That is, it will work the same way it works when the symlinked module is in a subdirectory of your project.

benlangfeld:
@iarna This looks great, and would resolve a large number of headaches for my team using a monorepo. What can I do to help this make progress?

zkat:
@benlangfeld not much right now. We're working towards an implementation now. You can track #16244, which is what this is going to be included in.

benlangfeld:
Ok, thank you @zkat. If there's anything you can make use of warm bodies for to help with this, you know where we are :)

iarna:
@benlangfeld Yeah, it's my current task. =) The actual implementation isn't too tricky, most of the work has been in tests.

fis-cz:
Just a question. why you deprecate the file:?

could not be like the file: will still have the same behavior (copy files to local repo) while link: would create a junction or a symlink?

I would say both of them could be handy in some cases.

zkat:
@fis-cz because `file:` on directories was not a well-thought-out feature and has very little actual utility. We repeatedly run into folks surprised by its behavior. It requires extensive maneuvering in order to be used during development, as well.

For folks who want to do the package-up-then-install workflow, `file:` on tarballs is still available, so they can run `npm pack ./my-dir` to create a tarball, then `npm i ./my-dir-tarball.tgz`.

fis-cz:
@zkat Thanks for explanation, makes sense to me ;)

whitecolor:
But there is case where symlinks are inappropriate. Like when you use the same node_modules on a different system like native and docker/windows WSL. Haven't you thought about it?

akosyakov:
@zkat How will you set up multiple local packages with common dependencies? We've tried to use `npm link` at the beginning but it has a lot of drawbacks like duplicate dependencies and referenced js/ts files are not resolved properly. Switching to local path dependencies was an improvement since there was no difference in `node_modules` structure in dev and prod mode. We had to use watching though to synchronize changes in dependencies, like [here](https://github.com/theia-ide/theia/blob/master/examples/browser/package.json#L15) for `theia-core`. It sounds to me that with npm5 it won't be possible. What is a proper way?

whitecolor:
@akosyakov I currently use [yalc](https://github.com/whitecolor/yalc) as a simple way to work with local packages that are in development phase. It acts like a simple local registry where you "publish" your package and it can automatically propagate changes to all installations, And not symlinks involved, which can be really cumbersome at least for the current state of affairs.

akosyakov:
@whitecolor thx, I will give it a try

nicojs:
For anyone looking to get the old functionality back: I've created a small package which does this: https://www.npmjs.com/package/install-local. It simply runs an `npm pack` followed by an `npm install <tarball file>`.

@akosyakov @whitecolor I've looked at `yalc` as well, but i've read that it still uses `npm link` under the hood. I'll might give it a shot for future projects. Thanks for sharing.

akosyakov:
@nicojs nice

`yalc add` install packaged modules from the local repository. if you do `yalc link` then, yes, you have same issues as with `npm link`.

We've given a try to `yalc`: https://github.com/theia-ide/theia/pull/156#issuecomment-304532963 We don't really have an interest in having the repository for local packages but rather want to set up a local development env for several independent node packages and have fast turnarounds to build them. Right now we are experimenting with [our own tooling](https://github.com/theia-ide/theia/blob/master/config/local-dependency-manager/README.md). For example for [this package](https://github.com/theia-ide/theia/blob/master/examples/browser/package.json) one can run `npm run localinstall` to install declared [local dependencies](https://github.com/theia-ide/theia/blob/master/examples/browser/package.json#L17) above remote and then `npm run watch:localdeps` to start `watch` script on all local dependencies and synchronize changes.

wclr:
> I've looked at yalc as well, but i've read that it still uses npm link under the hood.

No, it doesn't use `npm link`, but it allows to work in the same manner, when you don't want to touch your `package.json` (add dependency). But actually `npm link` it is kind of not very consistent fashion of replacing dependencies with other versions. 

[Yalc](https://github.com/whitecolor/yalc) was made to be able consistently to add code of **dependency packages** in your multiple projects. It not for managing dependencies of monorepos or something.  it just adds dependency package from local repository to your project in your node_modules. Consistently means that you doesn't use symlinks, and install all the dependency packages, and I usually keep this code in the project repo and commit it, to stay fully consistent and not to wary about publishing (in npm or git) my dependency packages. Linking capabilities are added more like experiment, I don't use them.

MikeMatusz:
As @shadowmint mentioned on #16990, the symlink doesn't honor the `files` section of the `package.json` and this can prove problematic.  #10343 has been around for awhile because you end up with a duplicate node_modules folder.  This change basically makes all of the open issues around symlinks a lot more severe.  I had been avoiding using them specifically because of the issues, and now if I want to use npm@5 I'm forced to deal with them.  I'd love to take advantage of all the improvements in the new version, and I'd even prefer symlinks if they behaved the same as an actual install, but until they do I think this is a deal breaker for my team. :(

Would it be possible to symlink only the files/directories that match the pattern of `files` || !`.npmignore` (i.e. the ones that would be present on an actual install)?  Or are there cases where that would create an insane number of links or be impractical for other reasons?

wesleytodd:
I just made my first attempt to run our companies applications with `npm@5` and ran into a few snags with what I think is this feature (was led here by [this post](http://blog.npmjs.org/post/161081169345/v500)).  My two questions are:

1. Local dependencies don't seem to be getting an `npm i` in them, atleast not a full install.  It looks like deps shared in the root project are not being installed in the symlinked package.  The old `npm link` did run a full install, but it appears that now we have to manually run it in each local dep.  Is that correct?  If so, it seems rather weird because an `npm i` in the root project does not result in a runnable application.
2. Is there a built in way to do the "old" mechanism?  In certain circumstances we have code that relies on singletons (I have long since been pushing our team away from the pattern, but alas, never soon enough).  It would appear we either need to stall on `npm@4` or publish these to our internal npme (aka stop using local deps).  Is this the planned way forward, or are there doc somewhere to ease this transition that I did not find?

Also of note: we are in the process of upgrading from `node@4`, so cannot use the `--preserve-symlinks` flag which I believe would give us something closer to the behavior we desire.

nicojs:
If anyone is looking for the old functionality of `npm i file:`, please take a look at https://www.npmjs.com/package/install-local. It should work for all versions of npm since version 2. I will use that for all my projects where i want an actual local install, instead of an npm link. 

@wesleytodd might be useful for your usecase as well.

cchamberlain:
`npm` 5 has a few issues on Azure app service as it does not (and will likely never) support symbolic links. See https://github.com/projectkudu/KuduSync/issues/33#issuecomment-310811410

In addition to local packages breaking, npm scripts that reference dev dependency installed executables do not operate correctly (`rimraf`, `cross-env`, `webpack` in scripts all fail, unable to find the executable). Guessing this is probably related to a change in `node_modules/.bin` that involves symbolic links? Seems likely there are other cloud services that will suffer similar breakage.

wclr:
symlinks are bad. npm don't cares.

donaldpipowitch:
Will npm support a feature similar to `yarn`'s workspaces _natively_ in the future? Is there an issue which I could track? I can't find one.

I'd like to see a way to resolve packages locally without symlinks, with flat dependencies and which respects `.gitignore`/`.npmignore`/`"files"`. E.g. so they are installed like a regularly published package or basically like `git`-based packages, but locally.

wclr:
@donaldpipowitch 

> I'd like to see a way to resolve packages locally without symlinks, with flat dependencies and which respects .gitignore/.npmignore/"files". E.g. so they are installed like a regularly published package or basically like git-based packages, but locally.

Look at [yalc](https://github.com/whitecolor/yalc)

donaldpipowitch:
> Look at yalc

Yeah, thanks. Just wanted to ask if there are plans to do something similar *natively* in npm?

nicojs:
> Yeah, thanks. Just wanted to ask if there are plans to do something similar natively in npm?

There is something *kind of* native. Use `npm pack` in the dependency folder and `npm install ../sibling/sibling.0.0.0.tar.gz` in the dependent folder.

This is what [install-local](https://www.npmjs.com/package/install-local) does for you.

wclr:
> Yeah, thanks. Just wanted to ask if there are plans to do something similar natively in npm?

I have doubts about `npm`, `yarn` maybe will do something like that but pretty sure not soon, though is really kind of simple task. If we don't have this natively we have to make/use some *hack* before the feature is supported.

donaldpipowitch:
Okay, thank you both.

shadowmint:
Given that the open issues around this are just being ignored at the moment
(with all the other npm 5 things that need fixing I guess), I wouldn't hold
your breath for any kind of quick resolution.

I recommend you just stick with npm 4 for now (and probably the foreseeable
future).

iarna:
@donaldpipowitch
> Will npm support a feature similar to yarn's workspaces natively in the future? Is there an issue which I could track? I can't find one.

Yes, no issue as yet, I need to translate their RFC into something that makes sense more broadly (it was rather yarn specific, to the point that it was hard to follow as an implementor w/o deep yarn knowledge). But once I do, we'll track that in new RFC issue. Assuming they haven't strayed from the RFC they shared with us a few months back, I anticipate our implementation being compatible. The idea is that tools like [`lerna`](https://github.com/lerna/lerna) will get to defer more of the thinking about how things install to the package managers.

At its heart it's pretty simple: `workspaces` property of `package.json`, an array of globs. For all folders that match the globs, install their deps when running `npm install` at the top level and cross-link them when they request any version of a module w/ the same name.

That's currently scheduled for _after_ fixing `npm link`'s behavior as it relates to file dependencies.

iarna:
The spec here actually landed in e563d619c, so I'm going to close this PR. I encourage folks looking for further discussion around this implementation to open a new issue or comment on an existing one.

donaldpipowitch:
Thank you for the explanation!

UnrememberMe:
FYI, the current implementation breaks one of the mono-repo scenario as discovered in the Pants issue linked above.

The scenario is:
1) inside of a mono-repo, we created a wrapper for a commonly used tool, say webpack. 
2) include the wrapper module as a dependency on another module and trying to `npm run` the wrapper module to invoke webpack.

This scenario works in npm3 because webpack as a dependency for the wrapper module will be installed in node_module/.bin. For npm5, a symlink will be created for the wrapper module and when the wrapper trying to invoke webpack, webpack will not be on $PATH.

Basel78:
Hello,
I have LibA with ModuleA and LibB with ModuleB that use ModuleA from LibA
also I have Project that use Both LibA & LibB
how best to import LibA in LibB and both in Project if I am using the following versions
LibA & LibB both are angular4
Project is angular-cli
What I want to achieve here is to avoid the relative path of module in my code, 
1- LibA as external to LibB so it wont be imported once once bundle the libraries,
2- LibA & LibB as externals and will only install them in Project.

BTW: I used  symlink first to reference LibA in LibB but was only able to import ModuleA in relative path whivh what i want to avoid it, since that would mean n Project I would have LibA, and LibB with Liba under node_modules with relative path to ModuleA
and on the Project level will have libA & LibB in node_modules with Import to ModuleA & ModuleB in relative path.
now I am using npm pack and npm install file://path/to/module-a/moduleA-1.2.3.tgz inside LibB
but import { ModuleA } from "LibA"; is giving "error cannot find ModuleA"

also I have noticed that after using the npm pack that the referenced LibA appears in LibB-> node_modules folder but the "d.ts & .ts" files is dist folder does not appear anymore, even though it appear in folder in LibA ->dist.

UnrememberMe:
to @dotchev,  sorry I did not actively monitor this issue.  To answer your question, in a mono-repo, a reasonable setup for a tool might be:
1) create a wrapper (component A) for a commonly used tool, say eslint / webpack etc and supply some base/common configs, and expose a new command line
a simple example can be 
mylint.sh =>
```
... overlay $1 on top of eslint.base.config if any and generate a new combined config
eslint --config=generated_config
```
2) other projects (component B) within a mono-repo might reference the wrapper instead of the original tool via a file link and call `mylint my-overlay-lint-config`. 

with npm5, eslint will no longer be in .bin for component B, and invoking `mylint.sh` directly will fail to execute if running under component B directory.
We can certainly package component A into a private npm registry.  However, one advantage of mono-repo is codes can reference their dependencies directly without needing external artifact storages.

octogonz:
Hi @iarna,

We're having an issue with the "file://" specifier not being handled correctly by NPM 5.  We have an automated tool that generates a **package.json** with references like this:

```json
"dependencies": {
    "yargs": "~4.6.0",
    "z-schema": "~3.18.3",
    "project1": "file:./projects/project1.tgz",
    "project2": "file:./projects/project2.tgz",
```

The "project1.tgz" archive contains a **package.json** file like this:

```json
{
  "name": "project1",
  "version": "0.0.0",
  "private": true,
```

When NPM 4 installs it, the **node_modules/project1/package.json** looks like this, as expected:
```json
  "version": "0.0.0"
```

However, NPM 5 for some reason writes it like this:
```json
  "version": "file:projects/project1.tgz"
```

Was this an intentional design change?  Is it an NPM 5 bug?

It's causing trouble because [read-package-tree](https://www.npmjs.com/package/read-package-tree) chokes on this JSON value. We get an error like this:

```
ERROR: Failed to parse package.json for project1: Invalid version: "file:projects/project1.tgz"
```

octogonz:
I'm pretty sure this is an NPM bug.  I've opened #19006

adrian-gierakowski:
@iarna what happened to this proposal? has it been rejected? 

btw. looks like something similar has been [implemented in yarn some time ago](https://github.com/yarnpkg/yarn/pull/3359)

zkat:
@adrian-gierakowski [it was merged](https://github.com/npm/npm/blob/latest/doc/spec/file-specifiers.md) and implemented as part of [`npm@5.0.0`](https://github.com/npm/npm/releases/tag/v5.0.0)

PR #15901, opened 2017-03-01
install: Use EXDEV aware move instead of rename

iarna:
This will allow moving across devices and moving when filesystems don't
support renaming directories full of files.

Fixes: #9863 (hopefully)


lib/utils/rename.js
```diff
@@ -1,16 +0,0 @@
-'use strict'
```
graingert: might be worth leaving this as an empty file so it doesn't break anyone using:

```docker
# Fix bug https://github.com/npm/npm/issues/9863
RUN cd $(npm root -g)/npm \
  && npm install fs-extra \
  && sed -i -e s/graceful-fs/fs-extra/ -e s/fs\.rename/fs.move/ ./lib/utils/rename.js
```

legodude17:
Build failed, guessing it was a test using `lib/utils/rename.js`. Travis web client is not cooperating.

lib/utils/rename.js
iarna: I'm grumpy about it but yeah, since this is fixing things for exactly the folks who would have had that patch, we probably shouldn't break the work around. =D

iarna: We'll remove the hack w/ npm@5, since it's a breaking bump anyway.

iarna:
@legodude17 Ack, I hit the wrong button and deleted your comment, but this is failing because I missed adding `lib/utils/move.js`. All the tests should already be updated.

legodude17:
@iarna That's ok. By the way, where did appveyor go? I don't see it.

zkat (approved):
🐑

PR #15922, opened 2017-03-02
fixed link package name

yadimon:


KenanY:
I believe `"redis"` is correct, see https://github.com/npm/npm/pull/7538.

yadimon:
then other `node-redis` names should be renamed to `redis`?

KenanY:
Sure, it would probably be less confusing that way. We could even use a different module entirely.

iarna:
As @KenanY points out, `redis` is the name of the module, so I'm going to close this. But I would be interested in a PR that makes the docs more consistent in this respect (even using a different module).

PR #15966, opened 2017-03-06
Fix readme on Caret description

wanghaofei90:
All other descriptions are correct, except for this line.

### DIFF
--- a/doc/misc/semver.md
+++ b/doc/misc/semver.md
@@ -222,7 +222,7 @@ additive (but non-breaking), according to commonly observed practices.
 
 * `^1.2.3` := `>=1.2.3 <2.0.0`
 * `^0.2.3` := `>=0.2.3 <0.3.0`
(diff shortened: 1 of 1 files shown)


iarna:
The existing documentation is correct. This is one of the quirks of the `^` operator that isn't widely known. If you look a [few lines up](https://github.com/npm/npm/pull/15966/files#diff-86daa0cfb5cf2a8920042149fb2ad7eaR209) you'll see the behavior for 0.x and 0.0.x releases described:

> Allows changes that do not modify the left-most non-zero digit in the [major, minor, patch] tuple. In other words, this allows patch and minor updates for versions 1.0.0 and above, patch updates for versions 0.X >=0.1.0, and no updates for versions 0.0.X.

(The tests are failing because that document has to exactly match the [`semver`](https://github.com/npm/node-semver#caret-ranges-123-025-004) module's documentation.)

wanghaofei90:
Thanks for your explanation.

PR #15967, opened 2017-03-06
Limit concurrent metadata fetches and actions

iarna:
This is an old (circa 2015) branch that adds some basic concurrency limits.


lib/fetch-package-metadata.js
```diff
@@ -39,7 +40,9 @@ function andLogAndFinish (spec, tracker, done) {
   }
 }
 
-module.exports = function fetchPackageMetadata (spec, where, tracker, done) {
+module.exports = limit(fetchPackageMetadata, 10)
```
legodude17: I think this should be a configurable number.

lib/install/actions.js
```diff
@@ -62,7 +63,7 @@ Object.keys(actions).forEach(function (actionName) {
     function andDone (cb) {
       return andFinishTracker(log, andAddParentToErrors(pkg.parent, andHandleOptionalDepErrors(pkg, cb)))
     }
-  }
+  }, 10)
```
legodude17: Same here. Configuration is good!

lib/install/actions.js
```diff
@@ -2,7 +2,8 @@
 var validate = require('aproba')
 var chain = require('slide').chain
 var asyncMap = require('slide').asyncMap
-var iferr = require('iferr')
+var log = require('npmlog')
```
legodude17: Why is this added, if this file was already using `log`?

lib/install/actions.js
```diff
@@ -2,7 +2,8 @@
 var validate = require('aproba')
 var chain = require('slide').chain
 var asyncMap = require('slide').asyncMap
-var iferr = require('iferr')
```
legodude17: Wait, was this dep used before this PR?

legodude17 (commented):
I think I might be too nitpicky...

Also, I have no authority, just some suggestions.

lib/install/actions.js
iarna: this was from a bad rebase, fyi

iarna:
I'm -1 on making this configurable. It's in our power to choose reasonable limits that neither cause npm to eat your router nor slow down your installs.

You will find that I'm of the opinion that configuration has to carry its own weight in terms of value. I will tend toward not having configuration without justification. Configuration has a cost:

1. It leads to bad UX. See Gnome 1, for example.
2. It makes it harder to teach all the ways to use the software. Any configuration has to be documented and the longer the documentation the harder it is to internalize, so it implicitly makes existing documentation harder to understand.
3. It makes it harder, as a programmer, to understand the behavior of the software as configuration options are global state and you need to be aware of all of them. It also complicates debugging because a user might have a different value.

legodude17:
@iarna If you don't want that, then I won't argue, just thought it was a good idea.

zkat (approved):
yyeeyyy

lib/npm.js
```diff
@@ -51,6 +51,13 @@
 
   npm.commands = {}
 
+  // TUNING
```
legodude17: What do you mean "TUNING"?

zkat: as in, give us a single place to play around with numbers so we can find a good balance to ship with. We're not gonna make this user-configurable.

legodude17: Ok. That sounds like a good plan.

PR #15968, opened 2017-03-07
extract: support corguments

iarna:
This replaces the package manifest in the ideal tree with one read off disk and merged with what we got from the registry


zkat (changes requested):
The changes basically match the stuff I've been testing with, so LGTM on that front. This should really get some tests, though.

iarna:
hmm, I guess a test where we inject a corgument and assert that we got the package.json version in the installed package…

zkat:
@iarna https://github.com/zkat/pacote/blob/latest/test/util/mock-tarball.js would this help

iarna:
I usually just make a teeny-tiny minimal tarball and embed it with `tacks`

PR #15969, opened 2017-03-07
fetch-package-metadata: Support the _hasShrinkwrap registry metadata

iarna:
For faster installs, if it's set to `false` then we don't have to extract the tarball and look for one before continuing.


soldair:
as of now _hasShrinkwrap is either true or undefined. is this ok? :fearful:

iarna:
Summarizing in air conversations: Yeah, the code here makes it pretty clear why "true or undefined" doesn't work for us. What we actually need to know is which packages DON'T have shrinkwraps and we have to distinguish between "registry doesn't support _hasShrinkwrap" and "no modules have shrinkwraps".

If `undefined` could be JSON encoded then we could have worked with this, but it can't. (`JSON.stringify` strips any properties whose value is `undefined`.)

iarna:
This landed in `4.5.0`

Issue #15933, opened 2017-03-03
npm search returns empty Object in callback


MexXxo:
#### I'm opening this issue because:

  - [ ] npm is crashing.
  - [ ] npm is producing an incorrect install.
  - [ X ] npm is doing something I don't understand.
  - [ ] Other (_see below for feature requests_):

#### What's going wrong?
npm search returns empty Object in callback

#### How can the CLI team reproduce the problem?
It's happening using npm programmatically in node.

```
    npm.commands.search(['someModule'], function(err, data){
        console.log('returned');
        if (err) {
            console.log(err);
            return;
        }
        console.log(data); //data is not the search result instead it's an Object
    });
```

Looking to the code, it comes from here: https://github.com/npm/npm/blob/latest/lib/search.js#L87

legodude17:
Maybe instead, the search function should return the output stream?

MexXxo:
As the callback comes after the stream has ended, it makes more sense to return the output all together. 
Depending if we use json or not as output, I would return a string, that the component that receives the callback can parse itself, also I would add the json output flag as argument to the search (currently it is only taken from config). 
I will create today a pull request with a solution, if this is okay.

legodude17:
@MexXxo that's good. Then the core team can review it.

MexXxo:
#15973 - PR here

PR #15973, opened 2017-03-07
returning the searchResults as a string in the callback

Flowr-es:
With this change, the callback would return a string instead an empty object (see Issues: #15933  )
The string is the same that is outputted to the console.
I took a look into the tests, but it seems there are no tests for the commands API at all. 
If needed I can create one.

### DIFF
--- a/lib/search.js
+++ b/lib/search.js
@@ -71,9 +71,11 @@ function search (args, cb) {
     parseable: npm.config.get('parseable'),
     color: npm.color
   })
+  var searchResults = '';
   outputStream.on('data', function (chunk) {
     if (!anyOutput) { anyOutput = true }
     output(chunk.toString('utf8'))
+	searchResults += chunk.toString('utf8')
   })
 
   log.silly('search', 'searching packages')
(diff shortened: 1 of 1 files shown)


lib/search.js
```diff
@@ -71,9 +71,11 @@ function search (args, cb) {
     parseable: npm.config.get('parseable'),
     color: npm.color
   })
+  var searchResults = '';
   outputStream.on('data', function (chunk) {
     if (!anyOutput) { anyOutput = true }
     output(chunk.toString('utf8'))
+	searchResults += chunk.toString('utf8')
```
legodude17: Going to want to append a newline to this. `output` is a `console.log`, which has newlines.

legodude17: Outdent this by 2 spaces

legodude17 (commented):
Nice Job! We can see if the CI is green.

iarna (changes requested):
Using npm internal libraries from other applications is unsupported and so we aren't going to make changes solely to improve that experience. Internal APIs are not a part of our contract and we can and will change them in ways that would be breaking if you were depending on them without a major version bump.

The supported ways to do this are:
Using `child_process` to call `npm search --json` or `npm search --parseable` and parse the output.
And implementing the [search API](https://github.com/npm/registry/blob/master/docs/REGISTRY-API.md#get-v1search) directly (it's very simple).

PR #15976, opened 2017-03-08
install as if this is fresh and there is no node modules yet

iarna:
This is a version of `npm install` that will always check the registry for updates to all of your direct and transitive dependencies. It's kind of like `npm update` at all depth levels.

Before this could land it would need to have either a better name or get it folded into a command line option. Possibly we might replace `npm update` and `npm outdated` with this.


legodude17:
I would suggest the name be `reinstall`.

iarna:
I'm closing this to clean out the PR queue.

PR #15977, opened 2017-03-08
dedupe rewrite

iarna:
This is the rewrite of `npm dedupe` that no one asked for.

The existing version looks for changes it can make to your existing tree and generates a list of `move` actions to take to move modules to the top of the tree.

This new version just moves them in the idealtree directly and lets `diff-trees` figure out the move commands. This pushes the smarts into the installer itself, keeping them in one place as much as possible.


iarna:
Plan: This is going into `npmc` to increase our confidence in the changes. Though at the moment it might not be much of a risk, considering the state of `npm dedupe`.

PR #16013, opened 2017-03-12
Execute preinstall before tree is calculated

### LINKED ISSUE #10379: preinstall execution order in npm@3.x

palmerj3:
This PR attempts to fix https://github.com/npm/npm/issues/10379 by executing preinstall before the tree is calculated and an ideal tree generated.

Functionally this does appear to work but I'm looking for feedback and testing.

@iarna @zkat would love your feedback on this if you have time. Thanks!

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -242,8 +242,17 @@ Installer.prototype.run = function (_cb) {
     }
   }
 
+  var preinstallSteps = []
   var installSteps = []
   var postInstallSteps = []
+
+  if (!this.dryrun) {
+    preinstallSteps.push(
+      [this.newTracker(log, 'runTopLevelLifecycles', 2)],
+      [this, this.runPreinstallTopLevelLifecycles]
+    )
+  }
+
   installSteps.push(
     [this.newTracker(log, 'loadCurrentTree', 4)],
     [this, this.loadCurrentTree],
@@ -264,9 +273,6 @@ Installer.prototype.run = function (_cb) {
     [this, this.debugActions, 'decomposeActions', 'todo'])
   if (!this.dryrun) {
     installSteps.push(
-      [this.newTracker(log, 'runTopLevelLifecycles', 2)],
-      [this, this.runPreinstallTopLevelLifecycles],
-
       [this.newTracker(log, 'executeActions', 8)],
       [this, this.executeActions],
       [this, this.finishTracker, 'executeActions'])
@@ -289,32 +295,35 @@ Installer.prototype.run = function (_cb) {
     [this, this.printInstalled])
 
   var self = this
-  chain(installSteps, function (installEr) {
-    if (installEr) self.failing = true
-    chain(postInstallSteps, function (postInstallEr) {
-      if (self.idealTree) {
-        self.idealTree.warnings.forEach(function (warning) {
-          if (warning.code === 'EPACKAGEJSON' && self.global) return
-          if (warning.code === 'ENOTDIR') return
-          var output = errorMessage(warning)
-          output.summary.forEach(function (logline) {
+  chain(preinstallSteps, function (preinstallEr) {
+    if (preinstallEr) self.failing = true
+    chain(installSteps, function (installEr) {
+      if (installEr) self.failing = true
+      chain(postInstallSteps, function (postInstallEr) {
+        if (self.idealTree) {
+          self.idealTree.warnings.forEach(function (warning) {
+            if (warning.code === 'EPACKAGEJSON' && self.global) return
+            if (warning.code === 'ENOTDIR') return
+            var output = errorMessage(warning)
+            output.summary.forEach(function (logline) {
+              log.warn.apply(log, logline)
+            })
+            output.detail.forEach(function (logline) {
+              log.verbose.apply(log, logline)
+            })
+          })
+        }
+        if (installEr && postInstallEr) {
+          var msg = errorMessage(postInstallEr)
+          msg.summary.forEach(function (logline) {
             log.warn.apply(log, logline)
           })
-          output.detail.forEach(function (logline) {
+          msg.detail.forEach(function (logline) {
             log.verbose.apply(log, logline)
           })
-        })
-      }
-      if (installEr && postInstallEr) {
-        var msg = errorMessage(postInstallEr)
-        msg.summary.forEach(function (logline) {
-          log.warn.apply(log, logline)
-        })
-        msg.detail.forEach(function (logline) {
-          log.verbose.apply(log, logline)
-        })
-      }
-      cb(installEr || postInstallEr, self.getInstalledModules(), self.idealTree)
+        }
+        cb(preinstallEr || installEr || postInstallEr, self.getInstalledModules(), self.idealTree)
+      })
     })
   })
 }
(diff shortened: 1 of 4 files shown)


lib/install.js
```diff
@@ -208,7 +208,7 @@ function Installer (where, dryrun, args) {
   this.todo = []
   this.progress = {}
   this.noPackageJsonOk = !!args.length
-  this.topLevelLifecycles = !args.length
+  this.topLevelLifecycles = !!args.length
```
legodude17: Just wondering, why did you change this?

palmerj3: Because preinstall was never being called at all without this and I think it was a bug to begin with given how other args.length are being handled. This is the code in reference that was always triggered with args.length being the way it was https://github.com/palmerj3/npm/blob/0ee8d4248b4294c347923b8cff41f89dfe597ac9/lib/install.js#L541

zkat:
@palmerj3 thanks a lot for tackling this! Ping us when it's ready! I'm gonna /cc @iarna cause she made the original call to move this up. :)

We have a nice window for a breaking change like this coming up (and I'm pretty sure it's a breaking change -- correct me if I'm wrong), so if you get it done in the next ~2 weeks, we'd be able to land it as part of `npm@5`. It might be longer than that, because that depends on other factors, but hopefully it's still incentive :)

lib/install.js
iarna: Your `preinstall` top level lifecycle shouldn't run if you run `npm install foobar`. It should only run if you're installing the top level package, eg `npm install`.

palmerj3: @iarna ah ok good point :)

lib/install.js
```diff
@@ -242,8 +242,15 @@ Installer.prototype.run = function (_cb) {
     }
   }
 
+  var preinstallSteps = []
   var installSteps = []
   var postInstallSteps = []
+
+  preinstallSteps.push(
```
iarna: Why introduce a new phase here? Why not just make this the start of the install steps? Either way, this needs to not run of `this.dryrun` is true.

palmerj3: @iarna A new phase is optional and I'm fine removing it. Before getting this to work I had a few other steps there and it made some sense. Removing now and pushing a few tweaks!

palmerj3: Actually let me edit that... it does make some sense because, in the case where you want to bail before install, then I think having it organized this way will allow for this whereas having everything chained in a single install step would mean you'd wait longer for that feedback?

palmerj3:
@zkat thanks! :)  I'll work on the tweaks mentioned and write a few tests then get your feedback again.

iarna (changes requested):
With a couple of tweaks I think we should be able to take this.

iarna:
Other work that will need to be done:

* [ ] Travis should be green (all tests passing)
* [ ] It would be nice to have a test that can only pass if `preinstall` runs early.

palmerj3:
Ok cool! I'm working on making the tests green now.

palmerj3:
@iarna @zkat I got the tests to pass. If you get some time please review.. the main complication is the use of [tacks](https://github.com/iarna/tacks) in the tests in order to generate fake package.json files. So this latest iteration handles that. I would much prefer to not have a code path just for tests so if you have any suggestions I would love to hear.

In the meantime I'll try to work on a test that will only pass if preinstall is run early.

palmerj3:
@zkat I'm not sure this would be a breaking change... I'll totally leave that up to you all to decide. Really the only way this would be breaking is if devs attempted to manipulate the tree preinstall, it failed, and they are now relying on that :)

selvaa89:
Any idea when this PR will be merged? Our repos mutates package.json on preinstall script for various reasons. Because of this issue we are stuck with 2.x!

palmerj3:
@zkat @iarna looking at the other tests I don't see any sort of stubbing libs being used like sinon.. so I'm not totally sure how I could reasonably write a test for this if I can't spy on the "loadCurrentTree" method. Thoughts?

iarna:
@palmerj3 We stub with `requireInject`, but I would be inclined to go with a functional test for this. For example, you could:

1. Pass in a registry that can't work. (eg, `https://127.0.0.11:1111`) 
2. Have a `preinstall` that moves a dependency into `node_modules`
3. Try to run the install. It should succeed because its dependency should be in place due to the preinstall. If the new behavior is not in place then it will fail with a network error.

palmerj3:
@iarna thank you! I will take that approach.

palmerj3:
@iarna I took your advice pretty much verbatim and wrote a test that fails on "latest" and passes with this PR to verify preinstall fires early.

Thanks again for the advice and direction!

test/tap/install-parse-error.js
```diff
@@ -39,7 +39,6 @@ test('failing to parse package.json should be error', function (t) {
     function (err, code, stdout, stderr) {
       if (err) throw err
       t.equal(code, 1, 'exit not ok')
-      t.similar(stderr, /npm ERR! Failed to parse json/)
```
legodude17: Why was this line removed?

scripts/changelog.js
```diff
@@ -75,7 +75,6 @@ function main () {
     if (/^---$/.test(line)) {
       print_commit(commit)
     } else if (m = line.match(/^([a-f0-9]{7,9}) ([a-f0-9]+) (?:[(]([^)]+)[)] )?(.*?) [(](.*?)[)]/)) {
-
```
legodude17: Any idea why this was changed?

legodude17 (commented):
😄

test/tap/install-parse-error.js
palmerj3: Because I wasn't getting the same error message. Was an error none-the-less but not with that specific text. And I don't think this is a valuable assertion to have and has the risk of breaking across node versions.

scripts/changelog.js
palmerj3: I only changed it because the linter was complaining about it

test/tap/install-parse-error.js
legodude17: Ok, fair enough.

scripts/changelog.js
legodude17: I am surprised the linter complained about code that was already there.

palmerj3: Someone else sent a fix for the same issue but it hasn't been merged. - https://github.com/npm/npm/pull/16110

lib/install.js
```diff
@@ -536,8 +545,21 @@ Installer.prototype.runPreinstallTopLevelLifecycles = function (cb) {
...
+  }
+
+  // Construct initial node for top of the tree
+  var rootNode = {
+    package: p
+  }
+
   steps.push(
-    [doOneAction, 'preinstall', this.idealTree.path, this.idealTree, trackLifecycle.newGroup('preinstall:.')]
+    [doOneAction, 'preinstall', process.cwd(), rootNode, trackLifecycle.newGroup('preinstall:.')]
```
iarna: `process.cwd()` isn't safe for this, as you can run `npm install` in a subdir of a project and it'll install at the project root.

test/tap/install-parse-error.js
iarna: This assertion needs to stand. The error message it's checking for is from npm and so Node.js version doesn't matter.

If this assertion is failing it means that something is unexpected about the error message. The install failed but what error did it report?

scripts/changelog.js
iarna: It has as of last Friday—if you rebase on to the latest `release-next` you should see it go away.

lib/install.js
iarna: Chains are run in sequence and stop after the first error. The only reason to have multiple chains is if you want to run a later chain even though an earlier one failed.

lib/install.js
```diff
@@ -242,8 +242,17 @@ Installer.prototype.run = function (_cb) {
     }
   }
 
+  var preinstallSteps = []
```
iarna: https://github.com/npm/npm/pull/16013#discussion_r108286379

lib/install.js
iarna: What you want here is `this.where` I believe.

scripts/changelog.js
palmerj3: @iarna this PR is toward "latest" - should I switch to release-next?

lib/install.js
palmerj3: Ok cool I'll strip that out.

palmerj3:
Thank you again @iarna for the guidance! I'll make the changes requested and look further into the json error.

xkr47:
@palmerj3 ping?

palmerj3:
I haven't had time recently to make the final changes but I hope to soon. I wish I had received feedback all at once instead of in two phases but alas changes are required still and I hope to get to them soon.

zkat:
This was done as part of `npm@5.0.0`. In the rush, we shoved in our own code, but thank you for filing this PR, and I hope `npm@5` works like you need it to, now 🎉

PR #16014, opened 2017-03-12
Replace http: URL in npm-shrinkwrap with https:

divergentdave:
Per http://blog.npmjs.org/post/154400916805/avoid-http-urls-in-shrinkwrap-files.

This is affecting me because I'm using [wintersmith](https://www.npmjs.com/package/wintersmith), which pulls in npm as a dependency.

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -1590,7 +1590,7 @@
         "are-we-there-yet": {
           "version": "1.1.2",
           "from": "are-we-there-yet@>=1.1.2 <1.2.0",
-          "resolved": "http://registry.npmjs.org/are-we-there-yet/-/are-we-there-yet-1.1.2.tgz",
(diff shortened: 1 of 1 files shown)


watilde (approved):

legodude17:
Are there any other instances of `http://`?

divergentdave:
There's just the one in this branch.

legodude17:
@divergentdave Little odd that only one of them was using `http://`, but ok.

iarna:
Huuh, good catch. Because we bundle all of our production dependencies, the shrinkwrap is never used for anything other than development dependencies so it coincidentally did not pose a security risk. But it didn't belong in there.

This also needed to include a new install of the effected module as without it, npm would throw out this change the next time anyone ran `npm shrinkwrap` or `npm install --save`.

divergentdave:
This change is needed for cases where npm is required as a dependency. For
example, the `wintersmith` package depends on npm.

iarna:
This shipped in [v4.5.0](http://blog.npmjs.org/post/158793343690/npm-v450-2017-03-24)!

PR #16040, opened 2017-03-14
Bundle dependencies inside bundled scope deps

iarna:
We had a bug in the files and ignores code that resulted in the dependencies of bundled scoped modules that were weren't flattened would not be included in the bundle.

Due to npm@3+'s self-healing behavior, modules built this way were still installable, but they lost the advantage of the bundle for those modules.


lib/utils/tar.js
```diff
@@ -136,8 +136,10 @@ BundledPacker.prototype.applyIgnores = function (entry, partial, entryObj) {
 
     // the package root.
     var p = this.parent
-    // the package before this one.
+    // the directory before this one.
     var pp = p && p.parent
+    // the directory before that (if this is scoped)
+    if (pp && pp.basename[0] === '@') pp = pp && pp.parent
```
iarna: Incidentally, `fstream-npm` doesn't require a matching patch. This patch is required because `this.isBundled` does not entirely mesh with`fstream-npm` as I didn't understand `fstream-npm`'s filtering method very well when I wrote it. (For instance, I didn't know that `applyIgnores` would get called recursively.)

zkat (approved):
👍

lib/install/action/extract.js
```diff
@@ -68,14 +68,11 @@ function stageBundledModule (bundler, child, staging, parentPath, next) {
   var stageFrom = path.join(parentPath, 'node_modules', child.package.name)
   var stageTo = moduleStagingPath(staging, child)
 
-  // if this module bundles its own deps and it's not coming from OUR bundle then
-  // we're not responsible for staging its children.
-  if (!child.fromBundle && child.package.bundleDependencies) return finishModule()
```
iarna: Skipping walking the kids like this would have resulted in breaking modules that had ina rather complicated set of interleaved bundled dependencies.

If one module with bundles depended on another with bundles and we ended up deciding to replace the second module, but not all of its children, then some of the children would come from the first module's bundled deps and some would come from the second.

lib/utils/error-message.js
```diff
@@ -268,7 +268,6 @@ function errorMessage (er) {
       detail.push([
         'enoent',
         [
-          er.message,
```
iarna: It's already included in the short message and we never print the detail without the short message.

lib/install/action/extract.js
```diff
@@ -68,7 +69,9 @@ function stageBundledModule (bundler, child, staging, parentPath, next) {
 
   function moveModule () {
     if (child.fromBundle) {
-      return move(stageFrom, stageTo, iferr(next, updateMovedPackageJson))
+      return mkdirp(path.dirname(stageTo), iferr(next, function () {
```
iarna: We need to mkdirp so that `.staging/@scope` is created before we call move. A separate bug was masking the error so that we didn't know that it had failed until `finalize` failed.

zkat (approved):
🐑 if tests pass

iarna:
Landed in 4.4.3 as 6754dabb6

PR #16051, opened 2017-03-15
docs: Add section to TROUBLESHOOTING

legodude17:
Adds a section to `TROUBLESHOOTING.md` about the `Cannot read property 'path' of null` and such type errors.

### DIFF
--- a/TROUBLESHOOTING.md
+++ b/TROUBLESHOOTING.md
@@ -197,7 +197,11 @@ If your install fails and it says something like `Cannot read property 'path' of
 
 ### Steps to Fix
 
-0. Follow the steps in [Proxies and Network](#proxy-and-networking-issues) steps.
+0. Open up your `npm-debug.log`
+0. Look near the bottom, above the error part
+0. Compare the contents there with the rest of this guide
+0. If you find something, then follow those instructions
+0. If you dont't find anything, open an [issue](https://github.com/npm/npm/issues/new).
 
 ### Examples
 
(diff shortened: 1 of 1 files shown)


TROUBLESHOOTING.md
```diff
@@ -190,3 +191,23 @@ ENOGIT
 ### Related Issues
 
 * [#11095](https://github.com/npm/npm/issues/11095)
+
+## Cannot Read property
+If your install fails and it says something like `Cannot read property 'path' of null`, then it is most likely a network issue in disguise.
```
iarna: I'm... not really comfortable putting this in as a general troubleshooting item because that error can be caused by a number of things. (Ultimately it is, of course, an error in npm itself.) I might feel better about it if it were to cite versions where it's been known to happen?

zkat: I would tell people to pop open their error logs and look for the error details near the bottom of the log, and compare them with more expanded examples. Both of the examples seem to be referring to `ETIMEDOUT` errors -- can we help clarify that in here?

zkat (changes requested):

TROUBLESHOOTING.md
legodude17: Maybe narrow it to just `Cannot read property 'path' of null`? Have you ever seen an instance of that where the problem is not with the internet?

iarna: Yes, absolutely.

legodude17: @zkat Did what you suggested.

legodude17:
This feels kind of pointless with the new forum.

PR #16129, opened 2017-03-22
On SIGINT, ensure process stays alive until spawned child_process exits.

jaridmargolin:
Currently npm run scripts can get caught in a strange state where the child process is still operating but the parent process has returned control to the shell.

This PR aims to solve two uses cases:
1. It waits to exit until the child async SIGINT handlers have completed.
2. It allows the second SIGINT call to kill the process.


iarna (approved):
This looks good to me.

iarna:
Thanks for this! This shipped in [v4.5.0](http://blog.npmjs.org/post/158793343690/npm-v450-2017-03-24)!

jaridmargolin:
This moved through way faster than I anticipated. Thanks @iarna!

PR #16204, opened 2017-03-30
utils: remove `lib/utils/rename.js`

watilde:
It was used as a stub file to not break a monkey patching hack, but it's no longer required. It's removed with npm@5.

Fixes: npm#15965


zkat (approved):
LGTM!

zkat:
...and into the integration branch it goes! Thanks!

Issue #11342, opened 2016-01-29
npm ERR! Version not changed


harshadbankar:
I am using maven front end plugin for npm and node scipts, but npm version is failing with "[ERROR] npm ERR! Version not changed", when my package.json is has same version as that of command line version value. Any idea on how to override the package.json version if already exists same version?

KenanY:
@harshadbankar What would you like to happen when you run `npm version 1.0.0` where `1.0.0` is already the `"version"` defined in your `package.json`?

harshadbankar:
I just want npm not to complain about this on console, as it makes my maven front end build to fail, and in maven front end plugin, option of "failOnError" is not working. 


Thanks for reply @KenanY

thybzi:
If you had already commited new version number into your `package.json`, just type `npm publish`, and npm will get your updated version of module

harshadbankar:
Hi , i am not pushing this to git, just updating my local package.json, npm publish not working for me

romero83:
Hi,

You can do it by using com.google.code.maven-replacer-plugin replacer maven plugin.

```
...
<build>
      <plugins>
	<plugin>
            <groupId>org.codehaus.mojo</groupId>
            <artifactId>build-helper-maven-plugin</artifactId>
            <version>${build-helper-maven-plugin.version}</version>
            <executions>
               <execution>
                  <phase>validate</phase>
                  <id>parse-version</id>
                  <goals>
                     <goal>parse-version</goal>
                  </goals>
                  <configuration>
                     <propertyPrefix>parsedVersion</propertyPrefix>
                  </configuration>
               </execution>
            </executions>
         </plugin>
         <plugin>
            <groupId>com.google.code.maven-replacer-plugin</groupId>
            <artifactId>replacer</artifactId>
            <version>1.5.3</version>
            <executions>
               <execution>
                  <phase>generate-sources</phase>
                  <goals>
                     <goal>replace</goal>
                  </goals>
               </execution>
            </executions>
            <configuration>
               <file>package.json</file>
               <replacements>
                  <replacement>
                     <token>"version": "(.*)",</token>
                     <value>"version": "${parsedVersion.majorVersion}.${parsedVersion.minorVersion}.${parsedVersion.incrementalVersion}",</value>
                  </replacement>
               </replacements>
            </configuration>
         </plugin>
...
```

lucastheisen:
[...]

PR #16213, opened 2017-03-31
fix for npm ERR! Version not changed #11342

lucastheisen:
Simple feature add with accompanying unit tests.  Fully backwards compatible.

### DIFF
--- a/test/tap/version-from-git.js
+++ b/test/tap/version-from-git.js
@@ -142,7 +142,7 @@ test('npm version from-git with an existing version', function (t) {
   }
 
   function checkVersion (er) {
-    t.equal(er.message, 'Version not changed')
+    t.ok(er.message.match(/Version not changed.*/))
     t.done()
   }
 })
--- a/test/tap/version-from-git.js
+++ b/test/tap/version-from-git.js
@@ -142,7 +142,7 @@ test('npm version from-git with an existing version', function (t) {
   }
 
   function checkVersion (er) {
-    t.ok(er.message.match(/Version not changed.*/))
+    t.like(er.message, /Version not changed/)
     t.done()
   }
 })
--- a/doc/cli/npm-version.md
+++ b/doc/cli/npm-version.md
@@ -78,6 +78,14 @@ and tag up to the server, and deletes the `build/temp` directory.
 
 ## CONFIGURATION
 
+### allow-same-version
+
+* Default: false
+* Type: Boolean
+
+Prevents throwing an error when `npm version` is used to set the new version 
+to the same value as the current version.
+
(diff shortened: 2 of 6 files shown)


test/tap/version-from-git.js
```diff
@@ -142,7 +142,7 @@ test('npm version from-git with an existing version', function (t) {
   }
 
   function checkVersion (er) {
-    t.equal(er.message, 'Version not changed')
+    t.ok(er.message.match(/Version not changed.*/))
```
iarna: tap has a verb for doing regexp compares, using that this would look like:
```
t.like(er.message, /Version not changed/)
```
(And the trailing `.*` is uneccessary)

lucastheisen:
Sure, I can totally fix that.  I was actually trying to keep things consistent with existing code so I looked for an example and found [version-lifecycle.js](https://github.com/npm/npm/blob/latest/test/tap/version-lifecycle.js#L32)...  Can have an updated patch later today.

lucastheisen:
@iarna, I made the update you requested (to use _like_).  Also, noticed i was missing the cleanup in my new test file, so added that.  Should be good to go.

On a side note, while testing locally I cancelled the tests in the middle of processing (as the full suite takes a long time).  I happened to cancel after `builtin-config.js` started, but before its cleanup.  This left the temp files in `test/tap/builtin-config/...`.  At that point, when I tried to run again, the `standard` linter failed due to a bug in one of the modules downloaded for `builtin-config.js` (file that bug [here](https://github.com/alexindigo/asynckit/issues/46)).  I am guessing that, due to your CI process you always start from a clean checkout so you would not run into this issue.  But as a _pull request contributor_, this was a frustrating problem.  Perhaps it would make more sense to create a `<GIT_PROJECT_ROOT>/tmp` folder for all temp files to separate it from the actual test folder so that you could avoid linting leftover test files (or better yet, change your test script to `rimraf tmp && mkdirp tmp && standard && npm run test-tap` or the like)?  Also, it would be nice to have more stuff in the `.gitignore`, specifically `node_modules` (you already have `/node_modules`).  Anyway, thanks for considering this request.

iarna:
Yeah, `builtin-config` is particularly guilty of this, and it happens occasionally to me too. It's annoying, but updating all of the tests to do their work in a more clean way hasn't bubbled up to the top of anyone's priority list yet. It's I'd like to do though.

iarna (approved):

zkat:
This has already been merged as part of `npm@4.6.0` a while back. Closing.

lucastheisen:
@iarna , @zkat , thanks, this greatly simplified my build workflow.

Issue #16210, opened 2017-03-31
Export NODE_EXE from the npm shell script


lucastheisen:
#### What's the feature?
Export NODE_EXE from the `bin/npm` shell script.

#### What problem is the feature intended to solve?
[node_sass](https://www.npmjs.com/package/node-sass) requires:

> Some users have reported issues installing on Ubuntu due to node being registered to another package. Follow the official NodeJS docs to install NodeJS so that **#!/usr/bin/env node** correctly resolved.

But npm uses the version of node in the same folder as `npm` or `npm.cmd`.  If your global version differs, then it node-sass fails because it has an accompanying binary compiled per version of node.  If you were to export the `NODE_EXE` variable you create, then other packages that need to spawn a `node` process could use the same instance as the `npm` command that spawned it.

To reproduce:
Install one version of node globally (say 6.9.5), then create a per project node installation folder (by unpacking one of these: https://nodejs.org/dist/v7.8.0/ into your project folder).  Add [node_sass](https://www.npmjs.com/package/node-sass) as a dependency of your project (from the root folder for your project, run: `.\node\npm install node-sass --save-dev`).  Then create an empty file called `foo.scss`, and add a script to your `package.json`: 
```
  "scripts": {
    "sass": "node-sass foo.scss foo.css"
  }
```
When you run this `.\node\npm run sass`, you will see an error like:

> Error: Missing binding /tmp/foo/node_modules/node-sass/vendor/_buildinfo_/binding.node
> Node Sass could not find a binding for your current environment: _buildinfo_ with Node.js 6.x

If, however, you were to export `NODE_EXE`, i could request that the node-sass project check for that value.

#### Is the absence of this feature blocking you or your team? If so, how?
Not really, but i currently need a workaround that prepends my project node folder to PATH.  This is rather inconvenient given that the `npm `script itself detects the proper node instance.

#### Is this feature similar to an existing feature in another tool?
Its similar to what you are already doing (detecting the location of NODE_EXE).  It just enables anything spawned by `npm `to leverage that detection.

#### Is this a feature you're prepared to implement, with support from the npm CLI team?
Sure, but its just a one line addition to your `bin/npm` script.

lucastheisen:
I created a corresponding [issue on node-sass](https://github.com/sass/node-sass/issues/1940) that would make use of this fix.

lucastheisen:
[...]

PR #16214, opened 2017-03-31
fix for Export NODE_EXE from the npm shell script #16210

lucastheisen:


lucastheisen:
Never mind... just found [--scripts-prepend-node-path](https://github.com/npm/npm/releases/tag/v4.0.0), works like a charm!

PR #14568, opened 2016-11-09
Change `npm_config_node_gyp` to use shell script on Windows

ghostoy:
On non-Windows platforms, custom node-gyp can be used by npm by
`export npm_config_node_gyp=$(which node-gyp)`. But on Windows,
the implementation expected the real js script of node-gyp rather
than the shell script. So you have to set the environment to
`path\to\global\node_modules\node-gyp\bin\node-gyp.js`. This patch
changed the launch script of node-gyp by invoking the script set
to `npm_config_node_gyp` directly instead of invoking as node script.

Fixed #14543

Issue #12986, opened 2016-06-07
Remove `_token` support altogether`


zkat:
Reading through the code, it looks like a couple of places (and `npm-registry-client`) still handle the "extremely legacy" `_token` login token (known as `couchToken` in `n-r-c`).

As part of a breaking change, we could just take this out and just have a little less to worry about.

PR #16244, opened 2017-04-04
release: npm@5.0.0

zkat:
## NOTE: This PR is a WIP. npm@5 beta is currently underway. See #16510 for deets.

Wowowowowow npm@5!

This release marks months of hard work for the young, scrappy, and hungry CLI team, and includes some changes we've been hoping to do for literally years. npm@5 takes npm a pretty big step forward, significantly improving its performance in almost all common situations, fixing a bunch of old errors due to the architecture, and just generally making it more robust and fault-tolerant. It comes with changes to make life easier for people doing monorepos, for users who want consistency/security guarantees, and brings semver support to git dependencies. See below for all the deets!

### Breaking Changes

* One binary to rule them all: `./cli.js` has been removed in favor of `./bin/npm-cli.js`. In case you were doing something with `./cli.js` itself. (#12096)

* Stub file removed (#16204)

* Existing npm caches will no longer be used: you will have to redownload any cached packages. There is no tool or intention to reuse old caches. (#15666)

* `npm install ./packages/subdir` will now create a symlink instead of a regular installation. `file://path/to/tarball.tgz` will not change -- only directories are symlinked. (#15900)

* npm will now scold you if you capitalize its name. seriously it will fight you.

* The "extremely legacy" `_token` couchToken has been removed. (#12986)

* [npm will `--save` by default now](https://twitter.com/maybekatz/status/859229741676625920). Additionally, `package-lock.json` will be automatically created unless an `npm-shrinkwrap.json` exists. (#15666)

* git dependencies support semver (#15308, #15666)

* git dependencies with `prepare` scripts will have their `devDependencies` installed, and `npm install` run in their directory before being packed.

* `npm cache` commands have been rewritten and don't really work anything like they did before. (#15666)

* `--cache-min` and `--cache-max` have been deprecated. (#15666)

* running npm while offline will no longer insist on retrying network requests (#15666)

* The contents of `~/.npm` have completely changed.

* A new loglevel, `notice`, has been added and set as default.

* package locks no longer exclude `optionalDependencies` that failed to build. This means package-lock.json and npm-shrinkwrap.json should now be cross-platform. (#15900)

* shrinkwrap and package-lock no longer exclude `devDependencies`. You can still prevent their installation with `--prod`/`--only=prod`. (#15900)

* If you generated your package lock against registry A, and you switch to registry B, npm will now try to [install the packages from registry B, instead of A](https://twitter.com/maybekatz/status/862834964932435969). If you want to use different registries for different packages, use scope-specific registries (`npm config set @myscope:registry=https://myownregist.ry/packages/`). Different registries for different unscoped packages are not supported anymore.

* shrinkwrap and package-lock no longer warn and exit without saving the lockfile

* Local tarballs can now only be installed if they have a file extensions `.tar`, `.tar.gz`, or `.tgz`.

### Feature Summary

There's a bunch of exciting stuff going in here! You can track all the changes planned for `npm@5` on [its project tracker](https://github.com/npm/npm/projects/1).

#### Installer changes

* A new, standardised lockfile feature meant for cross-package-manager compatibility (`package-lock.json`), and a new format and semantics for shrinkwrap. (#16441)

* `--save` is no longer necessary. All installs will be saved by default. You can prevent saving with `--no-save`. Installing optional and dev deps is unchanged: use `-D/--save-dev` and `-O/--save-optional` if you want them saved into those fields instead. Note that since npm@3, npm will automatically update npm-shrinkwrap.json when you save: this will also be true for `package-lock.json`. (#15666)

* Installing a package directory now ends up creating a symlink and does the Right Thing™ as far as saving to and installing from `npm-shrinkwrap.json` goes. If you have a monorepo, this should make things much easier to work with, and probably a lot faster too. 😁 (#15900)

* git dependencies now support semver-based matching: `npm install git://github.com/npm/npm#semver:^5` (#15308, #15666)

* git dependencies with `prepare` scripts will now [have their devDependencies installed, and their prepare script executed](https://twitter.com/maybekatz/status/860363896443371520) as if under `npm pack`.

* node-gyp now supports `node-gyp.cmd` on Windows (#14568)

* npm no longer blasts your screen with the whole installed tree. Instead, you'll see a summary report of the install that is much kinder on your shell real-estate. Specially for large projects. (#15914):
```
$ npm install
npm added 125, removed 32, updated 148 and moved 5 packages in 5.032s.
$
```

* `--parseable` and `--json` now work more consistently across various commands, particularly `install` and `ls`.

* Indentation is now [detected and preserved](https://twitter.com/maybekatz/status/860690502932340737) for `package.json`, `package-lock.json`, and `npm-shrinkwrap.json`. If the package lock is missing, it will default to `package.json`'s current indentation.

#### Publishing

* New [publishes will now include *both* `sha512`](https://twitter.com/maybekatz/status/863201943082065920) and `sha1` checksums. Versions of npm from 5 onwards will use the strongest algorithm available to verify downloads. (https://github.com/npm/npm-registry-client/pull/157)

#### Cache Rewrite!

We've been talking about rewriting the cache for a loooong time. So here it is. Lots of exciting stuff ahead. The rewrite will also enable some exciting future features, but we'll talk about those when they're actually in the works. #15666 is the main PR for all these changes. Additional PRs/commits are linked inline.

* package metadata, package download, and caching infrastructure replaced

* It's a bit faster. [Hopefully it will be noticeable](https://twitter.com/maybekatz/status/865393382260056064). 🤔

* With the shrinkwrap and package-lock changes, tarballs will be looked up in the cache by content address (and verified with it)

* Corrupted cache entries will [automatically be removed and re-fetched](https://twitter.com/maybekatz/status/854933138182557696) on integrity check failure.

* npm CLI now supports tarball hashes with any hash function supported by Node.js. That is, it will [use `sha512` for tarballs from registries that send a `sha512` checksum as the tarball hash](https://twitter.com/maybekatz/status/858137093624573953). Publishing with `sha512` is added by https://github.com/npm/npm-registry-client/pull/157 and may be backfilled by the registry for older entries.

* remote tarball requests are now properly cached. This means that even if you're missing the `integrity` field in your `npm-shrinkwrap.json`, npm will be able to install from the cache.

* Downloads for large packages are streamed in and out of disk. npm is now able to install packages of """any""" size without running out of memory. Support for publishing them is pending (due to registry limitations).

* [Automatic fallback-to-offline mode](https://twitter.com/maybekatz/status/854176565587984384). npm will seamlessly use your cache if you are offline, or if you lose access to a particular registry (for example, if you can no longer access a private npm repo, or if your git host is unavailable).

* A new `--prefer-offline` option will make npm skip any conditional requests (304 checks) for stale cache data, and *only* hit the network if something is missing from the cache.

* A new `--prefer-online` option that will force npm to revalidate cached data (with 304 checks), ignoring any staleness checks, and refreshing the cache with revalidated, fresh data.

* A new `--offline` option will force npm to use the cache or exit. It will error with an `ENOTCACHED` code if anything it tries to install isn't already in the cache.

* A new `npm cache verify` command that will garbage collect your cache, reducing disk usage for things you don't need (-handwave-), and will do full integrity verification on both the index and the content. This will also be hooked into `npm doctor` as part of its larger suite of checking tools.

* The new cache is *very* fault tolerant and supports concurrent access.
  * Multiple npm processes will not corrupt a shared cache.
  * Corrupted data will not be installed. Data is checked on both insertion and extraction, and treated as if it were missing if found to be corrupted. I will literally bake you a cookie if you manage to corrupt the cache in such a way that you end up with the wrong data in your installation (installer bugs notwithstanding).
  * `npm cache clear` is no longer useful for anything except clearing up disk space.

* Package metadata is cached separately per registry and package type: you can't have package name conflicts between locally-installed packages, private repo packages, and public repo packages. Identical tarball data will still be shared/deduplicated as long as their hashes match.

* HTTP cache-related headers and features are "fully" (lol) supported for both metadata and tarball requests -- if you have your own registry, you can define your own cache settings the CLI will obey!

* `prepublishOnly` now runs *before* the tarball to publish is created, after `prepare` has run.

* ...and more!

### For release task list

* [x] First downstreaming to Node.js 8 (for 5/9)
* [x] Update docs for new/changes features
  * [x] `prepare` on git dependency installation
  * [x] `package-lock.json` vs `shrinkwrap`
  * [x] Auto-save
  * [x] sha512 on new publishes
* [x] Land file specifier branch (including idempotent package-lock and many other fixes)
  * [ ] Last work on file specifier branch (complete spec tests)
  * [ ] details! links to links, 
  * [ ] details! removing links
* [ ] Improve behavior for shrinkwrap/package-lock
  * [x] warn if both files are there, preferring shrinkwrap
  * [x] Add new fields from spec: `created-with`, `shrinkwrap-version`, `package-integrity`, 
  * [x] full new install semantics around dependency fields, based on shrinkwrap version
  * [x] warn if the shrinkwrap is an incompatible version
    * [ ] for npm@5
  * [x] update link spec doc to match shrinkwrap
  * [ ] Stop generating shrinkwrap/lockfiles off the filesystem -- calculate the tree w/ the installer
* [x] Make `npm update` save files as right type: devDeps end up getting saved as deps
* [ ] Leave uninstallable optionals _in the tree_ in such a way that they don't get installed by `diff-trees` but they do end up in the lock.
* [ ] Ensure `npm install --production` doesn't remove dev deps from the lock.
* [x] Add `notice` and `timing` log levels directly to npmlog
* [x] Land patch about running `prepare` when installing git dependencies.
* [x] Fix cross-protocol redirects: https://github.com/zkat/make-fetch-happen/issues/26


siddharthkp:
@zkat Will any of these flags feature in the defaults?

Some folks bump into addition flags but it doesn't reach everyone (example: `--cache-min`)
`--prefer-offline` is something that would make npm installs faster for everyone and probably should be a default.

I understand that it breaks expectations on how people think npm works right now. But, do you think major version is probably the time to change this expectation?

zkat:
@siddharthkp There's no significant speed difference between `--prefer-offline` and `--prefer-online` when I actually test it (I'm surprised too). And I mean that -- the two installs on the repo I do my usual benchmarks are within 1 second of each other, for a minute-long install. Those 304 check are _very fast_, and pacote/make-fetch-happen are both very very fast at what they do when it comes to networking.

I see you mentioned that you think users would expect `--prefer-offline` to be the default, but I don't think that's true in practice: Many/most users install newer versions of packages, for example, by doing `npm i -S pkg`. In this case, using `--prefer-offline` will not upgrade anything: `latest` is already locally cached. `pacote.manifest()` will be satisfied with that, and the user will receive a very stale version of the package manifest. The intention behind `--prefer-offline` is to give users an option to reduce network usage in-general, if they're aware of the consequences of doing do.

Now, if you're looking for a speedup, the biggest one you can get with `npm@5` is to use `npm-shrinkwrap.json` -- as of this patch, shrinkwrap will include content addresses for all package data, which means that as long as you have a warm cache, the network will not be used at all, and the entire package manifest phase will be completely skipped (because we don't need manifests if we have a shrinkwrap). The speedup from that is often on the order of 2x a non-shrinkwrap cached `--prefer-offline` install.

So, no, I probably won't use npm@5 to make `--prefer-offline` the default, since I'm not sure there's a reason to, in practice. Making npm-shrinkwrap be a default, though... that's in the works. Maybe not for npm@5 but I'm pushing for it for sure.

siddharthkp:
> There's no significant speed difference between --prefer-offline and --prefer-online when I actually test it (I'm surprised too).

😮 
What were the network conditions like in your benchmarks?

I had [a conversation](https://twitter.com/siddharthkp/status/824389590807048192) with @iarna a while back (npm 3) where she educated me about the network requests to the registry even when using cache + the effect of latency on it. (Significant for emerging nations - prominently 2G and latency heavy nations - Australia and such)

Are you saying pacote nullifies that effect (or at least makes it not noticeable) because that would be 💯 

> I see you mentioned that you think users would expect --prefer-offline to be the default

Ouch! My bad, I meant `--prefer-online`. Honest mistake.


> The speedup from that is often on the order of 2x a non-shrinkwrap cached --prefer-offline install.

This is another interesting part 🤔
Is this because npm still needs to resolve the dependency tree with `--prefer-offline` which means network requests? Not the same with shrinkwrap.

> Making npm-shrinkwrap be a default, though... that's in the works. 

💯 Is there a github issue that I can upvote for this?

zkat:
> What were the network conditions like in your benchmarks?
I'm on a Bay Area network, which is admittedly absurdly good. If the difference is so small now, though, I imagine it wouldn't be so big: even at 1/10th the speed, that would still be barely 10 seconds added (to a 60-second installation).

> Are you saying pacote nullifies that effect? (or at least makes it not noticeable) because that would be 💯
I really can't say that for sure right now. tbh, the difference is so surprisingly small that I think I'm gonna have to spend some time crawling through tests and making sure I don't have some bug. That code definitely has tests attached, though...

>Ouch! My bad, I meant --prefer-online. Honest mistake.
Ah. As far as this one goes, pacote obeys cache age settings from a given registry. This is generally happy news for all involved (if we hit the registry less without degrading user experience, that's a win!), and those settings tend to be fairly low: the main npm registry sets `Cache-Control: max-age=300`, which is in seconds (5 minutes). After those 5 minutes, we'll do the 304 checks again. This seems reasonable to me, specially with the `--prefer-offline` escape hatch. Anyway, no worries. They're kinda easy to get confused 😁 

 I _would_ care more if it weren't for shrinkwrap being the main target of optimization, which brings me to...

> Is this because npm still needs to resolve the dependency tree
Pretty much this. If we don't have the exact hashes for the tarballs we want to end up installing, npm needs to repeat the process of `lookup by name --> resolve version --> get shasum + dependency names from version --> repeat for all recursive dependencies`. That's a pretty heavy process.

> 💯 Is there a github issue that I can upvote for this?
Not quite yet. There's [a github project for npm@5](https://github.com/npm/npm/projects/1#card-2362037) and one of my TODOs is to write a proposal for how shrinkwrap-by-default would work. From what I've talked with @iarna, though, we need to fix other stuff (like `npm update`) before we even consider this. So, it'll probably be specced out after npm@5 is out and we're ready to start thinking about npm@6.

siddharthkp:
If shrinkwrap by default is likely to come in npm6, does it make sense to make `--prefer-offline` a default in npm5?
Or is that going in the wrong direction?

Kuroljov:
Any suggestions why yarn is so fast? I really expect npm@5 to be as fast as yarn. How can I help?

siddharthkp:
@Kuroljov With npm-shrinkwrap, npm will be in serious competition with yarn.

@zkat Can we start playing around with that somewhere? (Can contribute to your benchmarks?)

zkat:
@Kuroljov This release addresses a lot of performance issues. Right now, npm@5 installs are running anywhere between 2x-5x faster than npm@4. There's a few optimizations left that I don't know if we'll have ready by npm@5, but that number is bound to get significantly bigger (as in, 30%+ from one of those changes)

Some sample benchmarks:

```
react boilerplate install (including lifecycle scripts):
npm@4 i (no cache): 147.23s
npm@4 i (warm cache): 145.70s
npm@5 i (no cache): 74.30s
npm@5 i (warm cache): 70.85s
npm@5 i (no cache, no shrinkwrap): 99.83s
```

```
~/code/npm-website $ npm i --ignore-scripts
warm cache:
npm@4.4.0     - 150s (?!?!?!? 🤔)
npm@5.0.0 - 52s
npm@5.0.0 w/ shrinkwrap   - 38s
```

zkat:
@siddharthkp expect a beta release of npm@5 in the coming weeks. I can't give you the exact date, but I'm currently in the process of reviving npm's massive test suite after such a big change and fixing all my little messups along the way.

What I can tell you is that if you make https://npm.im/make-fetch-happen and https://npm.im/cacache faster and better, it'll be directly reflected on the performance of npm@5. Once https://npm.im/pacote gets its `@2` release (probably shortly before `npm@beta`), the same will apply to that. Those first two projects are relatively stable and usable, though, and I think people can definitely already start contributing significantly to them. There's also https://npm.im/ssri but I don't think that's much of a bottleneck at all rn.

I'll make an announcement somewhere dev-oriented once `npm@beta` is out, cause I definitely want early adopters to help me find all the things that broke during the rewrite -- because it's a big one.

zkat:
@siddharthkp coming back to this: what are the network conditions where you're at? (since you asked). I'd like to get a better, more realistic perspective on how npm5 handles slow and/or unstable networks and special configurations like proxies, which is all fairly hard to simulate unless you've been deep in that situation, imo. There's now a (semi-closed) beta of npm5 that you can install with `npm i -g 'npm/npm#release-beta-5` that will give you an `npm5` binary that lives alongside your standard installation. It would be super awesome if folks who are willing to look into weirdness tried it out and gave feedback. I very much want it to be a smooth, seamless experience even in bad network conditions: the worst thing that should happen on unreliable networks is that your download will be slightly slower. Everything else should Just Work™.

siddharthkp:
I'm sitting in India, and have a pretty good connection in the city I'm in. Most smaller cities don't. 

Can benchmark with network throtelling though? (don't know to simulate flakiness)

I think you need someone in Australia for testing high latencies

zkat:
@siddharthkp deffo australia 😭 

Feel free to test away, though. I'm really hoping you'll see some real improvement regardless of connection quality :)

iarna:
We actually have a registry mirror in Australia, so I'd expect it to be plenty fast there. (Geo-ip is used to pick the one closest, network-wise, to you.)

jdalton:
I'm super interested in the perf on Windows, out of the box yarn loses much of its perf wins on Windows unless you deny list some directories in Windows Defender.

zkat:
@jdalton try it out! I'll poke around on windows later and see how it compares to npm@4 probably. Getting some bugfixes out of the way 😁

zkat:
@jdalton ok I ran the numbers, against react-native, which seems to be a popular one to try. These are both done assuming a corresponding lockfile, since I think that's a fair comparison when it comes to npm5 in particular (I'd bench without a lockfile for npm4, which I'm not even trying for this).
```
cold cache:
npm@5 - 45s
yarn@0.23.2 - 75s

warm cache:
npm@5 - 40s
yarn@0.23.2 - 43.5s
```

That's neat! huh.

zkat:
(incidentally, I haven't done anything involving Windows Defender so idk)

jdalton:
@zkat See [the windows install notice](https://github.com/yarnpkg/website/commit/06b063e5e78afa7ddcb56c75cb1e5e12a749f7da) for more info.

code-chris:
Is it still planned to ship this with Node 8?

zkat:
@code-chris that's the intention, yes. We've talked to the node core folks and we're aiming to have a usable version by May 4th. Take that with a grain of salt, though. Deadlines love slipping, and the goal here is to have a good, solid release. 👍

felixfbecker:
Can't wait to try this out - could we get a prerelease (alpha/beta/rc)?

corbinu:
@felixfbecker 

> `npm i -g 'npm/npm#release-beta-5`

felixfbecker:
Would prefer an official prerelease on npm over a git branch dependency.

My use case is programmatic usage and I'm currently using yarn (which feels like beta software in a way too and is 0.x) because I need the speed. This may make be switch back.

zkat:
@felixfbecker May 4th is when there will probably be an official registry-installable beta :) Probably.

corbinu:
@felixfbecker Totally understood sorry but I am just an excited observer also :)

@zkat Thanks that great news!

Yes I agree yarn still has some serious bugs (in my experience especially around global installs). In just a few tests on my laptop it looks like yarn is 2 times faster than npm4 but only 30% faster than npm5. 

Not to get on my soap box but FB kind has a history of this (drop HHVM rather than contribute to PHP ..... PHP 7.0 takes all the good parts and is just as fast)

francoispluchino:
@zkat On Windows 10, the command `npm5 install` throw the exception: `This is most likely not a problem with npm itself and is related to insufficient space on your system`.

**Node:** v7.9.0 64bit
**NPM:** v5.0.0-beta.9

I have more than 200go of free on my disk, and I have no problem with `npm@4.5.0`.

zkat:
@corbinu try doing `npm shrinkwrap` -- npm5 is intended to be used with a lockfile

@francoisfrisch awesome! Can you post the debug log in a gist and link it?

corbinu:
@zkat Thanks I will try that for sure!

francoispluchino:
@zkat Logs are [here](https://gist.github.com/francoispluchino/3b5d86e59eda60bdcbe72a2333a97679).

The bug happens when I use a dependency not referenced in the registry.

Example:
```json
{
  "name": "npm5-bug",
  "dependencies": {
    "jquery": "^2.1.0",
    "footable": "git+ssh://git@github.com/fooplugins/FooTable.git#2.0.3"
  }
}
```
It's good with `npm@4.5.0`.

zkat:
@francoisfrisch thanks! This looks like an issue with spawning git. I noticed your `cwd` is `D:\test`. I wonder if that has anything to do with it 🤔 .

I'll reboot into Windows and see if I can hunt this one down. Woo!

zkat:
@francoisfrisch pretty sure I've got a fix. Thanks again. Windows and uid stuff don't get along 😂

corbinu:
@zkat I ran some more detailed tests (ran each test 3 times). They were run on 

- 2012 Macbook Air i7 with Sierra
- Node v7.9.0
- npm 4.6.1
- npm5 5.0.0-beta.9 
- yarn 0.23.1

They are with and without shrinkwrap (though I do want to ask yarn says to use lock files for applications and libraries are we supposed to use shrinkwrap for all npm projects including libraries?)

npm 4 at times is actually more like 160% to 460% times slower, but npm 5 is 40% to 140%. Keep up the good work I am excited! I did also include the install times for when already installed which I know is less important. I think whats going on there is npm is doing much more checking that the tree is correct.

I may right a small tool that would automate this kind of benchmarking for any module. (In Rust actually as I am trying to practice that language.) I have attached my deps below. Let me know if I should post later results in a separate issue or here.

| Installer | Status      | Time     | Shrinkwrap/locked | Pref  |
|-----------|-------------|----------|-------------------|-------|
| npm4      | cache clean | 01:33.48 | N                 | 336%  |
| npm5      | cache clean | 01:00.12 | N                 | 216%  |
| npm4      | cache clean | 01:10.51 | Y                 | 253%  |
| npm5      | cache clean | 00:37.80 | Y                 | 136%  |
| yarn      | cache clean | 00:27.82 | Y                 | 0%    |
| npm4      | cache hot   | 01:18.38 | N                 | 461%  |
| npm5      | cache hot   | 00:40.04 | N                 | 236%  |
| npm4      | cache hot   | 01:12.56 | Y                 | 427%  |
| npm5      | cache hot   | 00:31.64 | Y                 | 186%  |
| yarn      | cache hot   | 00:17.00 | Y                 | 0%    |
| npm4      | installed   | 00:12.93 | N                 | 1372% |
| npm5      | installed   | 00:15.26 | N                 | 1619% |
| npm4      | installed   | 00:13.24 | Y                 | 1405% |
| npm5      | installed   | 00:16.34 | Y                 | 1734% |
| yarn      | installed   | 00:00.94 | Y                 | 0%    |


```
"dependencies": {
    "axios": "^0.16.0",
    "es6-promise": "^4.1.0",
    "lodash": "^4.17.4",
    "store2": "^2.5.0",
    "vue": "^2.2.2",
    "vue-axios": "^2.0.1",
    "vue-my-dropdown": "^2.0.3",
    "vue-resource": "^1.2.1",
    "vue-select": "^2.1.0"
  },
  "devDependencies": {
    "@corbinu/eslint-plugin-corbinu": "^2.0.0",
    "autoprefixer": "^6.7.2",
    "babel-core": "^6.22.1",
    "babel-eslint": "^7.2.3",
    "babel-loader": "^7.0.0",
    "babel-plugin-transform-runtime": "^6.22.0",
    "babel-preset-env": "^1.2.1",
    "babel-preset-es2015": "^6.24.0",
    "babel-preset-stage-2": "^6.22.0",
    "babel-register": "^6.22.0",
    "chalk": "^1.1.3",
    "connect-history-api-fallback": "^1.3.0",
    "copy-webpack-plugin": "^4.0.1",
    "css-loader": "^0.28.0",
    "eslint": "^3.14.1",
    "eslint-friendly-formatter": "^2.0.7",
    "eslint-loader": "^1.6.1",
    "eslint-plugin-html": "^2.0.0",
    "eventsource-polyfill": "^0.9.6",
    "express": "^4.14.1",
    "extract-text-webpack-plugin": "^2.0.0",
    "file-loader": "^0.11.1",
    "friendly-errors-webpack-plugin": "^1.1.3",
    "function-bind": "^1.1.0",
    "html-webpack-plugin": "^2.28.0",
    "http-proxy-middleware": "^0.17.3",
    "opn": "^4.0.2",
    "optimize-css-assets-webpack-plugin": "^1.3.0",
    "ora": "^1.1.0",
    "rimraf": "^2.6.0",
    "semver": "^5.3.0",
    "typescript": "~2.2.0",
    "url-loader": "^0.5.7",
    "vue-loader": "^11.1.4",
    "vue-style-loader": "^3.0.0",
    "vue-template-compiler": "^2.2.1",
    "webpack": "^2.2.1",
    "webpack-bundle-analyzer": "^2.2.1",
    "webpack-dev-middleware": "^1.10.2",
    "webpack-hot-middleware": "^2.16.1",
    "webpack-merge": "^4.0.0"
  }
```

zkat:
@corbinu great! Thanks!

The profile is bound to change a little more in the next couple of weeks as other features and bugfixes land (see also #16441, for example). But this does seem like pretty good progress overall 👍

corbinu:
@zkat So correct me if I am wrong then but the answer to my above question 

> though I do want to ask yarn says to use lock files for applications and libraries are we supposed to use shrinkwrap for all npm projects including libraries?

Is shrinkwrap is for applications that are being distributed and project-lock will be for libraries/keeping developer installs in sync across machines?

Yes looks like great progress for sure! I will make sure to at least rerun as that commit alone should make npm 5 go from 40%-140% slower to 40%-90% :+1:

zkat:
@corbinu you should use shrinkwrap if you want your published library to have a very specific tree under your library. this is usually very rare.

corbinu:
so we can just pretty much always use project-lock... sweet!

felixfbecker:
I'm a bit sad to see that apparently yarn is still faster by ~10s.

@corbinu the benchmark is missing yarn without yarn.lock?

corbinu:
@felixfbecker I didn't include it as only time I know of that the lockfile is not supposed to be there is when yarn is first used to install a project. Where as shrinkwrap is opt in. If project-lock and shrinkwrap performance turn out to be the same once that drops. Then I will probably drop that column entirely. Unless you know of a reason somebody would blow away their lockfile?

felixfbecker:
@corbinu Exactly, when it is first used to install a project. That is the common case for my programmatic usage. I would appreciate it if you could include that case.

corbinu:
@felixfbecker No thats not what I am saying ... it would only be when first installing using yarn. Normal install of a project would already have it as would be checked into a git repo. For example a CI run should never create a lockfile. All projects are supposed to have a lock including libraries https://yarnpkg.com/blog/2016/11/24/lockfiles-for-all/

felixfbecker:
That's debatable, but still there are tenthiusands of projects out there who don't have a committed yarn.lock (they may use shrinkwrap or nothing). The "first installation" without lockfile is a valid use case (on https://sourcegraph.com it's the common case) and should be included for completeness of the benchmark

victorwpbastos:
Please, don't let `npm install --save` mess with indentation in `package.json`. This would be awesome!

sindresorhus:
@victorwpbastos Yarn preserves the indentation.

mathiasbynens:
@victorwpbastos See #5694 and others.

simonua:
@victorwpbastos, I'm back-seat-driving here, and this may not be relevant to your concern, but you can always exclude `package.json` and `npm-shrinkwrap.json` from formatting by using tools such as [EditorConfig](http://editorconfig.org/). It's actually quite common to see `.editorConfig` files in repos that specifically exclude those two files and maintain formatting at 2 spaces. Again, I'm sorry if this doesn't quite hit the mark, but I thought it may be valuable to know.

sindresorhus:
@simonua That's exactly *why* it's annoying. You have to add a rule explicitly for `package.json` to every single `.editorconfig` file, just because npm is stubborn.

simonua:
@sindresorhus I hear you. It's the prerogative of the creator, but fortunately, it is so easily mitigated. There are bigger fish to fry. =)

zkat:
@sindresorhus @simonua 

I talked to @iarna. We'll take a patch that will automatically detect and duplicate whatever indentation is in these files. Are either of you interested in sending such a patch our way? I can tell you where these files are written if you have a hard time finding them.

Thanks for bringing it up. Cheers!

simonua:
@zkat @sindresorhus, sounds awesome. Looks like [just a few locations](https://github.com/npm/npm/search?utf8=%E2%9C%93&q=JSON.stringify&type=) where it is writing out the file (and with the expected 2 space hard-coded indentation).

Does this capture the gist? 
1. On any open/read of `package.json` or `npm-shrinkwrap.json` determine the current indentation. Keep track of that indentation in context (maybe globally, if necessary)
1. On any write with hard-coded spaces, default to previously-determined indentation `|| 2`.

I'm not sure I can commit to it as I am tied up the next few days, but maybe this can help someone.

zkat:
@simonua I believe you *only* need to track down the places where they're written. I believe those places all _read_ the file _before_ writing it. So you can do the indentation calculation on that read. (if it's not already reading right before writing, go ahead and do a read just to figure out what indentation it's using and throw away the data).

drewfish:
At yahoo we have a custom npm client (alas) and it does preserve indenting when it modifies package.json. It uses the format-package-json npm package to do this. I've seen it used something like this:
```javascript
var format = require('format-package-json');
function write(data, cb) {
    var file = path.join(process.cwd(), "package.json");
    format.read(file, function (err, options) {
        format.write(file, data, options, cb);
    });
}
```

zkat:
@drewfish thanks! It looks like it uses https://npm.im/detect-indent (oh hey it's @sindresorhus ware!)

Using that lower-level one is probably more than enough to solve this particular bit :)

zkat:
Update: I just did it. `beta.42` now preserves indentation :) cc @sindresorhus @simonua

simonua:
You're on a roll, @zkat! Get some rest and enjoy your weekend. 😁

Cheers, 

Simon

felixfbecker:
Could someone explain the difference between package-lock.json and npm-shrinkwrap.json?

corbinu:
@felixfbecker zkat can correct me if I am wrong but based on the questions I asked and the PR my understanding is that package-lock.json will now be autogenerate and used all the time. Its purpose is the same as a yarn.lockfile. It will speedup installs and make sure that developers get the same versions of modules installed. So you just check it into your repo and enjoy the benefits. 

npm-shrinkwrap.json is manually generated and it will actually be used much less now. I am simply going to quote @zkat from above 

> you should use shrinkwrap if you want your published library to have a very specific tree under your library. this is usually very rare.

So basically npm is getting an autogenerated lockfile similar to yarn/bundler/cargo and npm-shinkwrap is now obsolete for most of the current use cases.

felixfbecker:
Afaik everything you just described applies to npm-shrinkwrap.json as well... I tested both on a project, and the generated npm-shrinkwrap.json on first sight looks identical to the generated package-lock.json. So what is the point of introducing a new file? Why not simply generate npm-shrinkwrap.json by default?

corbinu:
@zkat please correct me if I'm wrong but the difference is It is manually generated you choose how and when shinkwrap gets updated. Where as the new lockfile gets updated as NPM commands are run. So shrinkwrap would be used to manually pin the deps before a release for example. Those use cases are relatively small though. So to answer "why not simply autogenerate shrinkwrap" is because they would critically break a small number of users workflow. Sounds like you and I aren't those uses though so we can pretty much go forward pretending shrinkwrap doesn't exit anymore :)

felixfbecker:
I just ran `npm5 install` on our project and it updated our npm-shrinkwrap.json automatically
```diff
W
 npm-shrinkwrap.json | 5262 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++------------------------------------------------------------------------------------
 1 file changed, 2631 insertions(+), 2631 deletions(-)
```

corbinu:
Sounds like you found a bug I don't believe shinkwrap behavior is supposed to have changed like that :(

zkat:
@felixfbecker that is correct behavior. The two files are essentially identical. You can think of shrinkwrap as the "publishable mode" for package-lock.json. If you have a shrinkwrap, we'll interact with that file automatically, instead of generating a package lock. If you want to use the package-lock.json (which cannot be published), just `git mv npm-shrinkwrap.json package-lock.json`.

I know it seems a little weird, but a large part of this is due to backwards compatibility + a few people who actually want to _release_ things with a shrinkwrap attached to it for various reasons. Having a new filename for the autogenerated version both improved people's understanding of what the file does (we've technically had a "lockfile" since like npm@1!) and makes it easier for us to autogenerate the file without accidentally borking the ecosystem, since having an `npm-shrinkwrap.json` prevents deduplication.

joshuakelly:
When is an official release likely guys?

hawkrives:
@joshuakelly It's slated to ship with node 8, later this month.

zkat:
@joshuakelly literally none of the devs you're asking this from are guys ;)

That said, May 30th is our target date. We might have it ready a bit sooner.

rmg:
@felixfbecker said:
> I just ran `npm5 install` on our project and it updated our `npm-shrinkwrap.json` automatically

@zkat said:
> @felixfbecker that is correct behavior.

@zkat, can you clarify? Maybe you weren't responding to the latest comment by @felixfbecker, since your explanation better matches his _previous_ question?

We use `npm-shrinkwrap.json` _heavily_ in published packages, so I just want to know if that's an intended change in behaviour or a bug as @corbinu suggested.

zkat:
@rmg you can continue using `npm-shrinkwrap.json` as usual. Since npm now saves by default, it means your shrinkwrap will be updated automatically for all operations.

If you don't intend to _publish_ your `npm-shrinkwrap.json` -- that is, if you only ever use it when pulling from git, then you might want to consider renaming it to `package-lock.json`.

Both files are the same thing, but the format is somewhat changing (some fields added, some fields removed). That means that the first time you run `npm5`, your existing shrinkwrap will be rewritten to the new format. After that, you should only see diffs for exactly the things you add/remove.

The last bit is that, since `npm-shrinkwrap.json` and `package-lock.json` are essentially the same file, npm _will not write_ a `package-lock.json` if `npm-shrinkwrap.json` already exists. It will simply do the exact same operations, but on the latter file.

The reasons for still having shrinkwrap around are:
1. lockfiles should not be publishable by default (`package-lock.json` can never be published). Otherwise, we wouldn't be able to flatten. *some* users still want to enforce exact trees for their libraries: they can continue using `npm-shrinkwrap.json` for that.
2. `npm-shrinkwrap.json` is recommended if you're creating a CLI tool (or anything you're supposed to install with `-g`), since it'll guarantee the install will be reproducible for your users.
3. There exist legacy shrinkwraps out in the wild. These should continue to work as expected.
4. Using an `npm-shrinkwrap.json` instead of `package-lock.json` will allow you to *intentionally* force package tree shape for npm versions all the way back to `npm@1`. If you need users to install your thing exactly, from any npm version, you can use shrinkwrap.

Does that answer your question?

rmg:
@zkat as I'm someone who works almost exclusively with published packages and _not_ applications, `npm-shrinkwrap.json` currently does exactly what I want.

What worries me is that `npm install` (as in no args, just part of normal dev process) will _update_ it without being asked to. This would be a very confusing way to start working on back-porting a fix to a shrinkwrapped release branch:
```
$ git clone git@github.com:org/some-package.git
$ cd some-package
$ git checkout v1.2.x
$ npm install-test
...
$ git diff --stat
npm-shrinkwrap.json | 5262 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++------------------------------------------------------------------------------------
 1 file changed, 2631 insertions(+), 2631 deletions(-)
```

Part of my release process also involves stripping out the `resolved` properties from `npm-shrinkwrap.json` to remove references to any internal npm registries/proxies, which is probably something else that might need updating if the format is changing.

I'm fine with `package-lock.json` being different, since that's clearly intended for application developers.

zkat:
@rmg the only major difference for you here is the autosave behavior. `--no-save` will prevent it from changing the shrinkwrap if you explicitly don't want it to.

As far as switching registries: I just pushed `beta.49`, which makes npm agnostic about which registry is in your `resolved` field. It won't try to fetch that specific tarball unless your current registry setting matches it. So you should be able to have portable shrinkwraps without having to do anything special.

rmg:
> So you should be able to have portable shrinkwraps without having to do anything special.

It is also only advisory, even in the current releases. The sanitization is more for removing potentially sensitive information from published packages :-)

felixfbecker:
@zkat does the package-lock.json got any other tricks in its sleeve? For example, can it handle `optionalDependencies`? Right now if you have `fsevents` in your `optionalDependencies` and shrinkwrap on a Mac, it will be included in the npm-shrinkwrap.json. When you commit and run in your Linux CI the install will fail. On the other hand, if you install with `--no-optional` and shrinkwrap, no Mac will ever install `fsevents` because it's not in the npm-shrinkwrap.json and running any file watching script will heat up your Mac enough to significantly contribute to global warming.

zkat:
@felixfbecker it's not in yet, but optionalDeps are going to be saved into the shrinkwrap unconditionally, and handled on a per-platform basis from there.

svipben:
@zkat Hi, I have few questions.

1. So `package-lock.json` is the same as `yarn.lock`? 
2. I always publish `yarn.lock` to Git, it will be good practice to publish `package-lock.json` to Git too?
 
I'm considering to move away from Yarn to NPM, since it have better performance (talking about NPM 4 and NPM 5) and it's installed together with Node.js.

zkat:
@svipben hi! thanks for your questions:

> So `package-lock.json` is the same as `yarn.lock`?

No, it contains information that `yarn.lock` does not contain, which adds the ability for us to generate deterministic trees across versions (or, if supported, across package managers). The file still has the same purpose, though: to represent a specific interpretation of the manifest (`package.json`) such that different developers will end up with identical `node_modules`. You can read more about it over at https://github.com/npm/npm/pull/16441.

> I always publish `yarn.lock` to Git, it will be good practice to publish package-lock.json to Git too?

Yes. This is the expected workflow. There is an additional note: If you are developing a _command line tool_, such as `grunt-cli`, `webpack`, etc, you should consider using the `npm shrinkwrap` command to switch to `npm-shrinkwrap.json` and add it to your published files. This file is essentially the same as `package-lock.json` but allows you to actually publish it: that way, global installations will have the same determinism guarantees that you get during development.

It is not recommended you use `npm-shrinkwrap.json` for regular JS libraries, since that will defeat tree flattening algorithms/deduplication, as well as prevent you from getting automatic bugfixes on transitive dependencies (which will still only happen when you update the parent lib, and which you can still manage/track using the lockfile)

svipben:
@zkat Thank you very much for information! :+1:

simonua:
The ability to install all node modules from a hot cache and with a proper _package-lock.json_ is awesome! I disconnected my laptop from the network, wiped my local _node_modules_ folder, then ran an install. Everything was put in place properly and working with no network access needed. Very cool!!

corbinu:
Just committed package-lock.json to all my repos! Going to try to just use npm5 the test of this week :) Amazing work developers!

On a side note I spent the last 3 days looking for an informal gender neutral alternative to "guys" which is only gender neutral when spoken aloud and came up with "people" or "comrades". If a suggested list of pronouns for the community could be added to the code of conduct I know I at least would appreciate it. Thanks!

zkat:
@corbinu "folks" and "y'all" are my go-to faves. Sometimes "friends", "FELLOW DEVELOPERS", etc :)

Hope you have a good time with it the rest of this week! We're almost over the finish line! 🎉  And thanks for the `package.json`. I've used it for testing/benchmarking a _lot_.

simonua:
Ditto on using npm5 latest beta. I already have packages that are installed, shrinkwrapped, and and consumed by build VMs running 4.5.0. There has not been any regression, and I look forward to upgrading the servers after npm5 will have been released a little while and take advantages of its features. Great job, folks!

corbinu:
@zkat Thanks I am actually going to point out folks a few places as that seems to have been missed (gotta love the English language as if it isn't complicated enough we find a hole and decide to fill it differently for speaking and writing haha)

No problem :) If you like I would be happy to try to put together a canary in the gold mine type repo for benchmarking/testing?

zkat:
This was tagged and released yesterday as `npm@5` 🎉 thanks everyone!

olegkorol:
Hi,

what is the new format of `npm-shrinkwrap.json`?
After updating npm to v. 5 it won't `npm install` anymore...

Thanks in advance!

PR #16296, opened 2017-04-09
Preserve the same name convention for command in docs

desfero:
Change all occurence of `<cmd>` to `<command>`, so now we will have the same name used in the whole project.


iarna (approved):
Looks good, thank you!

PR #16336, opened 2017-04-13
Don't ham-it-up when expecting JSON

bdukes:
I turned on the ham-it-up configuration setting, and then noticed that I received an error when running the npm-windows-upgrade module, because it was trying to parse the JSON from calling `npm view npm versions --json` and didn't know what to do with the easter egg text at the end.

This change disables the ham-it-up text when the json option is present.

### DIFF
--- a/bin/npm-cli.js
+++ b/bin/npm-cli.js
@@ -83,7 +83,7 @@
     if (er) return errorHandler(er)
     npm.commands[npm.command](npm.argv, function (err) {
       // https://www.youtube.com/watch?v=7nfPu8qTiQU
-      if (!err && npm.config.get('ham-it-up') && !npm.config.get('json')) {
+      if (!err && npm.config.get('ham-it-up') && !npm.config.get('json') && !npm.config.get('parseable')) {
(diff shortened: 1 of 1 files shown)


bin/npm-cli.js
```diff
@@ -83,7 +83,7 @@
     if (er) return errorHandler(er)
     npm.commands[npm.command](npm.argv, function (err) {
       // https://www.youtube.com/watch?v=7nfPu8qTiQU
-      if (!err && npm.config.get('ham-it-up')) {
+      if (!err && npm.config.get('ham-it-up') && !npm.config.get('json')) {
```
zkat: Could you also check for [`parseable`](https://docs.npmjs.com/misc/config#parseable) here? I think with that, you'll be covered for output-related stuff 👍

zkat (changes requested):
Thanks for this! I'm amused people are using this. It's in my default config too ;)

LGTM with a single comment. I'll approve the PR once that little thing's changed and we'll include it in this week's release 🎉

bin/npm-cli.js
bdukes: I've added that check, thanks @zkat!

zkat (approved):

zkat:
LGTM! 🎉

PR #16364, opened 2017-04-17
Shorten the ELIFECYCLE error message

### LINKED ISSUE #8821: run-scripts are too noisy while used in development
Hi,

As many other users, I'm starting to use `npm` as a development tool for simple projects. Now that the ecosystem provides enough matured `cli` tools to do most of the things we want, `gulp` or `grunt` can be overkill for small projects.

**tl;dr: `run-scripts` seems to be mostly addressed to package users, not developers. As developers, we would like to be able to use `npm` for development, and run in silent mode.**

As you already know, a `run-script` that exits with a non-zero error code triggers additional logging and a `npm-debug.log`. These are not necessary in a _development flow_ usage, as you:
- expect to understand what's going wrong, because you are the developer, not the user.
- expect tools to fail during development (lint, test, build, ...).
- process failures and report them with the adapted tool.

To counter this, developers can use the `|| true` trick which is **very dangerous**, or manually add `--silent` which is unnecessary verbose.

I totally understand the effort of communicating to the user that some script has failed and that is not related to `npm`. But I don't understand it anymore when it's addressed to the developer itself.
It seems that `run-scripts` usage is evolving in a way that was perhaps not intended by `npm`.

It would be awesome if we had a control on that log level somehow. Multiple solutions could be explored.
## dev-scripts

Would be the brother of `run-script` and invoked like this: `npm dev`. It would run in silent mode. The name can change, but it would be clear that these kind of scripts are used for development purposes only. It's like `devDependencies` but for scripts.
## more shortcuts

`test` for example is available as `npm test` and runs in silent mode (note that `npm run test` runs in verbose mode). Adding more shortcuts targeting most development use cases could do the job: `lint`, `build`, ...
## `package.json`: additional config variable

Like the existing `loglevel` variable but for `run-scripts`.
## `package.json`: deeper script property

Adding the possibility to specify a nested object like this for scripts:

``` json
"scripts": {
  "lint": {
    "command": "eslint lib/*.js",
    "loglevel": "silent"
  }
}
```
## support of an environment variable

Something like `NPM_RUNSCRIPT_LOGLEVEL=silent npm run lint` would be slient. We could setup our dev / ci environment accordingly.

---

We can choose one or several of those propositions, ~~but my personal taste would go in `dev-scripts` as it would be a clean and simple way of giving more control to the developer~~.

Thanks!

---

Refs: #5452, #5809, #3270.

j-f1:
Should fix #8821.

### DIFF
--- a/lib/utils/error-message.js
+++ b/lib/utils/error-message.js
@@ -33,7 +33,7 @@ function errorMessage (er) {
         '',
         [
           '',
-          'Failed at the ' + er.pkgid + ' ' + er.stage + " script.",
+          'Failed at the ' + er.pkgid + ' ' + er.stage + ' script.',
           'This is probably not a problem with npm. There is likely additional logging output above.'
         ].join('\n')]
       )
--- a/lib/utils/error-message.js
+++ b/lib/utils/error-message.js
@@ -33,16 +33,7 @@ function errorMessage (er) {
         '',
         [
           '',
-          'Failed at the ' + er.pkgid + ' ' + er.stage + " script '" + er.script + "'.",
-          'Make sure you have the latest version of node.js and npm installed.',
-          'If you do, this is most likely a problem with the ' + er.pkgname + ' package,',
-          'not with npm itself.',
-          'Tell the author that this fails on your system:',
-          '    ' + er.script,
-          'You can get information on how to open an issue for this project with:',
-          '    npm bugs ' + er.pkgname,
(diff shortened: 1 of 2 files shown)


iarna:
Thank you, this looks good!

iarna (approved):

legodude17:
If this was merged shouldn't it be closed? Or am I being a fool here?

j-f1:
This PR has been merged. Thanks @legodude17!

PR #16414, opened 2017-04-22
Fix spurious warning

zkat:
I'm not writing a test for this one 'cause it's mostly internal stuff that had no visible effect except a random warning that would pop up during save.


iarna (approved):
LGTM. 🐑

PR #16441, opened 2017-04-25
spec: Describe npm-shrinkwrap.json and package-lock.json

iarna:
Here is my proposal for how shrinkwraps in npm@5 will work. We will also be introducing a new file, `package-lock.json` that will fill the same role as `npm-shrinkwrap.json` in projects that don't otherwise have a shrinkwrap.

The TLDR is:

You'll use an `npm-shrinkwrap.json` if you want to publish it so that users of your module will consume it.
Otherwise you'll use a `package-lock.json`. `package-lock.json` files will never be published but are suitable for checking into git.

This is open for feedback for the next two weeks. (I mean, we'll take feedback whenever, but it won't be immediately actionable beyond that.)

### DIFF
--- a/doc/spec/package-lock.md
+++ b/doc/spec/package-lock.md
@@ -1,0 +1,302 @@
+# package-lock and npm-shrinkwrap
+
+`npm` can have one of two different lock files:
+
+* `package-lock.json`, which is ordinarily always present and is never published.
+* `npm-shrinkwrap.json`, which is created with `npm shrinkwrap` and usually published.
+
+You can only have one of them and in the event that you have both,
+`npm-shrinkwrap.json` takes precedence.  The files are exactly the same
+format and in fact all the `npm shrinkwrap` command does is rename your
+`package-lock.json`.
+
+Through the rest of this document we will refer to the package-lock and
+`package-lock.json` but everything also applies to `npm-shrinkwrap.json`.
+
+## File Format
+
+### name
+
+The name of the package this is a package-lock for. This must match what's in `package.json`.
+
+### version
+
+The version of the package this is a package-lock for. This must match what's in `package.json`.
+
+### createdWith *(new)*
+
+An object with the the following properties:
+
+* `installer` - The package manager used to create this `package-lock`.
+* `runtime` - The runtime used to run the installer.
+* `platform` - The platform used to run the installer.
+* `arch` - The architecture used to run the installer.
+
+The values of `installer` and `runtime` must be `name@version`.  The name part must be
+a valid npm package name.  The version part must be a valid semver version
+string. The values of `platform` and `arch` are strings. For example:
+
+```
+{
+  "installer": "npm@5.0.0",
+  "runtime": "nodejs@8.0.0",
+  "platform": "darwin",
+  "arch": "x64"
+}
+```
+
+Installers should not make decisions based on this field.  And if they do,
+they should be limited to working around known bugs in very specific
+versions.  The inteded purpose of this field is in helping with isolating
+the source of bugs.
+
+### lockfileVersion *(new)*
+
+An integer version, starting at `1` with the version number of this document
+whose semantics were used when generating this `package-lock.json`.
+
+### packageIntegrity *(new)*
+
+This is a
+[subresource integrity](https://w3c.github.io/webappsec/specs/subresourceintegrity/)
+value created from the `pacakge.json`. No preprocessing of the
+`package.json` should be done.  Subresource integrity strings can be
+produced by modules like
+[`ssri`](https://www.npmjs.com/package/ssri).
+
+### preserveSymlinks *(new)*
+
+Indicates that the install was done with the environment variable
+`NODE_PRESERVE_SYMLINKS` enabled. The installer should insist that the value of this
+property match that environment variable.
+
+### dependencies
+
+A mapping of package name to dependency object.  Dependency objects have the
+following properties:
+
+#### version *(changed)*
+
+This is a specifier that uniquely identifies this package and should be
+usable in fetching a new copy of it.
+
+* bundled dependencies: Regardless of source, this is a version number that is purely for informational purposes.
+* registry sources: This is a version number. (eg, `1.2.3`)
+* git sources: This is a git specifier with resolved committish. (eg, `git+https://example.com/foo/bar#115311855adb0789a0466714ed48a1499ffea97e`)
+* http tarball sources: This is the URL of the tarball. (eg, `https://example.com/example-1.3.0.tgz`)
+* local tarball sources: This is the file URL of the tarball. (eg `file:///opt/storage/example-1.3.0.tgz`)
+* local link sources: This is the file URL of the link. (eg `file:libs/our-module`)
+
+#### integrity *(new)*
+
+This is a [Standard Subresource
+Integrity](https://w3c.github.io/webappsec/specs/subresourceintegrity/) for
+this resource.
+
+* For bundled dependencies this is not included, regardless of source.
+* For registry sources, this is the `integrity` that the registry provided, or if one wasn't provided the SHA1 in `shasum`.
+* For git sources this is the specific commit hash we cloned from.
+* For remote tarball sources this is an integrity based on a SHA512 of
+  the file.
+* For local tarball sources: This is an integrity field based on the SHA512 of the file.
+
+#### resolved
+
+* For bundled dependencies this is not included, regardless of source.
+* For registry sources this is path of the tarball relative to the registry
+  URL.  If the tarball URL isn't on the same server as the registry URL then
+  this is a complete URL.
+
+#### link *(new)*
+
+If this module was symlinked in development but had semver in the
+`package.json` then this is the relative path of that link.
+
+Discussion of the semantics of this will go in the symlinks RFC.
+
+Implementation note: To be implemented post npm@5.
+
+#### bundled *(new)*
+
+If true, this is the bundled dependency and will be installed by the parent
+module.  When installing, this module will be extracted from the parent
+module during the extract phase, not installed as a separate dependency.
+
+#### dev
+
+If true then this dependency is either a development dependency ONLY of the
+top level module or a transitive dependency of one.  This is false for
+dependencies that are both a development dependency of the top level and a
+transitive dependency of a non-development dependency of the top level.
+
+#### optional
+
+If true then this dependency is either an optional dependency ONLY of the
+top level module or a transitive dependency of one.  This is false for
+dependencies that are both an optional dependency of the top level and a
+transitive dependency of a non-optional dependency of the top level.
+
+All optional dependencies should be included even if they're uninstallable
+on the current platform.
+
+#### from *(deprecated)*
+
+This is a record of what specifier was used to originally install this
+package.  This should not be included in new `package-lock.json` files.
+
+#### dependencies
+
+The dependencies of this dependency, exactly as at the top level.
+
+## Generating
+
+### `npm init`
+
+If neither a `package-lock.json` nor an `npm-shrinkwrap.json` exist then
+`npm init` will create a `package-lock.json`.  This is functionally
+equivalent to running `npm shrinkwrap` after the current init completes and
+renaming the result to `package-lock.json`.
+
+### `npm install --save`
+
+If either an `npm-shrinkwrap.json` or a `package-lock.json` exists then it
+will be updated.
+
+If neither exist then a `package-lock.json` should be generated.
+
+If a `package.json` does not exist, it should be generated.  The generated
+`package.json` should be empty, as in:
+
+```
+{
+   "dependencies": {
+   }
+}
+```
+
+If the user wants to get a default package name/version added they can run `npm init`.
+
+### `npm shrinkwrap`
+
+If a `package-lock.json` exists, rename it to `npm-shrinkwrap.json`.
+Refresh the data from the installer's ideal tree.
+
+The top level `name` and `version` come from the `package.json`.  It is an
+error if either are missing or invalid.
+
+`createdWith` should be filled in from `npm` itself.
+
+#### dependencies.dev
+
+This is `true` if this dependency is ONLY installed to fulfill either a top
+level development dependency, or one of its transitive dependencies.
+
+Given:
+```
+B (Dev) → C
+```
+
+Then both B and C would be `dev: true`.
+
+Given:
+```
+A → B → C
+B (Dev) -> C
+```
+
+Then all dependencies would be `dev: false`.
+
+#### dependencies.optional
+
+This is `true` if this dependency is ONLY ever either an optional dependency
+or a transitive dependency of optional dependencies.
+
+Given:
+```
+A (Opt) → B → C
+```
+
+Then all three of A, B and C would be flagged as optional.
+
+Given:
+```
+A (Opt) → B → C
+D → C
+```
+
+Then A and B would be flagged as optional, but C would not be.
+
+Given:
+```
+A (Opt) → B → C
+D → A
+```
+
+Then none would be flagged as optional.
+
+## Installing
+
+If the `packageIntegrity` in the `package-lock.json` differs from the one
+computed from the `package.json` then places where the `package.json` is
+incompatible with the `package-lock.json` a new module should be installed.
+That is, while the `package-lock.json` ordinarily defines the state of your
+project, if your `package.json` is edited independently it will take
+precedence.
+
+The `package-lock.json` describes the exact tree that `npm` should create.
+Any deviation between the `package.json` and the shrinkwrap/lock should
+result in a warning be issued.  This includes:
+
+* Modules in `package.json` but missing from the `package-lock.json`
+* Modules in the `package-lock.json` but missing from the `package.json`.
+* Modules in `package.json` whose specifiers don't match the version in `package-lock.json`.
+
+Warn if the `lockfileVersion` in the `package-lock.json` is for a different
+major version than we implement.
+
+Module resolution from package-lock data works as such:
+
+* If install was run with `--resolve-links` and a dependency has a `link`
+  property then a symlink is made using that.  If the version of the
+  destination can not be matched to the package-lock and/or the package.json
+  then a warning will be issued.
+
+* Otherwise, if a `integrity` is available then we try to install it from the cache using it.
+
+If `integrity` is unavailable or we are unable to locate a module from the `integrity` then:
+
+* If `lockfileVersion` is set:
+  * Install using the value of `version` and validate the result against the
+    `integrity`.
+* Otherwise, try these in turn and validate the result against the `integrity`:
+  * `resolved`, then `from`, then `version.
+  * `from` can be either `package@specifier` or just `specifier`.
+
+Regardless of how the module is installed the metadata in the installed
+module should be identical to what it would have been if the module were
+installed w/o a package-lock.
+
+## Implied Changes To Other Commands
+
+### `npm rm --save`
+
+Currently if you ask to remove a package that's both a direct and a
+transitive dependency, we'll remove the package from `node_modules` even if
+this results in a broken tree.  This was chosen at the time because we felt
+that users would expect `npm rm pkgname` to be equivalent of
+`rm -rf node_modules/pkgname`.
+
+As you are no longer going to be allowed to put your `node_modules` in a
+state that's not a valid package-lock, this means this behavior is no longer
+valid.  Instead we should follow normal rules, removing it from the
+dependencies for the top level but only removing the module on disk if
+nothing requires it any more.
+
+## Additional fields / Adding new fields
+
+Installers should ignore any field they aren't aware of.  It's not an error
+to have additional properities in the package-lock or lock file.
+
+Installers that want to add new fields should either have one added via RFC
+in the npm issue tracker and an accompanying documentation PR, or should prefix
+it with the name of their project.
--- a/doc/spec/package-lock.md
+++ b/doc/spec/package-lock.md
@@ -23,33 +23,6 @@ The name of the package this is a package-lock for. This must match what's in `p
 
 The version of the package this is a package-lock for. This must match what's in `package.json`.
 
-### createdWith *(new)*
-
-An object with the the following properties:
-
-* `installer` - The package manager used to create this `package-lock`.
-* `runtime` - The runtime used to run the installer.
-* `platform` - The platform used to run the installer.
-* `arch` - The architecture used to run the installer.
-
-The values of `installer` and `runtime` must be `name@version`.  The name part must be
-a valid npm package name.  The version part must be a valid semver version
-string. The values of `platform` and `arch` are strings. For example:
-
-```
-{
-  "installer": "npm@5.0.0",
-  "runtime": "nodejs@8.0.0",
-  "platform": "darwin",
-  "arch": "x64"
-}
-```
-
-Installers should not make decisions based on this field.  And if they do,
-they should be limited to working around known bugs in very specific
-versions.  The inteded purpose of this field is in helping with isolating
-the source of bugs.
-
 ### lockfileVersion *(new)*
 
 An integer version, starting at `1` with the version number of this document
@@ -183,8 +156,6 @@ Refresh the data from the installer's ideal tree.
 
 The top level `name` and `version` come from the `package.json`.  It is an
 error if either are missing or invalid.
-
-`createdWith` should be filled in from `npm` itself.
 
 #### dependencies.dev
 


spec/shrinkwrap-and-lock.md
```diff
@@ -0,0 +1,273 @@
...
+* `npm update`
+* `npm rm`
+* `npm shrinkwrap`
+
+## File Format
+
+`package-lock.json` and `npm-shrinkwrap.json` are JSON documents described
+below.  The difference between a `package-lock.json` and an
+`npm-shrinkwrap.json` is that the former is never published. The `npm shrinkwrap` command
+renames the `package-lock.json` (or if missing, generates one.)
```
jesstelford: Renames to `npm-shrinkwrap.json`?

zkat: @jesstelford correct.

spec/shrinkwrap-and-lock.md
```diff
@@ -0,0 +1,273 @@
...
+This is a [Standard Subresource
+Integrity](https://w3c.github.io/webappsec/specs/subresourceintegrity/) for
+this resource.
+
+* For bundled dependencies this is not included, regardless of source.
+* For registry sources, this is the `integrity` that the registry provided, or if one wasn't provided the SHA1 in `shasum`.
+* For git sources this is the specific commit hash we cloned from.
+* For remote tarball sources this is an integrity based on a SHA512 of
+  the file.
+* For local tarball sources: This is an integrity field based on the SHA512 of the file.
```
jesstelford: Interesting that there will be potentially 3 different formats for integrity (SHA1, SHA512, the subresource Integrity format) - would this potentially increase complexity in consuming applications (such as npm, and potentially yarn)?

felixrieseberg:
Hey Rebecca, fantastic effort 👏 I'm so happy you're putting some focus on shrinkwraps, they're core to releases we make. Thank you!

One thing I'd like to call out: *Cross-platform shrinkwraps*. You cover that with "All optional dependencies should be included even if they're uninstallable on the current platform.", but I'd like to see this be a reliable feature of shrinkwraps.

npm is possibly the best cross-platform installation tool available right now. At Slack and in other projects, I use shrinkwraps to describe the exact dependency tree of a cross-platform project at compilation time. We currently merge shrinkwraps generated on each platform by hand, but we'd love to be able to generate the "one true master shrinkwrap" automatically.

spec/shrinkwrap-and-lock.md
zkat: @jessmartin there is exactly one format: subresource integrity. That standard supports any algorithm attached to tarballs. Integrity fields can also support multiple different hashes for the same data (with different algorithms). So, if you have SRI support, your tool will automatically pick up `sha512` from registries that serve those.

zkat:
@felixrieseberg This spec also covers the shrinkwrap behavior. The difference between `package-lock.json` and `npm-shrinkwrap.json` is that the latter is meant for publication, and the former is meant for devs to commit. The rest is the same (modulo the fact that this new version of the lockfile has different fields -- older npms will still be able to install with shrinkwraps generated by this though)

vinkla:
Great idea :tada: I think this will make it easier for newcomers to understand what the lock file is.

> We will also be introducing a new file, `package-lock.json` that will fill the same role as `npm-shrinkwrap.json` in projects that don't otherwise have a shrinkwrap.

What about naming the lock file to `package.lock` similar to how [Composer](https://getcomposer.org) handles lock files? IMO it would make it even easier to understand.

iamstarkov:
i know @zkochan worked hard to get reliable lock file for `pnpm`. he might has some input

iamstarkov:
@vinkla `package.lock` seems like a nice filename

zkat:
@vinkla The current name is clearer about the file format. `require('./package-lock.json')` will give you a parsed JSON object with this. Other tools that look at the extension will work unaltered (for example, webpack, etc).

vinkla:
@zkat yeah, I understand the file format issue and I agree it makes it clearer what kind of content the file contains. Though, in my opinion it is clearer that the file ending with `lock` locks the dependencies in the `package.json` file. IMO it also makes the file tree look cleaner.

```
package.json
package.lock
```

I think, for newcomers `package.json` and `package-lock.json` will take longer to connect than what `package.json` and `package.lock` would. 

I might be the wrong person to discuss this though. I've gotten used to the naming convention from Composer. This is just an idea.

spec/shrinkwrap-and-lock.md
```diff
@@ -0,0 +1,273 @@
...
+   }
+}
+```
+
+If the user wants to get a default package name/version added they can run `npm init`.
+
+### `npm shrinkwrap`
+
+If a `package-lock.json` exists, rename it to `npm-shrinkwrap.json`.
+Refresh the data from the installer's ideal tree.
```
jesstelford: Could the "ideal tree" and the "currently installed tree" differ? Could this lead to a different dependency tree for a co-worker / on deployment / etc?

zkat: @jesstelford the two main cases where these trees might deviate are:

* `npm i --only=prod` - this removes `dev: true` deps
* optional dependencies (when they fail to build on certain platforms)

spec/shrinkwrap-and-lock.md
iarna: Yes, I'll clarify that!

spec/shrinkwrap-and-lock.md
```diff
@@ -0,0 +1,273 @@
...
+* Otherwise, if a `integrity` is available then we try to install it from the cache using it.
+
+If `integrity` is unavailable or we are unable to locate a module from the `integrity` then:
+
+* If `shrinkwrap-version` is set:
+  * Install using the value of `version` and validate the result against the
+    `integrity`.
+* Otherwise, try these in turn and validate the result against the `integrity`:
+  * `resolved`, then `from`, then `version.
+  * `from` can be either `package@specifier` or just `specifier`.
```
jesstelford: Should this downgrade path produce a warning?

zkat: It could, but also this would mean you get warnings for published npm packages that publish an `npm-shrinkwrap.json`. I don't feel like it has that much value if it will be adapted next time someone runs a new install.

iarna:
@felixrieseberg Regarding cross platformness, I'll make sure it's clearer, but yes, ALL optional dependencies, ALL dev dependencies should be in lock/shrinkwrap files even if you installed with `--no-optional` and `--no-dev`.

jesstelford (approved):
Excellent stuff - looking forward to it!

spec/shrinkwrap-and-lock.md
jesstelford: Interesting, I hadn't thought through those extra use-cases. Thanks for the explanation 👍

spec/shrinkwrap-and-lock.md
iarna: Sure, if you have optional dependencies that couldn't be installed, or you're running with `--production` then your currently installed tree wouldn't have those, but the ideal tree would

spec/shrinkwrap-and-lock.md
jesstelford: Interesting - a nested shrinkwrap is honoured? I thought the top level shrinkwrap would include all the nested dependencies already, and as such would be generated with the tooling being run by the developer (not the 3rd party module author)?

The specific case I was thinking about is when deploying to production involves an `npm i`: The downgrade could potentially mean MITM, brining in a semver minor when you wanted to lock it to semver patch ranges only, etc. All of which could introduce either breaking changes, or worse; hidden security holes.

zkat: cc @iarna but I'm pretty sure the toplevel lockfile overrides literally everything else. Nested shrinkwraps are only used during installation of that child package.

zkochan:
What about some of configurations found in npmrc?

Shouldn't they be included in the lock file? I mean things like: `no-bin-links`, `ignore-scripts`. Probably there are more. Configs that can change the way dependencies are installed

spec/shrinkwrap-and-lock.md
iarna: That's right. The deeper shrinkwrap would be consulted when first installing the module (when adding it to your project) but it won't be consulted ever again as all of its information will end up in your top level shrinkwrap.

iarna:
@zkochan No, I believe those belong in a project-local `.npmrc`. There is one that that I think does belong in a lock/shrinkwrap and that's node's `preserve-symlinks` (env: `NODE_PRESERVE_SYMLINKS`).

novemberborn:
How would one update a `package-lock.json` if new dependencies were added by manually editing `package.json`? Presumably running `npm install` would install those dependencies, but it won't update `package-lock.json`.

spec/shrinkwrap-and-lock.md
```diff
@@ -0,0 +1,273 @@
...
+
+All optional dependencies should be included even if they're uninstallable
+on the current platform.
+
+#### from *(deprecated)*
+
+This is a record of what specifier was used to originally install this
+package.  This should not be included in new `npm-shrinkwrap.json` files.
+
+#### dependencies
```
zkochan: Is there any reason to have this nesting?

For pnpm we have a flat mapping, `<pkgId>: <resolution>`. Like this:

```yaml
  /ansi-regex/2.1.1: c3b33ab5ee360d86e0e628f0468ae7ef27d654df
  /ansi-styles/2.2.1: b432dd3358b634cf75e1e4664368240533c1ddbe
  /chalk/1.1.3:
    dependencies:
      ansi-styles: 2.2.1
      escape-string-regexp: 1.0.5
      has-ansi: 2.0.0
      strip-ansi: 3.0.1
      supports-color: 2.0.0
    resolution: a8115c55e4a702fe4d150abd3872822a7e09fc98
```

We create one copy of the lockfile inside node_modules and one outside. Comparing the two lockfiles can very easily give us valuable info about the state of the dependency tree.

We can just compare the two lockfiles and say if we need to run `npm install`. We can look what packages are missing from the inner lockfile and tell what can be pruned.

Maybe with this structure it is possible to do these manipulations as well. But IMHO, with a flat key/value structure algorithms become simpler.

spec/shrinkwrap-and-lock.md
```diff
@@ -0,0 +1,273 @@
...
+* For remote tarball sources this is an integrity based on a SHA512 of
+  the file.
+* For local tarball sources: This is an integrity field based on the SHA512 of the file.
+
+#### resolved
+
+* For bundled dependencies this is not included, regardless of source.
+* For registry sources this is path of the tarball relative to the registry
+  URL.  If the tarball URL isn't on the same server as the registry URL then
+  this is a complete URL.
```
zkat: @zkochan the intention is to be able to switch registries and have it continue to work. So, `npm config get registry` :)

This is something the CLI's already been doing since forever and a day by munging tarball URLs internally, but we thought it might be nicer to make the "relative-ness" of these tarball URLs more explicit.

spec/shrinkwrap-and-lock.md
iarna: The layout describes the structure of the `node_modules` direcotry to be created and insures that you'll get the same result every time. Without this you're at the mercy of the algorithm of the installer. If that changes, you suddenly have a different layout. While modules _shouldn't_ depend on particular layouts, we know from hard experience that they _do_.

iarna:
@novemberborn You'll note that the `package-lock` has a integrity hash generated from the `package.json`. Ah, I didn't mention that when describing install semantics, I'll correct that.

spec/shrinkwrap-and-lock.md
```diff
@@ -0,0 +1,273 @@
+# The Problem We're Solving
+
+In the near future we would like to make the experience of using
+shrinkwrap as pleasant for our users as possible.  However, behavior of
+shrinkwrap has never been specified.
+
+## Managing the two files
+
+If at any time both `package-lock.json` and an `npm-shrinkwrap.json` exist
+then the former should be unlinked with a warning the latter used. This should happen on any command
```
legodude17: I would suggest putting an `and` there:
> ...with a warning _and_ the latter used...

spec/shrinkwrap-and-lock.md
legodude17: Does it generate a `package-lock.json` or an `npm-shrinkwrap.json`?

spec/shrinkwrap-and-lock.md
```diff
@@ -0,0 +1,273 @@
...
+This is non-normative and intended to help in isolating the cause of errors.
+
+### shrinkwrap-version *(new)*
+
+An integer version, starting at `1` with the version number of this document
+whose semantics were used when generating this `npm-shrinkwrap.json`.
+
+### package-integrity *(new)*
+
+A hash of the `package.json` that was in use when this shrinkwrap was
```
legodude17: What algorithm?

spec/shrinkwrap-and-lock.md
legodude17: @zkat By switch registries, what do you mean?

spec/shrinkwrap-and-lock.md
```diff
@@ -0,0 +1,273 @@
...
+If true, this is the bundled dependency and will be installed by the parent
+module.  When installing, this module will be extracted from the parent
+module during the extract phase, not installed as a separate dependency.
+
+#### dev
+
+If true then this dependency is either a development dependency ONLY of the
+top level module or a transitive dependency of one.  This is false for
+dependencies that are both a development dependency of the top level and a
+transitive dependency of a non-development dependency of the top level.
```
legodude17: What about if it is a dev dep and a dep of the top project?

spec/shrinkwrap-and-lock.md
```diff
@@ -0,0 +1,273 @@
...
+
+#### optional
+
+If true then this dependency is either an optional dependency ONLY of the
+top level module or a transitive dependency of one.  This is false for
+dependencies that are both an optional dependency of the top level and a
+transitive dependency of a non-optional dependency of the top level.
```
legodude17: What about if it is in optional deps and deps of the parent?

spec/shrinkwrap-and-lock.md
```diff
@@ -0,0 +1,273 @@
...
+
+All optional dependencies should be included even if they're uninstallable
+on the current platform.
+
+#### from *(deprecated)*
```
legodude17: So this is for backwards compat?

spec/shrinkwrap-and-lock.md
```diff
@@ -0,0 +1,273 @@
...
+
+This is a record of what specifier was used to originally install this
+package.  This should not be included in new `npm-shrinkwrap.json` files.
+
+#### dependencies
+
+The dependencies of this dependency, exactly as at the top level.
```
legodude17: Couldn't this lead to an infinite loop? Like, top is `A`. `A` depends on `B` which depends on `C` which depends on `B`. The installer could handle this by flattening the tree, but what about the shrinkwrap? Does it map out the flattened tree?

spec/shrinkwrap-and-lock.md
```diff
@@ -0,0 +1,273 @@
...
+
+If neither a `package-lock.json` nor an `npm-shrinkwrap.json` exist then
+`npm init` will create a `package-lock.json`.  This is functionally
+equivalent to running `npm shrinkwrap` after the current init completes and
+renaming the result to `package-lock.json`.
+
+### `npm install --save`
+
+If either an `npm-shrinkwrap.json` or a `package-lock.json` exists then it
+will be updated.
```
legodude17: What about if both exist?

spec/shrinkwrap-and-lock.md
```diff
@@ -0,0 +1,273 @@
...
+
+* Otherwise, if a `integrity` is available then we try to install it from the cache using it.
+
+If `integrity` is unavailable or we are unable to locate a module from the `integrity` then:
+
+* If `shrinkwrap-version` is set:
+  * Install using the value of `version` and validate the result against the
+    `integrity`.
+* Otherwise, try these in turn and validate the result against the `integrity`:
+  * `resolved`, then `from`, then `version.
```
legodude17: I think you missed the closing backtick on `version.`.

spec/shrinkwrap-and-lock.md
```diff
@@ -0,0 +1,273 @@
...
+
+### `npm rm --save`
+
+Currently if you ask to remove a package that's both a direct and a
+transitive dependency, we'll remove the package from `node_modules` even if
+this results in a broken tree.  This was chosen at the time because we felt
+that users would expect `npm rm pkgname` to be equivalent of
+`rm -rf node_modules/pkgname`.
+
+As you are no longer going to be allowed to put your `node_modules` in a
```
legodude17: What would `npm install` or `npm shrinkwrap` do if you manually edit your `node_modules` to put it in an invalid state?

legodude17 (commented):
Fast installs for everyone!

spec/shrinkwrap-and-lock.md
zkat: This one is up to the individual installers, I think. npm reads and verifies and fixes your trees. Others rely exclusively on a hash in a dotfile file in `node_modules` and trust the user. I think it's ok to assume this is out of scope for this RFC

spec/shrinkwrap-and-lock.md
zkat: https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity a string that fits this spec. So. Whichever.

spec/shrinkwrap-and-lock.md
zkat: It just means that you're likely to run into it because of older shrinkwraps (which can be processed by this spec!)

So. Don't add a field called `from`.

novemberborn:
@iarna 

> You'll note that the package-lock has a integrity hash generated from the package.json. Ah, I didn't mention that when describing install semantics, I'll correct that.

Sure. Once it's out of sync though, how would I update the `package-lock.json` file?

indexzero:
:1st_place_medal: well articulated & sensible. Really happy to see `npm` moving towards supporting lockfiles, but also introducing the semantics that `yarn` lacks. That is differentiating between publishable lockfiles (in this case `npm-shrinkwrap.json`) and non-publishable lockfiles for performance optimization of the average case (`package-lock.json`).

spec/shrinkwrap-and-lock.md
```diff
@@ -0,0 +1,273 @@
...
+
+#### integrity *(new)*
+
+This is a [Standard Subresource
+Integrity](https://w3c.github.io/webappsec/specs/subresourceintegrity/) for
+this resource.
+
+* For bundled dependencies this is not included, regardless of source.
+* For registry sources, this is the `integrity` that the registry provided, or if one wasn't provided the SHA1 in `shasum`.
+* For git sources this is the specific commit hash we cloned from.
```
zkat: I think we should put the sha512 of the generated tarball here anyway.

It might not be of much use, and the installer can understand enough about this to skip it for git deps if they don't match, *but* -- this is still useful if folks have already built a tarball locally, cause they'd still be able to do content-addressed lookups this way.

Once we have a tarball packer that can do sequential packs on the same data that have the same checksum, we'll be able to use this more often. We can just avoid *verifying* the generated tarballs with this, due to those differences.

Furthermore, the hash we cloned from is already in `resolved`. There's no need for this duplication.

spec/shrinkwrap-and-lock.md
```diff
@@ -0,0 +1,273 @@
...
+
+Discussion of the semantics of this will go in the symlinks RFC.
+
+Implementation note: To be implemented post npm@5.
+
+#### bundled *(new)*
+
+If true, this is the bundled dependency and will be installed by the parent
+module.  When installing, this module will be extracted from the parent
+module during the extract phase, not installed as a separate dependency.
```
zkat: I'd be a little more specific here and say it will be extracted from the root bundler (the first package in the parent chain that _doesn't_ have `bundled` on it). Bundles can be nested.

zkat (commented):
Added some comments around git and bundle deps stuff

spec/shrinkwrap-and-lock.md
Daniel15: @legodude17 - I think "switch registries" means switching from the main upstream registry (registry.npmjs.org) to a mirror - For example, CloudFlare's mirror at https://www.npmjs.cf/, Taobao's mirror for Chinese users at https://npm.taobao.org/, or an internal mirror for corporate environments.

spec/shrinkwrap-and-lock.md
```diff
@@ -0,0 +1,273 @@
...
+As you are no longer going to be allowed to put your `node_modules` in a
+state that's not a valid shrinkwrap, this means this behavior is no longer
+valid.  Instead we should follow normal rules, removing it from the
+dependencies for the top level but only removing the module on disk if
+nothing requires it any more.
+
+## Additional fields / Adding new fields
+
+Installers should ignore any field they aren't aware of.  It's not an error
+to have additional properities in the shrinkwrap or lock file.
```
Daniel15: Small nit: Typo "properities" → "properties"

spec/shrinkwrap-and-lock.md
```diff
@@ -0,0 +1,273 @@
...
+
+Installers that want to add new fields should either have one added via RFC
+in the npm issue tracker and an accompanying documentation PR, or should prefix
+it with the name of their project.
```
Daniel15: This should specify the format of the prefix - For example, it could be underscore separated (`{project}_{fieldName}`) or colon separated (`{project}:{fieldName}`) or something else altogether.

zkat: What would you suggest?

CSS folks have a pretty "standard" familiar syntax: `-brwsr-my-property:`

I do like `:` or `::` as separators though. :)

Daniel15: Honestly, I don't think it matters exactly what the format is, as long as it's documented. Just go with whatever you think looks best 😄  I do think `:` looks like a nice separator though.

zkat: awesome! 🎉

alloy:
Has there been any thought given into making the lockfile shareable with yarn?

zkat:
@alloy Yes. We honestly believe that this `package-lock.json` syntax can be used (and written) by yarn. The goal here is that different people on the same team can use different package managers with their own advantages, but all would be able to generate *identical* trees, regardless of which tool you use, or even which version of the tool you use.

This format defines a specific _tree_ -- something that all npm-compatible package managers would need to have as a data structure internally somewhere. This format also includes the tarball checksum, the resolved field to get the exact version remotely, the version #. The tree structure defines which node depends on what. I believe all current package managers with lockfiles would be able to do their required work with this information. There's also an allowance for extending stuff with package manager-specific metadata that they could use for this or that.

alloy:
Awesome, thanks for the info and great stuff 👏

zkat:
(if it's not clear yet, it's _really important_ to us that we don't end up with a lockdown situation in any direction: much like npm goes out of its way to support alternative registries, and the registry allows non-npm-cli clients to connect to it, and we document all our APIs. We'll keep doing our best to encourage the entire ecosystem's growth. In the end, npm is here for the sake of the community)

cpojer:
One issue we ran into at Facebook before building Yarn was that npm-shrinkwrap wasn't stable and small changes often produced 10,000+ line changes in the shrinkwrap file. I would suggest sorting the `package-lock.json` file using `json-stable-stringify` and adding that to the spec.

Further, we noticed people weren't ever looking at shrinkwrap changes, which is what lead us to create a separate, human-readable format for yarn's lockfiles. Can you elaborate on why you are suggesting to use JSON as the preferred format? Thanks!

zkat:
@cpojer Hi!

1. `package-lock.json` is already sorted. Specifying that in the spec for clarity is a *great* idea.

2. A huge reason for npm-shrinkwrap giving you conflicts and nonsense diffs was the `from` field, as well as inconsistencies in how `version` was written (because we'd end up writing `resolved` into `version`. You can see the sorts of diffs that you get now here:
screen shot 2017-05-02 at 11 14 05

3. JSON is a format that's natively readable by node.js, and shares a format with package.json. With the simplified, standardized fields, this would bring the JSON format to the same information level as something like `yarn.lock`. For something that is intended to be used across any package manager, I think it's the right thing to not require a particular non-built-in format. Furthermore, `package-lock.json`, through its nesting, defines a *specific* tree to build -- my understanding is that formats like `yarn.lock` require a specific treebuilding algorithm (which can change between versions!) to build identical trees. In contrast, someone could symlink `package-lock.json` to `npm-shrinkwrap.json` for a CI build of their library using npm@2, and they'll get the exact same tree that yarn may have generated. It means that you can use yarn's trees in npm! :)

iarna:
> One issue we ran into at Facebook before building Yarn was that npm-shrinkwrap wasn't stable and small changes often produced 10,000+ line changes in the shrinkwrap file. I would suggest sorting the package-lock.json file using json-stable-stringify and adding that to the spec.

npm-shrinkwrap has been sorted and its structure stable since npm@3. The primary source of the churn was the `from` field. The secondary one is folks running `git pull` and then `npm shrinkwrap` without an intervening `npm install` (which was was possible when the changes had been semver compatible). (The latter will be addressed by changes to the implementation within npm.)

Basically, none of the complaints I've seen about npm-shrinkwrap were related to the file format itself.

The main reason we didn't consider yarn.lock is that it does not actually describe a physical tree and the shape of the installed tree can, will and has changed between implementations and versions of implementations. That you will always get exactly the same tree is a guarantee we wanted to be able to make.

pmuellr:
> This format defines a specific tree ...  The tree structure defines which node depends on what.

Looking only at [GH/zkat/cacache/package-lock.json](https://github.com/zkat/cacache/blob/a962e98dd3178eec31a2ff589128ac3b45c4f5db/package-lock.json), and not the code, have some questions.

It looks like the lock file contents mirror the directories npm creates during an `npm install` - eg, flattened, with same package/different versions handled via deeper directories in the packages that depend on them.  Eg, the different versions of `wordwrap` used in the referenced lock file.

This seems like an implementation detail.  Has any thought been given to showing the actual dep structure instead?  More like the output of `npm ls --json`.  An "ideal" directory tree could be calculated from that, and having the structure would be nice for dependency analysis tooling.  As an example, if you'd like to do some kind of interesting dependency analysis of a package, and had the lock file with that structure, you could just read it.  Without the structure, you'd have to start at the package file (or lock file), then get deps for dependent packages from npm.

As an example of dep analysis, see https://github.com/pmuellr/npmls2dg - you can drag'n'drop the output of `npm ls --json` on the page and get deps graphed.  If the lock file had the dep structure, you could drop the lock file, instead of having to run `npm ls --json` at all.  Having the lock files checked in means all this dep info would be available directly in GH without having to touch an npm registry at all.

The resulting file would be larger than the current lock file, as there would be duplication information in the tree.  I'm not sure in the end the size matters that much, but might be interesting to see if a hybrid structure would allow for some cost savings - eg, store a separate map of package/version => info, and ref the dependencies as just the package/version key of that map, sort of thing.

spec/shrinkwrap-and-lock.md
destroyerofbuilds: A similar issue is documented here - https://github.com/yarnpkg/yarn/issues/2566

Being able to switch registries for packages specified in the lockfile will be really helpful.

zkat:
@pmuellr it's not an implementation detail in practice: the actual physical arrangement of the tree has a huge effect on actual behavior and is one of the biggest reasons for having a lockfile (aside from just being more specific about versions). That's why it's important to have the final structure meant for the disk. Different algorithms will flatten the tree using different heuristics (if any!) so this matters.

pmuellr:
> the actual physical arrangement of the tree has a huge effect on actual behavior and is one of the biggest reasons for having a lockfile

I'm very likely missing an important piece here, perhaps what "actual behavior" means. Seems like given the deps structure from `npm ls --json`, you can calculate a disk structure (or structures, you can of course imagine multiple strategies for this - I want npm2 style! :-) ).  Maybe that's not enough tho.

I guess I basically just feel like this may be a missed opportunity at having some **really interesting** dep information persisted with projects as a file - info that you can only get today via server interaction.  The info is calculated, then tossed in the bit bucket.  Sad! (heh)

Could there be some way of doing both?  Maybe serialize the dep structure, but add a property for the resultant disk path it resides at?

zkat:
@pmuellr 

> you can of course imagine multiple strategies for this - I want npm2 style! :-)

I see one of the points of a lockfile to be precisely to ensure that different tools using different algorithms arrive at the exact same tree structure, not just the same general versions for things.

> Could there be some way of doing both?

I think this is the sort of thing that would be a good use-case for extracting the tree generator bits of npm into a standalone library, and/or having `npm ls` be able to work off the logical tree (not just the physical tree). The latter has other use-cases: for example, finding dependencies when an install failed, which is basically impossible right now. This would allow your analysis tool to run `npm ls --logical --json` on a project with a `package-lock.json`. This is a separate thing from the lockfile format, though, so if we're gonna talk about that, best to make a separate feature request.

legodude17:
From #16580:
The package-lock.json might want to include nodejs version.

spec/shrinkwrap-and-lock.md
```diff
@@ -0,0 +1,273 @@
...
+
+This is a specifier that uniquely identifies this package and should be
+usable in fetching a new copy of it.
+
+* bundled dependencies: Regardless of source, this is a version number that is purely for informational purposes.
+* registry sources: This is a version number. (eg, `1.2.3`)
+* git sources: This is a git specifier with resolved committish. (eg, `git+https://example.com/foo/bar#115311855adb0789a0466714ed48a1499ffea97e`)
+* http tarball sources: This is the URL of the tarball. (eg, `https://example.com/example-1.3.0.tgz`)
+* local tarball sources: This is the file URL of the tarball. (eg `file:///opt/storage/example-1.3.0.tgz`)
+* local link sources: This is the file URL of the link. (eg `file:libs/our-module`)
```
zkat: What's the logic for making these relative/absolute?

spec/shrinkwrap-and-lock.md
```diff
@@ -0,0 +1,273 @@
...
+string. For example: `npm@5.0.0`
+
+This is non-normative and intended to help in isolating the cause of errors.
+
+### shrinkwrap-version *(new)*
+
+An integer version, starting at `1` with the version number of this document
+whose semantics were used when generating this `npm-shrinkwrap.json`.
+
+### package-integrity *(new)*
```
zkochan: what will this be used for? to check if dependency specs have changed and whether they should be re-resolved?

iarna:
So, updates:

* Normalizing names toward `package-lock`.
* camelCase
* A description of how `packageIntegrity` impacts install behavior and how it is computed.

spec/shrinkwrap-and-lock.md
iarna: I think that's an error, in that in the file-specifier RFC we assert they'll all be relative.

spec/shrinkwrap-and-lock.md
iarna: I actually just added docs for how this impacts installation, but yes, that's essentially correct. It saves people who manually edit their `package.json` files (which is a thing).

doc/spec/package-lock.md
```diff
@@ -0,0 +1,290 @@
...
+string. For example: `npm@5.0.0`
+
+This is non-normative and intended to help in isolating the cause of errors.
+
+### lockfileVersion *(new)*
+
+An integer version, starting at `1` with the version number of this document
+whose semantics were used when generating this `package-lock.json`.
+
+### packageIntegrity *(new)*
```
zkochan: why not just duplicating the specifiers in `package-lock`? Something like

```
specs: {
  foo: '^1.0.0',
  bar: '2.0.0'
}
```

dependencies, devDependencies, optionalDependencies can be merged into one property. In that case, package managers will know exactly which dependencies should be updated and skip those which did not change. It can also print some info message that "this or that does not match the spec from `package.json`"

iarna:
More changes:

* Expand `createdWith` into an object with installer version, Node.js version, platform and architecture.
* Revert changes to `packageIntegrity`, it's back to being a hash of the `package.json`.
* NEW PROPERTY: `preserveSymlinks` records if the user ran with `NODE_PRESERVE_SYMLINKS` enabled. If they did then all future work with the tree (in Node.js or npm) MUST have it enabled too, or the resulting tree won't be usable. `NODE_PRESERVE_SYMLINKS` changes how the Node.js module loader works in regards to symlinks and results in our needing to construct trees differently.

doc/spec/package-lock.md
iarna: Ok, so @zkat and I talked a bit more about this in person:

The conclusion we came to is that this should just hash the `package.json` directly. The problem here is that potentially deps aren't the only thing that should trigger a change in lockfile. Keeping the `package.json` opaque keeps us nicely future proof.

iarna: I hasten to add, I'm not saying the behavior should be "act like there was no lockfile present" but more along the lines of taking the package.json into account. Ordinarily that's not done.

legodude17:
@iarna What is the reason for ditching `created-with`?

zkat:
@legodude17 https://github.com/npm/npm/issues/16464#issuecomment-301954806 has a detailed write-up about it.

legodude17:
Yes, I found that right after I posted the comment.

doc/spec/package-lock.md
```diff
@@ -0,0 +1,273 @@
+# package-lock and npm-shrinkwrap
+
+`npm` can have one of two different lock files:
+
+* `package-lock.json`, which is ordinarily always present and is never published.
+* `npm-shrinkwrap.json`, which is created with `npm shrinkwrap` and usually published.
+
+You can only have one of them and in the event that you have both,
```
jsdnxx: I think it read clearer with the same meaning as "You should only have one of them" rather than "You can only have one of them"

Andarist:
im probably not that familiar with the shrinkwrap, but why exactly is there a need for 2 different files (shrinkwrap and package-lock)? The TLDR is not quite sufficient for a layman

kourge:
As I read it, `package-lock.json` can be thought of as an "automatic" version of `npm-shrinkwrap.json`: if you don't shrinkwrap a package you're maintaining, you'll still benefit from a subset of what shrinkwrapping brings to the table.

The material difference comes from publishing: if you shrinkwrap package `foo`, publish it, and have someone install `foo`, they will be able to see your shrinkwrapped dependencies by cracking open `./node_modules/foo/npm-shrinkwrap.json`, whereas `package-lock.json` is meant to be committed to your repo and not published. This contrast is very similar to `dependencies` and `devDependencies`: if someone installs your published package, they'll only pull in `dependencies` of said package, but if someone is in the source code repo of your package and runs `npm install`, they'll pull in both sets of those dependencies.

zkat:
what @kourge said is all correct.

On a more human level, the experience the CLI team has had is that no one understands what a shrinkwrap is. People just don't get it. That's why people are all like "npm didn't have a lockfile before!"

So we decided to use a new package name, and that happened to resolve a lot of other semantic issues we were having. We really didn't want to have `npm-shrinkwrap.json` getting published by default because it'd ruin the ecosystem (by preventing flattening and propagation of patch updates).

jsdnxx:
@zkat this makes sense to me, but from conversations I've heard, I think there's some general confusion on 

>  it'd ruin the ecosystem (by preventing flattening and propagation of patch updates

especially from people who mostly work at the application level and aren't publishing library packages. This could be an opportunity for a lightning talk or something setting up the problem and how changing the semantics helps.

The biggest win, to me, in npm5 is having `--save` by default, in line with what package managers in other ecosystems do.

zkat:
@jsdnxx alas, app devs just don't need to worry about it. We just need something that works for both app devs and lib devs.

zkat:
This PR was merged as part of npm@5, so I'm gonna close this. A rewritten version landed in our `latest` branch yesterday.

Might be worth noting that we dropped `packageIntegrity` at the last minute, but otherwise this all applies. Thanks all for engaging, and we'll still be open to further discussion on the format going forward!

doc/spec/package-lock.md
Daniel15: @jsdnxx - They mean different things though. "You can" implies that it's a requirement, whereas "you should" implies that it's a good idea but not mandatory. I guess it depends on how strong the suggestion/requirement should be.

zkat:
wait no wrong branch lol

evolve2k:
A few people have been debating the naming convention of calling the lock file package-lock.json

## The case for .lock ##
This follows expectation of naming convention for a number of other package lock files for example: yarn.lock, Gemfile.lock, 

The extension .lock sends the clearest message that this is a system generated file to be committed and handled like they do other .lock files.

## The case for .json ##
The file is in the .json format and signifies that the contents of the file are in json format. This approach ensures pragmatic syntax management by text editors, and since it is indeed .json why would you give up .json as the file extension and loose that natural formatting assitance. 
.lock sends no message as to the format of the content of the file. 

## Summary ##
This issue has already been solved in other domains. 
Chaining both extensions would allow the best of both worlds.

## The case for .lock.json ##
This happily meets both objectives. It's possible to manage as a .lock file without having to do regex hunting for the more awkward -lock pattern and also with a trailing .json the text editors are happy. The file conforms to both extensions so use both. There is precedents for this.

Please can we rename this file as package.lock.json

Andarist:
@zkat @kourge 
It still really buffles me

> We really didn't want to have npm-shrinkwrap.json getting published by default because it'd ruin the ecosystem (by preventing flattening and propagation of patch updates).

isnt shrinkwrap now exactly for publishing purposes?

Also still dont get why the need for 2 files when they are exactly the same (release note mentioning shrinkwrap being alias for lock). Why the same couldnt be both:
- committed to the repo
- published
- interpretated differently when installing package dependencies and when installing package as a dependency

zkat:
@Andarist I'm afraid I don't know how to convey the reasoning behind this more clearly. I've stated the reasoning above.

@evolve2k I don't find the grep argument particularly convincing. You still need a trailing wildcard to match `.lock`. I am generally pretty opposed to spending much time on relatively minute details, and if a technical decision needs to be made on something, it really should be *very significant* to warrant additional attention. There doesn't seem to be particular reactji consensus that this is either a very important issue, nor particularly a clear-cut preference in your favor. I hope that's sufficient to address this. I probably won't discuss it further unless a really startling issue with it comes up. Consider that at this point, with npm@5 in prerelease, the file is already being used in the wild, and to counter your point, we've now added the cost of adding _another_ file for thousands of people to switch to (and try to figure out why they have both in their directory)

monolithed:
How to update a package-lock.json file if new dependencies were added by manually editing package.json???

Andarist:
@zkat 
i might not be familiar enough how things are working under the hood and how do u resolve dependencies, but this

> We really didn't want to have npm-shrinkwrap.json getting published by default because it'd ruin the ecosystem (by preventing flattening and propagation of patch updates).

seems a little bit odd. First of all I think u have made mistake here and ment `package-lock.json` (if not then im confused even more, but probably entirely my fault). How published `package-lock.json` would ruin things when published `npm-shrinkwrap.json` is not when they are from what I understood the same file (regarding contents) - only under other names.

AlecHFerguson:
I stumbled upon this PR because I was trying to understand `package-lock.json` since I hadn't seen it.  First off, thanks @iarna for writing documentation!  This will make the project way more usable for the community.

After reading this doc, there are a couple of things I'm not clear on.  I think they can all be categorized as, _What's new in NPM version 5?_  Consider adding a heading which details the differences with the new version?
1) What purpose specifically does `package-lock.json` serve?  It sounds like it is not shipped along with the packaged module; so then why bother creating this file?  Is it because certain bugs arose when using the non-locked `package.json`?  Consider clarifying for those of us just now upgrading.
2) _Should_ `package-lock.json` be checked into version control?  A comment above says, "Is suitable for checking in"... but what is the recommended practice?
3) Are there any backwards compatibility issues with creating this file?  For dinosaurs like me still on npm version 3.X?  I cloned a repo that had this file (thus must have been on version 5.X).  It seemed like installing packages and even changing package versions worked just fine.  But doing so did not change this file.  Do you foresee any issues if one developer is on npm version 5 but other developers are on earlier versions?

Hope that helps.  And thanks for documenting, you're the best!

monolithed:
I want to understand what the lock file is, but I can't while we cannot update the package-lock.json file when new dependencies were added by manually editing package.json. it's weird.

zkochan:
I think it would be nice to be able to detect which dependencies cannot be installed on the current system using just the lockfile. Currently, the `package.json` has to be downloaded in order to check the `engines` field. Having the engines field in the lockfiles would save us a few redundant HTTP requests.

zkat:
@zkochan part of the issue is that the _definition_ of "failed optional dep" is literally "tried to install it and it failed". The engines field has always been an FYI and we can't really treat it as normative, lest a lot of other stuff break, I think. /cc @iarna cause I'm pretty sure that's right.

zkochan:
@zkat unless [engine-strict](https://docs.npmjs.com/misc/config#engine-strict) is set to `true`

> If set to true, then npm will stubbornly refuse to install (or even consider installing) any package that claims to not be compatible with the current Node.js version.

isaacs:
It's well after the time for comments on this particular aspect of the bikeshed, and what I wanted already happened, but `.lock` as a file extension tells me it's a file system mutex lock. I know yarn and gem don't use it that way, but I've interacted with many programs over the years that do. (For example, this is what a `.lock` file in an npm <4 cache is.)

The `.json` extension says "this contains meaningful data".

gauravmuk:
Is there any way that I can **not** generate package-lock.json every time on npm install with npm@5?

EDIT: got the answer, --no-package-lock

doc/spec/package-lock.md
```diff
@@ -0,0 +1,273 @@
...
+### lockfileVersion *(new)*
+
+An integer version, starting at `1` with the version number of this document
+whose semantics were used when generating this `package-lock.json`.
+
+### packageIntegrity *(new)*
+
+This is a
+[subresource integrity](https://w3c.github.io/webappsec/specs/subresourceintegrity/)
+value created from the `pacakge.json`. No preprocessing of the
```
sea-witch, line 35: Typo: pacakge --> package  :)

zkat:
This has been merged 👍

PR #16452, opened 2017-04-27
feat: add "npm org" command

chrisdickinson:
This adds support for adding, removing, and listing org memberships from the command line!

```
$ npm org add npmcorp chrisdickinson admin
$ npm org set npmcorp chrisdickinson developer
$ npm org rm npmcorp chrisdickinson
$ npm org ls npmcorp
```

### DIFF
--- a/lib/config/cmd-list.js
+++ b/lib/config/cmd-list.js
@@ -32,6 +32,7 @@ var affordances = {
   'login': 'adduser',
   'add-user': 'adduser',
   'author': 'owner',
+  'ogr': 'org',
   'home': 'docs',
   'issues': 'bugs',
   'info': 'view',
@@ -71,6 +72,7 @@ var cmdList = [
   'owner',
   'access',
   'team',
+  'org',
   'deprecate',
   'shrinkwrap',
 
--- a/lib/org.js
+++ b/lib/org.js
@@ -1,0 +1,55 @@
+'use strict'
+
+var mapToRegistry = require('./utils/map-to-registry.js')
+var npm = require('./npm')
+var output = require('./utils/output.js')
+
+module.exports = org
+
+org.subcommands = ['set', 'rm', 'add']
+
+org.usage =
+  'npm org set orgname username [developer | admin | owner]\n' +
+  'npm org add orgname username [developer | admin | owner]\n' +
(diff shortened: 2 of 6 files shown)


zkat:
@chrisdickinson Could you add some tests? https://github.com/npm/npm/blob/latest/test/tap/team.js is the corresponding test for team.

zkat (approved):
LGTM 👍

iarna (changes requested):
I'm of the opinion that cli commands should never emit JSON unless explicitly asked to with `--json` and that we should come up with some human readable output format for these commands.

There also need to be docs for these new commands.

lib/org.js
```diff
@@ -0,0 +1,55 @@
...
+
+function org (args, cb) {
+  // Entities are in the format <scope>:<org>
+  var cmd = args.shift()
+  var orgname = args.shift()
+  var username = args.shift()
+  var role = args.shift()
+  return mapToRegistry('/', npm.config, function (err, uri, auth) {
+    if (err) { return cb(err) }
+    try {
```
watilde: Probably, we can get rid of this `try-catch` like the other implementations.

node_modules/npm-registry-client/index.js
```diff
@@ -60,6 +60,7 @@ function RegClient (config) {
   client.get = require('./lib/get')
   client.initialize = require('./lib/initialize')
   client.logout = require('./lib/logout')
+  client.org = require('./lib/org')
```
watilde: Somehow we're missing this module file. Let's run:
> $ scripts/dep-update npm-registry-client 8.2.0

lib/config/cmd-list.js
```diff
@@ -32,6 +32,7 @@ var affordances = {
   'login': 'adduser',
   'add-user': 'adduser',
   'author': 'owner',
+  'ogr': 'org',
```
zkat: 👹

watilde: might be fine as an affordance?

zkat:

zkat:
Additionally, chris, a new version of this might want to follow the implementation of `npm hook`, and use `npm-registry-fetch` plus a `libnpmorg` or something.

PR #16474, opened 2017-04-29
test: fix/skip test/tap/lifecycle-signal.js

addaleax:
The tests in `test/tap/lifecycle-signal.js`, as well as the features
they are testing, are partially broken. This moves them from
being skipped in CI to being disabled only for certain platforms.

In particular, because `npm` spawns its lifecycle scripts in a
shell, signals are not necessarily forwarded by the shell and
won’t cause scripts to exit; also, shells may report the signal
they receive using their exit status, rather than terminating
themselves with a signal.


iarna:
Thanks for your patience, this is merged into `release-next` now and will be going out later today.

Issue #16243, opened 2017-04-04
binding.gyp + ignore-scripts + publish = broken?


bnoordhuis:
#### I'm opening this issue because:

  - [ ] npm is crashing.
  - [x] npm is producing an incorrect install.
  - [x] npm is doing something I don't understand.
  - [ ] Other (_see below for feature requests_):

#### What's going wrong?

I published updates to three native modules earlier today (buffertools, heapdump, iconv) and all three exhibit the same issue: `npm install` doesn't invoke node-gyp to build them.  Previous releases work alright.

I discovered that the `.scripts` property in (e.g.) `.npm/heapdump/0.3.8/package/package.json` is an empty object, whereas in previous releases it has a 'test' and 'install' key (`"install": "node-gyp rebuild"`.)

Some sleuthing suggests the `npm.config.get('ignore-scripts')` check in lifecycle() in lib/lifecycle.js is responsible for that.  I had `ignore-scripts=true` set in my .npmrc; after removing that and publishing again, packages install okay again and the `.scripts` object is no longer empty.

Question 1: is this the intended behavior?  I sure hope not!

Question 2: is this documented anywhere?  I couldn't find anything in the documentation; the bug tracker didn't turn up anything either.

### supporting information:

 - `npm -v` prints: 4.2.0
 - `node -v` prints: v8.0.0-pre (today's master)
 - `npm config get registry` prints: https://registry.npmjs.org/
 - Windows, OS X/macOS, or Linux?: OS X, Linux
 - Network issues:
   - Geographic location where npm was run:
   - [ ] I use a proxy to connect to the npm registry.
   - [ ] I use a proxy to connect to the web.
   - [ ] I use a proxy when downloading Git repos.
   - [ ] I access the npm registry via a VPN
   - [ ] I don't use a proxy, but have limited or unreliable internet access.
 - Container:
   - [ ] I develop using Vagrant on Windows.
   - [ ] I develop using Vagrant on OS X or Linux.
   - [ ] I develop / deploy using Docker.
   - [ ] I deploy to a PaaS (Triton, Heroku).

bnoordhuis:
Ping.  It's been three weeks and it doesn't look like this issue has even been triaged yet.

legodude17:
Sorry, somehow this slipped through my notifications. I don't believe this is intended behavior, but I am not a dev.

silentroach:
Nice way

PR #16476, opened 2017-04-29
lifecycle: move ignore-scripts handling to actual place of use

addaleax:
Instead of overriding `pkg.scripts` when `ignore-scripts` is in
effect, test the option when actually determining whether there is a
script to invoke. Otherwise, published packages will not contain
the proper `scripts:` fields when `ignore-scripts` is `true`.

Fixes: https://github.com/npm/npm/issues/16243

If you have any pointers for writing a test for this, I’d love to look into that. :)

### DIFF
--- a/lib/utils/lifecycle.js
+++ b/lib/utils/lifecycle.js
@@ -51,10 +51,6 @@ function lifecycle (pkg, stage, wd, unsafe, failOk, cb) {
   log.info('lifecycle', logid(pkg, stage), pkg._id)
   if (!pkg.scripts) pkg.scripts = {}
 
-  if (npm.config.get('ignore-scripts')) {
-    log.info('lifecycle', logid(pkg, stage), 'ignored because ignore-scripts is set to true', pkg._id)
-    pkg.scripts = {}
-  }
   if (stage === 'prepublish' && npm.config.get('ignore-prepublish')) {
     log.info('lifecycle', logid(pkg, stage), 'ignored because ignore-prepublish is set to true', pkg._id)
     delete pkg.scripts.prepublish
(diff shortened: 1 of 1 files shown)


iarna:
It was doing what?! Oh geez! Good catch!

zkat (approved):
ahahaha I just ran into this code today. this is so much better.

addaleax:
Rebased :)

mikesherov:
This would need to be re-rebased since my change that completely skips lifecycle events when scripts are empty.

mikesherov:
Can you please reopen this PR against npm/lifecycle as it's now been extracted to a separate repo?

addaleax:
Maybe, if I find the time, but honestly, having this just lie around for a couple months in here makes me inclined to just say “let anybody who wants to do it”. :) It’s not like this patch is rocket science.

I’m closing this then, as it doesn’t seem to apply anymore.

mikesherov:
I think that's fair to close it. Having PRs lie around definitely have that effect on me too. Mind if I port it to npm/lifecycle then for you? You'll still retain credit, but I don't want to copy your patch without permission.

Lmk, thanks!

addaleax:
@mikesherov Sure – go ahead, I don’t even care about the credit :)

mikesherov:
Thanks!

iarna:
This is merged into the latest `npm-lifecycle` as [b136ac8](https://github.com/npm/lifecycle/commit/b136ac859628a129540250114f35498f809447a6) which will be included in the next npm release.

iarna:
(I magiced the credit in 😉)

Issue #13789, opened 2016-08-29
npm on windows can't run scripts with ".", relative paths "../"


whitecolor:
Npm 3.3.12

On windows npm can not run scripts like

```
"build": "../node_modules/.bin/tsc",
```

Error:

```
'..' is not recognized as an internal or external command,                                                                                                                                            
operable program or batch file.                                                                                                                                                                                                      
```

Should replace it with `..\\`

```
"build": "..\\node_modules\\.bin\\tsc",
```

But it is not crossplatfrom, I think can be fixed
#### How can the CLI team reproduce the problem?

place in script task something with "../" or "./" on WINDOWS.

This seem to be a like issue in 2014:
https://github.com/npm/npm/issues/4403

This is really very limiting.

whitecolor:
I believe it should be fixed something around here https://github.com/npm/npm/blob/19254de7b10c84a8a146ee044b95cf740f84ba50/lib/utils/lifecycle.js#L238

derekgreer:
Is there a cross-environment work around for this issue?

I can prefix commands with ${PWD} (unix) or ${CD} for windows, but neither is compatible in all environments.  If an npm environment variable such as npm_package_location existed then that would help tremendously, but no such variable seems to exist.

whitecolor:
the most cross-platfrom is to use WSL on windows 10, and BTW there was a PR in `yarn` that solved OP issue.

derekgreer:
@whitecolor Thanks for the recommendation, but I'm really more interested in simple, node-based solutions to the issue until the npm team can address this.

It does appears that the cross-run package can be used to execute commands with a relative path.

whitecolor:
https://github.com/yarnpkg/yarn is actually kind of node-base solution 😄 and not more complex then npm. I wanted make a PR here, but [I did make it in yarn](https://github.com/yarnpkg/yarn/pull/2082) eventually.

derekgreer:
When I stated that I was looking for a simple, node-based solution, that was only in reference to running node with WSL.   I guess I should have said "npm-based solution" to be specific.  Yes, yarn, gulp, and grunt, etc. are all node-based alternatives, but I was really just inquiring about work-arounds to addressing this issue within the context of using npm scripts.

whitecolor:
>, yarn, gulp, and grunt 

👍 

[...]

PR #16498, opened 2017-05-02
Fix issue on Windows with relative paths

ghost:
Hi,

We faced an issue this week with NPM (3.10.8) while executing the`command `npm run lint with the following task in my package.json:

```json
"lint": "../../node_modules/.bin/eslint.cmd --color '{src,test}/**/*.js'",
```

The error is the following:

```shell
C:\Users\jbaudin\workspace\frint\packages\frint-model>npm run lint

> frint-model@1.3.0 lint C:\Users\jbaudin\workspace\frint\packages\frint-model
> ../../node_modules/.bin/eslint --color '{src,test}/**/*.js'

'..' is not recognized as an internal or external command,
operable program or batch file.

npm ERR! Windows_NT 10.0.14393
npm ERR! argv "C:\\Program Files\\nodejs\\node.exe" "C:\\Program Files\\nodejs\\node_modules\\npm\\bin\\npm-cli.js" "run" "lint"
npm ERR! node v6.9.1
npm ERR! npm  v3.10.8
npm ERR! code ELIFECYCLE
npm ERR! frint-model@1.3.0 lint: `../../node_modules/.bin/eslint --color '{src,test}/**/*.js'`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the frint-model@1.3.0 lint script '../../node_modules/.bin/eslint --color '{src,test}/**/*.js''.
npm ERR! Make sure you have the latest version of node.js and npm installed.
npm ERR! If you do, this is most likely a problem with the frint-model package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     ../../node_modules/.bin/eslint --color '{src,test}/**/*.js'
npm ERR! You can get information on how to open an issue for this project with:
npm ERR!     npm bugs frint-model
npm ERR! Or if that isn't available, you can get their info via:
npm ERR!     npm owner ls frint-model
npm ERR! There is likely additional logging output above.
npm WARN Local package.json exists, but node_modules missing, did you mean to install?

npm ERR!     C:\Users\jbaudin\workspace\frint\packages\frint-model\npm-debug.log
```

My fix build an absolute path when a relative one is specified. It fix also the issue #13789 

I would like to add some unit-tests for Windows however I don't really know from where to start.

There are for now 32 tests failing on Windows 10 with Node v6.9.1. Is there already something going on for fixing the tests for this OS?

### DIFF
--- a/lib/utils/lifecycle.js
+++ b/lib/utils/lifecycle.js
@@ -263,11 +263,13 @@ function runCmd_ (cmd, pkg, env, wd, stage, unsafe, uid, gid, cb_) {
     shFlag = '/d /s /c'
     conf.windowsVerbatimArguments = true
 
-    // check if it's a path to a binary
-    var isRelativePath = cmd.indexOf('/') !== -1
+    // check if it's a relative path
+    var isRelativePath = cmd.indexOf('./') !== -1
 
     if (isRelativePath) {
-      cmd = path.resolve(path.join(wd, cmd))
+      // if yes then build and resolve command path
+      var winCmdPath = path.join(wd, cmd)
+      cmd = path.resolve(winCmdPath)
     }
   }
 
--- a/lib/utils/lifecycle.js
+++ b/lib/utils/lifecycle.js
@@ -258,22 +258,24 @@ function runCmd_ (cmd, pkg, env, wd, stage, unsafe, uid, gid, cb_) {
   var sh = 'sh'
   var shFlag = '-c'
 
-  var cmdAndFlags
-
   if (process.platform === 'win32') {
     sh = process.env.comspec || 'cmd'
     shFlag = '/d /s /c'
     conf.windowsVerbatimArguments = true
-    cmdAndFlags = [shFlag, 'node', cmd]
-  } else {
-    cmdAndFlags = [shFlag, cmd]
+
+    // check if it's a path to a binary
+    var isRelativePath = cmd.indexOf('/') !== -1
+
+    if (isRelativePath) {
+      cmd = path.resolve(path.join(wd, cmd))
+    }
   }
 
   log.verbose('lifecycle', logid(pkg, stage), 'PATH:', env[PATH])
   log.verbose('lifecycle', logid(pkg, stage), 'CWD:', wd)
   log.silly('lifecycle', logid(pkg, stage), 'Args:', [shFlag, cmd])
 
-  var proc = spawn(sh, cmdAndFlags, conf)
+  var proc = spawn(sh, [shFlag, cmd], conf)
 
   proc.on('error', procError)
   proc.on('close', function (code, signal) {
--- a/lib/utils/lifecycle.js
+++ b/lib/utils/lifecycle.js
@@ -263,7 +263,27 @@ function runCmd_ (cmd, pkg, env, wd, stage, unsafe, uid, gid, cb_) {
     shFlag = '/d /s /c'
     conf.windowsVerbatimArguments = true
 
-    cmd = path.normalize(cmd)
+    // check if we have a composed command
+    // e.g.: `"test": "node ../test/test"`
+    if (cmd.indexOf(' ') !== -1) {
+      // split the command
+      var cmdChunks = cmd.split(' ')
+
+      cmd = cmdChunks.map(function (chunk) {
+        // let's assume that a path starts with either `./` or `../`
+        // maybe using `lstat` or `lstatSync` could be cleaner
+        if (chunk.indexOf('./') !== -1) {
+          // return the normalized path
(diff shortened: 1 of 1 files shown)


jackTheRipper:
Hi, I updated the PR. Building an absolute path is not a reliable solution.

Using (normalize)[https://nodejs.org/api/path.html#path_path_normalize_path] from the Node Path api looks better.

lib/utils/lifecycle.js
```diff
@@ -262,6 +262,28 @@ function runCmd_ (cmd, pkg, env, wd, stage, unsafe, uid, gid, cb_) {
...
     conf.windowsVerbatimArguments = true
+
+    // check if we have a composed command
+    // e.g.: `"test": "node ../test/test"`
+    if (cmd.indexOf(' ') !== -1) {
+      // split the command
+      var cmdChunks = cmd.split(' ')
+
+      cmd = cmdChunks.map(function (chunk) {
+        // let's assume that a path starts with either `./` or `../`
```
mAiNiNfEcTiOn: Isn't this a wrong assumption? What if the path is absolute (E.g. `/my/folder/and/file.js` ?

Definitely using `path.parse()` (with `path.isAbsolute()`) would allow you to check if it's a valid path, and then `fs.statSync()` to check if it exists.

lib/utils/lifecycle.js
```diff
@@ -262,6 +262,28 @@ function runCmd_ (cmd, pkg, env, wd, stage, unsafe, uid, gid, cb_) {
     sh = process.env.comspec || 'cmd'
     shFlag = '/d /s /c'
     conf.windowsVerbatimArguments = true
+
+    // check if we have a composed command
+    // e.g.: `"test": "node ../test/test"`
+    if (cmd.indexOf(' ') !== -1) {
+      // split the command
+      var cmdChunks = cmd.split(' ')
```
mAiNiNfEcTiOn: Are you sure you want to do a `split()` like that?

Consider the whole set of options including the repl flag:

```bash
node -e "var myVar = './stuff'; console.log(myVar);"
```

lib/utils/lifecycle.js
jackTheRipper: Well under Windows `/my/folder/and/file.js` doesn't exist. If you wanna use an absolute path you could only do `C:\\my\\folder\\and\\file.js`.

Regarding the relative path my aim is not to change the behaviour of npm. 

The question about does the file exist or not will be managed by the `spawn` [here](https://github.com/npm/npm/pull/16498/files#diff-8c44327c5a50975dee6ded9e9789e96eR293)

lib/utils/lifecycle.js
jackTheRipper: What would you do instead of a `split(' ')` ?

mAiNiNfEcTiOn: Pass it on an _argv_ like library?

iarna:
@jackTheRipper Hey, sorry for taking so long to get back to you, you had the misfortune to file this just as we were entering the npm@5 crunch.

> `"lint": "../../node_modules/.bin/eslint.cmd --color '{src,test}/**/*.js'",`

So if you wrote that as:

> `"lint": "..\\..\\node_modules\\.bin\\eslint.cmd --color '{src,test}/**/*.js'",`

Then it would work. Your run scripts are run with `cmd` and a quirk of `cmd` is in its command lookup you have to use backslashes (forward slashes aren't accepted). This, despite forward slashes otherwise being acceptable to the Windows API. ¯\\\_(ツ)\_/¯

ghost:
Hi @iarna , thank you for your feedback.

Following what you are suggesting, should I have a set of commands for Linux and another one for Windows?

The aim of my PR is to be able to use the *nix notation for all the platform in order to avoid things such as `npm run build:nix` on a platform and `npm run build:win` on another

iarna:
@jackTheRipper  Ah, I had assumed you only cared about Windows because that command was already Windows specific: The  `.cmd` suffix is only used on Windows.

I think what we're seeing here is that `npm` does not currently support calling commands that aren't either global or dev deps. That is, there's no way to provide a specific path to something in a crossplatform manner.

The place to start for this would be to write a command that can be installed as a dev dep that takes one of these command lines as an argument and executes it in a consistent way on Windows, Linux and Mac. That will give you a result that will work regardless of npm version (or even with non-npm installers) and it gives a safe place to iterate on this while you figure out exactly what will be needed to actually make it work in a truely cross platform manner.

If after that tool has been tested in actual practice and the implementation seems sound, then I'd love to merge it into npm.

I'm going to close this issue, but thank you for the work you've done here, and I hope you give writing this a helper a try.

PR #16523, opened 2017-05-04
Fix interpreter command to /usr/bin/nodejs

metux:
The command name 'node' is very ambiguous conflicts with other packages
(eg. AX25 tools). Therefore distros have to rename it. The obvious
consensus is calling it 'nodejs'.

### DIFF
--- a/bin/npm-cli.js
+++ b/bin/npm-cli.js
@@ -1,4 +1,4 @@
-#!/usr/bin/env node
+#!/usr/bin/nodejs
 ;(function () { // wrapper in case we're in module_context mode
   // windows: running "npm blah" in this folder will invoke WSH, not node.
   /*global WScript*/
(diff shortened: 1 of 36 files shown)


zkat:
We can't possibly land something like this unless the entire ecosystem comes along for the ride.

If you want to see this change happen, you'll first have to convince the folks over at https://github.com/nodejs/node. Their installer, which is our primary distribution target after the registry itself, is what determines the binary name, and thus what our scripts need to target.

I understand the difficulty that Linux distros end up having to go through in order to support things like `node` binaries, in light of those conflicts, but our hands are tied at npm. If you are interested in packaging up `npm` for distro use, consider something like the excellent https://npm.im/pkg, which seems to work pretty well from my own vague testing. That would allow you to generate a standalone `npm` binary that would no longer rely on having `node` in `$PATH`

All that said, this specific patch itself is something we would not take: it does a straight-up rewrite, including in our dependencies, which we do not edit, ourselves. As such, I'm going to close this PR. Thanks for taking the time to submit it, and let me know if you have any questions about submitting something like this to Node Core's consideration (because any changes to this on our end are reliant on them).

PR #16533, opened 2017-05-05
feat(git): git repos with prepare scripts are now built on install

zkat:
Not entirely sure how I feel about this one, tbh. That install actually takes a while, and it's hard to figure out wtf logging stuff should actually be done (if any).

It doesn't feel great and I'm not sure how to make it so.


zkat:
Alright, this is "done".

PR #16546, opened 2017-05-06
remove temporary files

metux:


legodude17:
How is this a temporary file?

metux:
seems like an editor backup, thus temporary

legodude17:
Ah, yeah.

iarna:
Thanks for putting this together. As of npm@5 this is no longer applicable as we've removed `normalize-git-url` from npm.

PR #16547, opened 2017-05-06
read-package-json.js: fix permissions

metux:
The script isn't directly executable, therefore the +x flag is wrong.


iarna (commented):
I'd never even noticed that there, tbh. It's been sitting largely untouched since 2011. It was probably intended to ship as an additional binary next to npm, but that never happened…

I'm inclined to either just remove it, or add a shebang so it _can_ be directly executed.

iarna:
Thanks for pointing this out. I removed `read-package-json.js` with 2ef0670fb which is now in `release-next`.

PR #16548, opened 2017-05-06
fix permissions of documents

metux:
Documents obviously aren't executable, so drop +x flag


iarna:
Thanks! This is a good catch! I've merged this to `release-next` as f9c18779f.

PR #16550, opened 2017-05-06
Makefile: fix doc build tools dependencies

metux:
The build tools for the documentation need to be built/installed
before the documents, even with parallel builds.

Make has a simple mechanism which was made exactly for that:
target dependencies.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -87,12 +87,16 @@ doc-clean:
     html/doc \
     man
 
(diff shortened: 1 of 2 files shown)


iarna:
Thanks for your patience and for this. This is is landed on `release-next` as 0756d687d and will be in the next npm release.

PR #16602, opened 2017-05-11
Add support for long option to view output

andytompkins:
This PR adds support for the "--long" option for view output.
Issue was filed here: https://github.com/npm/npm/issues/16590
Without the option the behavior is exactly the same as before. With the long option the array output will not be truncated (and in my use case, all versions will be shown).

### DIFF
--- a/lib/view.js
+++ b/lib/view.js
@@ -259,7 +259,22 @@ function printData (data, name, cb) {
         if (npm.config.get('json')) {
           msgJson[msgJson.length - 1][f] = d
         } else {
-          d = util.inspect(d, false, 5, npm.color)
+          if (npm.config.get('long')) {
+            d = util.inspect(d, {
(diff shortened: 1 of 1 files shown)


legodude17:
I think that by default is should just JSON stringify it and print it.

iarna:
As a rule, changes need tests and docs. I'm prepared to forgo tests for this, but adding a mention to the `view` docs would be nice.

iarna (changes requested):
As a rule, changes need tests and docs. I'm prepared to forgo tests for this, but adding a mention to the `view` docs would be nice.

iarna:
We're closing this pull request as it has gone sixty days without activity after we requested changes.  In our experience pull requests that go this long are unlikely to land.  If you do write docs for this, please don't hesitate to open a new pull request.

PR #16633, opened 2017-05-15
Parallelize tarball extraction

zkat:


lib/install/action/extract-worker.js
```diff
@@ -0,0 +1,16 @@
...
+module.exports = (args, cb) => {
+  const parsed = typeof args === 'string' ? JSON.parse(args) : args
+  const spec = parsed[0]
+  const extractTo = parsed[1]
+  const opts = parsed[2]
+  if (!opts.log && opts.loglevel) {
+    opts.log = npmlog
+    opts.log.level = opts.loglevel
+  }
+  extract(spec, extractTo, opts).nodeify(cb)
```
rmg: `extract` here may be returning a "vanilla" `Promise` from https://www.npmjs.com/package/promise-retry which is not a Bluebird promise, so it won't have a `nodeify` method.

This is spewing out all over the place with the current canary version (5.0.2-canary.10) which appears to have this code in it.
```
/Users/ryan/.nvs/node/8.0.0/x64/lib/node_modules/npmc/lib/install/action/extract-worker.js:15
  extract(spec, extractTo, opts).nodeify(cb)
                                 ^

TypeError: extract(...).nodeify is not a function
    at module.exports (/Users/ryan/.nvs/node/8.0.0/x64/lib/node_modules/npmc/lib/install/action/extract-worker.js:15:34)
    at handle (/Users/ryan/.nvs/node/8.0.0/x64/lib/node_modules/npmc/node_modules/worker-farm/lib/child/index.js:41:8)
    at process.<anonymous> (/Users/ryan/.nvs/node/8.0.0/x64/lib/node_modules/npmc/node_modules/worker-farm/lib/child/index.js:47:3)
    at emitTwo (events.js:125:13)
    at process.emit (events.js:213:7)
    at process.nextTick (internal/child_process.js:755:12)
    at _combinedTickCallback (internal/process/next_tick.js:95:7)
    at process._tickCallback (internal/process/next_tick.js:161:9)
```

zkat: oop. Well, that's fun. Thanks for catching this. :)

zkat:
@rmg I've pushed a change to wrap it in a bluebird promise, and I've made a patch over in pacote to also make sure we return a `Bluebird` object from the `retry`. Thanks 🎉

PR #16634, opened 2017-05-16
Convert Windows newlines in hashbang CLI scripts on install

### LINKED ISSUE #2800: Unix/Linux line endings in CLI need to be enforced
`bin/mocha` and `bin/_mocha` need to have Unix line endings. Currently Git's autocrlf setting can screw this up. (See also #2799 for an example of when this goes wrong.) Our tests don't check for it either.

We will need to either check for or fix the line endings in these files as part of publishing. Or, if doing it on publish isn't feasible (prepublish vs. prepublishOnly seems to be a bit of an issue?), on version (most of our other publishing prep is in versioning already anyway...).

Ideally, I'd like to just run a Node command and convert the line endings, so it doesn't matter what Git did (or the user, etc) -- this isn't something that should have to be fixed manually when it is detected.

### LINKED ISSUE #12371: bin scripts should always have Unix line endings
See #2097 for context and discussion. When publishing from Windows, depending on how developers set up Git, it's possible that binary scripts will end up in the tarball with `CR LF` line endings, instead of the `LF` expected by Unix / OS X. This can cause failures when those scripts are invoked. The logic that Git applies to figure out whether a file is binary or not is surprisingly subtle, and this is the sort of thing that can cause really hard to debug issues if not handled correctly. This is the heuristic @iarna and I have come up with for figuring out which files need their line endings cleaned up:
1. Is it linked to from `bin`?
2. Is the first line a shebang?
#2 is necessary in the case in which `bin` links to a native binary, which we don't want to mangle during the publish or installation processes.

To be a complete solution, this would need to be some kind of stream transform that gets applied at both publish and install time – publish, so that newly-published packages have the correct line endings, and install, so that non-Windows users running installs with the fixed version of the CLI aren't bitten by legacy packages with the wrong line endings in their scripts.

### LINKED ISSUE #13203: npm has a line ending problem
I realize this issue has been addressed multiple times in the past, but I am bringing it back up because as it stands, here is what is happening:
- The NPM Package manager repository is becoming littered with cross-platform incompatible packages, severely hindering NPM's viability as a cross-platform solution.

The problem is that the Windows git-scm default install sets the core.autocrlf setting to true. This, by default, replaces the line endings with CRLF when you do a git clone. This means that every time I clone the repository and theN publish it to NPM, I have now published a broken package. Multiply this times a million and you get the idea. Tons of "no such file or directory" errors on a cli !# /usr/bin/env node line.

NPM is the perfect abstraction layer / pipe to have a one in - one out solution for the problem. It should not be left up to the windows user to always have to deal with this problem, or even be aware of how line endings work on windows vs linux. The end user should not have to deal with these kind of technical details. It is this kind of problem that hurts NPM way more than it helps them, and expecting the windows users to continually manage this problem on their own is just unrealistic, as well as continues to hurt NPM.

This is a major problem, and I can't believe I wasted so many hours on it.

ScottFreeCode:
Fixes #12371 -- mostly... It does not include the packaging/publishing side, only the installing side. (However, it does include a couple util functions, `isHashbangFile` and `dos2Unix`, that may be reusable for the packaging/publishing side.)
Fixes #13203 (duplicate of #12371)
Fixes https://github.com/mochajs/mocha/issues/2800

(Corrected, see below.) ~Or at least, it would fix these things if the rename action in the newline conversion step worked. I set it up to stream the converted file to a temporary and then rename because I wasn't sure what would happen if I tried to stream from and to the same file (plus the effect would be atomic, which is always helpful), but the test keeps getting EPERM errors on the rename and then more EPERM errors trying to do something else to the file after that. I tried a few different things to see if I could fix it -- use graceful-fs instead of fs, unlink the file before renaming so it's only very-nearly-atomic, use the stream close event instead of the stream finish event, do it before chmodding the file permissions instead of after... none of these seem to solve the issue though.~

~So I'm posting this in case anybody knows what I'm doing wrong and can get it working without too much additional trouble -- hopefully this at least saves a little implementation effort...?~

(Let me know if this needs to be squashed manually rather than by GitHub's squash merge option.)


ScottFreeCode:
Hmm.

So, the test failures on Travis are due to my misunderstanding the use of buffer and `bytesRead` in `fs.read`. Simply removing [this line](https://github.com/ScottFreeCode/npm/blob/5d7e698dc65e8c921a13c21d3b5ec5397bda86e3/lib/utils/is-hashbang-file.js#L11) should correct those.

I also noticed that in `convert-line-endings` I could be using `fs-write-stream-atomic`, not fiddling with an explicit temporary file and just passing `cb` to the `finish` handler. I guess that should be switched along with using `graceful-fs` instead of `fs` in that module.

*I can go ahead and add those real quick if it's not going to get in the way of reviewers.*

Even with those updates made locally, I still can't figure out why Windows is getting EPERM issues though (and I've tried additional tweaks to mode settings such as using `stats.mode & 0o777` or removing the mode parameter to the write stream altogether, both before and after the chmod action) -- and, somewhat alarmingly, Travis isn't showing those errors, so there's potential here for bugs to slip through...

ScottFreeCode:
Fixed!!

Apparently I'd forgotten to close the opened file after checking for the hashbang. I guess Windows is pickier about not messing with open files (come to think of it, I recall having heard that Unix/Linux will just keep using open file handles to the old file while pointing the name to a new file...).

Removed the offending line that was breaking other tests too.

ScottFreeCode:
The test added in this PR was based on https://raw.githubusercontent.com/npm/npm/d46015256941ddfff1463338e3e2f8f77624a1ff/test/tap/install-cli-production.js if I recall correctly; to handle the NPM 5 updates, it may be necessary to update the test added in this PR in accord with whatever updates occurred to `install-cli-production.js`...

lib/utils/convert-line-endings.js
```diff
@@ -0,0 +1,42 @@
...
+        },
+        flush: function (done) {
+          if (previousChunkEndedInCR) {
+            this.push('\r')
+          }
+          done()
+        }
+      }))
+      .on('error', cb)
+      .pipe(fsWriteStreamAtomic(file, { mode: stats.mode }))
```
iarna, line 36: Despite the docs for `fs-write-stream-atomic`, it doesn't actually take a `mode` argument.

iarna (approved):
This has landed as 0ef320cb4. I rewrote `isHashbangFile` and `convertLineEndings` to use promises (which we're switching toward more generally).

iarna:
This is in `release-next` and will, as is implied, be in the next release. Publish-side stuff has to wait for `tarv3` to land, as it completely changes how we bundle up files, but that's landing any moment now.

lib/utils/convert-line-endings.js
ScottFreeCode: Huh, that's surprising -- good catch!

ScottFreeCode:
Thanks!!

> Publish-side stuff has to wait for  tarv3  to land, as it completely changes how we bundle up files, but that's landing any moment now.

Cool! I'm really glad that that's being looked at -- I realized later that, while the installation side is easier to implement since there's already a bin-links step in installation, it will probably take a publication-level fix to entirely resolve the issue for projects like Mocha that tend to support testing against rather old versions of Node/npm. But then I hadn't had time to go back and figure out where to put a publishing-side fix. So thanks for staying on the ball with this!

marcosscriven:
Unfortunately this break binary shebang files - I've raised an issue here https://github.com/npm/npm/issues/19151

Issue #15727, opened 2017-02-10
Incorrect extraneous package detection


stouf:
#### I'm opening this issue because:

  - [ ] npm is crashing.
  - [ ] npm is producing an incorrect install.
  - [X] npm is doing something I don't understand.
  - [ ] Other (_see below for feature requests_):

#### What's going wrong?

The repository I'm currently working on has a dependency declared in its `package.json` file as such:

```json
{
  "dependencies": {
    "name-of-my-private-lib": "git+https://XXXXXXXXX:x-oauth-basic@github.com/username/my-private-lib.git#vx.y.z"
  }
}
```

`name-of-my-private-lib` also depends on `mocha`, that is being declared as followed:

```json
{
  "dependencies": {
    "mocha": "=2.3.4"
  }
}
```

The repository I'm currently working on uses `npm shrinkwrap`.
Recently, I removed both the `node_modules` folder and the `npm-shrinkwrap.json` file, ran the `npm install` command and then ran the `npm shrinkwrap` command but that one failed, telling me that there are some extraneous packages. Usually, I run `npm prune`, run `npm shrinkwrap` again and it works, but unfortunately not this time.
Here are the 2 things I've noticed:

- with npm v4.1.2, `npm ls` lists all my packages and flags `mocha@2.3.4` as extraneous package but `npm prune` does not remove it.
- under the same conditions, running `npm prune` with npm v4.0.5 will remove `mocha@2.3.4` whereas npm v4.1.2 will not

See the following:

```
$ npm ls
└─┬ name-of-my-private-lib@0.8.0 (git+https://XXXXXXX:x-oauth-basic@github.com/username/my-private-lib.git#1234567890987654322)
  ├─┬ i18next@1.10.2
  │ ├─┬ cookies@0.6.2
  │ │ ├── depd@1.1.0
  │ │ └── keygrip@1.0.1
  │ ├── i18next-client@1.10.2
  │ └── json5@0.2.0
  ├── mocha@2.3.4 extraneous
  ├─┬ moment-timezone@0.3.1
  │ └── moment@2.10.6
  └─┬ pdfmake@0.1.18
    ├── lodash@3.1.0
    ├── lodash-node@3.1.0
    └─┬ pdfkit@0.7.1
      ├─┬ linebreak@0.1.2
      │ └── unicode-trie@0.1.2
      └── png-js@0.1.1

[...]

Issue #15669, opened 2017-02-03
`npm prune --production` deletes dependencies found in both `dependencies` and `devDependencies`


eschwartz:
#### I'm opening this issue because:

  - [ ] npm is crashing.
  - [ ] npm is producing an incorrect install.
  - [X] npm is doing something I don't understand.
  - [ ] Other (_see below for feature requests_):

#### What's going wrong?

Situation
- You have the same dependency listed in your `package.json`'s `dependencies` and `devDependencies`
- You install your dependencies
- You run `npm prune --production` using npm **v4.1.2**

What actually happens:
- The dependency is deleted from your `node_modules` folder

What I expect to happen:
- The dependency should _not_ be deleted, as it is a production dependency.

Note the using npm v4.05 results in expected behavior. 

#### How can the CLI team reproduce the problem?

Create a `package.json` file like so:

```
{
  "devDependencies": {
    "mongodb": "^2.2.21"
  },
  "dependencies": {
    "mongodb": "^2.2.21"
  }
}
```

Using npm v4.1.2 run:

```
$ npm install
$ npm prune --production
```

### supporting information:

 - `npm -v` prints: 4.1.2
 - `node -v` prints: v4.3.2, also tried with v6.9.1
 - `npm config get registry` prints: https://registry.npmjs.org/
 - Windows, OS X/macOS, or Linux?: Linux
 - Network issues:
   - Geographic location where npm was run:
   - [ ] I use a proxy to connect to the npm registry.
   - [ ] I use a proxy to connect to the web.
   - [ ] I use a proxy when downloading Git repos.
   - [ ] I access the npm registry via a VPN
   - [ ] I don't use a proxy, but have limited or unreliable internet access.
 - Container:
   - [ ] I develop using Vagrant on Windows.
   - [ ] I develop using Vagrant on OS X or Linux.
   - [x] I develop / deploy using Docker. 
     I tried this on my Ubuntu 16.04 machine, and also [in a container running Linux](https://github.com/lambci/docker-lambda/blob/master/nodejs4.3/build/Dockerfile)
   - [ ] I deploy to a PaaS (Triton, Heroku).

doug-wade:
I... why do you have mongo listed as both a devDependency and a regular dependency?  Why not remove it from your devDeps?

eschwartz:
> why do you have mongo listed as both a devDependency and a regular dependency

It's a developer mistake.

[...]

Issue #15646, opened 2017-02-01
npm prune corrupts non-extraneous packages


Lodin:
#### I'm opening this issue because:

  - [ ] npm is crashing.
  - [ ] npm is producing an incorrect install.
  - [x] npm is doing something I don't understand.
  - [ ] Other (_see below for feature requests_):

#### What's going wrong?
I created a PR to the [create-elm-app](https://github.com/halfzebra/create-elm-app), and the travis build went wrong in some reason. After exploring I have understood that the `npm prune` removes all files from the `node_modules/.bin` directory. It happens only with `node v7.5.0` and `npm 4.1.2`, so the build for `node v6.9.5` and `npm v3.10.10` was successful. `node v7.4.0` works well too. 

It works not only under travis Debian, but under my Windows 10 as well.

Logs: 
* The travis log for failed build can be found [here](https://travis-ci.org/halfzebra/create-elm-app/jobs/197337880)
* The travis log for successful build can be found [here](https://travis-ci.org/halfzebra/create-elm-app/jobs/197337881)

#### How can the CLI team reproduce the problem?
Under the `node v7.5.0` and `npm 4.1.2`
```sh
$ npm install -g yarn
$ git clone https://github.com/Lodin/create-elm-app.git
$ cd create-elm-app
$ git checkout webpack-update
$ yarn
$ npm link ../create-elm-app
$ npm prune
$ npm test
```
After the last command it should fail with following error:
```sh
sh: 1: mocha: not found
npm ERR! Test failed.  See above for more details.
```
### supporting information:

 - `npm -v` prints: 4.1.2
 - `node -v` prints: 7.5.0
 - `npm config get registry` prints: https://registry.npmjs.org/
 - Windows, OS X/macOS, or Linux?: Windows 10 and Debian Linux
 - Network issues:
   - Geographic location where npm was run:
   - [ ] I use a proxy to connect to the npm registry.
   - [ ] I use a proxy to connect to the web.
   - [ ] I use a proxy when downloading Git repos.
   - [ ] I access the npm registry via a VPN
   - [ ] I don't use a proxy, but have limited or unreliable internet access.
 - Container:
   - [ ] I develop using Vagrant on Windows.
   - [ ] I develop using Vagrant on OS X or Linux.
   - [ ] I develop / deploy using Docker.
   - [ ] I deploy to a PaaS (Triton, Heroku).

PR #15090, opened 2016-12-01
New npm@3 style prune command

iarna:
This eliminates one of the last three remaining uses of `readInstalled` in the code base. (The other two being `rebuild` and `utils/completion/installed-deep`.)

PR #16637, opened 2017-05-16
Fix prune, add tests to demonstrate brokenness in npm 4.1.0+

exogen:
This adds two test cases to demonstrate how npm 4.1.0+ (including the latest version) fails to prune in certain cases (likely introduced in #15090). A fix is included, but will need to be reviewed carefully (I'm not sure whether cycle detection or checking `userRequired` is necessary, or if that's already taken care of by `isExtraneous`).

`prune` misbehaving is likely the cause of some of the funny business on display in issues #15727, #15669, #15646.

My interest in this stems from the fact that broken pruning causes [postinstall-build](https://github.com/exogen/postinstall-build) to break on the latest versions of npm ever since 4.1.0.

### The Tests

* Call `npm prune --production` on a module with ONLY these dependencies:

  ```json
  {
    "devDependencies": {
      "test-package-with-one-dep": "0.0.0",
      "test-package": "0.0.0"
    }
  }
  ```

  ...which looks like this before pruning:

  ```console
  ├── test-package@0.0.0
  └─┬ test-package-with-one-dep@0.0.0
    └── test-package@0.0.0  deduped
  ```

  `npm prune --production` SHOULD remove both packages, resulting in an empty `node_modules` tree, but it fails to remove `test-package` (because it's a prod dep of `test-package-with-one-dep`, but that shouldn't matter because `test-package-with-one-dep` itself is a dev dep).

* Call `prune --production` with the same package in both `dependencies` and `devDependencies`. The package SHOULD NOT be removed.

### The Fix

In `--production` mode, packages are now pruned if they are only reachable from dev dependencies (the actual logic is: NOT reachable from prod or opt dependencies).

/cc @iarna

### DIFF
--- a/test/tap/prune-dev-dep-cycle.js
+++ b/test/tap/prune-dev-dep-cycle.js
@@ -1,0 +1,79 @@
+'use strict'
+var fs = require('fs')
+var path = require('path')
+var test = require('tap').test
+var Tacks = require('tacks')
+var File = Tacks.File
+var Dir = Tacks.Dir
+var common = require('../common-tap.js')
+var testdir = path.join(__dirname, path.basename(__filename, '.js'))
+
+var fixture = new Tacks(
+  Dir({
+    node_modules: Dir({
+      'a': Dir({
+        'package.json': File({
+          _requested: {
+            rawSpec: 'file:///mods/a'
+          },
+          dependencies: {
+            'b': 'file:///mods/b'
+          },
+          name: 'a',
+          version: '1.0.0'
+        })
+      }),
+      'b': Dir({
+        'package.json': File({
+          _requested: {
+            rawSpec: 'file:///mods/b'
+          },
+          dependencies: {
+            'a': 'file:///mods/a'
+          },
+          name: 'b',
+          version: '1.0.0'
+        })
+      })
+    }),
+    'package.json': File({
+      name: 'test',
+      version: '1.0.0',
+      devDependencies: {
+        'a': 'file:///mods/a'
+      }
+    })
+  })
+)
+
+function setup () {
+  cleanup()
+  fixture.create(testdir)
+}
+
+function cleanup () {
+  fixture.remove(testdir)
+}
+
+test('setup', function (t) {
+  setup()
+  t.end()
+})
+
+test('prune cycle in dev deps', function (t) {
+  common.npm(['prune', '--production'], {cwd: testdir}, function (err, code, stdout, stderr) {
+    if (err) throw err
+    t.is(code, 0, 'prune finished successfully')
+    t.equal(stderr,
+      '- a@1.0.0 node_modules/a\n' +
+      '- b@1.0.0 node_modules/b\n')
+    var dirs = fs.readdirSync(testdir + '/node_modules').sort()
+    t.same(dirs, [])
+    t.end()
+  })
+})
+
+test('cleanup', function (t) {
+  cleanup()
+  t.end()
+})
--- a/test/tap/prune-with-dev-dep-duplicate.js
+++ b/test/tap/prune-with-dev-dep-duplicate.js
@@ -14,12 +14,14 @@ var pkg = path.resolve(__dirname, 'prune')
 var cache = path.resolve(pkg, 'cache')
 
 var json = {
-  name: 'prune-with-only-dev-deps',
+  name: 'prune-with-dev-dep-duplicate',
   description: 'fixture',
   version: '0.0.1',
   main: 'index.js',
+  dependencies: {
+    'test-package': '0.0.0'
+  },
   devDependencies: {
-    'test-package-with-one-dep': '0.0.0',
     'test-package': '0.0.0'
   }
 }
@@ -59,7 +61,7 @@ test('npm install', function (t) {
 
 test('verify installs', function (t) {
   var dirs = fs.readdirSync(pkg + '/node_modules').sort()
-  t.same(dirs, [ 'test-package', 'test-package-with-one-dep' ].sort())
+  t.same(dirs, [ 'test-package' ].sort())
   t.end()
 })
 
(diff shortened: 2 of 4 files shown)


exogen:
```js
    var isChildDev = function (parent) { return isDev(parent, childName) }
    return child.requiredBy.every(isChildDev)
```

This appears to be the logic at fault. If I understand correctly, it only prunes if `child` is a `devDependency` in every package it's required by, but that's not right: it should prune if it's only required by other `devDependencies`.

exogen:
I came up with a fix and updated the PR description. There are now two test cases; CI shows that they both demonstrate the failure until this fix is applied. 😄

exogen:
Some testing shows that this does indeed need to take cycles into account.

IMO it would be nice if this behavior was baked into `isExtraneous` in the first place (by adding a `production` flag). In production mode, devDeps should be considered extraneous. Then prune's job is to simply remove anything marked extraneous...no additional logic.

test/tap/prune-dev-dep-cycle.js
```diff
@@ -0,0 +1,79 @@
...
+  t.end()
+})
+
+test('prune cycle in dev deps', function (t) {
+  common.npm(['prune', '--production'], {cwd: testdir}, function (err, code, stdout, stderr) {
+    if (err) throw err
+    t.is(code, 0, 'prune finished successfully')
+    t.equal(stderr,
+      '- a@1.0.0 node_modules/a\n' +
+      '- b@1.0.0 node_modules/b\n')
```
iarna, line 69: I wouldn't expect this to ever pass because that output is printed to `stdout` not `stderr`.

test/tap/prune-with-dev-dep-duplicate.js
```diff
@@ -0,0 +1,116 @@
...
+
+test('npm install', function (t) {
+  common.npm([
+    'install',
+    '--cache', cache,
+    '--registry', common.registry,
+    '--loglevel', 'silent',
+    '--production', 'false'
+  ], EXEC_OPTS, function (err, code, stdout, stderr) {
+    t.ifErr(err, 'install finished successfully')
```
iarna, line 55: I know you likely just copied this from another test, but it was wrong there too. =D

The test label says 'install finished successfully' but that's not what not getting an `err` value means. `err` will ONLY be set if something catastrophic happened like we couldn't open a new process. This would be better written as:

```
if (err) throw err
```

(The story of how any tests started doing the `ifError` thing is an interesting bit of code archaeology, but the long and short of it is that this is never an appropriate check in this context. It was introduced due to either errors when converting between two types of test, or by someone who cargo culted it from another similar but quite different context.)

test/tap/prune-with-dev-dep-duplicate.js
```diff
@@ -0,0 +1,116 @@
...
+
+test('npm prune', function (t) {
+  common.npm([
+    'prune',
+    '--loglevel', 'silent',
+    '--production'
+  ], EXEC_OPTS, function (err, code, stderr) {
+    t.ifErr(err, 'prune finished successfully')
+    t.notOk(code, 'exit ok')
+    t.notOk(stderr, 'Should not get data on stderr: ' + stderr)
```
iarna, line 95: This is not `stderr` that is being checked here, this is `stdout`.

I'm honestly not sure what the expected behavior for this one is? In my testing with the bleeding edge, it removes this:

```
    #     {
    #       "action": "remove",
    #       "name": "test-package",
    #       "version": "0.0.0",
    #       "path": "node_modules/test-package"
    #     }
```
Is that correct, or wrong?

test/tap/prune-with-only-dev-deps.js
```diff
@@ -0,0 +1,116 @@
...
+
+test('npm prune', function (t) {
+  common.npm([
+    'prune',
+    '--loglevel', 'silent',
+    '--production'
+  ], EXEC_OPTS, function (err, code, stderr) {
+    t.ifErr(err, 'prune finished successfully')
+    t.notOk(code, 'exit ok')
+    t.equal(stderr,
```
iarna, line 93: This variable holds STDOUT, not STDERR.

test/tap/prune-with-only-dev-deps.js
```diff
@@ -0,0 +1,116 @@
...
+  t.end()
+})
+
+test('npm prune', function (t) {
+  common.npm([
+    'prune',
+    '--loglevel', 'silent',
+    '--production', 'false'
+  ], EXEC_OPTS, function (err, code, stdout, stderr) {
+    t.ifErr(err, 'prune finished successfully')
```
iarna, line 72: Same comments about ifErr apply

iarna (changes requested):
With and without your changes I'm seeing this test failure:
```
EDIT: never mind this, I misread my own edited output =D
```

iarna (approved):

iarna:
Ok, so don't worry about making any changes to this, I wanted to give you the feedback, but I've gone and made them myself when I was making this merge clean w/ the branch that will be release-next tomorrow. (or released tomorrow, not sure =D)

This is what'll be merged: https://github.com/npm/npm/compare/lock-missing-dep-fixes...iarna/pr/16637

exogen:
@iarna Thanks for the feedback! I was indeed confused about the stderr/stdout discrepancies myself when making these changes – the code is copied from existing tests that had the same mistake. And yes, I left this with new test cases that were failing, never got around to solving it completely.

test/tap/prune-with-dev-dep-duplicate.js
exogen: If I understand correctly, that is wrong: `test-package` should remain, because the result of `npm prune --production` should be "anything NOT required by production `dependencies` is removed", but `test-package` is in `dependencies`, so removing it would break the install.

iarna:
This was merged into `release-next` as 8a2d739fb and 0712ac695.

PR #16686, opened 2017-05-22
utils/link: Make less race prone

iarna:
Previously it was gently-rming it's target before making a new target
directory.  gently-rm vacuums, removing empty directories.  This was fine
until you run more than one `link` command concurrently, at which point it's
entirely possible for one invocation to `mkdir` the directory, then
immediately have another one `gently-rm` it, at which point the symlink
fails.  Races!


lib/install.js
```diff
@@ -400,26 +400,25 @@ Installer.prototype.loadAllDepsIntoIdealTree = function (cb) {
...
+    steps.push(
+      [prefetchDeps, this.idealTree, depsToPreload, cg.newGroup('prefetchDeps')])
+  }
+  if (this.prod) {
+    steps.push(
+      [loadDeps, this.idealTree, cg.newGroup('loadDeps')])
+  }
+  if (this.dev) {
+    steps.push(
+      [loadDevDeps, this.idealTree, cg.newGroup('loadDevDeps')])
```
zkat: wait wait how were we even fetching things at all before?! D:

lib/install.js
```diff
@@ -581,7 +580,14 @@ Installer.prototype.saveToDependencies = function (cb) {
   validate('F', arguments)
   if (this.failing) return cb()
   log.silly('install', 'saveToDependencies')
-  saveRequested(this.args, this.idealTree, cb)
+  if (
+    this.differences.length ||
+    (this.idealTree.removedChildren || []).length
+  ) {
+    saveRequested(this.idealTree, cb)
```
zkat: I think we need to save unconditionally, because:
1. I think this generates the initial `package-lock.json` even when you have an up-to-date `node_modules`.
2. We want to update the `packageIntegrity` even if it's something else that changed its hash

lib/prune.js
```diff
@@ -59,7 +59,7 @@ Pruner.prototype.loadAllDepsIntoIdealTree = function (cb) {
   var toPrune = this.idealTree.children.filter(shouldPrune).map(getModuleName).filter(matchesArg).map(nameObj)
 
   steps.push(
-    [removeDeps, toPrune, this.idealTree, null, cg.newGroup('removeDeps')],
+    [removeDeps, toPrune, this.idealTree, null],
```
zkat: is this gonna get rid of the `removing: <pkg>` logs?

lib/uninstall.js
```diff
@@ -27,17 +24,18 @@ uninstall.completion = require('./utils/completion/installed-shallow.js')
...
-  })
 
-  var where = npm.config.get('global') || !args.length
+  const where = npm.config.get('global') || !args.length
             ? path.resolve(npm.globalDir, '..')
             : npm.prefix
 
+  args = args.filter(function (a) {
+    return path.resolve(a) !== where
+  })
```
zkat: I'd rather warn users if they do this. This is totally weird of them to do.

lib/utils/link.js
```diff
@@ -43,6 +43,10 @@ function ensureFromIsNotSource (from, to, cb) {
   })
 }
 
+function unlinkAndIgnoreErrors (file, cb) {
+  fs.unlink(file, () => cb())
```
zkat: `npm yolo`

iarna:
@kat yeah, the branch this was merging _into_ got rebased and repushed and that made the commit list wonky. It's just the one commit you yoloed on =)

lib/utils/link.js
zkat: `test/tap/unit-link.js` needs to be updated, since it mocks `fs`. It's failing with `fs.unlink is not a function` right now.

zkat (changes requested):
needs a unit test fix. All other failures are acceptable because those tests are failing in the link-specifier branch itself.

lib/utils/link.js
iarna: ah yeah…

zkat:
This was already merged in and included in npm@5.0.0. Closing.

Issue #6543, opened 2014-10-22
npm scripts shell select


XVincentX:
Good morning sirs,
I would like to understand if npm scripts under script section of _packages.json_ are run using the shell listed into configuration 

> npm config get shell

From my tests (SO: Windows), looks like that even if **sh** have been set (--global flag), cmd.exe is being used and so it is unable to perform my tasks.

I've made an overview of documentation and source code too, but I was not able to figure out.

Thank you,
V.

majgis:
I was very surprised to find that when using cygwin, git bash, or PowerShell, `npm run` still executes all commands in cmd.exe.   This behavior seems like a serious flaw with npm.

majgis:
I deleted this [if statement](https://github.com/npm/npm/blob/master/lib/utils/lifecycle.js#L17) and executed `npm run`.  It behaved as I expected under cygwin, git bash and powershell.

majgis:
I can see two issues with using NPM_CONFIG_SHELL for run scripts on windows:

1. This would be a breaking change
2. The args for spawn are different for cmd vs bash, so the shell type would still need to be identified on windows.

I explored the environment variables output from `process.env` across PowerShell, git bash and cygwin and I couldn't find anything consistent to hook into.

**Proposal:**
Create a new npm config setting called `posix`.
* The command `npm config set posix true` sets posix=true in .npmrc
* The default value in windows is `false`, and it is `true` on all other platforms
* Those using cygwin, powershell, git bash or other POSIX shell can set NPM_CONFIG_POSIX=true
*  Any other behavior, in addition to run scripts, where it would make sense to flip to the POSIX default behavior should also be changed.
* Existing projects need only add the following to their package.json to make their project POSIX friendly, and make it explicit what syntax is being used:

        "config": {
             "posix":true
        }

XVincentX:
I have no knowledge at all on nom code (so I am not the right person to evaluate your proposal), but I like your solution, actually.

othiym23:
If – and this is a big if – a way can be found to support this that doesn't _ever_ break things for Windows users following the path of least resistance (i.e. running npm from `cmd.exe`, using the npm included by default with Node, installed via an official Node.js or io.js installer), then I could see it landing.

[...]

PR #16687, opened 2017-05-22
feat: allow custom shell for run-script

mmkal:
Addresses https://github.com/npm/npm/issues/6543

Maybe I'm under-thinking the problem, but this seems to work for me. I can set
`command-shell = C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe`
or
`command-shell = C:\Windows\System32\bash.exe`
in my `.npmrc` and it just works.

~I'm guessing tests would go in https://github.com/npm/npm/blob/latest/test/tap/run-script.js, but I'd be interested to hear thoughts on the approach before I go that far.~

### DIFF
--- a/doc/misc/npm-config.md
+++ b/doc/misc/npm-config.md
@@ -271,13 +271,6 @@ It is _not_ the path to a certificate file (and there is no "certfile" option).
 If false, never shows colors.  If `"always"` then always shows colors.
 If true, then only prints color codes for tty file descriptors.
 
-### command-shell
-
-* Default: `null`
-* Type: path
-
-The shell to use for scripts run with the `npm run` command.
-
 ### depth
 
 * Default: Infinity
@@ -838,6 +831,13 @@ in to a private registry for the first time:
 `npm login --scope=@organization --registry=registry.organization.com`, which
 will cause `@organization` to be mapped to the registry for future installation
 of packages specified according to the pattern `@organization/package`.
+
+### script-shell
+
+* Default: `null`
+* Type: path
+
+The shell to use for scripts run with the `npm run` command.
 
 ### scripts-prepend-node-path
 
--- a/doc/misc/npm-config.md
+++ b/doc/misc/npm-config.md
@@ -271,6 +271,13 @@ It is _not_ the path to a certificate file (and there is no "certfile" option).
 If false, never shows colors.  If `"always"` then always shows colors.
 If true, then only prints color codes for tty file descriptors.
 
+### command-shell
+
+* Default: `null`
+* Type: path
+
+The shell to use for scripts run with the `npm run` command.
+
 ### depth
 
 * Default: Infinity
--- a/lib/config/defaults.js
+++ b/lib/config/defaults.js
@@ -129,9 +129,6 @@ Object.defineProperty(exports, 'defaults', {get: function () {
     cert: null,
 
     color: true,
-
-    'command-shell': null,
-
     depth: Infinity,
     description: true,
     dev: false,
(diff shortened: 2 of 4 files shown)


mmkal:
@othiym23 @isaacs @majgis any thoughts on this?

misha-kaletsky-zocdoc:
@zkat ?

mmkal:
@zkat ?

mmkal:
I'm going to keep trying to bump this, as it would greatly improve development experience on Windows... @KenanY @zkat @watson @iarna

kenany:
I'll defer to the other members to review this.

zkat (changes requested):
Okay! We got to discussing this, and we decided to take this PR, pending one change:

Could you change this to `script-shell`? It's a bit of a bikeshed, but we decided this is a more fitting name in general.

Once that change happens, and the test suite passes, we'll merge this in! 🎉  Thanks a bunch for the PR!

zkat (approved):

zkat:
Cool beans 🐻

mmkal:
@zkat thanks!

danielbayley:
@zkat Could we be able to set the shell somewhere in `package.json`, rather than necessarily having to have an `.npmrc` file to utilise this feature?

Maybe something like:
~~~ json
"config": {
  "script_shell": "/bin/zsh"
}
~~~
or even
~~~ json
"scripts": {
  "shell": "/bin/zsh"
}
~~~

mmkal:
@danielbayley that should already work on npm 5.1.0+ if you use 
```
  "config": {
    "script-shell": "/bin/zsh"
  }
```

ie use a hyphen rather than an underscore.

See https://docs.npmjs.com/misc/config - that's a valid way to set config values.

iarna:
@mmkal 

The `config` section of `package.json` is used as a way of passing values to run scripts, but it's not used by npm itself to load configuration. Sorry, those docs could be clearer.

`publish-config` _is_ but only during publication.

The usual approach would be to have this as a project-local `.npmrc`, which is already supported. Just drop one in next to your `package.json`. Be aware that setting this to something truely exotic (like `tcsh`) may break some dependencies as this setting is per-project, not per-module. That is, we don't read `.npmrc` files from modules when we install them.

mmkal:
Ah, sorry, didn't realise. In that case I agree with @danielbayley !

danielbayley:
Also, I tried to utilise this with `npm config set script-shell node_modules/.bin/shx`

But for some reason it doesn't work… I'm not sure if the problem is with npm or [shx], so I posted on both and linked…

~~~ ini
# .npmrc
script-shell = node_modules/.bin/shx
~~~
~~~ js
// package.json
"scripts": {
  "test": "echo something" // fails! :/
}
~~~
~~~ sh
> npm test
Error: Missing ShellJS command name

npm ERR! Test failed.  See above for more details.
~~~

[shx]: https://github.com/shelljs/shx/issues/117

mmkal:
I suspect that is because shx is not a real shell, just a wrapper for a small number of commands. So invoking it like `shx -c "echo something"` (which is effectively what npm does) won't work. It'd probably be easier to fix in shx than npm - they'd just need to support the -c flag. But there'd still be a lot of missing shell features. Maybe a better solution would be another npm module that ships a small node executable script acting as a shell wrapper. The wrapper could use powershell on windows and bash on *nix.

danielbayley:
@mmkal Thanks for the tip-off. I fixed it in https://github.com/shelljs/shx/pull/118.

@iarna Thanks for the clarification. I have it working with `.npmrc`, but for many projects that otherwise don’t need that file, the ability to set the shell somewhere in `package.json` would definitely be preferable.

Issue #16083, opened 2017-03-17
outdated: exitcode still 1 with --silent flag


fearphage:
`npm outdated` changed to exiting with exit code 1 when things are outdated. According to #14013, the `--silent` flag should get the classic functionality of always exiting with exit code 0 (whether up to date or not). I am not able to confirm that however. I don't see any tests for that case either.

Node version: 7.7.1
NPM vesion: 4.1.2

```
➜  unity git:(starting-point) ✗ node --version
v7.7.1
➜  unity git:(starting-point) ✗ npm --version
4.1.2
➜  unity git:(starting-point) ✗ npm outdated > /dev/null; echo $?
1
➜  unity git:(starting-point) ✗ npm outdated --silent > /dev/null; echo $?
1
➜  unity git:(starting-point) ✗ npm --silent outdated > /dev/null; echo $?       
1
➜  unity git:(starting-point) ✗ npm --silent outdated --silent > /dev/null; echo $?
1
```

rtaft:
Also seeing this behavior in 4.1.1.  An exit code of 1 indicates an error.  Given that 'outdated' is an expected option and returns correctly, I don't understand why 0 is not the default return value.

vsubbotskyy:
npm@4.4.4
node@v7.8.0
same issue

wwalser:
The implementation of ```npm outdated```, a function called ```outdated```, has a parameter called ```silent```. This probably lead the implementor to think that this functionality was already taken care of and tested. Unfortunately the ```silent``` parameter isn't a CLI parameter, it's only used by the ```npm update```.

See it's original implementation here: https://github.com/npm/npm/commit/82894b62b6389aa10029bf5592ae8e41124a55ce

One piece of trouble for landing a PR for this is that the function parameter and CLI parameter, though both called ```silent``` have differing functionality. CLI ```--silent``` is meant to allow for normal execution of ```npm outdated```, including all of this other parameters. The function parameter on the other hand mutes all output.

I believe I'll try to land a PR for this but the above causes me to question the original proposed implementation, especially the name of the flag used.

rtaft:
That's one way to address an issue, ignore it and have the bot close it automatically.

PR #16703, opened 2017-05-24
Fix for #16083, add support for --silent flag with npm outdated.

wwalser:
Fix for #16083 which is a bug introduced in #14013.

This PR is a strict fix for what was said to be the intended behavior introduced in #14013. ```npm outdated --silent``` was suppose to maintain the old behavior of returning a success status code even if wanted or latest were out of date.

As an aside, this seems to me like the wrong implementation. It introduces a second meaning of the word "silent" and it may even clash with existing behavior, though I couldn't find an invocation where this was the case.

It seems to me that reusing silent is problematic. The real desire, I believe, is to be able to use ```npm outdated``` with some nuance. For example see facebook/fbjs/issues/213, where an out of date "wanted" may be a reasonable case for an error but "latest" may not be.

**Edit**: As @KenanY correctly points out, the collision of a command config option and loglevel shorthand causes problems. I suggest either dumping this feature and letting downstream consumers sort out the exit code for themselves _or_ creating a new option to ```npm outdated``` with name other than ```--silent```. I would prefer the later and I'm willing to change the PR to make that happen. Open to suggestions for what the config option should be called, ```---quiet```?


watilde:
+1 on this. This is what I expected, but I missed it at the last PR. We need to update `test/tap/outdated-json.js` to fix the ci error.
> https://github.com/npm/npm/blob/latest/test/tap/outdated-json.js#L83

KenanY:
As it stands, this breaks the `--silent => --loglevel silent` alias. Perhaps https://github.com/npm/npm/pull/16703/files#diff-3d20499d58f14c6f1edfe93d8ba8a8a2R115 should just check if `loglevel === 'silent'`?

wwalser:
@KenanY I must be doing something wrong, I'm not able to make the ```--silent``` flag work with outdated on my system installed npm. 

screen shot 2017-05-24 at 10 38 04 am

wwalser:
Ah, thanks for those details. I am probably missing something still as this doesn't actually break that feature. It just changes the return code to match what was described in #14013.

I mentioned my hesitation with the overlapping flag in my OP. Is that what you're describing?

Here I show a run with an ```.npmrc``` containing ```loglevel=http```.

example-with-silent

KenanY:
Okay, so oddly enough, this doesn't break the feature I mentioned **for `npm outdated` specifically**. I'm not sure yet why there's a difference, but my example above _does_ demonstrate the issue I'm talking about (at least on my machine 😄).

wwalser:
@KenanY You're right.

I haven't read enough of the code to tell why exactly. It makes sense that having something that is both a parameter and a shorthand causes a collision but why it functions properly for ```outdated``` is a mystery to me.

iarna:
@wwalser This patch is breaking things because `--silent` is already an option, see the `shorthands` in `lib/config/defaults.js`.

wwalser:
@iarna Yes, that's discussed in the comments here, in the bug, and in the original change that introduced the bug. My proposed solution is to rename the flag or drop this fix and accept the knock on regressions. I'm happy with any decision. The bug this caused for me has been fixed down stream by ignoring the exit code.

zkat:

PR #16749, opened 2017-05-27
Remove slow assertion from module-name util

mikesherov:
Was playing around and profiling the new npm5 release and it's still a drop sluggish in cases where it could be much faster (like when it already has package-lock.json). I can provide benchmarks if necessary to get this landed, just point me in the right direction. In my tests, it speeds up install by approx 25%!

The assertion appears to just be an abundance of caution, as I checked all call sites and it seems like an object is guaranteed in each of those cases. Let me know your thoughts, thanks for considering the patch and thanks for everything you all do!

P.S. Excuse the `patch-1` branch. I tested this on my computer but am recreating the pr on my phone!

### DIFF
--- a/lib/utils/module-name.js
+++ b/lib/utils/module-name.js
@@ -1,6 +1,5 @@
 'use strict'
 var path = require('path')
-var validate = require('aproba')
 
 module.exports = moduleName
 module.exports.test = {}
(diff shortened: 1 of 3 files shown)


iarna:
Similarly changing that to `validate('O', [tree])` should also be much faster. But yeah, the function takes only one argument, we probably don't need to validate it. =p

zkat (approved):
lol ok this is ridiculously fast. Tested it locally. Wow.

Thank you!

styfle:
@zkat @iarna  Has the npm team considered switching from runtime-type checking to compile-time type checking with TypeScript or Flow?

PR #16750, opened 2017-05-27
Talk less, complete more

aredridel:
Keeps extraneous output from breaking completion.


zkat:
"Just plop this in" they said

"it'll be a cute, quick easter egg" they said

I can't believe how much trouble this has caused ahahahah

zkat (approved):
So how'd you do it? How'd you patch this thing so fast? 😍

aredridel:
It was my shell's dying wish before it passed.

Issue #16746, opened 2017-05-26
npm publish --tag regression?


jsg2021:
#### I'm opening this issue because:

  - [ ] npm is crashing.
  - [ ] npm is producing an incorrect install.
  - [x] npm is doing something I don't understand.
  - [ ] Other (_see below for feature requests_):

#### What's going wrong?
`npm publish --tag alpha` did not use the alpha tag and instead used `latest`.
npm 3.x, 4.x worked fine.

#### How can the CLI team reproduce the problem?

run `npm publish --tag alpha` 

### supporting information:

 - `npm -v` prints: 5.0.0
 - `node -v` prints: 6.10.3
 - `npm config get registry` prints: private-sinopia-url
 - Windows, OS X/macOS, or Linux?: macOS
 - Network issues:
   - Geographic location where npm was run: Norman, OK - US
   - [x] I use a proxy to connect to the npm registry. (Sinopia)
   - [ ] I use a proxy to connect to the web.
   - [ ] I use a proxy when downloading Git repos.
   - [ ] I access the npm registry via a VPN
   - [ ] I don't use a proxy, but have limited or unreliable internet access.
 - Container:
   - [ ] I develop using Vagrant on Windows.
   - [ ] I develop using Vagrant on OS X or Linux.
   - [ ] I develop / deploy using Docker.
   - [ ] I deploy to a PaaS (Triton, Heroku).

zkat:
Thanks for the report! I'll take a look at why this isn't being picked up. 👍

zkat:
Hah, found it. Thanks!

zkat:
No, I take it back. I can't actually reproduce this. Any time I publish with `--tag` on `npm@5.0.0`, I get a new version under said tag 🙃 

Can you do `npm pub <pkg> --timing --tag foo` and post the debug log that prints out? (it's not actually erroring, `--timing` is just mostly a debug tool for the npm CLI devs right now, pardon the confusion).

jsg2021:
Here ya go
```
0 info it worked if it ends with ok
1 verbose cli [ '/Users/jgrimes/.nvm/versions/node/v6.10.3/bin/node',
1 verbose cli   '/Users/jgrimes/.nvm/versions/node/v6.10.3/bin/npm',
1 verbose cli   'publish',
1 verbose cli   '--timing',
1 verbose cli   '--tag',
1 verbose cli   'alphatest' ]
2 info using npm@5.0.0
3 info using node@v6.10.3
[...]

PR #16762, opened 2017-05-28
publish: adapt config for publish RegClient

zkat:
Fixes: #16746

There were apparently _two_ places that use `RegClient` directly, and I missed fixing the passed-in config stuff for the one in `get-publish-config`.

This is the last one, so this particular thing shouldn't cause any more issues.


popomore:
Publish a module with a wrong tag on 5.0.0, thx for this PR.

PR #16768, opened 2017-05-28
npm-config list -l option documented

happylynx:
`-l` option of `npm config list` documented.

Previously it was only mentioned in output of `npm config list`:

    ; "npm config ls -l" to show all defaults.

### DIFF
--- a/doc/cli/npm-config.md
+++ b/doc/cli/npm-config.md
@@ -6,7 +6,7 @@ npm-config(1) -- Manage the npm configuration files
     npm config set <key> <value> [-g|--global]
     npm config get <key>
     npm config delete <key>
-    npm config list
+    npm config list [-l]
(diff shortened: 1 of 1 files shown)


doc/cli/npm-config.md
```diff
@@ -48,7 +48,7 @@ Echo the config value to stdout.
 
     npm config list
 
-Show all the config settings.
+Show all the config settings. `-l` show all settings including non-customized config pairs.
```
zkat: I would probably reword this a bit to something like: `Use -l to also show defaults.`

zkat (changes requested):
Just a tiny nit, but this looks good. Thanks for taking care of it!

zkat (approved):
LGTM! Thanks 💃

Issue #16780, opened 2017-05-29
Child process is exited prematurely when running a view command


ncuillery:
#### I'm opening this issue because:

  - [ ] npm is crashing.
  - [ ] npm is producing an incorrect install.
  - [ ] npm is doing something I don't understand.
  - [x] Other (_see below for feature requests_):

#### What's going wrong?

When using NPM 5: if I run the `npm view` command in a child process (either `exec` or `spawn`) the child process exits with code 0 right after the reception of the first chunk of data.

#### How can the CLI team reproduce the problem?

Considering this simple Node program:
```javascript
const exec = require('child_process').exec;

const child = exec('npm view npm@latest');

child.stdout.on('data', data => {
  console.log('DATA', data.length);
});

child.stderr.on('data', data => {
  console.log('ERROR', data.length);
});

child.on('exit', code => {
  console.log('EXIT', code);
});
```

Output when NPM 4.6.1 is installed:
```
$ node debug-exec.js      
DATA 8192
DATA 8192
DATA 8189
DATA 8191
DATA 3817
EXIT 0
```

Output when NPM 5.0.0 is installed:
```
$ node debug-exec.js      
DATA 8192
EXIT 0
```

There is no problem with these commands (the whole output is received whatever the NPM version):
 - `ls -l node_modules`
 - `curl -s https://github.com/ncuillery`

### supporting information:

 - `npm -v` prints: either `4.6.1` or `5.0.0`
 - `node -v` prints: `v7.1.0`
 - `npm config get registry` prints: `https://registry.npmjs.org/`
 - Windows, OS X/macOS, or Linux?: macOS Sierra 10.12.4
 - Network issues:
   - Geographic location where npm was run:
   - [ ] I use a proxy to connect to the npm registry.
   - [ ] I use a proxy to connect to the web.
   - [ ] I use a proxy when downloading Git repos.
   - [ ] I access the npm registry via a VPN
   - [ ] I don't use a proxy, but have limited or unreliable internet access.
 - Container:
   - [ ] I develop using Vagrant on Windows.
   - [ ] I develop using Vagrant on OS X or Linux.
   - [ ] I develop / deploy using Docker.
   - [ ] I deploy to a PaaS (Triton, Heroku).

ncuillery:
Additional notes:

#### Same bug using the exec's callback argument:

```javascript
const exec = require('child_process').exec;

exec('npm info npm@latest', (err, stdout, stderr) => {
[...]

PR #16791, opened 2017-05-30
view: wait until write completes to call cb

zkat:
Fixes: https://github.com/npm/npm/issues/16780


rosen-vladimirov (approved):

zkat:
node is using this patch now, so I'm just gonna merge it, tbh

PR #16799, opened 2017-05-30
Document package-lock.json never being allowed in tarballs

sonicdoe:
See [lib/utils/tar.js#L106-L107](https://github.com/npm/npm/blob/v5.0.0/lib/utils/tar.js#L106-L107).


zkat (approved):
LGTM 👍

PR #16804, opened 2017-05-30
Use tar v3 for package creation. Remove lib/utils/tar.js

isaacs:
This moves all of the logic for package file inclusion to the `npm-packlist` module, which in turn uses `npm-bundled` to determine the list of top-level packages that are bundled.

It'd be pretty easy to have packlist take a list of the bundled packages, rather than get the list itself. (That's how the internal classes work anyway.)

Major changes:

1. `.npmignore` and `.gitignore` files are ignored by default.
2. The previous list of "ignored unconditionally" path names are now merely "ignored by default", and may be included by being in a package.json `files` list, or a `!include` rule in an ignore file.
3. If the `bundledDependencies` field is an object instead of an array, it's used in the pack process without printing a complaint about it.
4. `.gitignore` files are not renamed to `.npmignore`.  They're both ignored by default, so there isn't much point.

Other than that, the logic for how ignore files should be identical.  It's lifted directly from `fstream-npm` and `fstream-ignore`, and the current set of tests are only changed to account for the semantic changes above, and to pass on my dev environment.  (For some reason, my environment variables were making some of the git tests to fail.)


isaacs:
The next step after this is to make node-gyp use tar v3, so that fstream is gone entirely.

zkat:
RIP in peace fstream end of an era

martinheidegger:
Docker needs this! Docker & NPM 5 don't play well together 😣

zkat:
@martinheidegger we're holding off on merging this patch until I'm able to finish integration of tar3 into pacote. That is lower priority than fixing a number of other critical bugs but it's coming soon and most of the work is already done.

martinheidegger:
@zkat The package-lock has been particularly interesting to me for its use with docker. Though for some reason, it turns out that NPM extracts packages with the wrong permissions in some docker contexts that are hard, if not impossible, to work around. It is my hope (as I read somewhere about the changes) that this PR fixes it 🙏

zkat:
@martinheidegger are you still seeing that with npm@5.0.4? That should've been fixed by that version.

martinheidegger:
Busy at another project right now, I cant recall if it was 5.0.3 or 5.0.4. Will test it later today (5 hours?!)

martinheidegger:
Just had an hour and ran the install, used both 5.0.4 and npmc@5.0.4-canary.5 still getting an einval error.

Here are the issues for reference:
https://github.com/npm/npm/issues/16896
https://discuss.circleci.com/t/getting-a-weird-einval-error-with-npm-on-circleci-2-0/13465/4

iarna:
This has been merged into release-next!

Issue #16528, opened 2017-05-04
npm5 Authentication Failure (E401) for 3rd Party Private Packages Registry


simonua:
#### I'm opening this issue because:

  - [ ] npm is crashing.
  - [ ] npm is producing an incorrect install.
  - [x] npm is doing something I don't understand.
  - [ ] Other (_see below for feature requests_):

#### What's going wrong?
We have successfully used Microsoft's Visual Studio Team Services Package Management for npm with all versions npm 4. When trying out 5.0.0-beta.36, we immediately get E401 error codes.

[npm debug log](https://gist.github.com/simonua/124625a26f5f48e9c2ad8de61d692f3a)

#### How can the CLI team reproduce the problem?
Unfortunately, due to the setup with a 3rd party registry, I cannot provide details for reproduction of the error.

### supporting information:

 - `npm -v` prints: 5.0.0-beta.36
 - `node -v` prints: v7.1.0
 - `npm config get registry` prints: https://connectionseducation.pkgs.visualstudio.com/_packaging/ConnectionsEducationNpm/npm/registry/
 - Windows, OS X/macOS, or Linux?: Windows 7
 - Network issues:
   - Geographic location where npm was run:
   - [ ] I use a proxy to connect to the npm registry.
   - [ ] I use a proxy to connect to the web.
   - [ ] I use a proxy when downloading Git repos.
   - [ ] I access the npm registry via a VPN
   - [ ] I don't use a proxy, but have limited or unreliable internet access.
 - Container:
   - [ ] I develop using Vagrant on Windows.
   - [ ] I develop using Vagrant on OS X or Linux.
   - [ ] I develop / deploy using Docker.
   - [ ] I deploy to a PaaS (Triton, Heroku).

Supplemental information for our .npmrc files:
```
registry=https://connectionseducation.pkgs.visualstudio.com/_packaging/ConnectionsEducationNpm/npm/registry/
always-auth=true
//connectionseducation.pkgs.visualstudio.com/_packaging/ConnectionsEducationNpm/npm/registry/:_authToken=<removed>
```

simonua:
@zkat @zarenner, ping

zarenner:
@simonua Thanks for bringing this to our attention. Did a quick investigation, looks like a couple issues in https://github.com/zkat/pacote

registry-key.js has "path: parsed.path" when I believe it should be "pathname: parsed.path", thus the path segment (which VSTS has) is getting dropped.

[...]

PR #16879, opened 2017-06-01
Added support for always-auth and _auth. Both for root and scopes #16528

jozemlakar:
I had added the loading of auth config when always-auth and _auth are set.
I had added the sending of authorization header in the mentioned case. 
I have tested this against artifactory, when set globally. Someone else might want to test it against other registries or when scopes are used.


lib/config/pacote.js
```diff
@@ -46,9 +46,17 @@ function pacoteOpts (moreOpts) {
...
+    opts.auth.alwaysAuth = true;
+    if (!!npm.config.get('_auth')) {
+      opts.auth._auth=npm.config.get('_auth');
+    }
+  }
+
   npm.config.keys.forEach(function (k) {
     const authMatch = k[0] === '/' && k.match(
-      /(.*):(_authToken|username|_password|password|email|always-auth)$/
+      /(.*):(_authToken|username|_password|password|email|always-auth|_auth)$/
```
snirad: this regex assume we use 
//registry.npmjs... type of syantax.

could you please change it like so ?

/((.*):)?(_authToken|username|_password|password|email|always-auth|_auth)$/

snirad (commented):
- sorry for the confusion , now i see that it works even without adding this regex, but maybe it is still needed, im not sure.

node_modules/pacote/lib/fetchers/registry/fetch.js
```diff
@@ -94,6 +95,8 @@ function getHeaders (uri, registry, opts) {
       `${auth.username}:${auth.password}`, 'utf8'
     ).toString('base64')
     headers.authorization = `Basic ${encoded}`
+  } else if (shouldAuth && auth._auth) {
+    headers.authorization = `Basic ${auth._auth}`
   }
```
zkat: These changes should be files as a separate PR against `pacote`.

lib/config/pacote.js
```diff
@@ -46,9 +46,17 @@ function pacoteOpts (moreOpts) {
     Object.assign(opts, ownerStats)
   }
 
+  if (!!npm.config.get('always-auth')) {
```
zkat: I'd rather not coerce to boolean here. Let `if` do its usual work?

lib/config/pacote.js
```diff
@@ -46,9 +46,17 @@ function pacoteOpts (moreOpts) {
...
+    if (!opts.auth) { opts.auth = {} }
+    opts.auth.alwaysAuth = true;
+    if (!!npm.config.get('_auth')) {
+      opts.auth._auth=npm.config.get('_auth');
+    }
+  }
+
```
zkat: Please make sure the code passes `standard` linting.

lib/config/pacote.js
```diff
@@ -46,9 +46,17 @@ function pacoteOpts (moreOpts) {
     Object.assign(opts, ownerStats)
   }
 
+  if (!!npm.config.get('always-auth')) {
+    if (!opts.auth) { opts.auth = {} }
+    opts.auth.alwaysAuth = true;
+    if (!!npm.config.get('_auth')) {
```
zkat: This can't just be for `_auth`. All of the other auth parameters listed below should be made available as toplevel settings.

zkat (changes requested):
So far so good! A few things to improve first 👍

lib/config/pacote.js
snirad: how about adding this line just below, and making authMatch ; let instead of const.
that should hold for all cases, and you can remove that if condition you added just for _auth

`
  npm.config.keys.forEach(function (k) {
    let authMatch = (k[0] === '/' && k.match(/(.*):(_authToken|username|_password|password|email|always-auth|_auth)$/))
   if(!authMatch && /(_authToken|username|_password|password|email|always-auth|_auth)$/.test(k)){
        authMatch =[,registryKey(opts.registry), k]
}
    if (authMatch) {
      const nerfDart = authMatch[1]
      const key = authMatch[2]
      const val = npm.config.get(k)
.........................

`

- also we will need to import registryKey functoin as wel

`function registryKey (registry) {
  const parsed = url.parse(registry)
  const formatted = url.format({
    host: parsed.host,
    pathname: parsed.pathname,
    slashes: parsed.slashes
  })
  return url.resolve(formatted, '.')
}`

what do you think,
@zkat , @jozemlakar

jozemlakar: sounds good. I am currently working on tests on a change in https://github.com/zkat/pacote and will then do config. Whatever the final solution, it should be the same for scoped and global settings, so my first one is no good.

node_modules/pacote/lib/fetchers/registry/fetch.js
jozemlakar: Added https://github.com/zkat/pacote/pull/96
This has to be verified and then tagged so we can upgrade dependency in npm. Then I will try and fix the config.

lib/config/pacote.js
jozemlakar: So, looked at this issue and here is where I stand:
- PR created for pacote. I think that should clear. Once that lands we can update the dependency in npm
- There is only one general test for config/pacote in npm, so the first thing I need to do is create a test for config parsing
- Fix config parsing so we get all options (including _auth) both on global and scoped.

jozemlakar:
I think this is now complete. It adds support for all auth options, uses options in the same manner when scoped and global.

zkat (approved):

zkat:
Awesome work! Yay tests! I'll merge this once the test suite passes. Thanks a bunch 🎉

zkat:
@jozemlakar the tests seem to be failing right now :< https://travis-ci.org/npm/npm/builds/240593681?utm_source=github_status&utm_medium=notification

jozemlakar:
@zkat seems that registry configured in test is no longer used but rather goes to default public registry. Will look into it.

jozemlakar:
@zkat Build was broken with https://github.com/npm/npm/pull/17076
when merged without running tests first https://github.com/npm/npm/pull/16879/commits/c8528297ada8a542385efbefa6ca0d776bbb6342

Without that change the config is
```
{ type: 'remote',
  registry: undefined,
  where: undefined,
  raw: '@scoped/underscore@http://localhost:1337/scoped-underscore/-/scoped-underscore-1.3.1.tgz',
  name: '@scoped/underscore',
  escapedName: '@scoped%2funderscore',
  scope: '@scoped',
  rawSpec: 'http://localhost:1337/scoped-underscore/-/scoped-underscore-1.3.1.tgz',
  saveSpec: 'http://localhost:1337/scoped-underscore/-/scoped-underscore-1.3.1.tgz',
  fetchSpec: 'http://localhost:1337/scoped-underscore/-/scoped-underscore-1.3.1.tgz',
  gitRange: undefined,
  gitCommittish: undefined,
  hosted: undefined }
```

after change the config is
```
{ type: 'version',
  registry: true,
  where: undefined,
  raw: '@scoped/underscore@1.3.1',
  name: '@scoped/underscore',
  escapedName: '@scoped%2funderscore',
  scope: '@scoped',
  rawSpec: '1.3.1',
  saveSpec: null,
  fetchSpec: '1.3.1',
  gitRange: undefined,
  gitCommittish: undefined,
  hosted: undefined }
```

As a result, npm is sending the request to public npm registry instead of the mock.

I have no idea why this was changed or how this passed existing tests. Please have a look at it.

jozemlakar:
@zkat if you need me to check something or try something out, let me know. My knowledge of shrinkwrap is next to zero, so I really can't handle the changes you made on my own.

jozemlakar:
@zkat in case I was not understood. Commit https://github.com/npm/npm/commit/c8528297ada8a542385efbefa6ca0d776bbb6342 breaks existing tests i.e. shrinkwrap-scoped-auth.js
When you fix that I can fix other tests I created based on that one.

jozemlakar:
@snirad you are absolutely correct! I remove the lines, thank you.

jozemlakar:
@zkat this is now ready. Once i pulled all the changes from release-next and synced my tests with the ones you changed tests succeed.

snirad:
@zkat  hi kat :) 

any chance we get to see it on 5.0.4 ?

zkat:
This has been merged into `release-next` and is available in `npmc@5.0.4-canary.4`!

PR #16899, opened 2017-06-02
install: print spec for each requested package

### LINKED ISSUE #16897: npm5: installing globally no longer prints versions
npm@4:
```
$ npm i -g depcheck
/Users/ryan/.nvs/node/7.10.0/x64/bin/depcheck -> /Users/ryan/.nvs/node/7.10.0/x64/lib/node_modules/depcheck/bin/depcheck
/Users/ryan/.nvs/node/7.10.0/x64/lib
└─┬ depcheck@0.6.7
  ├─┬ babel-traverse@6.24.1
  │ ├─┬ babel-code-frame@6.22.0
... and the remainder of a very informative tree
```

npm@5:
```
$ npm i -g depcheck
/Users/ryan/.nvs/node/8.0.0/x64/bin/depcheck -> /Users/ryan/.nvs/node/8.0.0/x64/lib/node_modules/depcheck/bin/depcheck
updated 1 package in 3.567s
```

## Ask

I agree with the decision to _not_ dump the entire tree to the screen, but I think the trimming might have been a bit much. The first line of the tree was probably 99% of the usefulness of the tree since it shows what version of the requested package was actually installed.

Can we have npm at least list the versions of the packages that were given as arguments when installing?

npm@PLEASE:
```
$ npm i -g depcheck
depcheck@0.6.7
/Users/ryan/.nvs/node/8.0.0/x64/bin/depcheck -> /Users/ryan/.nvs/node/8.0.0/x64/lib/node_modules/depcheck/bin/depcheck
updated 1 package in 3.567s
```

rmg:
For every package that is given as an argument to install, print the
name and version that was actually installed.

Resolves #16897

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -741,6 +741,9 @@ Installer.prototype.printInstalledForHuman = function (diffs, cb) {
     }
   })
   var report = ''
+  if (added && this.args.length) {
+    report += this.args.map((p) => p.name + '@' + p.version).join('\n') + '\n'
+  }
   var actions = []
   if (added) actions.push('added ' + packages(added))
   if (removed) actions.push('removed ' + packages(removed))


zkat:
It would be really nice to have something like this for `uninstall` as well! But that can be a separate PR. Thanks for this! This is pretty much exactly what @iarna mentioned this should do when I brought up the issue. This is also UI-y enough that I'm ok with it not having tests. Cheers!

zkat (approved):
LGTM. This is pretty much what I would've asked for 🐑

Issue #16868, opened 2017-06-01
npm@5.0.3: When proxied, `strict-ssl=false` or custom CAfile isn’t working


colinrotherham:
#### I'm opening this issue because:

  - [ ] npm is crashing.
  - [ ] npm is producing an incorrect install.
  - [x] npm is doing something I don't understand.
  - [ ] Other (_see below for feature requests_):

#### What's going wrong?
Running `npm install` on **npm@4.6.1** works no problem
Running `npm install` on **npm@5.0.3** throws *UNABLE_TO_VERIFY_LEAF_SIGNATURE*

We're using an internal registry via `npm config set registry` which is behind a corporate proxy. The registry is protected using a self-signed root CA certificate.

We know our custom registry has an invalid SSL certificate.
Setting `strict-ssl=false` should workaround this.

The corporate root CA is trusted locally (Windows and Mac keychain etc) but for npm we add a custom certificate bundle via `npm config set cafile …` and `strict-ssl=false` as a workaround.

#### How can the CLI team reproduce the problem?
Try running `npm install` to a custom registry with an invalid SSL certificate.
To avoid any errors, ensure `strict-ssl=false` is set.

It works on npm@4.6.1 but throws an error on npm@5.0.3.

Included is my `npm-debug.log` (sensitive info replaced):
https://gist.github.com/colinrotherham/ce6d21ba35271e356f9255a5c12c3d94

Thanks

### supporting information:

 - `npm -v` prints: 5.0.3
 - `node -v` prints: v8.0.0
 - `npm config get registry` prints: Sorry, can't disclose this
 - Windows, OS X/macOS, or Linux?: macOS
 - Network issues:
   - Geographic location where npm was run:
   - [x] I use a proxy to connect to the npm registry.
   - [x] I use a proxy to connect to the web.
   - [x] I use a proxy when downloading Git repos.
   - [ ] I access the npm registry via a VPN
   - [ ] I don't use a proxy, but have limited or unreliable internet access.
 - Container:
   - [ ] I develop using Vagrant on Windows.
   - [ ] I develop using Vagrant on OS X or Linux.
   - [ ] I develop / deploy using Docker.
   - [ ] I deploy to a PaaS (Triton, Heroku).

colinrotherham:
As mentioned in the linked gist above, our `.npmrc` file contains:

```
strict-ssl=false
cafile=/path/to/certificates/cacert.pem
```

[...]

PR #16917, opened 2017-06-02
Add certificate authority config to pacoteOpts

colinrotherham:
Custom certificate config (e.g. via `npm config set cafile`) isn't making their way into `make-fetch-happen`, this appears to be an `npm@5` regression.

See the discussion here: https://github.com/npm/npm/issues/16868

### DIFF
--- a/lib/config/pacote.js
+++ b/lib/config/pacote.js
@@ -17,9 +17,12 @@ function pacoteOpts (moreOpts) {
   const ownerStats = calculateOwner()
   const opts = {
     cache: path.join(npm.config.get('cache'), '_cacache'),
+    ca: npm.config.get('ca'),
+    cert: npm.config.get('cert'),
     defaultTag: npm.config.get('tag'),
     dirPacker: pack.packGitDep,
     hashAlgorithm: 'sha1',
+    key: npm.config.get('key'),
     localAddress: npm.config.get('local-address'),
     log: log,
     maxAge: npm.config.get('cache-min'),


colinrotherham:
This also requires a patch in `pacote` which I've requested here: https://github.com/zkat/pacote/pull/95

//cc @zkat

colinrotherham:
@zkat As mentioned on https://github.com/zkat/pacote/pull/95 the options `ca`, `certfile` and `key` are now passed through.

colinrotherham:
@zkat I'm reading the Node documentation here:
https://docs.npmjs.com/misc/config#cert

>It is not the path to a certificate file (and there is no "certfile" option).

Maybe we should have used `cert` and not `certfile` as the option name mentioned here:
https://github.com/zkat/pacote/pull/95#issuecomment-305855764

wildlyinaccurate:
Hey, thanks so much for putting the effort into fixing this! We weren't able to get npm@5 working with our internal registry but I hadn't had the time to look into it properly yet.

I've followed the trail of PRs from this issue and there were two things I needed to change before everything worked:

0. https-proxy-agent needs the extra options like `key` and `cert`
0. pacote should use `cert` instead of `certfile` for the make-fetch-happen opts

With those changes that, npm@5 works beautifully with our registry :sparkles:

zkat:
Oops, that was totally my bad. I don't know why I got this confused with `certfile` or whatnot. It should indeed be `cert`. The rest of this looks good and it sounds like @wildlyinaccurate has confirmed the changes. 🎉 

I'll take this as soon as those changes are made 👍

zkat:
(I'll fix things pacote-side to use `cert` instead of `certfile` right quick)

colinrotherham:
@zkat Done.

I've matched `cert` in npm's **pacoteOpts** to `cert` in **pacote** (previously `certfile`) to align with https://github.com/zkat/pacote/commit/a45880d50beb0dcf89e0b2112f101a7d301b745e

Have we missed any other options?

zkat:
I think that's the lot of 'em! Thanks a ton!

colinrotherham:
Great, thanks @zkat. This helps us out a lot, glad to assist.

zkat:
@colinrotherham I've pushed this out in the canary as `npmc@5.0.2-canary.9`, if you wanna start using it right away.

wildlyinaccurate:
@zkat `npmc@5.0.2-canary.9` works for my use case (using the --registry, --key, --cafile, and --cert options). You and @colinrotherham are both AMAZING thank you!

Edit: Holler if you want a hand setting up some sort of regression suite to catch stuff like this?

zkat:
@wildlyinaccurate the proxy and http(s) agent stuff is pretty tricky to test, but it would be great to write more tests! It would also be nice to have more thorough test over at pacote and make-fetch-happen (specially the latter) to make sure all this stuff is working spiffy -- forgetting to pass options along is less worrisome than fundamental missing bits.

This particular part of the rewrite was something we figured was going to be some level of broken: we knew we needed to replace our network layer and move away from `request` (for perf and specific-need reasons), but that library took care of a _lot_ of stuff with _very_ mature code. And we just plain didn't have enough time/understanding/coverage of these bits to have confidence in it. We ended up jumping right in anyway and yolo or whatnot :)

Please lmk if you need anything or have any questions. I'm more than happy to try and point to places to look for things or fill in context.

PR #16918, opened 2017-06-02
Fix and unify prepublish deprecation messages

Hirse:
see #16685 

Use the same message for `prepare` and `pack` and end all lines with `.`:
```
As of npm@5, `prepublish` scripts are deprecated.
Use `prepare` for build steps and `prepublishOnly` for upload-only.
See the deprecation note in `npm help scripts` for more information.
```

npm


zkat:
I removed some spurious test crud that snuck its way in and merged this into release-next as 678d8864b

PR #16929, opened 2017-06-02
fix(pacote): Generate and pass npm-session header value to pacote

### LINKED ISSUE #16889: npm-session header missing in npm@5

  - [ ] npm is crashing.

#### What's going wrong?
npm previously sent the header npm-session with a unique ID per invocation of npm. This was useful for registry services to be able to correlate server requests across client sessions.

npm@5 appears to no longer send this header, presumably because of the change to pacote.

### supporting information:

 - `npm -v` prints: 5.0.1

zarenner:
Partially fixes #16889 along with https://github.com/zkat/pacote/pull/98

### DIFF
--- a/lib/config/pacote.js
+++ b/lib/config/pacote.js
@@ -2,12 +2,16 @@
 
 const Buffer = require('safe-buffer').Buffer
 
+const crypto = require("crypto")
 const npm = require('../npm')
 const log = require('npmlog')
 let pack
 const path = require('path')
 
 let effectiveOwner
+
+const npmSession = crypto.randomBytes(8).toString('hex')
+log.verbose('npm-session', npmSession)
 
 module.exports = pacoteOpts
 function pacoteOpts (moreOpts) {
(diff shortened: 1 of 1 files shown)


lib/config/pacote.js
```diff
@@ -2,13 +2,16 @@
...
 
+const crypto = require("crypto")
 const npm = require('../npm')
 const log = require('npmlog')
 let pack
 const path = require('path')
 
 let effectiveOwner
 
+const npmSession = crypto.randomBytes(8).toString('hex')
```
zarenner, line 13: @zkat 
Previously npm-registry-client verbosely logged the value in initialize. As written this loses that (value isn't logged).

Is there a similar place that makes sense to generate and verbosely log the value instead of here? If not, it's not a big deal.

zkat: @zarenner I think you can put `log.verbose(...)` right there for now, tbh.

zarenner: Done, thanks.

zkat (approved):
LGTM

PR #16937, opened 2017-06-03
quick fix for possibility that `req` is null

forivall:
fixes the bug seen in the following log

i have no idea what caused it

or if this is the correct fix

and i want to leave work instead of triaging this

https://gist.github.com/forivall/249c1966dd8ac335bca4a6177e41132d

i used `npm shrinkwrap --production`. that might be it.

### DIFF
--- a/lib/shrinkwrap.js
+++ b/lib/shrinkwrap.js
@@ -107,7 +107,7 @@ function shrinkwrapDeps (deps, top, tree, seen) {
       return
     }
     var pkginfo = deps[moduleName(child)] = {}
-    var req = child.package._requested || getRequested(child)
+    var req = child.package._requested || getRequested(child) || {}
     if (req.type === 'directory' || req.type === 'file') {
       pkginfo.version = 'file:' + path.relative(top.path, child.package._resolved || req.fetchSpec)
     } else if (!req.registry && !child.fromBundle) {


zkat:
I'm more concerned about why this would be null. I'll leave this PR open for now, but I'm pretty sure we have a bug floating around about this. I'd rather find the root cause.

zkat:
You know what? I'm just gonna take this. Root cause be damned. Thank you, @forivall 👍

BridgeAR:
@zkat doesn't this potentially open the doors for follow up errors and therefore does this really improve the situation?

zkat:
@BridgeAR it seems to have fixed things in practice, and tbh I'm ok putting a bandaid over it for now: we're going to be taking a closer look at how we deal with trees to clean it up. At the very least, this patch will make the error more obvious, wherever it is.

stefanpenner:
@zkat I believe root cause of this problem may be related to:

* #16994
* #17096
* #17610
* #17056

> At the very least, this patch will make the error more obvious, wherever it is.

The bandaid without root cause (or test) seems to be obscuring the problem.

I've left some additional info I've discovered in a recent spelunk: https://github.com/npm/npm/issues/16994#issuecomment-313729008

PR #16945, opened 2017-06-03
Added suppor for auth methods both root and scopes #16528 #16879

snirad:
added support for all auth methods for root and scopes
don't need to wait for changes in 
https://github.com/zkat/pacote/blob/latest/lib/fetchers/registry/fetch.js

### DIFF
--- a/lib/config/pacote.js
+++ b/lib/config/pacote.js
@@ -6,6 +6,7 @@ const npm = require('../npm')
 const log = require('npmlog')
 let pack
 const path = require('path')
+const url = require('url')
 
 let effectiveOwner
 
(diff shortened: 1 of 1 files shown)


jozemlakar:
Added PR https://github.com/npm/npm/pull/16879 which also adds test for the cases. I whink that is a better PR to merge. But if not, at least use the tests from that PR.

snirad:
@jozemlakar Agree that this need more testing, but I don't like the Idea of having 2 different request, one for global and one for scoped, I want both to have the same body. this simplify pacote, so it doesn't need to 'know' how to deal with each request, it also simplify testing & debugging. but this is just my opinion.

jozemlakar:
@snirad agreed! You will notice that in https://github.com/npm/npm/pull/16879/commits/bfed855f4645671785cd3c433d01ed572ec5b9b4 i submitted yesterday there is only one implementation. I would rather not say mine is better than yours. As I said, use tests from my implementation and whatever of both implementations and get npm fixed!

zkat:
This was superseded by #16879. Thanks for the contrib, though! 🎉

PR #16966, opened 2017-06-04
Fix wrong symlink direction in link documentation

izri16:
In npm-link documentation it is written that after running
```npm-link package-name``` symlink is created from global
folder to node_modules of the current project.

However the symlink is created from node_modules of the current
folder to the global folder, otherwise there would be no
way to achieve linked module. So I swapped the direction
in the documentation.

### DIFF
--- a/doc/cli/npm-link.md
+++ b/doc/cli/npm-link.md
@@ -18,8 +18,8 @@ link` command was executed. (see `npm-config(7)` for the value of `prefix`). It
 will also link any bins in the package to `{prefix}/bin/{name}`.
 
 Next, in some other location, `npm link package-name` will create a
-symbolic link from globally-installed `package-name` to `node_modules/`
-of the current folder.
+symbolic link from `node_modules/` of the current folder
+to globally-installed `package-name`.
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by iarna — CHANGES_REQUESTED** [review]
I'm finding both versions of these docs confusing. (Actually I find the original less confusing, but I'm having trouble articulating _why_.)

Maybe it would be best to rewrite the entire paragraph to be clearer. I was looking at how `man ln` describes it…
```
ln TARGET LINK_NAME 

create a link to TARGET with the name LINK_NAME
```
We create a link to the globally-installed `package-name` with the name `node_modules/package-name`.

I guess it's just that lonely `node_modules/` that's mixing it up for me?

izri16:
Could you somehow describe what is confusing on PR version? I find it quite understandable.

iarna:
@rizip1 I guess because if I'm creating that link by hand the order is: `ln -s /global/dir/pkgname node_modules/pkgname` and your documentation puts that in the opposite order.

zkat:

PR #17001, opened 2017-06-05
adds an alias for `rum` -> `run`

gesa:
Because I got one of those fancy ergonomic split keyboards and i cannot be the only one who keeps typing `npm rum`. especially since the last letter in npm makes for a nice pattern with rum.

.

.

.

[Edited by @zkat: unacceptable content]
<details>
```javascript
'npm'.toUpperCase() 
```
</details>


zkat (approved):
LGTM

Issue #16925, opened 2017-06-02
npm@5: Incorrect unmet peer dependency error


jhermsmeier:
#### I'm opening this issue because:

  - [ ] npm is crashing.
  - [ ] npm is producing an incorrect install.
  - [ ] npm is doing something I don't understand.
  - [x] Other: It is displaying an incorrect error

#### What's going wrong?

Dependencies are correctly installed, but a warning is emitted during installation, and an error when listing modules even though all requirements are properly satisfied. This behavior can be observed with both `npm@5.0.0` and `npm@5.0.1`, and regardless of the depth of the dependency in the tree.

#### How can the CLI team reproduce the problem?

1. Install `w3cjs` with `npm install w3cjs`
2. A warning will be emitted:
  ```
  npm WARN superagent-proxy@0.3.2 requires a peer of superagent@>= 0.15.4 && < 2 but none was installed.
  ```
3. Check whether `w3cjs` actually declares `superagent` as a dependency:
  ```
  $ npm info w3cjs dependencies
  
  { commander: '~2.6.0',
    superagent: '~0.21.0',
    'superagent-proxy': '^0.3.1' }
  ```
4. List dependencies with `npm ls`
5. An unmet peer dependency error will be displayed, even though the module in question *has been installed* by the previous step and meets `superagent-proxy`'s peer dependency requirement of `"superagent": ">= 0.15.4 && < 2"`:
```
  $ npm ls --depth 2
npm-peer-dep-test@0.0.0 /Users/Jonas/Code/npm-peer-dep-test
└─┬ w3cjs@0.3.0
  ├── commander@2.6.0
  ├─┬ UNMET PEER DEPENDENCY superagent@0.21.0
  │ ├── component-emitter@1.1.2
  │ ├── cookiejar@2.0.1
  │ ├── debug@2.6.8
  │ ├── extend@1.2.1
  │ ├── form-data@0.1.3
  │ ├── formidable@1.0.14
  │ ├── methods@1.0.1
  │ ├── mime@1.2.11
  │ ├── qs@1.2.0
  │ ├── readable-stream@1.0.27-1
  │ └── reduce-component@1.0.1
  └─┬ superagent-proxy@0.3.2
    ├── debug@2.2.0
    └── proxy-agent@1.1.1

npm ERR! peer dep missing: superagent@>= 0.15.4 && < 2, required by superagent-proxy@0.3.2
```

### supporting information:

 - `npm -v` prints: 5.0.1
[...]

PR #17007, opened 2017-06-06
deps: use relaxed semver range matching for compatibility

zkat:
It seems like, somehow, we were using relaxed semver range matching before. This enables that again, which fixes at least two current issues, possibly others, having to do with people putting the most amazing garbage in their `package.json`.

Fixes: #16925
Fixes: #16931

### DIFF
--- a/lib/install/deps.js
+++ b/lib/install/deps.js
@@ -74,7 +74,7 @@ function doesChildVersionMatch (child, requested, requestor) {
     return false
   }
   try {
-    return semver.satisfies(child.package.version, requested.fetchSpec)
+    return semver.satisfies(child.package.version, requested.fetchSpec, true)
   } catch (e) {
     return false
   }


lib/install/deps.js
```diff
@@ -74,7 +74,7 @@ function doesChildVersionMatch (child, requested, requestor) {
     return false
   }
   try {
-    return semver.satisfies(child.package.version, requested.fetchSpec)
+    return semver.satisfies(child.package.version, requested.fetchSpec, true)
```
ghost, line 77: The `loose` option does not seem to be [documented on node-semver](https://github.com/npm/node-semver), so this change would be relying on undocumented behavior.

Also, you mentioned this change allows people to put _garbage _ in their `package.json` files, which doesn't seem like a good investment in terms of what code should go into `npm`.

Lastly, this change does not have any unit tests to prevent a future regression.

BridgeAR: It is actually documented and you find multiple descriptions as e.g.:

```All methods and classes take a final loose boolean argument that, if true, will be more forgiving about not-quite-valid semver strings. The resulting output will always be 100% strict, of course.```

I personally think it would be a good thing to just go ahead and document this breaking change instead of loosening the strictness but that's just me. It could of course also be stricter in a future major, I think it would still be fine to just document it now as breaking changes are expected for 5.0 anyway.

ghost: > It is actually documented and you find multiple descriptions as e.g.:

Ugh. Thank you for pointing that documentation out. I jumped down to the function definition, `satisfies(version, range)`, and totally skipped over the introduction paragraph. It's definitely a documented feature.

zkat: @destroyerofbuilds This is [Postel's law](https://en.wikipedia.org/wiki/Robustness_principle) in action, and it's a pretty good principle. It is much better for npm to keep _working_ even when users get "creative". And, as @BridgeAR point out, the "loose" option is documented and has been there for a while -- and npm has been using it for a long time. This patch isn't a new thing. It's fixing a regression because for some reason we _stopped_ using loose here.

PR #17038, opened 2017-06-07
Clarify that `||` is an or

camjc:
The current language confused a few of my colleagues,
leading some to infer that || was marking a range instead of separating range-sets.
If there is a clearer way to clarify this I'm very happy to do that instead.

### DIFF
--- a/doc/misc/semver.md
+++ b/doc/misc/semver.md
@@ -81,7 +81,7 @@ Comparators can be joined by whitespace to form a `comparator set`,
 which is satisfied by the **intersection** of all of the comparators
 it includes.
 
-A range is composed of one or more comparator sets, joined by `||`.  A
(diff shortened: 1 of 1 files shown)


kenany:
Perhaps this should go to [node-semver](https://github.com/npm/node-semver) first?

camjc:
@KenanY sure, what's the relationship between the two repos?

kenany:
@camjc AFAICT, `doc/misc/semver.md` is mostly just a carbon copy of `node-semver`'s `README`.

iarna:
I very much appreciate the clarifying doc patches! 

@KenanY is correct, it's actually makes npm's test fail to update this w/o updating `node-semver` first. I'm going to close this and we can update the docs here when we update to a `node-semver that has patched docs.

PR #17060, opened 2017-06-07
fix(git): git repos with prepare scripts always install with dev flag

### LINKED ISSUE #17059: NODE_ENV=production does not install devDependencies of a dep with `prepare`

  - [ ] npm is crashing.

#### What's going wrong?
Adding a `prepare` script in npm5 states that it will install devDependencies of that package and run that `prepare` script. If you have NODE_ENV set to production then it does not actually install `devDependencies` and building of that package will fail. Probably spent about 8hrs trying to figure out why my dist folder didn't exist on Heroku for a package I depend on :)

1) Depend on a package that has devDependencies and builds based on those
2) Run the following:
```
export NODE_ENV=production
npm i
```

### supporting information:

 - `npm -v` prints: 5.0.3
 - `node -v` prints: 8.0.0
 - Windows, OS X/macOS, or Linux?: Linux or macOS
 - Container:
   - [x] I deploy to a PaaS (Triton, Heroku).

Feature arrived in: https://github.com/npm/npm/pull/16533

intellix:
Force devDependencies to always install with --dev flag in case NODE_ENV is production

Related: #16533
Closes #17059

### DIFF
--- a/lib/pack.js
+++ b/lib/pack.js
@@ -163,6 +163,8 @@ function packGitDep (manifest, dir) {
       const child = cp.spawn(process.env.NODE || process.execPath, [
         require.main.filename,
         'install',
+        '--dev',
+        '--prod',
         '--ignore-prepublish',
         '--no-progress',
         '--no-save'


lib/pack.js
```diff
@@ -163,6 +163,7 @@ function packGitDep (manifest, dir) {
       const child = cp.spawn(process.env.NODE || process.execPath, [
         require.main.filename,
         'install',
+        '--only=dev',
```
zkat: This will only install `devDependencies` and not `dependencies`. Did you want to use `--dev`?

zkat (changes requested):
Thanks for finding this!

lib/pack.js
intellix: I did originally have `--dev` but when I used it locally it told me that it's deprecated and that i should be using `--only=dev` so I assumed the name was misleading :)
Guess not and will revert back to --dev

intellix:
Addressed comments :) (not sure if you get a notification for that)

lib/pack.js
```diff
@@ -163,6 +163,7 @@ function packGitDep (manifest, dir) {
       const child = cp.spawn(process.env.NODE || process.execPath, [
         require.main.filename,
         'install',
+        '--dev',
```
iarna, line 166: I think there should be a `--prod` here as well to protect us against users who were doing an `--only=dev` install previously.

intellix: SGTM and solved :)

zkat (approved):

PR #17074, opened 2017-06-07
change config docs

KennethKinLum:
The `--flag` doesn't need to be at the *end* of the command. It just need to have "no value" for it to be `true`.


iarna (approved):
Good improvement!

(Again, we'll merge w/o the AUTHORS part since that's produced during publication.)

iarna:
Thanks for this, this was merged as c9b534a14.

PR #17076, opened 2017-06-07
shrinkwrap: lean on from field for better forward-compat

zkat:
This patch, I think, will finally make most folks' shrinkwrap upgrade experience work out: right now, if you have a legacy shrinkwrap, it treats the `resolved` field (with the package tarball) as the version. This means everyone ends up with `"version": "https://registry.npmjs.org....1.2.3.tgz"`.

This patch should fix that.


iarna (approved):
Looks great!

PR #17077, opened 2017-06-08
feat: add why command

cyberhck:
`npm why <package name>` now show why was a specific package installed

Please help me write tests and review :) (very new to npm code base, first PR)


cyberhck:
Never mind, this is not required (maybe we should create a alias of `why` to `ls`)?

PR #17104, opened 2017-06-09
error-handler: write an empty string and wait for callback to wait fo…

zkat:
…r output

I think this'll catch any remaining bits where we were exiting the npm process before stdout was flushed. I could do the same thing for stderr but I feel like that's... not as important?


iarna (approved):
Oh, this is an excellent strategy. I like!

iarna:
Merged as e2f815f87!

PR #17120, opened 2017-06-09
build: annihilate prefer-global warning

zkat:


legodude17:
May I ask the motivation behind this?

iarna (changes requested):
This'll also need a docs update I believe. Also tests update!

ljharb (approved):
Yay! Basically nothing should be installed globally.

zkat:
@iarna updated!

@legodude17 in the earlier days of npm, things were much more likely to not just be global, but _needed_ to be global, for some reason or another. In recent years, people have been moving towards `devDependency`-based workflows for most things related to globals. Since then, folks have been getting spurious warnings about requiring `-g` even for packages that work perfectly well installed on a per-project basis and used through `npx` or `npm run-script`. This patch is intended to continue pushing the community towards enabling workflows centered around local-only binaries.

I'm not considering a breaking change, 'cause it's just a warning, and it's been more of an annoyance than anything for a lot of people in recent times.

legodude17:
@zkat Thanks for the explanation! I agree.

iarna:
Merged as 57225d394!

PR #17142, opened 2017-06-12
Adding the ability to control running of git hooks when running `npm version`

faazshift:
At my company, we have a sizable set of tests that run on each commit. Unfortunately, that makes running `npm version` take a while, since the version commit triggers that commit hook. This PR adds the ability to control that behavior. Setting the new option, `git-tag-hooks`, to `false` will cause the `-n` option to be passed in the flags for the version commit, disabling the running of the `pre-commit` and `commit-msg` hooks. The default is to run the hooks, as this has been the case historically.


iarna (changes requested):
I was going to wonder if a more general way to pass arguments to git would appropriate, but this _is_ the only place we commit anything, so I guess I'm ok with it.

I would like a basic test for this... maybe pull the flag computation out into a function and write a unit test for that?

faazshift:
@iarna I've added tests, and in the process revamped the constructing of arguments for the version commit.

Among changes, I renamed the option to `commit-hooks`, since that seemed more appropriate to me.

lib/version.js
```diff
@@ -278,14 +278,23 @@ function checkGit (localData, cb) {
   })
 }
 
+module.exports.buildCommitArgs = buildCommitArgs
+function buildCommitArgs (args) {
+  args = args || [ 'commit' ]
+  npm.config.get('commit-hooks') || args.push('-n')
```
iarna, line 284: I would much prefer to see this as:

```
  if (!npm.config.get('commit-hooks')) args.push('-n')
```

iarna (approved):
In future, please rebase your changes on to our latest rather than merging our latest into your branch. The merges make landing things substantially more difficult.

Thanks for putting this together! I've landed this in release-next as 2dec601c6.

faazshift:
@iarna Makes sense. I'll keep that in mind for future contributions.

PR #17205, opened 2017-06-14
Npm is not consistent on arguments [was 'Why not'?]

gdassori:
If isntall is install... why udpate shouldn't be update ?


DimaRGB:
do you mean `'isntall': 'install',` ?

jbcpollak:
Accurev also supports this alias

fearphage:
FYI, the title of this issue is not descriptive at all. It doesn't tell you anything about what's inside.

zkat (approved):
ok

PR #17249, opened 2017-06-16
Fix self-reference in package-lock.json docs

not-an-aardvark:
The "see also" section in the [package-lock.json manpage](https://docs.npmjs.com/files/package-lock.json) links to itself.

### DIFF
--- a/doc/files/package-lock.json.md
+++ b/doc/files/package-lock.json.md
@@ -127,5 +127,6 @@ The dependencies of this dependency, exactly as at the top level.
 ## SEE ALSO
 
 * npm-shrinkwrap(1)
+* npm-shrinkwrap.json(5)
 * package.json(5)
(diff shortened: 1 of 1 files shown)


doc/files/package-lock.json.md
```diff
@@ -127,6 +127,5 @@ The dependencies of this dependency, exactly as at the top level.
 ## SEE ALSO
 
 * npm-shrinkwrap(1)
-* package-lock.json(5)
```
iarna, line 130: I think this was supposed to be `npm-shrinkwrap.json(5)`

not-an-aardvark: Thanks, fixed.

iarna:
Thanks! This has been merged as f19bd3c8c!

PR #17250, opened 2017-06-16
Fix typo

Zarel:
image


watilde:
Good catch! We might want to update L16 as well.
> doc/files/npm-shrinkwrap.json.md:16:Additionally, if both `package-lock.json` and `npm-shrinwkrap.json` are present

Zarel:
Oh, good other catch!

iarna:
Thanks! This has been merged as 32fc6e41a!

PR #17319, opened 2017-06-19
docs: add --silent to npm-run-script

### LINKED ISSUE #13241: Docs: Show `--silent` flag usage for npm run script
#### What's the feature?

Hide/silence npm ERR output when using `npm run script` so that the only output on the console is the from the script that was executed, not from npm.

Many times a day, developers use `npm run build` and the "npm ERR!" output is filling up the console with useless information. The dev has to scroll up past all the junk to find the actual error message. See [this question](http://stackoverflow.com/q/38084112/266535) on stackoverflow for more details.

No

Typically there is a `--verbose` flag to get additional details and the default behavior is to have fewer details. Some tools use a `--silent` flag if the default is verbose.

Maybe

styfle:
This fixes #13241

### DIFF
--- a/doc/cli/npm-run-script.md
+++ b/doc/cli/npm-run-script.md
@@ -3,7 +3,7 @@ npm-run-script(1) -- Run arbitrary package scripts
 
 ## SYNOPSIS
 
-    npm run-script <command> [-- <args>...]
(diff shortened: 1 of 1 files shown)


iarna:
Thanks! This has been merged as d57d4f48c and will be in the next npm release.

Issue #17068, opened 2017-06-07
npm prune creates package-lock.json file


watson:
#### I'm opening this issue because:

  - [ ] npm is crashing.
  - [ ] npm is producing an incorrect install.
  - [x] npm is doing something I don't understand.
  - [ ] Other (_see below for feature requests_):

#### What's going wrong?

A package-lock.json file is created if you run `npm prune` in a directory that doesn't have one already.

This is what it will look like if there is no package.json and no modules installed in the directory:

```json
{
  "lockfileVersion": 1
}
```

This is what it will look like if there is a package.json in the same directory, but `npm install` have not yet been run:

```json
{
  "name": "foo",
  "version": "1.0.0",
  "lockfileVersion": 1
}
```

If modules are already installed in a `node_modules` folder those will be listed in the package-lock.json file as well - no matter if there is a local package.json file listing those dependencies or not.

#### How can the CLI team reproduce the problem?

1. Go to a directory without a package-lock.json file, e.g: `cd /tmp && mkdir foo && cd foo` 
1. Run `npm prune`
1. Check to see if a package-lock.json file have been created

### supporting information:

 - `npm -v` prints: `5.0.3`
 - `node -v` prints: `v8.0.0`
 - `npm config get registry` prints: `https://registry.npmjs.org/`
 - Windows, OS X/macOS, or Linux?: macOS 10.12.5

This issue is related to #17061

PR #17329, opened 2017-06-19
prune: stop creating `package-lock.json`

watilde:
A package-lock.json file was created when `npm prune` is executed, and it deleted the existed dependencies/devDependencies. `npm prune` shouldn not touch `package-lock.json` from the first.

related issues:
+ https://github.com/npm/npm/issues/17061
+ https://github.com/npm/npm/issues/17068

### DIFF
--- a/lib/prune.js
+++ b/lib/prune.js
@@ -5,6 +5,7 @@ module.exports.Pruner = Pruner
 
 prune.usage = 'npm prune [[<@scope>/]<pkg>...] [--production]'
 
+var path = require('path')
 var npm = require('./npm.js')
 var log = require('npmlog')
 var util = require('util')
@@ -16,12 +17,16 @@ var removeDeps = require('./install/deps.js').removeDeps
 var loadExtraneous = require('./install/deps.js').loadExtraneous
 var chain = require('slide').chain
 var computeMetadata = require('./install/deps.js').computeMetadata
+var exists = require('./install/exists.js')
 
 prune.completion = require('./utils/completion/installed-deep.js')
 
 function prune (args, cb) {
-  var dryrun = !!npm.config.get('dry-run')
-  new Pruner('.', dryrun, args).run(cb)
+  exists(path.join(npm.localPrefix, 'package-lock.json'), function (e) {
+    if (e) npm.config.set('save', false)
+    var dryrun = !!npm.config.get('dry-run')
+    new Pruner('.', dryrun, args).run(cb)
+  })
 }
 
 function Pruner (where, dryrun, args) {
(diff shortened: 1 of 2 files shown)


lib/prune.js
```diff
@@ -21,6 +21,8 @@ prune.completion = require('./utils/completion/installed-deep.js')
 
 function prune (args, cb) {
   var dryrun = !!npm.config.get('dry-run')
+  // prune should not touch the lock file
+  npm.config.set('save', false)
```
watson: I think the proper config option to use for this is `shrinkwrap`. The package-lock.json file is only not produced as a side-effect of setting `save` to `false`.

watson: Btw, isn't setting this config option a permanent thing? Or is it just for this command and this run?

watilde: I think `npm.config.set('shrinkwrap', false)` does not work because we need to avoid `saveToDependencies` at [lib/install.js#L294-L300](https://github.com/npm/npm/blob/90a25f65804d7d7dc351d01c53826b879139d8da/lib/install.js#L294-L300) to not make the lock file with `npm prune`, and the flag comes from here [lib/install/save.js#L133-L136](https://github.com/npm/npm/blob/latest/lib/install/save.js#L133-L136). The default value of the save option is `true` since npm@5, then we can turn off it at the prune class.

`npm.config.set` is to set a value to a config property in runtime. It doesn't overwrite `.npmrc`.

watson: Ah ok, that sounds right then 👍

iarna:
I'm not sure this is right… if `npm prune` is modifying the tree then it needs to update any existing `package-lock` or `shrinkwrap`. I think what you want is for it to not create one if one isn't already there, which will require some new instrumenting over in `save` I think.

iarna (changes requested):
I'm not sure this is right… if `npm prune` is modifying the tree then it needs to update any existing `package-lock` or `shrinkwrap`. I think what you want is for it to not create one if one isn't already there, which will require some new instrumenting over in `save` I think.

watilde:
That makes sense! Then I will update this process to:
+ `npm prune` doesn't make `package-lock.json` when it doesn't exist.

// correct me if I'm misunderstanding :)

watilde:
I've updated as I commented.

iarna:
@watilde Forgive me only asking this now, but why shouldn't prune create a `package-lock.json`? If you don't want one, wouldn't you be running with `package-lock = false` in your config?

jasonraimondi:
Running `npm prune --production` is updating my `package-lock.json` file on a prod deploy. It would be great if there was a way to flag `--no-lockfile` on a prune run to get this to stop updating.

Currently I am needing to reset lock changes on the step after the prune to stop the lockfile changes: `git checkout package-lock.json`

PR #17336, opened 2017-06-20
Docs: Remove note about prepublish not being run

Hirse:
See #16685

Removing the following note from the [docs](https://docs.npmjs.com/misc/scripts) as the described change is not actually correct.
> IMPORTANT: As of npm@5, prepublish will only be run for npm publish. This will make its behavior identical to prepublishOnly, so npm@6 or later may drop support for the use of prepublishOnly, and then maybe we can all forget this embarrassing thing ever happened.

Follows the change in https://github.com/npm/npm/pull/16918.


iarna:
Thanks! This has been merged as 87629880a.

Issue #3927, opened 2013-09-24
Is `npm build .` a feature or a bug?


indexzero:
@isaacs I remember years ago there was an outstanding issue (that was fixed) for the same problem in `npm link`. That is, you decided that the `.` in:

```
npm link .
```

was unnecessary and now we can all run:

```
npm link
```

The same exists now for `npm build` (I need the `.`). Is this intended? Obviously low priority.

joaomilho:
Can you explain why `npm build` "do nothing and have no output"?

DullReferenceException:
This has bit me and teammates multiple times. `npm build` is one of those things you'd expect to just work, like `npm test`. The fact that it does nothing, outputs nothing, and succeeds can cause interesting issues. For example, I recently had this fun thing in my `package.json`:

```
{
  "...": "...",
  "scripts": {
    "build": "babel --out-dir ./lib src",
    "prepublish": "npm build"
  }
}
```

This bug caused me a lot of head-scratching when my code changes weren't reflected after doing an `npm publish` and `npm install`. Since `npm build` does nothing but still succeeds, `prepublish` didn't do anything, and the published lib code was not the latest transpiled code.

Would it be unreasonable to either have `npm build` do the same thing as `npm run build`, or alternatively have it output an error and fail with an exit code?

lennym:
We've just spent "some time" trying to get to the bottom of something similar. It would be good to have a warning if `npm build` is called with no arguments when a `build` script is defined in package.json.

Happy to put together a PR if it helps.

PR #17351, opened 2017-06-20
Warn if running `npm build` without args and with a build script

lennym:
If a user mistakenly shorthands `npm build` for `npm run build` then they currently get no output or information since `npm build` without arguments is a no-op.

If `npm build` is called without arguments when a `build` script is defined in the package.json `scripts` section then output a warning message to indicate this.

See https://github.com/npm/npm/issues/3927

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -43,14 +43,19 @@ function build (args, global, didPre, didRB, cb) {
     global = npm.config.get('global')
   }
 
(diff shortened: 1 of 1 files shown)


iarna:
Thanks! This would definitely save folks a lotta trouble. It's been merged as 82df7bb16.

PR #17370, opened 2017-06-21
Update npm-folders.md

kierendixon:
Correct the default prefix config path for Windows operating systems


iarna:
Thank you! This was merged as f8e76d856!

Issue #15845, opened 2017-02-23
Change the default license to MIT


franciscop:
#### What's the feature?

The preferred license for NPM packages is most likely MIT. So make the life of developers a little bit easier by offering that as the default as the ISC is not so known and might be confusing/lead to higher % of errors.

Note: some licenses like GPL might be misrepresented because larger corporations avoid it probably more than smaller ones, but I do not have the tools to truly check what license is most used in NPM so we'll have to do with the "top NPM libraries" for my report.

#### What problem is the feature intended to solve?

The default license ISC is unknown to developers for NPM and not the most used one even when it's a default. You can check this by running `legally` for the top 10 packages:

```js
npm install legally -g
legally browserify grunt-cli bower gulp grunt express npm cordova forever less
```

![mit](https://cloud.githubusercontent.com/assets/2801252/23262893/9ee61c4a-f9dc-11e6-9c6b-84b27792f9f7.png)

#### Is the absence of this feature blocking you or your team? If so, how?

No

#### Is this feature similar to an existing feature in another tool?

No

#### Is this a feature you're prepared to implement, with support from the npm CLI team?

Yes

franciscop:
This issue is literally changing 3 characters and might improve the development workflow and legal for many teams. It is fixing a default that should never have been made a default to set another one that the community clearly prefers.

Please, fix this instead of closing it. I am making a PR with this functionality.

leonadler:
Setting the bot to close after 7 days is a bit overzealous, IMHO.

wangpung12:
Good

PR #17383, opened 2017-06-22
Change default `npm init` license to MIT

franciscop:
[See the issue here](https://github.com/npm/npm/issues/15845), this PR is changing the default license from ISC to MIT to match the clear user preferences and make developer's life easier.

### DIFF
--- a/lib/config/defaults.js
+++ b/lib/config/defaults.js
@@ -156,7 +156,7 @@ Object.defineProperty(exports, 'defaults', {get: function () {
     'init-author-email': '',
     'init-author-url': '',
     'init-version': '1.0.0',
-    'init-license': 'ISC',
+    'init-license': 'MIT',
     json: false,
     key: null,
     'legacy-bundling': false,


iarna:
It's pretty arbitrary, but there is rather a lot of ISC affection at npm. When you say defaulting to ISC might result in a "higher % of errors" what do you mean? I find ISC easier to grok if only because it has less text—MIT is needlessly wordy.

Because this feels arbitrary to me I'm not super inclined to change the default, but I'm open to being convinced otherwise.

Also, if you want to say "most developers" and have me take that assertion seriously, you're going to need better methodology than you did in your linked issue. A [quick registry follower](https://github.com/npm/registry-follower-tutorial) could gather data on the entire registry. But it should also counted by dev, not by module as some authors are unusually prolific and skew any stats that you do by-module.

franciscop:
@iarna thanks for answering, [I understand you are quite overloaded](http://blog.npmjs.org/post/161832149430/npm-the-npm-github-issue-tracker-and-you).

From a personal experience the only reason I cannot use `--yes` is because of the license, and I am pretty sure it is the same situation for many other developers. If I use it then I have to remember to change it to the MIT manually. This is at best a small annoyance, and at worst it results in developers using the ISC license by accident.

> as some authors are unusually prolific and skew any stats that you do by-module

My intention with this PR is to make authors life (including those prolific ones writing MIT code) easier when making NPM packages.

I agree that my search of top packages and its dependencies is flawed, it was a quick sanity test more than hard data. I am sorry but the reference package seems too complex for my original intention and based on NPM changes and not an absolute search. I could find a couple of analysis:

| License        | [Sept 2014 (partial set)](http://futurice.com/blog/npm-registry-in-numbers)        | [Nov 2013](https://gist.github.com/robertkowalski/7620849)
|------------------|----------------|------------------|
| MIT              | **12018**  | **20811**     |
| Unknown     | 4099          | 27785          |
| BSD             | 1120          | 5240            |
| Apache-2.0  | 687            | 263              |
| ISC              | **442**       | **31**           |

So back then **developers were 3,000-64,500% times more likely to want to use the MIT over the ISC license**. I'd argue now it's similar; I couldn't find a modern analysis, however I'm pretty sure the difference is smaller in *big* part due to the ISC being the default. **If you require hard *recent* evidence for merging this PR please let me know and I might be able to get it**, but I think these two analysis+the quick sanity test would be enough evidence that developers overwhelmingly prefer the MIT license.

junosuarez:
@franciscop: I recommend adding this to your `~/.npmrc` if you want `npm init -y` to default to MIT:

```
init.license="MIT"
```

franciscop:
@js-n that is awesome and just made it. However I still think the default should be MIT for the thousands of JS developers who prefer it.

BTW, why did you downvote? Any particular reason?

junosuarez:
To explain my downvote: I disagree that the existing default should be changed and agree that it's pretty much arbitrary.

iarna:
@franciscop 
> If you require hard recent evidence for merging this PR please let me know and I might be able to get it

This is a fair question. TBH, I don't think a volume argument is going to be very convincing to me on its own. I did have this question though:

> When you say defaulting to ISC might result in a "higher % of errors" what do you mean? 

This here is what you need to convince me of to take this.

Right now you don't like the default but I'm not seeing the harm in the current value, particularly as you can can configure npm to use whatever license you like by default.

franciscop:
@iarna 

Because people leave defaults on (this phenomenon even [has its own Wikipedia article](https://en.wikipedia.org/wiki/Default_effect_(psychology))). So the chance of mistakes is minimized when the most wanted default is used **and** it reduces cognitive load:

- **Minimize errors**: when most people prefer A but B is the default, there is bound to be a larger number of errors than if most people prefer A and A is the default. This is because people leave defaults on unconsciously (see the previously mentioned Wikipedia article) where they'd have better switched. Even the whole Ruby on Rails relies heavily on defaults (convention) as a core design concept.

- Besides that, there's also the **cognitive load**. ISC is unknown to many developers (compared to MIT), so providing a license like MIT that is not only wanted but understood is what makes sense. Taken to the extreme there are some well known examples of defaulting to a set of clothes [[1]](https://www.forbes.com/sites/jacquelynsmith/2012/10/05/steve-jobs-always-dressed-exactly-the-same-heres-who-else-does/#42dd2e965f53) [[2]](https://www.independent.co.uk/news/people/why-mark-zuckerberg-wears-the-same-clothes-to-work-everyday-a6834161.html). If you were to pick a different license, I would actually recommend Apache or GPL, both understood and next in the list of most used.

- **Easier** [nitpick]: when the developer makes the conscious effort to change the license, there are fewer words typed (but O(n), where n = number of npm packages) writing manually "MIT [enter]" than just [enter] or the mentioned `-y`. The number of people changing from ISC to MIT is much larger from the statistics found than the number of developers switching from MIT to ISC.

- [More reasons](https://en.wikipedia.org/wiki/Default_effect_(psychology)#Explanations).

I agree though that there are two reasons (that I can think of) to keep the ISC License, however IMHO they are quite weak compared to the points above:

- It is the default already. This should be a no-issue though from the points mentioned above and how popular the MIT is.

- npm, Inc prefers it and is actively pushing it. You are in the end of the day a private company who can choose whatever you want; but you also have a privileged position in Node.js ecosystem, so IMO in situations like this where there's no economic difference it'd be better to improve developer's experience with npm.

@js-n I'd argue that being 3,000-64,500% times more likely to use MIT over ISC is not *arbitrary* for the developers. Your tip worked wonders though, thanks!

nickserv:
👎 I'm biased because npm inspired me to prefer ISC over MIT, but I'd rather have the default license be easy to read than popular. If you'd rather have a more popular license for all your projects the config change looks like a great solution. I agree that cognitive load is important but I personally see that as an argument in favor of ISC. Because it's just the MIT without Berne Convention ambiguities I don't see how this could cause errors (even if you accidentally chose the wrong license the legal aspects would be almost the same anyway).

franciscop:
@nickmccurdy it is not a *popularity* contest what I'm arguing for... it is a user preference! The great majority of npm developers clearly *prefer* MIT, so going against your users wishes is just making things harder for them. I agree thought that how big an error is choosing ISC over MIT is arguable, but it *is* a mistake if chosen by accident nevertheless.

PS, the only reason I care now is because I care about npm and making developers job easier; my issue has been fixed thanks to js-n.

nickserv:
I understand a lot of users would prefer MIT (and I respect that this is about what would be easier for most developers), though I agree with @iarna that I don't don't see harm in leaving the default as ISC. I feel like this is more of a semantic argument because the licenses are so legally similar.

PR #17411, opened 2017-06-23
action: no need to promisify move

vramana:


iarna:
Thank you! This was one of those things that was one of those cleanup chores we had hanging around post release and having this as a contribution is a huge help. This is merged as a01ba2afa.

PR #17441, opened 2017-06-26
15% speed up: avoid building environment for empty lifecycle scripts

mikesherov:
Hello npm friends! Thanks again for everything you do! My second contribution to npm perf is alas, another one-liner. In my tests, this saves about 3 seconds on a run where 1800 modules are being installed, which is roughly 10-15% of the run.

It avoids building environment variables and checking working directories for non existent lifecycle scripts.

Please let me know if you have any questions.


iarna:
Thanks, this looks fantastic. Merged as c01876db2!

Issue #17168, opened 2017-06-13
npm version fails if package-lock.json is gitignored


felixfbecker:
#### I'm opening this issue because:

  - [x] npm is crashing.
  - [ ] npm is producing an incorrect install.
  - [ ] npm is doing something I don't understand.
  - [ ] Other (_see below for feature requests_):

#### What's going wrong?
If package-lock.json is .gitignored, npm version will abort at the `git add package-lock.json` step and not do a version commit + tag.

```
14:01 $ npm version major
v1.0.0
npm ERR! code 1
npm ERR! Command failed: git add /.../package-lock.json
npm ERR! The following paths are ignored by one of your .gitignore files:
npm ERR! package-lock.json
npm ERR! Use -f if you really want to add them.
npm ERR!
```

I know you recommend commiting package-lock.json even for libraries, but there are also a lot of reasons that speak against it (false sense of security, mostly targeting older versions, ...). It should be up for the user to decide, at most print a warning, not fail `npm version`.

#### How can the CLI team reproduce the problem?

- add package-lock.json to .gitignore
- run `npm version`

### supporting information:

 - `npm -v` prints: 5.0.3
 - `node -v` prints: 8.0.0
 - `npm config get registry` prints: standard
 - Windows, OS X/macOS, or Linux?: macos
 - Network issues:
   - Geographic location where npm was run:
   - [ ] I use a proxy to connect to the npm registry.
   - [ ] I use a proxy to connect to the web.
   - [ ] I use a proxy when downloading Git repos.
   - [ ] I access the npm registry via a VPN
   - [ ] I don't use a proxy, but have limited or unreliable internet access.
 - Container:
   - [ ] I develop using Vagrant on Windows.
   - [ ] I develop using Vagrant on OS X or Linux.
   - [ ] I develop / deploy using Docker.
   - [ ] I deploy to a PaaS (Triton, Heroku).

richardsimko:
I agree! We ran in to this issue as well and with the current instability of package-lock.json (Changing back and forth between sha1 and sha512, sometimes not installing all dependencies, sometimes not including all dependencies etc) it's not feasible to rely on it.

zkat:
https://github.com/npm/npm/pull/17505 should fix this. Cheers.

felixfbecker:
This is still happening in v5.3.0 and v5.4.0

justinbmeyer:
[...]

PR #17505, opened 2017-06-29
version: support ignored shrinkwrap and package-locks

zkat:
Fixes: https://github.com/npm/npm/issues/17168

### DIFF
--- a/lib/version.js
+++ b/lib/version.js
@@ -278,26 +278,33 @@ function checkGit (localData, cb) {
 }
 
 function _commit (version, localData, cb) {
-  var packagePath = path.join(npm.localPrefix, 'package.json')
-  var options = { env: process.env }
-  var message = npm.config.get('message').replace(/%s/g, version)
-  var sign = npm.config.get('sign-git-tag')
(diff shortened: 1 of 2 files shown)


iarna:
I wish the "I search and replaced var to const" part was a separate commit. It's v hard to see what your changes are as it is.

zkat:
@iarna `+exports.exec = BB.promisify(execGit)` is the only actual change in that file.

I can pull that out and just do it separately, tho 👍

lib/version.js
```diff
@@ -278,26 +278,33 @@ function checkGit (localData, cb) {
...
-    cb
-  )
+    }
+  }).nodeify(cb)
+}
+
+function addLocalFile (file, options, ignoreFailure) {
+  const p = git.exec(['add', path.join(npm.localPrefix, file)], options)
+  return ignoreFailure
+  ? p.catch(() => {})
```
iarna, line 306: Ah, this is the new bit! 👍

iarna (approved):

patrykkopycinski:
regression in `5.3.0`

Issue #17173, opened 2017-06-13
npm 5.0.3 behaves strangely if package.json is read-only


radcapitalist:
#### I'm opening this issue because:

  - [ ] npm is crashing.
  - [ ] npm is producing an incorrect install.
  - [x] npm is doing something I don't understand.
  - [ ] Other (_see below for feature requests_):

#### What's going wrong?
This is on Windows.  We don't all use git for source control.  I use Perforce.  So files sync'd from source control are read-only unless checked out.  This was not a problem before.  With npm 5.0.3, it is a problem.  If package.json is read-only, running "npm install" always produces this message:

c:\jmpdev\main\source\Server\webjmp>npm install
npm WARN saveError EPERM: operation not permitted, rename 'c:\jmpdev\main\source
\Server\webjmp\package.json.3869010878' -> 'c:\jmpdev\main\source\Server\webjmp\
package.json'

I *think* the install succeeds, and no file named "package.json.3869010878" is actually getting created.  But the spurious message is quite disconcerting.

#### How can the CLI team reproduce the problem?

My npm-debug.log file is zero-length.

For any node.js project, make package.json read-only and then run "npm install".

### supporting information:

 - `npm -v` prints:  5.0.3
 - `node -v` prints:  v8.1.0
 - `npm config get registry` prints:  https://registry.npmjs.org/
 - Windows, OS X/macOS, or Linux?:  Windows

zkat:
And here's a PR: https://github.com/npm/npm/pull/17506

PR #17506, opened 2017-06-29
save: only write package.json and locks if files have changed

zkat:
Fixes: https://github.com/npm/npm/issues/17173


iarna (approved):
LGTM

PR #17508, opened 2017-06-29
Lock missing dep fixes

### LINKED ISSUE #16839: Missing dependencies after running `npm install` a second time

  - [ ] npm is crashing.
  - [X] npm is producing an incorrect install.

#### What's going wrong?

When there is an existing lockfile, `npm5` does not install all necessary packages.  (This issue specifically affects any project that depends on `foundry-kue-scheduler`)

The initial invocation of `npm install` installs the correct set of dependencies, and creates a lockfile as I would expect.  

When I delete `node_modules`, and run `npm install` again (against the same lockfile), I get a different set of packages (several transitive dependencies are missing), _and_ npm alters the contents `package-lock.json`.

This seems to be a minimal test-case that shows how `npm install` is broken for any package that depends on `foundry-kue-scheduler`:

```shell
# scaffold a minimal project
$ mkdir npm5-test
$ cd npm5-test
$ npm init -y

# install/save my first/only dependency
$ npm i foundry-kue-scheduler

npm WARN deprecated redlock@2.1.1: possible critical bug, see https://github.com/mike-marcacci/node-redlock/issues/31
npm notice created a lockfile as package-lock.json. You should commit this file.
npm WARN npm5-test@1.0.0 No description
npm WARN npm5-test@1.0.0 No repository field.

added 203 packages in 11.601s

# npm installed my packages and creates a lockfile
$ ls -lh
total 88
drwxr-xr-x  194 andrew  staff   6.4K May 31 15:09 node_modules
-rw-r--r--    1 andrew  staff    38K May 31 15:09 package-lock.json
-rw-r--r--    1 andrew  staff   286B May 31 15:09 package.json

# one of my transitive dependencies is present
$ ls -lh node_modules/kue/package.json
-rw-r--r--  1 andrew  staff   2.0K May 31 15:09 node_modules/kue/package.json

# At this point, everything is OK.  
# NPM has installed the set of packages that I expected.
# I can run `npm install` again, and it correctly reports that there's nothing to do.

# now, let's remove node_modules and reinstall (as though I'm installing
# packages for the first time in a repo that has package-lock.json checked-in,
# and node_modules is listed in .gitignore )
$ rm -rf node_modules/
$ npm i

npm WARN npm5-test@1.0.0 No description
npm WARN npm5-test@1.0.0 No repository field.

added 192 packages in 4.73s

# note that fewer packages were installed, and npm MODIFIED the extant lockfile
$ ls -lh
total 88
drwxr-xr-x  185 andrew  staff   6.1K May 31 15:11 node_modules
-rw-r--r--    1 andrew  staff    36K May 31 15:11 package-lock.json
-rw-r--r--    1 andrew  staff   286B May 31 15:11 package.json

# also, my transitive dependency is gone.
$ ls node_modules/kue/package.json
ls: node_modules/kue/package.json: No such file or directory

# subsequent invocations of npm install make things even worse
$ npm i

npm WARN npm5-test@1.0.0 No description
npm WARN npm5-test@1.0.0 No repository field.

removed 190 packages in 3.026s

# ahhhhhh!
$ ls -lh
total 16
drwxr-xr-x  5 andrew  staff   170B May 31 15:15 node_modules
-rw-r--r--  1 andrew  staff   503B May 31 15:15 package-lock.json
-rw-r--r--  1 andrew  staff   286B May 31 15:15 package.json
```

Here's [a gist](https://gist.github.com/schmod/552a0397bf60af217243f0486ef03794/revisions) that shows the package structure, as well as the changes to `package-lock.json` between the first and second invocation of `npm install`.  

### supporting information:

 - `npm -v` prints: `5.0.0`
 - `node -v` prints: `v6.10.3`
 - Windows, OS X/macOS, or Linux?: macOS (also verified inside Docker w/ Linux)
 - Network issues:
 - Container:
   - [X] I develop / deploy using Docker.

### LINKED ISSUE #16866: package-lock.json file not updated after package.json file is changed

  - [ ] npm is crashing.

#### What's going wrong?
clean project with some deps in `package.json`.
you run `npm i`
modules are installed and `package-lock.json` file is created.
say you update module A in `package.json` file.
you run `npm i`. I would expect this updates the `package-lock.json` file but it doesn't. which results in module A not being updated.

any project. see description above.

### supporting information:

 - `npm -v` prints: 5.x.x
 - `node -v` prints: 7.x.x
 - `npm config get registry` prints: private one
 - Windows, OS X/macOS, or Linux?: macOS
 - Network issues:
 - Container:

iarna:
This has a handful of fixes:

1. It introduces a new `package-lock.json` field, called `requires`, which tracks which modules a given module requires.
2. It fixes #16839 which was caused by not having this information available, particularly when git dependencies were involved.
3. It fixes #16866, allowing the `package.json` to trump the `package-lock.json`.
4. It stops calling `fetch-package-metadata.addBundled` on fake children (this was harmless, because bundleDependencies would never be set, but also pointless.)
5. `npm ls` now loads the shrinkwrap, which opens the door to showing a full tree of dependencies even when nothing is yet installed. (It doesn't do that yet though.)

### DIFF
--- a/doc/spec/package-lock.md
+++ b/doc/spec/package-lock.md
@@ -36,7 +36,11 @@ property match that environment variable.
 
 ### dependencies
 
-A mapping of package name to dependency object.  Dependency objects have the
+These are the modules installed in the `node_modules`.  Some of these are
+dependencies some of these are transitive dependencies (that is,
+dependencies of our dependencies).
+
+This is a mapping of package name to dependency object.  Dependency objects have the
 following properties:
 
 #### version *(changed)*
@@ -108,9 +112,17 @@ on the current platform.
 This is a record of what specifier was used to originally install this
 package.  This should not be included in new `package-lock.json` files.
 
+#### requires
+
+This is a mapping of module name to version.  This is a list of everything
+this module requires, regardless of where it will be installed.  The version
+should match via normal matching rules a dependency either in our
+`dependencies` or in a level higher than us.
+
 #### dependencies
 
-The dependencies of this dependency, exactly as at the top level.
+Exactly like `dependencies` at the top level, this is a list of modules to
+isntall in the `node_modules` of this module.
 
 ## Generating
 
--- a/doc/spec/package-lock.md
+++ b/doc/spec/package-lock.md
@@ -28,15 +28,6 @@ The version of the package this is a package-lock for. This must match what's in
 An integer version, starting at `1` with the version number of this document
 whose semantics were used when generating this `package-lock.json`.
 
-### packageIntegrity *(new)*
-
-This is a
-[subresource integrity](https://w3c.github.io/webappsec/specs/subresourceintegrity/)
-value created from the `pacakge.json`. No preprocessing of the
-`package.json` should be done.  Subresource integrity strings can be
-produced by modules like
-[`ssri`](https://www.npmjs.com/package/ssri).
-
 ### preserveSymlinks *(new)*
 
 Indicates that the install was done with the environment variable
--- a/lib/install/inflate-shrinkwrap.js
+++ b/lib/install/inflate-shrinkwrap.js
@@ -26,7 +26,7 @@ module.exports = function (tree, swdeps, opts, finishInflating) {
   if (!npm.config.get('shrinkwrap') || !npm.config.get('package-lock')) {
     return finishInflating()
   }
-  tree.loaded = true
+  tree.loaded = false
   tree.hasRequiresFromLock = Object.keys(swdeps).every((d) => swdeps[d].requires)
   return inflateShrinkwrap(tree.path, tree, swdeps, opts).then(
     () => finishInflating(),
@@ -168,7 +168,7 @@ function fetchChild (topPath, tree, sw, requested) {
     var isLink = pkg._requested.type === 'directory'
     const child = createChild({
       package: pkg,
-      loaded: true,
+      loaded: false,
       parent: tree,
       fromShrinkwrap: requested,
       path: childPath(tree.path, pkg),
--- a/lib/install/inflate-shrinkwrap.js
+++ b/lib/install/inflate-shrinkwrap.js
@@ -88,7 +88,7 @@ function inflatableChild (onDiskChild, name, topPath, tree, sw, requested, opts)
     onDiskChild.swRequires = sw.requires
     tree.children.push(onDiskChild)
     return BB.resolve(onDiskChild)
-  } else if (opts.fakeChildren !== false && sw.version && sw.integrity) {
+  } else if (sw.version && sw.integrity) {
     // The shrinkwrap entry has an integrity field. We can fake a pkg to get
     // the installer to do a content-address fetch from the cache, if possible.
     return BB.resolve(makeFakeChild(name, topPath, tree, sw, requested))
--- a/lib/install/inflate-shrinkwrap.js
+++ b/lib/install/inflate-shrinkwrap.js
@@ -114,34 +114,28 @@ function makeFakeChild (name, topPath, tree, sw, requested) {
     _where: topPath,
     _args: [[requested.toString(), topPath]]
   }
-  let bundleAdded = BB.resolve()
-  if (Object.keys(sw.dependencies || {}).some((d) => {
-    return sw.dependencies[d].bundled
-  })) {
-    pkg.bundleDependencies = []
-    bundleAdded = addBundled(pkg)
-  }
-  return bundleAdded.then(() => {
-    const child = createChild({
-      package: pkg,
-      loaded: true,
-      parent: tree,
-      children: pkg._bundled || [],
-      fromShrinkwrap: true,
-      fakeChild: sw,
-      fromBundle: sw.bundled ? tree.fromBundle || tree : null,
-      path: childPath(tree.path, pkg),
-      realpath: childPath(tree.realpath, pkg),
-      location: tree.location + '/' + pkg.name,
-      isInLink: tree.isLink
-    })
-    tree.children.push(child)
-    if (pkg._bundled) {
-      delete pkg._bundled
-      inflateBundled(child, child, child.children)
+
+  if (!sw.bundled) {
+    const bundleDependencies = Object.keys(sw.dependencies || {}).filter((d) => sw.dependencies[d].bundled)
+    if (bundleDependencies.length === 0) {
+      pkg.bundleDependencies = bundleDependencies
     }
-    return child
+  }
+  const child = createChild({
+    package: pkg,
+    loaded: true,
+    parent: tree,
+    children: pkg._bundled || [],
+    fromShrinkwrap: true,
+    fakeChild: sw,
+    fromBundle: sw.bundled ? tree.fromBundle || tree : null,
+    path: childPath(tree.path, pkg),
+    realpath: childPath(tree.realpath, pkg),
+    location: tree.location + '/' + pkg.name,
+    isInLink: tree.isLink
   })
+  tree.children.push(child)
+  return child
 }
 
 function adaptResolved (requested, resolved) {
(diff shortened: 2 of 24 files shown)


lib/install/inflate-shrinkwrap.js
```diff
@@ -26,7 +26,8 @@ module.exports = function (tree, swdeps, opts, finishInflating) {
   if (!npm.config.get('shrinkwrap') || !npm.config.get('package-lock')) {
     return finishInflating()
   }
-  tree.loaded = true
+  tree.loaded = false
+  tree.hasRequiresFromLock = Object.keys(swdeps).every((d) => swdeps[d].requires)
```
zkat: shouldn't this be `.some()`? You only need to know if there was _one_, right?

package-lock.json
```diff
@@ -6,113 +6,163 @@
     "abbrev": {
       "version": "1.1.0",
       "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.1.0.tgz",
-      "integrity": "sha1-0FVMIlZjbi9W58LlrRg/hZQo2B8="
+      "integrity": "sha1-0FVMIlZjbi9W58LlrRg/hZQo2B8=",
+      "requires": {}
```
zkat: wait, was your solution to unconditionally add a `requires:{}`? Mehhhh. I don't like this kinda noop crud :\

lib/shrinkwrap.js
```diff
@@ -133,13 +127,28 @@ function shrinkwrapDeps (deps, top, tree, seen) {
     }
     if (childIsOnlyDev) pkginfo.dev = true
     if (isOptional(child)) pkginfo.optional = true
+    pkginfo.requires = {}
```
zkat: meh @ always having `requires`

lib/install/mutate-into-logical-tree.js
```diff
@@ -81,7 +81,12 @@ function translateTree_ (tree, seen) {
   pkg._dependencies = pkg.dependencies
   pkg.dependencies = {}
   tree.children.forEach(function (child) {
-    pkg.dependencies[moduleName(child)] = translateTree_(child, seen)
+    const dep = pkg.dependencies[moduleName(child)] = translateTree_(child, seen)
+    if (child.fakeChild) {
+      dep.missing = true
+      dep.optional = child.package._optional
+      dep.requiredBy = child.package._spec
```
zkat, line 88: I don't get this line...

lib/install/inflate-shrinkwrap.js
```diff
@@ -105,43 +105,40 @@ function makeFakeChild (name, topPath, tree, sw, requested) {
...
     _resolved: adaptResolved(requested, sw.resolved),
     _requested: requested,
     _optional: sw.optional,
     _integrity: sw.integrity,
     _from: from,
     _spec: requested.rawSpec,
     _where: topPath,
-    _args: [[requested.toString(), topPath]]
+    _args: [[requested.toString(), topPath]],
+    dependencies: sw.requires
```
zkat, line 118: this feels so mischievous 😈

zkat (commented):
Added some comments. Nothing really bad jumps out at me. I'll take a closer look once I've gotten some sleep and/or can actually talk through this and test it out.

lib/install/inflate-shrinkwrap.js
iarna: well... no, if any of them are missing then we can't assume we know anything about the relationships between the modules.

lib/install/mutate-into-logical-tree.js
iarna: This is where we're producing `read-installed`'s output for `ls` to consume. In it's format `requiredBy` is the spec that required the missing dep.

package-lock.json
iarna: Well, I'm open to alternatives.

lib/install/inflate-shrinkwrap.js
iarna: At one point during the process of writing this it was possible to produce an install that only had partial resolved information. Buuut…

That doesn't seem to be the case any more. So this can become `some` and we can stop including `requires` on every node.

Relatedly, I'm finding that `npm install xyz` or `npm rm xyz` when you have a legacy package-lock but nothing installed fails to properly upgrade the package-lock, eg:

```
$ node ~/code/npm rm null

added 190 packages and removed 1 package in 6.671s
```

zkat:
@iarna when I use `npm ls`, I notice that we get all those nice unmet dependencies, which is awesome. There's two things, tho:

1. All unmet dependencies are reported as required by the _root_ package, not by their actual logical parent.
2. I feel like it's redundant for us to spit out individual errors/warnings when we're already printing the tree that shows exactly which ones are missing/invalid/etc. We could probably just collapse that all into a _single_ warning/error that says "some dependencies are missing or invalid" before doing exit 1.

iarna:
@zkat If it's showing you anything beyond the first level that's a bug. (It doesn't do that for me in my examples…)

We can bikeshed the details of how this looks different showing the full tree later, imo.

doc/spec/package-lock.md
```diff
@@ -117,9 +112,17 @@ on the current platform.
...
+This is a mapping of module name to version.  This is a list of everything
+this module requires, regardless of where it will be installed.  The version
+should match via normal matching rules a dependency either in our
+`dependencies` or in a level higher than us.
+
 #### dependencies
 
-The dependencies of this dependency, exactly as at the top level.
+Exactly like `dependencies` at the top level, this is a list of modules to
+isntall in the `node_modules` of this module.
```
legodude17, line 125: Is that supposed to say `isntall`?

doc/spec/package-lock.md
```diff
@@ -28,15 +28,6 @@ The version of the package this is a package-lock for. This must match what's in
 An integer version, starting at `1` with the version number of this document
 whose semantics were used when generating this `package-lock.json`.
 
-### packageIntegrity *(new)*
```
legodude17, line 31: Why was this removed? I am confused.

test/tap/shrinkwrap-local-dependency.js
```diff
@@ -85,7 +85,7 @@ test('shrinkwrap uses resolved with file: on local deps', function (t) {
       t.comment(stderr.trim())
       t.equal(code, 0, 'npm exited normally')
       var data = fs.readFileSync(path.join(testdir, 'npm-shrinkwrap.json'), { encoding: 'utf8' })
-      t.deepEqual(
+      t.like(
```
legodude17, line 88: Is this an update to `tap` or a design choice?

legodude17 (commented):
:package:

doc/spec/package-lock.md
iarna: This was a hash of the `package.json` which would result in churn and merge conflicts on the `package-lock.json`. It was never implemented and the consensus of the discussion is that it was a bad idea.

doc/spec/package-lock.md
iarna: nope, ty

test/tap/shrinkwrap-local-dependency.js
iarna: `deepEqual` requires exact matching of the datastructures. `like` let's you compare an incomplete datastructure against a complete one. Where they share keys they have to match.

So this change was made to allow changes to the `package-lock` format that aren't related to this test to happen w/o breaking the test.

zkat (approved):
LGTM! 💯 🐑 🚀 💥 🔥 😱

ntwb:
Oh well, this didn't work, once Travis CI tests are run a second time with caching enabled npm continues to crash 😢 

• https://travis-ci.org/WordPress-Coding-Standards/eslint-plugin-wordpress/jobs/249240419

Reference: https://github.com/npm/npm/issues/17419

PR #17510, opened 2017-06-29
pack: never ignore COPYING files

strugee:
GNU's GPL usage guides recommend using this name, so never ignore it as it may include a copy of the GPL.

I couldn't find any documentation as to exactly how npm commit messages should be formatted (and in particular what should go before the `:`) so I guessed. Happy to amend if necessary.

### DIFF
--- a/lib/utils/tar.js
+++ b/lib/utils/tar.js
@@ -71,7 +71,7 @@ BundledPacker.prototype.applyIgnores = function (entry, partial, entryObj) {
     if (entry.match(/^readme(\.[^\.]*)$/i)) return true
 
     // license files should never be ignored.
-    if (entry.match(/^(license|licence)(\.[^\.]*)?$/i)) return true
+    if (entry.match(/^(license|licence|copying)(\.[^\.]*)?$/i)) return true
 
     // copyright notice files should never be ignored.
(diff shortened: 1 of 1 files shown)


iarna:
Your commit message is 💯

As for the content… this makes sense to me, though I think we maybe want to wait till we update to tar@3. With that we're replacing `lib/utils/tar.js` and all of `fstream` and it'll make doing this kind of thing way easier. (As it is, we couldn't accept this without also editing `fstream-npm`… it's all kinds of terrible.)

I'm going to leave this open until we do the tar update to make sure we remember to merge a version of it (or discuss it further, if someone disagrees with me).

strugee:
@iarna neato! Feel free to ping me and I can submit something like this for `tar@3`

zkat:
All of this logic now [lives in `npm-packlist`](https://github.com/npm/npm-packlist/blob/master/index.js#L134) -- would you be so kind as to submit this change over there? It'll propagate to npm itself on upgrade.

Since we now have a way forward (externally to npm proper), I'm going to close this. Thanks for your patience and for submitting this patch!

strugee:
Filed npm/npm-packlist#11! :tada:

Thanks @zkat!

PR #17479, opened 2017-06-28
docs: Add Submitting a pull request section

vramana:

PR #17513, opened 2017-06-29
pack: generate portable tarfiles by default

### LINKED ISSUE #17412: Normalize timestamps in `npm pack` tarballs

  - [ ] npm is crashing.

#### Supporting information:

 - `npm -v` prints: 5.0.3
 - `node -v` prints: 8.1.2

#### What's the feature?

Embed normalized timestamps (e.g. `0` or something) or better, no timestamps in tar files that `npm pack` generates.

This allows packages to be built such that they are bit-for-bit identical to previous builds of that package. This is called reproducible builds and has a lot of useful security properties: notably, you can reproduce the build yourself and get a stronger guarantee that the "binary" (package) you downloaded actually corresponds to the source code provided by upstream. More info (and better explanations) at https://reproducible-builds.org/.

No.

Debian has been pushing changes like this into lots of other upstream toolchains.

Seems pretty trivial, so sure.

strugee:
In particular this will produce reproducible packages out-of-the-box
for most npm packages. See https://reproducible-builds.org/

Fixes #17412 

Per #17479 I targeted this at `release-next` but am happy to change to `latest` if necessary.


strugee:
I should also call out the fact that this has no tests, because I don't really see a useful way to test this. Suggestions welcome.

iarna:
Hey! Thanks for putting this together. The version of `tar` that we use (`2.2.1`) does not accept the `portable` flag. Upgrading to `tar@3` is very nearly the top of our todo lists, but it hasn't quite gotten there yet. When we do, we'll be building _only_ portable tarballs. I don't anticipate giving a config option. As that work is going to be the tar upgrade here and in `pacote`, I'm going to close this PR, but we're definitely going to be doing this _soon_. Maybe 5.2?

strugee:
@iarna oh, weird! I tested before and after this patch and (IIRC; it was quite late) it seemed to make a difference. Maybe I'm misremembering this though :)

Anyway, I'm looking forward to this feature! Thanks for working on npm :tada:

PR #17528, opened 2017-06-29
Chore: remove unused deps.

aladdin-add:
should also update package-lock.json?


kenany:
Previously: https://github.com/npm/npm/pull/15103 (which wanted to remove a subset of the deps you're removing here).

aladdin-add:
thanks for the clarification.

can we give a warning to those packages using these deps -- so we can remove in next major release.
or already have done this?

iarna:
@Aladdin-ADD Thanks for giving this a try. As @KenanY pointed out we actually need these. One, `node-gyp` is actually a direct dependency. (We depend on its _binary_, not it's libs, so that's probably why you missed it.)

There's nothing wrong with packages using these deps. There's no reason to remove them. The only harm is that it's confusing to folks who are looking for unused deps.

That being said, the original concerns around `npm@1` being able to update are no longer a problem as `npm@1` shipped with versions of Node.js that we don't support any more (and that `npm@5` won't run on).

We're still have to have the nested `node_modules` however for a little while yet. Currently there's a bug in the Node.js installer code that when it goes to install `npm` it doesn't remove previous versions first. This combines with a flattened tree to make flat npm's installed with it break. The Node.js team is currently working on a replacement for that installer and when it's ready we'll be able to flatten out npm's dependencies entirely.

Issue #17327, opened 2017-06-19
npm version --no-git-tag-version -m "%s" doesn't commit the message passed


RinkAttendant6:
#### I'm opening this issue because:

  - [ ] npm is crashing.
  - [ ] npm is producing an incorrect install.
  - [ ] npm is doing something I don't understand.
  - [x] Other (_see below for feature requests_):

#### What's going wrong?

`npm --no-git-tag-version version patch -m "Releasing %s"` is not creating a commit. I would expect this to still create the commit but skip the tag only.

#### How can the CLI team reproduce the problem?

`run npm --no-git-tag-version version patch -m "Releasing %s"` on any npm package and observe no commit made.

### supporting information:

Reopening #13242.
#7186 seems to say after a code dive that this is what the code does as the commit and tag is a single function

#### What's the feature?

Make `--no-git-tag-version` not tag but still commit

#### What problem is the feature intended to solve?

Some users use GitHub releases or other mechanisms for managing tagging, but would like the commit functionality.

#### Is the absence of this feature blocking you or your team? If so, how?

No, but it requires I do a `git commit` in post-version and the flag is misleading. Also, it's not straightforward to put the version in the commit message as would typically be done by `npm version -m "Releasing %s"`

#### Is this feature similar to an existing feature in another tool?

Not aware of

iarna:
Changing `--no-git-tag-version` would break existing use-cases so it's not on the table.

I think it's possible to get the behavior you want without patching npm:

1. Having `git-tag-version = false` in the project `.npmrc`.
2. Adding a `postversion` script that looks like: `git add package.json package-lock.json && git commit -mv$npm_package_version` or (on windows) git add package.json package-lock.json && git commit -mv%npm_package_version%`

Does that do what you were aiming for?

RinkAttendant6:
Regarding https://github.com/npm/npm/issues/17327#issuecomment-313814070: That certainly looks like an option. Is there a way to make the one-line script there cross-platform compatible or do I need to write a small script to check `process.platform`?

For my specific case at the moment, it's fine because there is only one developer who uses Windows and he wouldn't be running `npm version` (at least not for our project.) Though it would be nice to maintain cross-platform compatibility if possible.

josephfrazier:
@RinkAttendant6, you might be able to use [cross-var](https://github.com/elijahmanor/cross-var), though I'm hoping we can still find a way for npm to do this directly (see https://github.com/npm/npm/pull/17586#issuecomment-313859875).

PR #17586, opened 2017-07-03
version: still create commit with --no-git-tag-version

josephfrazier:
This fixes https://github.com/npm/npm/issues/17327.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,9 @@ matrix:
       # then master
     - node_js: "8"
       env: DEPLOY_VERSION=testing
+      script:
+       - "standard"
+       - "node . run tap -- \"test/tap/*.js\" \"test/broken-under-nyc/*.js\""
 before_install:
   # required by test/tap/registry.js
   - "mkdir -p /var/run/couchdb"
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,25 +29,8 @@ notifications:
     slack: npm-inc:kRqQjto7YbINqHPb1X6nS3g8
 cache:
   directories:
-  - $HOME/.npm
-  - html/doc
-  - man
-  - node_modules/.bin
   - node_modules/.cache
-  - node_modules/deep-equal
-  - node_modules/marked
-  - node_modules/marked-man
-  - node_modules/npm-registry-couchapp
-  - node_modules/npm-registry-mock
-  - node_modules/require-inject
-  - node_modules/sprintf-js
-  - node_modules/standard
-  - node_modules/tacks
-  - node_modules/tap
 install:
-  - "node . rebuild --depth=0"
-  - "node . install --ignore-scripts"
-  - "node . prune"
-  - "make -j4 doc"
+  - "node . install"
 script:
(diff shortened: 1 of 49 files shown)


josephfrazier:
Hmm, not sure what's going on with the Travis tests. I *was* reproducing some failures locally (Ubuntu 17.04, Node v8.1.3) when I imitated the Travis build by running:

```bash
node . rebuild --depth=0
node . install --ignore-scripts
node . prune
node --version
make -j4 doc
node . run tap -- "test/tap/*.js" "test/broken-under-nyc/*.js"
```

However, after changing the last line to run only the failing/pending tests on Travis (in https://github.com/npm/npm/pull/17586/commits/b6812fcca4d54a2d063b4acc9b2eaa251d650b63), the previously-failing tests pass locally, while the pending ones remain pending.

josephfrazier:
I fixed the relevant tests in https://github.com/npm/npm/pull/17586/commits/ec5329ed4ff166173023dc2932e1068b1cf80629, but there's still one seemingly spurious failure: https://travis-ci.org/npm/npm/jobs/249762403#L551-L569

I haven't been able to replicate this one locally, even inside a Docker container built from the following Dockerfile (adapted from `.travis.yml`):

```docker
FROM node:8
RUN mkdir -p /var/run/couchdb
RUN git config --global user.email "you@example.com"
RUN git config --global user.name "Your Name"
COPY . /app/
WORKDIR /app
RUN node . rebuild --depth=0
RUN node . install --ignore-scripts
RUN node . prune
RUN make -j4 doc
ENTRYPOINT node . run tap -- test/tap/00-verify-ls-ok.js
```

iarna:
Your test error was caused by a travis issue that's been resolved in `release-next`, so I think that all the technical bits are addressed. I'll get with @zkat today and see what she felt like needed discussion.

iarna:
As written this breaks the existing use case for `git-tag-version` (which disables the commit and the tag) and folks rely on that, so we're not keen on changing it.

I think it's possible to get the behavior you want without patching npm:

1. Having `git-tag-version = false` in the project `.npmrc`.
2. Adding a `postversion` script that looks like: `git add package.json package-lock.json && git commit -mv$npm_package_version` or (on windows) git add package.json package-lock.json && git commit -mv%npm_package_version%`

Does that do what you were aiming for?

josephfrazier:
> As written this breaks the existing use case for git-tag-version (which disables the commit and the tag) and folks rely on that, so we're not keen on changing it.

Good point. I had considered that, but decided to propose the change anyway to see how it was received. Thanks for confirming that the behavior shouldn't change.

> I think it's possible to get the behavior you want without patching npm:

Thanks for the suggestion! This seems to work ok, but it's a bit clunky, given the fact that the script is not cross-platform (though [cross-var](https://github.com/elijahmanor/cross-var) might help with this). What if we introduced a new option/flag called `git-commit-version` that defaults to the value of `git-tag-version`, but can be overridden by explicitly setting `git-commit-version = true` in `.npmrc`, or passing `--git-commit-version` on the CLI? This should preserve the current `--no-git-tag-version` behavior, while giving users the ability to automatically commit as well.

PR #17590, opened 2017-07-03
install/deps: Computed required package in legacy npm compatible way

iarna:
Not having this meant that uninstalls of globals installed by npm@4 or
earlier would fail, due to _requested not having the fields we expected.

### DIFF
--- a/lib/install/deps.js
+++ b/lib/install/deps.js
@@ -254,11 +254,16 @@ exports.loadRequestedDeps = function (args, tree, saveToDependencies, log, next)
   }, andForEachChild(loadDeps, andFinishTracker(log, next)))
 }
 
+function isNotEmpty (value) {
+  return value != null && value !== ''
+}
+
 module.exports.computeVersionSpec = computeVersionSpec
 function computeVersionSpec (tree, child) {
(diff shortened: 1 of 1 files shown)


lib/install/deps.js
```diff
@@ -254,11 +254,16 @@ exports.loadRequestedDeps = function (args, tree, saveToDependencies, log, next)
...
+  return value != null && value !== ''
+}
+
 module.exports.computeVersionSpec = computeVersionSpec
 function computeVersionSpec (tree, child) {
   validate('OO', arguments)
   var requested
-  if (child.package._requested) {
+  var childReq = child.package._requested
+  if (childReq && (isNotEmpty(childReq.saveSpec) || (isNotEmpty(childReq.rawSpec) && isNotEmpty(childReq.fetchSpec)))) {
```
iarna, line 266: We insist that `fetchSpec` exist with `rawSpec` because it  indicates modern metadata, where as `rawSpec` alone does not.

zkat (approved):
Yay!

I'd literally forgotten that `rawSpec` existed in legacy and `saveSpec` didn't.

PR #17591, opened 2017-07-03
Fix spurious removal reporting

iarna:
Previously, if you asked to remove something that wasn't in your `node_modules` and wasn't in your `package.json` (or you didn't have a `package.json`) we would still report that we removed something even though there was no work for us to do.


zkat (approved):
LGTM! 🐑 🚀

zkat (approved):
LGTM!

Issue #16891, opened 2017-06-01
npm5 global install deletes itself in rare circumstance


brentvatne:
#### I'm opening this issue because:

  - [ ] npm is crashing.
  - [x] npm is producing an incorrect install.
  - [ ] npm is doing something I don't understand.
  - [ ] Other (_see below for feature requests_):

#### What's going wrong?

npm deletes itself when installing a dependency globally, where there is a directory in pwd that contains a `package.json` without a `name`. I noticed because I have `create-react-native-app` in `~/coding` and I tried to run `npm i -g create-react-native-app` there. At its root, the project has `package.json` [without a name](https://github.com/react-community/create-react-native-app/blob/master/package.json)

#### How can the CLI team reproduce the problem?

```
which npm # this will do what you expect
mkdir example
touch example/package.json
echo {} > example/package.json
npm i -g example
which npm # npm is now gone, magic
```

### supporting information:

 - `npm -v` prints: 5.0.0
 - `node -v` prints: 8.0.0
 - `npm config get registry` prints: https://registry.npmjs.org/
 - Windows, OS X/macOS, or Linux?: macOS
 - Network issues:
   - Geographic location where npm was run: Vancouver, Canada
   - (none of the network issues listed apply, removed for brevity)

PR #17592, opened 2017-07-03
validate-args: Check for name and version

iarna:
If you didn't have a name and you tried to install something it would end up linking to the `node_modules` where it was installing to, removing the `node_modules`.  This mostly showed up in `npm install -g .` in a package w/o a name field.

Fixes: #16891

### DIFF
--- a/lib/install/validate-args.js
+++ b/lib/install/validate-args.js
@@ -14,10 +14,21 @@ module.exports = function (idealTree, args, next) {
 
   asyncMap(args, function (pkg, done) {
     chain([
+      [hasMinimumFields, pkg],
       [checkSelf, idealTree, pkg, force],
       [isInstallable, pkg]
     ], done)
   }, next)
+}
+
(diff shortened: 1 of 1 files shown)


lib/install/validate-args.js
```diff
@@ -14,12 +14,23 @@ module.exports = function (idealTree, args, next) {
...
     chain([
+      [hasMinimumFields, pkg],
       [checkSelf, idealTree, pkg, force],
       [isInstallable, pkg]
     ], done)
   }, next)
 }
 
+function hasMinimumFields (pkg, cb) {
+  if (pkg.name === '' || pkg.name == null) {
```
zkat, line 25: maybe I'm being kinda golfy, but I'm also curious: Why not just `if (!pkg.name) {}`?

zkat (approved):
LGTM. Wondering about a single phrasing thing but the code is fine. Thaaaaaanks 💃

Issue #16126, opened 2017-03-22
Feature request: Option to only print linked dependencies in npm ls


richardsimko:
#### What's the feature?
Add an option to `npm ls` to only print dependencies which are linked, for example `npm ls --link-only`.

#### What problem is the feature intended to solve?
Currently there is no good way to get an overview of what packages are linked, while `npm ls` does show this information it would be nice to be able to filter out only the linked packages.

#### Is the absence of this feature blocking you or your team? If so, how?
No, it's mostly an annoyance.

#### Is this feature similar to an existing feature in another tool?
Not as far as I know, it can somewhat be replicated using grep, `npm ls | grep -e '->'` but it would be neat to have it built in.

#### Is this a feature you're prepared to implement, with support from the npm CLI team?
Yes, I have a working solution in place but it might need some review. Current option is `--link-only` which might also need changing.

richardsimko:
Fixed in #17616

PR #17601, opened 2017-07-04
Add possibility to ls only linked packages

richardsimko:
This is a fix for #16126 (Even though it was automatically closed by the bot) which I finally got around to implementing. The idea is to give users the ability to list only packages which have been linked using `npm link`, to get an overview of those (Useful for teams with many internal modules).

Also see this [Stackoverflow question](https://stackoverflow.com/q/24933955/1091402) (Not mine) for proof that there is demand for this feature.

I'm looking for feedback on the overall approach as well as the name of the config, are these OK? If so I will go ahead and update documentation as well as (Try to) add a test case for it.

### DIFF
--- a/doc/cli/npm-ls.md
+++ b/doc/cli/npm-ls.md
@@ -91,7 +91,7 @@ When "dev" or "development", is an alias to `dev`.
 
 When "prod" or "production", is an alias to `production`.
 
-### links
+### link-only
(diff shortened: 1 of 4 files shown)


richardsimko:
I pointed this to the wrong branch, see #17616 instead.

Issue #17338, opened 2017-06-20
package-lock.json contains os specific path when using local dependencies


reegnz:
#### I'm opening this issue because:

  - [ ] npm is crashing.
  - [ ] npm is producing an incorrect install.
  - [ X ] npm is doing something I don't understand.
  - [ ] Other (_see below for feature requests_):

#### What's going wrong?
package-lock.json does not contain an os-independent string for local dependencies, therefore package-lock.json is always changing when it was committed from windows and then building on linux.

I think the issue might be that the lock file contains an OS-normalized string, while it should actually contain some string that is independent of the OS and only gets translated when the package-lock.json is parsed.

Expected behaviour would be that the same package-lock.json is generated no matter the OS.

#### How can the CLI team reproduce the problem?
Have a package json dependency like this:
```
    "dependencies": {
        "some-local-dependency": "file:../myproject/module"
    }
```
This produces a package-lock.json on windows that has this relevant section generated:
```
    "dependencies": {
        "some-local-dependency": {
            "version": "file:\\myproject\\module"
        }
    }
```
On linux, the same package json results in the following package-lock.json (and therefore makes the committed package-lock.json dirty in git):
```
    "dependencies": {
        "some-local-dependency": {
            "version": "file:../myproject/module"
        }
    }
```

### supporting information:
 - `npm -v` prints:
5.0.1
 - `node -v` prints:
v6.10.3
 - `npm config get registry` prints:
https://registry.npmjs.org/

 - Windows, OS X/macOS, or Linux?:
windows and linux

 - Network issues:
   - Geographic location where npm was run:
   - [X] I use a proxy to connect to the npm registry.
   - [X] I use a proxy to connect to the web.
   - [X] I use a proxy when downloading Git repos.
   - [ ] I access the npm registry via a VPN
   - [ ] I don't use a proxy, but have limited or unreliable internet access.
 - Container:
   - [ ] I develop using Vagrant on Windows.
   - [ ] I develop using Vagrant on OS X or Linux.
   - [ ] I develop / deploy using Docker.
   - [ ] I deploy to a PaaS (Triton, Heroku).

reegnz:
Thanks!

zurgul:
@iarna 
Still reproducible. npm version 5.3.0.
The package chokidar@1.7.0 depends on fsevents. fsevents shall be installed only under MacOS.
Package-lock.json was created on macbook, thus "npm install" throws errors when run under Windows.

Standard8:
[...]

PR #17606, opened 2017-07-04
install: fix package.json and package-lock.json to use unix style paths (#17338)

Standard8:
This changes npm to write unix style paths regardless of platform into package.json and package-lock.json files.


iarna:
This looks great, thank you! Merged to `release-next` as 183d1058c!

PR #17616, opened 2017-07-05
Add possibility to ls only linked packages

richardsimko:
This is a fix for #16126 (Even though it was automatically closed by the bot) which I finally got around to implementing. The idea is to give users the ability to list only packages which have been linked using `npm link`, to get an overview of those (Useful for teams with many internal modules).

Also see this [Stackoverflow question](https://stackoverflow.com/q/24933955/1091402) (Not mine) for proof that there is demand for this feature.

### DIFF
--- a/doc/cli/npm-ls.md
+++ b/doc/cli/npm-ls.md
@@ -91,6 +91,13 @@ When "dev" or "development", is an alias to `dev`.
 
 When "prod" or "production", is an alias to `production`.
 
+### link
+
+* Type: Boolean
+* Default: false
+
+Display only dependencies which are linked
+
 ## SEE ALSO
 
 * npm-config(1)
--- a/lib/ls.js
+++ b/lib/ls.js
@@ -78,6 +78,8 @@ var lsFromTree = ls.fromTree = function (dir, physicalTree, args, silent, cb) {
 
   pruneNestedExtraneous(data)
   filterByEnv(data)
+  filterByLink(data)
+
   var unlooped = filterFound(unloop(data), args)
   var lite = getLite(unlooped)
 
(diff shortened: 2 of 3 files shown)


zkat:
Hey, we had to do a force-push to `release-next`. Would you kindly rebase your code onto the latest `release-next`? Cheers <3

richardsimko:
Done and done!

zkat:
@richardsimko hi!

Ok so we talked about it, and we're ok with this patch, with two changes:

1. We don't need a new option! I think it's 💯 to just reuse the existing `--link` option and just add a detail about what it means to `npm ls`.

2. It would be nice to have a test for this. You can probably crib something from `test/tap/link.js`, or even, I think, add a single test _inside_ that file to make sure `ls` is reporting what it should, here. Either way is fine by me.

_Edit: autocompleted the wrong name. Now fixed._

zkat (changes requested):
Requested some changes. See above comment. Otherwise, the idea is ok and we'll totally take it!

RichardLitt:
@richardsimko what do you think? I think tests sound good, and using `npm ls` also works.

zkat:
ahahah I'm sorry, @RichardLitt. Defeated by the autocomplete. This was totally meant for @richardsimko 😂

RichardLitt:
All good. It's a good PR!

richardsimko:
Great! I did consider re-using `--link` but I couldn't quite figure out if it made sense but I absolutely agree, I'll change it. Thanks for the pointer on where to put the tests.

richardsimko:
Done and done! Did I put the docs in the right place?

richardsimko:
The broken build is not my fault 😢  It complains about some styling error in a file from 183d1058cde1c360aaaf988eb95aa08e8e82c648 it seems.

richardsimko:
@zkat Have you had a chance to look at my changes? :)

zkat:
Oh. Yeah, this lgtm. 👍

zkat (approved):

PR #17629, opened 2017-07-06
remove: Make file: spec/link removal match specification

iarna:
When removing a link, dependencies that are installed inside of it should
not be removed if the link target is not inside the current project.

This fixes the "removing globally installed link deletes local deps"
problem, amongst others.

### DIFF
--- a/lib/install/deps.js
+++ b/lib/install/deps.js
@@ -671,6 +671,7 @@ var findRequirement = exports.findRequirement = function (tree, name, requested,
     return null
   }
   if (tree.isTop) return null
+  if (!preserveSymlinks() && /^[.][.][\\/]/.test(path.relative(tree.parent.realpath, tree.realpath))) return null
   return findRequirement(tree.parent, name, requested, requestor)
 }
 
--- a/lib/install/diff-trees.js
+++ b/lib/install/diff-trees.js
@@ -5,6 +5,7 @@ var npa = require('npm-package-arg')
 var flattenTree = require('./flatten-tree.js')
 var isOnlyDev = require('./is-only-dev.js')
 var log = require('npmlog')
+var path = require('path')
 
 function nonRegistrySource (pkg) {
   validate('O', arguments)
@@ -122,6 +123,8 @@ var diffTrees = module.exports._diffTrees = function (oldTree, newTree) {
   Object.keys(flatOldTree).forEach(function (flatname) {
     if (flatNewTree[flatname]) return
     var pkg = flatOldTree[flatname]
+    if (pkg.isInLink && /^[.][.][/\\]/.test(path.relative(newTree.realpath, pkg.realpath))) return
+
     toRemove[flatname] = pkg
     var pkgunique = getUniqueId(pkg.package)
     if (!toRemoveByUniqueId[pkgunique]) toRemoveByUniqueId[pkgunique] = []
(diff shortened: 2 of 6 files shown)


lib/install/node.js
```diff
@@ -41,13 +44,8 @@ var create = exports.create = function (node, template, isNotTop) {
...
     // by contrast, isInLink is true for only the things inside a link
-    if (node.isLink == null && isLink(node.parent)) {
-      node.isLink = true
-      node.isInLink = true
-    } else if (node.isLink == null) {
-      node.isLink = false
-      node.isInLink = false
-    }
+    if (node.isLink == null) node.isLink = isLink(node.parent)
+    if (node.isInLink == null) node.isInLink = isInLink(node.parent)
```
iarna, line 48: The way we were default before meant that nodes couldn't set their isInLink property on their own. This was stomping on values we were passing in from `install/deps`.

lib/install/inflate-shrinkwrap.js
```diff
@@ -135,6 +135,7 @@ function makeFakeChild (name, topPath, tree, sw, requested) {
     path: childPath(tree.path, pkg),
     realpath: childPath(tree.realpath, pkg),
     location: tree.location + '/' + pkg.name,
+    isLink: requested.type === 'directory',
```
iarna, line 138: Fake children need to have isLink markers too!

lib/install/diff-trees.js
```diff
@@ -122,6 +123,8 @@ var diffTrees = module.exports._diffTrees = function (oldTree, newTree) {
   Object.keys(flatOldTree).forEach(function (flatname) {
     if (flatNewTree[flatname]) return
     var pkg = flatOldTree[flatname]
+    if (pkg.isInLink && /^[.][.][/\\]/.test(path.relative(newTree.realpath, pkg.realpath))) return
```
iarna, line 126: Filter out any removes that would happen inside links to modules that aren't inside our project.

lib/install/deps.js
```diff
@@ -671,6 +671,7 @@ var findRequirement = exports.findRequirement = function (tree, name, requested,
     return null
   }
   if (tree.isTop) return null
+  if (!preserveSymlinks() && /^[.][.][\\/]/.test(path.relative(tree.parent.realpath, tree.realpath))) return null
```
iarna, line 674: Apparently something about our previously broken `isLink`/`isInLink` handling was papering over this bug. `findRequirement` needs to not count things above links-outside-our-project as viable dependencies.

(I realized I needed to do this because the spec tests started failing. =D)

test/tap/spec-local-specifiers.js
```diff
@@ -609,11 +609,133 @@ test('save behavior', function (t) {
...
+    const rmconf = {cwd: `${rmdir}/rmsymlink`, env: conf.env, stdio: conf.stdio}
+    return common.npm(['uninstall', 'dep1'], rmconf).spread((code, stdout) => {
+      t.is(code, 0, 'uninstall ran ok')
+      t.comment(stdout)
+      noFileExists(t, `${rmdir}/rmsymlink/node_modules/dep1`, 'removed symlink')
+      noFileExists(t, `${rmdir}/rmsymlink/dep1/node_modules/dep2`, 'removed transitive dep')
+      fileExists(t, `${rmdir}/rmsymlink/dep2`, 'original transitive dep still exists')
+    })
+  })
+  t.test("should not remove transitive deps if it's outside the package and --preserver-symlinks isn't set", t => {
```
zkat, line 729: typo: `--preserve-symlinks`

zkat (approved):
🐑 💯 🈂️

zkat:
Merged.

PR #17632, opened 2017-07-06
minor changes

natewatson999:
One of the scripts is strict code now. The only real changes are some use of the const keyword, semicolons, minor changes for readability, and a string.

### DIFF
--- a/lib/whoami.js
+++ b/lib/whoami.js
@@ -1,47 +1,47 @@
-var npm = require('./npm.js')
-var output = require('./utils/output.js')
-
-module.exports = whoami
-
-whoami.usage = 'npm whoami [--registry <registry>]\n(just prints username according to given registry)'
-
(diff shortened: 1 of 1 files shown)


zkat:
Hey! Thanks for taking the time to put together this PR, but:

1. Unless accompanied by other changes, or unless the lack of strict-mode is causing specific issues in that file, I don't think it's necessary to merge changes that only really add `use strict` at the top

2. The style changes are outright rejected: we use [`standard`](https://npm.im/standard) as the linter for all of npm, and any code submitted to the project must first pass that linter's checks before being accepted.

Because I don't think there's anything else other than these two things in this PR, I'm going to close this, but thanks again for the submission and I look forward to other contributions! I'm glad you took the time to read through some of npm's sources and poke at the code 💯

PR #17652, opened 2017-07-06
Fix link loops

iarna:
There are three things here, all related:
1. Make travis stop trying to do weird optimizations now that npm@5 is fast. Have travis only cache the npm cache.
2. Guard against cycles in the tree in the thing that adds in-tree-locations to nodes in the tree after loading.
3. More generally rewrite recursion guards to use `Set` where appropriate.


zkat (approved):
LGTM! (pending a standard-related change we talked about, and moving the travis patch directly into release-next)

PR #17653, opened 2017-07-06
Fix changelog issue links

statianzo:
Issue links are of the form `npm/npm/issues/:id`


zkat:
This is a duplicate of #17645, which has already landed in `release-next`! Jinx! /cc @schmod 👍

statianzo:
Thanks for the quick response @zkat and good work @schmod!

PR #17654, opened 2017-07-06
outdated: Show results for globals again

iarna:
Previously `isExtraneous` never flagged anything as extraneous w/o a
package.json.  `isExtraneous` was simplified in the refactorings in npm@5
due to `npm install` creating synthetic dependency data.  As `outdated`
didn't do that too it ended up ignoring all globals as extraneous.


zkat (approved):
🐑 🚀

PR #17668, opened 2017-07-07
Standard@10.0.2

vramana:


zkat:
@vramana yay! Thank you! lmk when this is ready to review!

vramana:
@zkat standard@10 has a bug where it simply crashes when I run it on the whole code base. Running it per directory seems to work. And standard is bit dumb assuming that every callback is node style error first callback. What should do in those cases should I disable the line causing the error or ping you how to re-write the function?

iarna:
@vramana `standard dirname` doesn't run any tests (standard has to take js files as an argument, not directories).  `standard "*.js" "bin/**/*.js" "test/**/*.js" "lib/**/*.js"` still crashes. (It's the lib one that actually causes the crash).

My recommendation is to do this as an update to `standard@9` which I can confirm does not crash. We an wait on `standard@10` till they fix their bug.

vramana:
@iarna Cool.

PR #17677, opened 2017-07-08
install: Fix deps getting moved on update

iarna:
This makes circular deps get pruned when removing obsolete dependencies (for instance, when removing an older version of a dep when updating).


zkat (approved):
LEGTUM

Issue #17639, opened 2017-07-06
package dependencies go missing when updating


SomeHats:
Thank you for npm 💜✨
I noticed this with 5.0.x and found it was still an issue in 5.1.0 so thought I'd let you know :)

#### I'm opening this issue because:
  - npm is producing an incorrect install.

#### What's going wrong?
[next](https://github.com/zeit/next.js)@2 depends on webpack 2. next@3 depends on webpack 3.

With my project on next@2, I can run `npm ls webpack` and get:
```
tmp@1.0.0 /Users/somehats/Projects/tmp
└─┬ next@2.4.6
  └── webpack@2.5.1
```
I can also successfully run `next build`.

If I run `npm i next@beta`, I get:
```
> uglifyjs-webpack-plugin@0.4.6 postinstall /Users/somehats/Projects/tmp/node_modules/uglifyjs-webpack-plugin
> node lib/post_install.js

npm WARN uglifyjs-webpack-plugin@0.4.6 requires a peer of webpack@^1.9 || ^2 || ^2.1.0-beta || ^2.2.0-rc || ^3.0.0 but none was installed.
npm WARN tmp@1.0.0 No description
npm WARN tmp@1.0.0 No repository field.

+ next@3.0.1-beta.9
added 56 packages, removed 105 packages, updated 1 package and moved 7 packages in 10.159s
```

Now, running `npm ls webpack` gives:
```
tmp@1.0.0 /Users/somehats/Projects/tmp
└── (empty)
```
and `next build` fails as it can't find webpack.

The only thing that seems to fix things is `rm -rf node_modules package-lock.json && npm i`. After that, `npm ls webpack` gives:
```
tmp@1.0.0 /Users/somehats/Projects/tmp
└─┬ next@3.0.1-beta.9
  └── webpack@3.0.0
```
and `next build` runs successfully.

#### How can the CLI team reproduce the problem?
Start with this package.json:
```json
{
  "name": "tmp",
  "version": "1.0.0",
  "dependencies": {
    "next": "2",
    "react": "^15.6.1",
    "react-dom": "^15.6.1"
  }
}
```
Run `npm i`. `npm ls webpack` should show webpack as above. Run `npm i next@beta`. Now `npm ls webpack` should give 'empty'.

### supporting information:

 - `npm -v` prints: 5.1.0
 - `node -v` prints: v8.1.2
 - `npm config get registry` prints: https://registry.npmjs.org/
 - Windows, OS X/macOS, or Linux?: macOS
[...]

PR #17678, opened 2017-07-08
install/deps: Stop flattening modules that require peer deps

iarna:
This is necessary because we don't know where their peer dep is installed.
In fact, it might not _be_ installed and that's still a valid tree.  By not
flattening it we can ensure that it's always able to load it regardless of where
it was required.

Fixes: #17639

### DIFF
--- a/lib/install/deps.js
+++ b/lib/install/deps.js
@@ -693,6 +693,10 @@ var earliestInstallable = exports.earliestInstallable = function (requiredBy, tr
   }
   if (tree.children.some(undeletedModuleMatches)) return null
 
+  // Peer deps can't be flattened because we don't know where they'll be installed.
+  // We can later add optimiations to flatten the dep if that's valid.
+  if (pkg.peerDependencies && Object.keys(pkg.peerDependencies).length) return null
+
   // If any of the children of this tree have conflicting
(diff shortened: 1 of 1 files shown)


zkat (approved):
lolsob

yvele:
@iarna @zkat This totally breaks my whole organisation 😭 

I have some packages that also acts like "_dependency presets_".. When you get one, you also get his **flattened** dependencies, and I'm depending on them.

In my use case, the problem is that many packages like [eslint-plugin-babel](https://github.com/babel/eslint-plugin-babel/blob/master/package.json) have peer dependencies and are not flattened anymore... I cannot use my "_dependency presets_" anymore, and instead should explicitly list ALL dependencies

Should I open an issue? Is my use of "_dependency presets_" not legit?

Edit: Looks like some other peeps have the same issue https://github.com/npm/npm/issues/17717

jsg2021:
Yep this totally breaks my team's entire workflow. 😭

doasync:
`"bundleDependencies": false` doesn't help (it would be a hotfix for my case)

zkat:
It's already reverted BTW. The patch is going out probably today

PR #17680, opened 2017-07-08
allows the option to extend npm commands

gabrielcsapo:
# Problem Statement

How do I run binaries in context, without using the scripts command?

# Solution

> allow npm to run global modules under the context of the binary

I have long wanted the feature to be able to extend npm to do more things with global modules.

For example if I install a module `npm-what`, the shorthand for running this should be `npm what`.

Git has had this functionality for a [while](https://github.com/git/git/blob/master/git.c#L528) and people have been extending git via this [method](https://coderwall.com/p/bt93ia/extend-git-with-custom-commands)

The previous help menu used to read as

```
Usage: npm <command>

where <command> is one of:
    access, adduser, bin, bugs, c, cache, completion, config,
    ddp, dedupe, deprecate, dist-tag, docs, edit, explore, get,
    help, help-search, i, init, install, install-test, it, link,
    list, ln, login, logout, ls, outdated, owner, pack, ping,
    prefix, prune, publish, rb, rebuild, repo, restart, root,
    run, run-script, s, se, search, set, shrinkwrap, star,
    stars, start, stop, t, team, test, tst, un, uninstall,
    unpublish, unstar, up, update, v, version, view, whoami

npm <cmd> -h     quick help on <cmd>
npm -l           display full usage info
npm help <term>  search for help on <term>
npm help npm     involved overview

Specify configs in the ini-formatted file:
    /Users/gabrielcsapo/.npmrc
or on the command line via: npm <command> --key value
Config info can be viewed via: npm help config

npm@4.0.5 /Users/gabrielcsapo/.nvm/versions/node/v4.7.0/lib/node_modules/npm
```

now will read as 

```
Usage: npm <command>

where <command> is one of:
    access, adduser, bin, bugs, c, cache, completion, config,
    ddp, dedupe, deprecate, dist-tag, docs, doctor, edit,
    explore, get, help, help-search, i, init, install,
    install-test, it, link, list, ln, login, logout, ls,
    outdated, owner, pack, ping, prefix, prune, publish, rb,
    rebuild, repo, restart, root, run, run-script, s, se,
    search, set, shrinkwrap, star, stars, start, stop, t, team,
    test, tst, un, uninstall, unpublish, unstar, up, update, v,
    version, view, whoami

npm <command> -h     quick help on <command>
npm -l           display full usage info
npm help <term>  search for help on <term>
npm help npm     involved overview


npm commands available from elsewhere on your $PATH
    check, what

Specify configs in the ini-formatted file:
    /Users/gabrielcsapo/.npmrc
or on the command line via: npm <command> --key value
Config info can be viewed via: npm help config

npm@5.1.0 /Users/gabrielcsapo/Documents/npm
```

This format is comparable with the `git --help -a` response

```
usage: git [--version] [--help] [-C <path>] [-c name=value]
           [--exec-path[=<path>]] [--html-path] [--man-path] [--info-path]
           [-p | --paginate | --no-pager] [--no-replace-objects] [--bare]
           [--git-dir=<path>] [--work-tree=<path>] [--namespace=<name>]
           <command> [<args>]

available git commands in '/Applications/Xcode.app/Contents/Developer/usr/libexec/git-core'

  add                       merge-octopus
 *** 

git commands available from elsewhere on your $PATH

  unstaged

'git help -a' and 'git help -g' list available subcommands and some
concept guides. See 'git help <command>' or 'git help <concept>'
to read about a specific subcommand or concept.
```

out

I am looking for feedback before completing the rest of the implementation as well as edge cases and tests, if this is something that is wanted for `npm`.

### DIFF
--- a/lib/config/cmd-elsewhere.js
+++ b/lib/config/cmd-elsewhere.js
@@ -1,0 +1,26 @@
+var path = require('path')
+var fs = require('fs');
+var log = require('npmlog')
+
+var npmconf = require('./core.js')
+var cmdElseWhere = {}
+
+Object.defineProperty(cmdElseWhere, 'cmdList', {
+    get: function get() {
+        try {
+            var bin = path.resolve(npmconf.defaults.prefix, 'bin');
+            var commands = fs.readdirSync(bin).filter(function(name) {
+                return name.substring(0, 4) == 'npm-'
+            }).map(function(name) {
+                return name.substring(4, name.length)
+            })
+            return commands;
+        } catch(ex) {
+            // silently ignore error
+            log.error('cmdElseWhere', ex.toString());
+            return [];
+        }
+    }
+});
+
+module.exports = cmdElseWhere
--- a/lib/help.js
+++ b/lib/help.js
@@ -12,6 +12,7 @@ var npm = require('./npm.js')
 var log = require('npmlog')
 var opener = require('opener')
 var glob = require('glob')
+var cmdElseWhere = require('./config/cmd-elsewhere').cmdList
 var cmdList = require('./config/cmd-list').cmdList
 var shorthands = require('./config/cmd-list').shorthands
 var commands = cmdList.concat(Object.keys(shorthands))
@@ -173,6 +174,10 @@ function npmUsage (valid, cb) {
     'npm -l           display full usage info',
     'npm help <term>  search for help on <term>',
     'npm help npm     involved overview',
+    '',
+    '',
+    'npm commands available from elsewhere on your $PATH',
+    '    ' + wrap(cmdElseWhere),
     '',
     'Specify configs in the ini-formatted file:',
     '    ' + npm.config.get('userconfig'),
--- a/lib/npm.js
+++ b/lib/npm.js
@@ -37,9 +37,11 @@
   var aliases = require('./config/cmd-list').aliases
   var cmdList = require('./config/cmd-list').cmdList
   var plumbing = require('./config/cmd-list').plumbing
+  var cmdElseWhere = require('./config/cmd-elsewhere').cmdList
   var output = require('./utils/output.js')
   var startMetrics = require('./utils/metrics.js').start
   var perf = require('./utils/perf.js')
+  var spawn = require('child_process').spawn
 
   perf.emit('time', 'npm')
   perf.on('timing', function (name, finished) {
@@ -172,6 +174,7 @@
   }
 
   npm.deref = function (c) {
+    if (cmdElseWhere.indexOf(c) !== -1) return c
     if (!c) return ''
     if (c.match(/[A-Z]/)) {
       c = c.replace(/([A-Z])/g, function (m) {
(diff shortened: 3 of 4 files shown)


nyteshade:
👍 I can see how this would be quite useful.

indutny:
I love the idea!

ammark47:
This is perfect 👍

dmikey:
👍

zkat:
As a general thing, the npm CLI has always been designed such that this sort of user-based extension isn't allowed, and that's been an _intentional_ choice. Keeping commands limited makes it easier to maintain compatibility in the future -- right now, adding new npm commands is only a `semver-minor` bump. If we were to add this patch, every single time we add a new command _or even an alias_, we would need to release a `semver-major` version of npm, since it's entirely possible we'll be seriously breaking someone's build. I think this applies regardless of whether the semantics of a feature like this are "always shadow existing commands" or "existing command always trumps custom one".

As an alternative, I'd like to point out that I'm currently working (today!) on getting [npx](https://npm.im/npx) integrated into npm, and included as one of the binaries in our standard distribution. I think npx is a better overall solution to this problem: it's more generalized, it allows more flexibility about how you install these things, and most importantly, it doesn't turn otherwise-benign semver-minor bumps into semver-major changes.

Going by your example, anyone using `npm@5.2.0` or later (or, anyone who's installed the standalone `npx` straight from the registry) will be able to do:

```
$ npx npm-check -u
...
```

And this will work in all the following cases:
1. If you've installed `npm-check` _locally in your current project_, it will use _that_ version.
2. If you've installed `npm-check` globally, it will use that one.
3. If you don't have `npm-check` installed _at all_, npx will install it into a temporary location and do a one-off run: you don't even have to explicitly install anything.

What are your thoughts? I think your PR addresses a very real need, but I also want to make sure we don't end up breaking other things in the process. Thank you so much, by the way, for putting this patch together and helping us have this conversation. <3

gabrielcsapo:
Thank you for responding and I am excited for `npx` to be merged into npm core. A lot of good solutions that I could see positively effecting the experience of using npm. 

While the core functionality of NPM is important to many people, I find the use cases that many people would like to add small utility functions to it. 

- https://www.npmjs.com/package/npmr
- https://www.npmjs.com/package/npm-what
- https://www.npmjs.com/package/npm-check
- https://www.npmjs.com/package/npm-name
- https://www.npmjs.com/package/npm-user
- https://www.npmjs.com/package/npm-run-path

Just a few examples of some `npm-${extra}` modules I found on the registry. 

In the use case of [yarn](https://yarnpkg.com/en/) the only reason that it was created in my opinion was the lack of extensibility to NPM, so instead of creating a module that modified the behavior of install, an entire module was born to serve the same purpose of `npm`. 

While I by no means think this PR solves all of the problems, it needs more work in the use case of running in context and adding extensibility apis, it might help users feel more inclined to creating modules that coexist with `npm` rather than work against it. 

To talk to the point of `npm` having to do major bumps I think a contract could be created with anyone using add on modules, to understand that minor bumps could cause issues or potential collisions. I was talking to @iarna about this situation and suggested that running global commands that do have collisions as `npm install --overrideGlobal` could be a potential solution for this problem. Since the user is accepting the override of a default function to `npm`.

gabrielcsapo:
@zkat with the release of npx as a npm runner, this functionality seems unlikely to be merged in, right?

zkat:
@gabrielcsapo Yeah I'm not terribly inclined to merge it. I don't think a "contract with anyone who creates addon modules" is a strong enough guarantee to head off problems in the future, and I don't like the fallback semantics that could get us into. I don't think an option like `--overrideGlobal` is that great an idea, either.

My strong preference is for npx and other such tools to facilitate other users to do their own extension. If what you want is an extra command, `npx my-command` vs `npm my-command` is literally the same thing, except the former is _designed_ for this sort of thing. npm users going forward are guaranteed to have `npx` installed now, after all.

What do you think?

gabrielcsapo:
Makes sense, thanks for reviewing the change 😊

zkat:
@gabrielcsapo and thanks for submitting a detailed, thought-out proposal!

bitmoji

Issue #6053, opened 2014-08-28
standalone npm-exec


othiym23:
Add a binary to `npm` called `npm-exec` that does basically what `npm run-script` does, but without the `nopt` / `npmconf` command-line parsing of `npm` itself. Improves the situation described in #3494.

garrett-gottlieb:
+1

rajsite:
+1

zkat:
For the sake of getting people's feelings about it -- do you think instead of something specifically like `npm-exec` (or _only_ having `npm-exec`), that something like `virtualenv`'s `activate` functionality would be desirable in general? Would that fulfill the needs this is meant to fulfill any better?

For those unfamiliar, I would imagine it to work something like this:

```sh
$ cd path/to/proj
$ npm i standard
$ npm activate
(proj)$ standard
(proj)$ npm deactivate
$ standard
command not found: standard
```

es128:
@zkat I don't see how your proposal would help with the type of use cases most people need this feature for. How would anything be accomplished with a single shell command? Something like `npm activate && standard` wouldn't work and is no improvement over the `npm-exec` ideas anyway.

zkat:
There's two workflows you're talking about here: uncommon, quick, one-off commands (that `npm-exec` shines in), and long-term, project-level workflows where it's slightly more verbose (where `activate` might be more helpful).

At least I remember spending quite a bit of time helping people make sure their `grunt`, `webpack`, etc commands were properly installed, globally, because `$(npm bin)/grunt` was kinda verbose and ugly to be typing constantly.

es128:
I was referring to both cases. Don't see how `npm activate` helps any more than (or even as much as) a bundled `npm-exec`. It's kind of a style preference, I guess? Perhaps you can elaborate further on what benefits you see with your idea.

bcomnes:
> npm activate
  npm deactivate

I'm opposed to replicating virtual env because getting away from global dependency hacks is what makes npm totally great!  I know this wouldn't bring global deps, but it would affect global shell state.  It would be better if we can keep things contained inside project folders, including shell command context.  I'm worried this would allow python devs coming to to node to be lazy and not learn why npm scripts, local by default and recursive dependencies are more reliable (and way less annoying) than virtual environments.  E.g. Npm activate then run this series of 5 different commands vs `npm i ; npm run build`

[...]

PR #17685, opened 2017-07-08
npx: bundle npx with npm itself

zkat:
npx is an npm package runner with a bunch of nice features. While
it is also available on the npm registry as `npx`, it's the sort
of thing that's better off being bundles with npm itself.

Fixes: #6053


rmg:
TL;DR - think very long and hard about committing to a 2nd bin in a package.

Have you considered making this a sub-command (maybe `npm x`/`npm exec`) instead of a separate bin? We found having multiple bins in a CLI package to be a support nightmare, especially when there was overlap with another package's bin.

The window would, admittedly, be short for people using both `npm` _and_ the separate `npx` on the same system, but it will definitely be a pain as long as people are juggling multiple versions of `npm` and/or `npx`.

Perhaps this behaviour has changed since v2 when I last fought with this, but at the time `npm i -g npmc` would have aborted if `npx` was _already_ installed. You would actually have to uninstall it first. And if you just deleted the conflicting bin and installed `npmc`, then `npx` would sometimes _fail_ to uninstall because the `npx` bin it finds doesn't "belong" to the `npx` package.

This caused a lot of support pain for the early days at StrongLoop when we made similar changes to our CLI at the time - npm has many many orders of magnitude more users. I don't want to see y'all face a repeat of all the 😢/😠  we saw!

zkat:
@rmg I've pushed up a patch that we should probably have in some form _anyway_, but at the very least it means that if installing `npx` fails, `npm x ...` will still work, and run basically identical code: the npx CLI will completely take over the current npm process, and do this way before we do any option parsing whatsoever.

I kinda like having `npx` around anyway, if only to have a consistent thing that even folks using npx without npm can call, people can put in READMEs, etc.

I think _in general_ it's not too worrisome for an `npx` binary to get installed from now on. Last time I looked (when originally making npx), there's no `npx` in any linux distro I checked, or on brew. It seems to be a pretty available name right now. I don't mind snatching it up 😏

DomT4:
> or on brew

Homebrew's `node` formula would need some reasonably minor rewriting to actually put the `npx` executable & manpage in the right place _if_ I'm reading this PR correctly, but I cannot imagine that being a major problem.

rmg:
@zkat it's not so much the name squatting aspect but the change in number/names of bins across upgrade/downgrade/install/uninstall. Obviously removing one is semver major. Adding one _also_ affects compatibility with an overall system and might be interpreted by some as semver major as well. I'm not sure about semver-major, but it is definitely a little more disruptive than adding a function to an API due to the essentially global namespace. Kinda like adding a method to the global scope in a browser library I suppose? 🤔

This is about when I would split the PR and land one with `npm x` only and move the discussion to a new PR that only adds the `npx` bin.

Ultimately we'll all live with whatever decision gets made :-)

zkat:
@rmg That's a really good point. I think it didn't really sink in what the experience would be switching back to pre-npx versions of npm.

The fact that it was so easy to integrate directly into the npm binary (using the same process-replacement trick that npx itself does) makes me agree with you more that maybe we should just not have `npx` as a binary. It means I'd be able to keep using the `npx` binary on the registry to keep people more up-to-date, or put up experimental features over there -- specially since breaking changes to npx potentially become semver-major changes to npm itself, and we need to be more careful about semver-majors overall.

junosuarez:
One added benefit of the `npm exec` approach is more continuity and overlap across ecosystems: in particular, for ruby developers, `bundle exec` behaves somewhat similarly and would be easy to explain to a developer who is familiar with one and learning the other, or to someone who is learning both ruby and node at the same time. Reduced cognitive load is a benefit for humans.

iarna:
@rmg Can you be more specific about the places you ran into these problems? There are technical challenges to merging npx into npm that we haven't yet fully addressed. And beyond that, we have the feeling that npm already has too many verbs. I'd like to see npm ship with a suite of related, simpler, command lines instead of just getting heavier and heavier weight.

zkat:
Recording a conclusion from our meeting today:

* We're going to start off by shipping `npx` as a standalone binary alongside npm, as planned. We understand the issues and are looking into ways to mitigate some of the upgrade/downgrade problems, but this is also one step towards doing more of this sort of thing in the future. We _want_ to be able to ship more than just the npm binary going forward.

* Integrating npx as `npm x`/`npm exec` requires a lot of work to iron out user experience issues, and that happens to involve refactoring a particularly ancient part of npm to get right. This is work we want to do, but it's not what we're going to focus on just yet. We'd rather ship something sooner rather than later. Specifically, we need to have more control over the way npm parses arguments and config files, and the way it builds the current "god object" we've been meaning to get rid of for a long time. This isn't an afternoon's worth of work (we think).

zkat:
Integration notes for myself, when putting this up for Node:

* This needs updated for the Windows installer to work: https://github.com/nodejs/node/blob/master/tools/msvs/msi/product.wxs#L210-L216
* And this is for the python-based *nix installer: https://github.com/nodejs/node/blob/23498f259db2d26c30969c0f66a380adf8725cd9/tools/install.py#L92-L99

iarna (approved):
Looks good!

rmg:
@iarna the best publicly visible example is https://github.com/strongloop/strongloop/issues/166 and it happens to show the actual errors I mentioned earlier about refusing to clobber files. This _was_ 3 full years ago, though, so it's possible the behaviour that caused trouble there has changed (which I would consider a regression if it now silently clobbers bins).

iarna:
@rmg Thank you, that's quite illustrative! If I remember correctly silently clobbering bins on upgrade was a consciously made breaking change in npm@3 to support the `gulp`/`gulp-cli` use case.

If two of your transitive deps share a bin name they get placed in different locations in the tree so they don't step on each other so ordinarily this only happens with globals.

rmg:
@iarna wow, I had no idea.

(delay in response because I had to reinstall npm after doing `npm i -g @rmg/not-npm`)

zkat:
This was already merged and included in `npm@5.2.0`.

refack:
Cross ref: https://github.com/nodejs/node/issues/13640#issuecomment-315718778

PR #17691, opened 2017-07-09
Missing --no-save flag for uninstall command?

timneedham:
Hi,

I'm using NPM V5.1.0 and could really have done knowing about the `--no-save` flag!


zkat (approved):
lgtm

Issue #17702, opened 2017-07-10
cache folder has wrong owner when using sudo


mohd-akram:
#### I'm opening this issue because:

  - [x] npm is doing something I don't understand.

#### What's going wrong?

When using sudo, npm creates `/root/.npm` owned by `SUDO_USER` instead of `root`.

#### How can the CLI team reproduce the problem?

`sudo npm install <package>`

### supporting information:

 - `npm -v` prints: 5.0.4
 - `node -v` prints: v8.1.3
 - `npm config get registry` prints: https://registry.npmjs.org/
 - Windows, OS X/macOS, or Linux?: FreeBSD 11

xcambar:
I can confirm this, as per the referenced ansible issue above.

Scriptkiddi:
also it does not respect the user when using sudo -u other_user

PR #17705, opened 2017-07-10
Remove special sudo handling

mohd-akram:
The special handling of commands called with `sudo` causes weird behavior (#17702) that is inconsistent with other applications. I also could not find any documentation mentioning this, leading to a lot of head-scratching. The fact that this behavior is not matched when using `su` adds to the confusion.


zkat:
See https://docs.npmjs.com/misc/config#unsafe-perm

The other reason besides scripts is that, if we detect you're in sudo mode, we don't want to write to the shared cache as root -- this was a source of a lot of confusion back in the npm@1 days, which is why this is around, and the concern is still there: users often run npm as sudo when, for example, they do a global installation. This would then cause them to have a bunch of perms errors when doing normal installs with a now-mixed cache.

If there's specific issues happening with this SUDO_UID dance, I think those are worth fixing (and some of them have been fixed recently!). It's likely that if you're seeing more issues surrounding this recently, it's because a lot of the cache and extraction-related code was rewritten and some of the old semantics may not have been duplicated as accurately as I'd hoped.

So, I'm not gonna take this PR or anything else that outright removes this feature from npm, as much as I understand that it's surprising to some people who have expectations around how *nix tools behave. In the end, it turned out to be a worse user experience _without_ this.

mohd-akram:
`unsafe-perm` is a separate option for scripts and addresses `root` safety, which is fine. It also behaves the same regardless of `su` or `sudo` from what I can tell. However, there is no mention in the documentation of the other subtle differences between these two.

Why use a shared cache to begin with? Just treat `root` as another user with their own cache dir. This already happens when using `su`. I don't see what problems changing behavior *just* when using `sudo` solves. From what I can tell, the only reason npm would have the bugs you mention is *because* these `sudo` checks were not being applied uniformly (which is still the case as linked in my pull request).

It seems like the previous assumptions that required this code are no longer true and need to be re-examined.

mtibben:
I agree @mohd-akram, I find the special-casing of root and sudo in npm to be confusing, inconsistent and buggy.

And I don't understand the use-case behind it - I have no doubt this was put in with good intentions, but it would be great to have a full explanation, rationale and implementation docs. Why use a shared cache? Why do we even want to run scripts as different users? Why special-case sudo? Why special-case root?

It seems to me npm is just being a bit too clever here, leading to complexity and bugs.

> as much as I understand that it's surprising to some people who have expectations around how *nix tools behave. In the end, it turned out to be a worse user experience without this.

Yes, it is surprising - but what exactly is the worse user experience without this?

PR #17706, opened 2017-07-10
Use semver to compare node versions in npm doctor

leo-shopify:
`npm doctor` suggests to upgrade the version of node if it is outdated.
The suggestion is to use the latest LTS version.
Comparing versions was done by comparing strings.
A lexicografical comparison can give incorrect results. e.g.:
'v6.10.0' > 'v6.9.0' => false

To correct the issue the comparison is done using `semver` which is a dependency `npm` aleady includes.


zkat (approved):
LGTM!

PR #17728, opened 2017-07-11
Document semver git urls in package.json docs.

sankethkatta:
Add documentation to `package.json` docs to match [`npm install` docs](https://docs.npmjs.com/cli/install) describing semver git urls.


zkat (approved):
🐑

Issue #17701, opened 2017-07-10
npm install crashes if a devDependency package name contains a ')'


jdiehl:
#### I'm opening this issue because:

  - npm is crashing.

#### What's going wrong?

Since npm 5.1.0 when a devDependecy contains a ')' character, `npm install`crashes with the following error message: `npm ERR! Invalid regular expression: /^[package-name-containing-)]@/: Unmatched ')'`.

This error does not occur with npm 5.0.4.

#### How can the CLI team reproduce the problem?

1. Create an empty npm project: `npm init -y`
2. Add a dev dependency that contains a ')' in the name: `npm install --save-dev "@-)/utils"`
3. Run: `npm install`

### supporting information:

 - `npm -v` prints: 5.1.0
 - `node -v` prints: v8.1.2
 - `npm config get registry` prints: https://registry.npmjs.org/
 - Windows, OS X/macOS, or Linux?: macOS
 - Network issues:
   - Geographic location where npm was run: None
 - Container: None

jdiehl:
Still happening with nom 5.2.0

jdiehl:
```
Invalid regular expression: /^@-)/utils@/: Unmatched ')'

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/jdiehl/.npm/_logs/2017-07-11T20_08_29_151Z-debug.log
```

jdiehl:
npm-debug.log:
```
0 info it worked if it ends with ok
1 verbose cli [ '/usr/local/Cellar/node/8.1.3/bin/node',
1 verbose cli   '/usr/local/bin/npm',
1 verbose cli   'install' ]
2 info using npm@5.2.0
3 info using node@v8.1.3
4 verbose npm-session bc523da8bfd4693f
5 silly install runPreinstallTopLevelLifecycles
6 silly preinstall x@1.0.0
7 info lifecycle x@1.0.0~preinstall: x@1.0.0
8 silly install loadCurrentTree
9 silly install readLocalPackageData
10 silly install loadIdealTree
11 silly install cloneCurrentTreeToIdealTree
12 silly install loadShrinkwrap
13 silly install loadAllDepsIntoIdealTree
14 verbose stack SyntaxError: Invalid regular expression: /^@-)/utils@/: Unmatched ')'
14 verbose stack     at RegExp (<anonymous>)
14 verbose stack     at doesChildVersionMatch (/usr/local/lib/node_modules/npm/lib/install/deps.js:66:80)
[...]

PR #17729, opened 2017-07-11
Fix for #17701 (install devDeps not working if organisation contains regex-unsafe characters)

jdiehl:
Replaced regular expression with simple string manipulation to avoid module names that contain unsafe characters (possible with an organization) to break things


zkat:

PR #17733, opened 2017-07-11
action: Split remove and unbuild actions into two

iarna:
Removing modules when there are transitive symlinks requires that it be
deepest first, working our way out.

Running uninstall lifecycles needs to be shallowest first working our way
in.

Previously these were all wrapped up in `lib/unbuild` and we couldn't win.
This splits them apart. This wasn't previously an issue because we had
no facility for creating transitive dependencies as symlinks.

This should also fix our last outstanding travis failure.


zkat (approved):
LGTM! 🐑

lib/install.js
```diff
@@ -520,7 +520,8 @@ Installer.prototype.executeActions = function (cb) {
     [lock, node_modules, '.staging'],
     [rimraf, staging],
     [doParallelActions, 'extract', staging, todo, cg.newGroup('extract', 100)],
-    [doReverseSerialActions, 'remove', staging, todo, cg.newGroup('remove')],
+    [doReverseSerialActions, 'unbuild', staging, todo, cg.newGroup('unbuild')],
+    [doSerialActions, 'remove', staging, todo, cg.newGroup('remove')],
```
fengmk2, line 524: This change will remove all exists packages on `node_modules` before install action start.

PR #17740, opened 2017-07-12
tar: ignore Vim backup files (Fix #17560)

### LINKED ISSUE #17560: Impossible to ignore README.md~
I just published my first package and I was surprised to see my `README.md~` Vim backup file was published even though there was a `*~` line in my `.npmignore`. I then noticed that the docs (https://docs.npmjs.com/misc/developers) say that README "and its variants" are never ignored, and I can see that making sense for README.md, README.txt etc, but I can't see any downside to ignoring README.md~. Would it be a bother to make this ignorable?

### supporting information:

 - `npm -v` prints:
3.10.8

 - `node -v` prints:
v7.0.0

https://registry.npmjs.org/

 - Windows, OS X/macOS, or Linux?:
Linux
 
- Network issues:
Australia

Gerhut:
In order to ignore Vim backup of README.md, move ignore rule in front of
READMEs condition.

### DIFF
--- a/lib/utils/tar.js
+++ b/lib/utils/tar.js
@@ -63,6 +63,29 @@ function BundledPacker (props) {
 inherits(BundledPacker, Packer)
 
 BundledPacker.prototype.applyIgnores = function (entry, partial, entryObj) {
+  // some files are *never* allowed under any circumstances
+  // (VCS folders, native build cruft, npm cruft, regular cruft)
+  if (entry === '.git' ||
+      entry === 'CVS' ||
+      entry === '.svn' ||
+      entry === '.hg' ||
+      entry === '.lock-wscript' ||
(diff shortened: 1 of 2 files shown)


zkat:
I'm hesitant to merge this right now, so we might wait a bit and this might need to be rewritten: for one, this needs a corresponding change in `fstream-npm` (because of horrible reasons), and for two, #16804 is about to completely replace this code, and it's gonna land ~soon.

iarna:
Ok, so as @zkat mentioned this is going to be need to be rewritten as we've eliminated `fstream` entirely from `npm`. This logic is now implemented in [npm-packlist](https://github.com/npm/npm-packlist), specifically [here](https://github.com/npm/npm-packlist/blob/master/index.js#L20).

I'm going to close this PR, but I hope you'll open it again over on `npm-packlist`.

Thanks!

PR #17742, opened 2017-07-12
Fixed lockfiles not being added to the "npm version" commit

### LINKED ISSUE #17707: package-lock modified but not committed to git repo during "npm version [major|minor|patch]

  - [ ] npm is crashing.

#### What's going wrong?

When using "npm version [major | minor | patch]" on a project in a git repo, npm will bump the version then commit the package.json changes to git.  However, package-lock.json also gets updated with the bumped version but is not included in the commit.  This leaves the git repo with unstaged changes after a version bump, requiring a separate commit or amend.

Create a node project inside a git repo.  Install a dependency to cause the creation of a package-lock.json.  Run "npm version patch", and observe the creation of a new git commit, and the presence of package-lock as an unstaged change under "git status".

### supporting information:

 - `npm -v` prints: 5.1.0
 - `node -v` prints: 6.9.1
 - `npm config get registry` prints: (internal corp proxy registry)
 - Windows, OS X/macOS, or Linux?: Windows 10

markpeterfejes:
Hey!

This small PR fixes #17707 by executing the `git add` commands sequentially instead of concurrently, which sometimes causes the aforementioned error.

I'm not really sure that this is the prettiest way to implement this, change suggestions are very welcome!

Thank you!

### DIFF
--- a/lib/version.js
+++ b/lib/version.js
@@ -297,16 +297,12 @@ function _commit (version, localData, cb) {
 }
 
 function stagePackageFiles (localData, options) {
-  return new BB(function (resolve, reject) {
-    addLocalFile('package.json', options, false).then(() => {
-      if (localData.hasShrinkwrap) {
-        addLocalFile('npm-shrinkwrap.json', options, false).then(resolve, reject)
-      } else if (localData.hasPackageLock) {
-        addLocalFile('package-lock.json', options, false).then(resolve, reject)
-      } else {
-        resolve()
-      }
-    })
+  return addLocalFile('package.json', options, false).then(() => {
+    if (localData.hasShrinkwrap) {
+      return addLocalFile('npm-shrinkwrap.json', options, false)
+    } else if (localData.hasPackageLock) {
+      return addLocalFile('package-lock.json', options, false)
+    }
   })
 }
 
--- a/lib/version.js
+++ b/lib/version.js
@@ -294,19 +294,16 @@ function _commit (version, localData, cb) {
       ], options)
     }
   }).nodeify(cb)
-
 }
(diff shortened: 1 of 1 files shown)


lib/version.js
```diff
@@ -282,25 +282,34 @@ function _commit (version, localData, cb) {
...
         'tag', npm.config.get('tag-version-prefix') + version,
-        flag, message
+        flagForTag, message
       ], options)
     }
   }).nodeify(cb)
 }
 
+function stagePackageFiles (localData, options) {
+  return new BB(function (resolve, reject) {
```
zkat: You don't need a promise constructor here. Just:

```
return addLocalFile('package.json', options, false).then(() => {
  if (localData.hasShrinkwrap) {
    return addLocalFile('npm-shrinkwrap.json', options, false)
  } else if (localData.hasPackageLock) {
    return addLocalFile('package-lock.json', options, false)
  }
})
```

zkat (changes requested):
needs a small mostly-stylistic change

lib/version.js
markpeterfejes: I'm doing this right now! Is there a possibility that this could be merged to the 5.3.0 relese today?

zkat: probably! I'll take a look in the AM and see if this patch is ready to merge. 👍

markpeterfejes: Thank you very much!

markpeterfejes:
Should be ready for release :)

zkat (approved):
changes look good, everything seems to work. gj, and thanks for the contrib! 🎉

PR #17748, opened 2017-07-12
finalize: When rolling back use symlink project-relative path

iarna:
Previously it was using the realpath (the path w/o any symlinks in it) which
proceeded to confusing `fs-vacuum` as that path wasn't necessarily in the
base path (our module).

(Partially)
Fixes: #17637


zkat (approved):
🐑

zkat:
bitmoji

PR #17777, opened 2017-07-14
decompose-actions: Update should remove the OLD pkg, add the new

iarna:
This is a fix for 7f28a77f3. Things worked prior to 7f28a77f3 because previously we were running `npm.commands.unbuild` and that was reading the `package.json` off disk on its own.

### DIFF
--- a/lib/install/decompose-actions.js
+++ b/lib/install/decompose-actions.js
@@ -46,7 +46,7 @@ function addSteps (decomposed, pkg, done) {
 }
 
 function updateSteps (decomposed, pkg, done) {
-  removeSteps(decomposed, pkg, () => {
+  removeSteps(decomposed, pkg.oldPkg, () => {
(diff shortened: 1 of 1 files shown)


lib/install/decompose-actions.js
```diff
@@ -46,7 +46,7 @@ function addSteps (decomposed, pkg, done) {
 }
 
 function updateSteps (decomposed, pkg, done) {
-  removeSteps(decomposed, pkg, () => {
+  removeSteps(decomposed, pkg.oldPkg, () => {
```
zkat, line 49: whaaaat. OwO what's this.

zkat: ah. Got it. We were reading the pkgjson magically. Gotcha.

zkat (approved):
ok

PR #17785, opened 2017-07-14
updating naming convention for certificates

lgp1985:
where it states PEM included Windows naming as  (Windows calls it "Base-64 encoded X.509 (.CER)")


zkat (approved):
This seems pretty handy! Thank you!

PR #17810, opened 2017-07-17
feat: allow custom shell opts for run-script

madarche:
Usage:
```shell
$ npm config set script-shell /bin/bash
$ npm config set script-shell-opts "-o pipefail -c"
```

This PR is related to #16687 

The aim of this PR is to be able to specify custom shell options such as "-o pipefail" for bash or any other shell, so that npm can return error on pipe-chained commands which are very common when one uses npm as a build tool with lots of npm scripts.

### Example

```json
"scripts": {
    "build:js": "browserify -g envify -e www/js/main.js | uglifyjs -c -m | tee www/bundle/bundle.js | gzip > www/bundle/bundle.js.gz",
  }
```

The above `browserify` command may fail, but the other piped commands (`uglifyjs`, `tee`, `gzip`) won't and thus the `build:js` script won't fail as a whole and a CI will consider the `npm run build:js` a success and will proceed to the next action.

If `npm config set script-shell-opts "-o pipefail -c"` was supported, the `build:js` npm script will produce an error and the CI will fail, providing a much safer situation.

### Reference

And for information here is [another reference of `pipefail` usage on stackoverflow](https://stackoverflow.com/questions/4472318/error-handling-with-chained-commands-pipes-in-a-bash-script).

Cheers


lib/utils/lifecycle.js
```diff
@@ -265,9 +265,11 @@ function runCmd_ (cmd, pkg, env, wd, stage, unsafe, uid, gid, cb_) {
   var shFlag = '-c'
 
   var customShell = npm.config.get('script-shell')
+  var customShellOpts = npm.config.get('script-shell-opts')
 
   if (customShell) {
     sh = customShell
+    shFlag += ' ' + customShellOpts
```
mmkal: shouldn't this just be `shFlag = customShellOpts`? It's conceivable that you wouldn't want `-c` used at all.

madarche:
@mmkal thanks for your interest and your remark. Please tell me if you agree with the current state of this PR.

madarche:
PR rebased onto `latest`

danielbayley:
Would it not be better to simply allow `npm config set script-shell '/bin/bash -o pipefail -c'`?

madarche:
@danielbayley: Doing `npm config set script-shell '/bin/bash -o pipefail -c'` doesn't work (please try it yourself) because of the way the shell is spawned:
```javascript
var proc = spawn(sh, [shFlag, cmd], conf, opts.log)
```

That's why I've proposed this PR.

zkat:

Issue #17048, opened 2017-06-07
package-lock.json ordering is locale dependent


unilynx:
#### I'm opening this issue because:
  - [x] npm is doing something I don't understand.

#### What's going wrong?

The packages in package-lock.json are reordered depending on the locale of the user using npm. This is causing unnecessary updates to the package-lock.json file, which causes needless updates in versioning systems.

#### How can the CLI team reproduce the problem?
(this might depend on the default locale settings on OSX, 
```
export LANG=en_US.UTF-8
mkdir /tmp/bla
cd /tmp/bla
npm init
npm install string_decoder
npm install string-w-dth
LANG="en_US.UTF-8" LC_ALL="en_US.UTF-8" npm install
cat package-lock.json # string_decoder appears before string-width
LANG="" LC_ALL="" npm install
cat package-lock.json # string_decoder now appears after string-width
```

### supporting information:

 - `npm -v` prints:
5.0.3
 - `node -v` prints:
v8.0.0
 - `npm config get registry` prints:
https://registry.npmjs.org/
 - Windows, OS X/macOS, or Linux?:
macOS Sierra

paulmelero:
Same here. Cannot get the same package-lock modules order in my machine as my co-workers. But I do not have any `LANG` env variable.

* node -v
v6.9.4

* npm -v
5.0.4

* npm config get registry prints:
https://registry.npmjs.org/

* Win10 64

LotharSee:
Experienced exactly the same troubles. At work, different versions of OSX come with different default locales, so our lock file isn't stable between machines.

To reproduce the locale-dependent lock file, create a directory with a `package.json`.
Done with npm 5.3.0.

```json
{
  "name": "example",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "author": "",
  "license": "ISC",
  "dependencies": {
    "fsevents": "1.1.2"
  }
}
```

Then compare the `package-lock.json` with:

- `LANG="" LC_ALL="" npm i`
- `LANG="en_US.UTF-8" LC_ALL="en_US.UTF-8" npm i`

They put `string_decoder` and `string-width` in a different order in `["dependencies"]["fsevents"]["dependencies"]`.



Also opened a PR to use a locale-agnostic sort instead. https://github.com/npm/npm/pull/17844

PR #17844, opened 2017-07-18
Make package-lock.json sorting locale-agnostic

LotharSee:
### Summary

Make `package-lock.json` sorting of its dependencies locale-agnostic.
Use `sort()` on the module names without providing a custom `compareFunction`. This is Unicode-based and locale-agnostic.

Fix https://github.com/npm/npm/issues/17048

### How to reproduce

Reason of the trouble: 

```bash
$ LANG='' node -e 'console.log("string_decoder".localeCompare("string-width"))'
1
$ LANG='en' node -e 'console.log("string_decoder".localeCompare("string-width"))'
-1
```

To reproduce the locale-dependent lock file, create a directory with a `package.json`.
Done with npm 5.3.0.

```json
{
  "name": "example",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "author": "",
  "license": "ISC",
  "dependencies": {
    "fsevents": "1.1.2"
  }
}
```

Then compare the `package-lock.json` with:

- `LANG="" LC_ALL="" npm i`
- `LANG="en_US.UTF-8" LC_ALL="en_US.UTF-8" npm i`

They put `string_decoder` and `string-width` in a different order in `["dependencies"]["fsevents"]["dependencies"]`.

### Implications

- It makes the package-lock more deterministic and no longer depends on user's locale, fixing #17048 
- It means it will potentially reorder dependencies of existing `package-lock.json` if user's locale was making `localeCompare` behave differently than the default `sort`. That would happen only once and for all but would need to properly appear in the changelog.

### DIFF
--- a/lib/shrinkwrap.js
+++ b/lib/shrinkwrap.js
@@ -101,7 +101,7 @@ function shrinkwrapDeps (deps, top, tree, seen) {
   if (!seen) seen = new Set()
   if (seen.has(tree)) return
   seen.add(tree)
-  tree.children.sort(function (aa, bb) { return moduleName(aa).localeCompare(moduleName(bb)) }).forEach(function (child) {
+  sortModules(tree.children).forEach(function (child) {
     if (child.fakeChild) {
       deps[moduleName(child)] = child.fakeChild
       return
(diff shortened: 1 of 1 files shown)


zkat (approved):
LGTM!

PR #17897, opened 2017-07-23
Add info about `files` and `npm pack`

davidjgoss:
Add advice about possibly using `files` in package.json instead of `.npmignore` and how to test whether your config's good by using `npm pack`.

Fixes npm/docs/#843


zkat (approved):
This seems fine to me. Thanks! 🐑

PR #17919, opened 2017-07-25
Fix `npm prune --production` when devDependencies contain bins

### LINKED ISSUE #17781: npm@5.3: npm prune --production fails

  - [ ] npm is crashing.

#### What's going wrong?
`npm prune --production` fails with `npm ERR! May not delete: /path/node_modules/.bin`
`NODE_ENV=production npm prune` fails with `npm ERR! May not delete: /path/node_modules/.bin`

Run `NODE_ENV=production npm prune` or  `npm prune --production` in any project which has a `node_modules/.bin` folder with npm@5.3.0

npm@5.2.0 is working as expected.

### supporting information:

 - `npm -v` prints: 5.3.0
 - `node -v` prints: v8.1.4
 - Windows, OS X/macOS, or Linux?: Linux 3.13.0-123-generic

fasterthanlime:
This closes #17781.

*Caveat: all that follows is based on my current understanding of npm after one morning's exploration into it, feel free to point out mistakes*

## The bug

When pruning, `unbuild.rmStuff` ends up being called.

Several async calls deeper, the `gentlyRmBinRoot` closure gets called (if removing the shims went fine) in `rmBins`:

```javascript
function rmBins (pkg, folder, parent, top, cb) {
  if (!pkg.bin) return cb()
  var binRoot = top ? npm.bin : path.resolve(parent, '.bin')
  asyncMap(Object.keys(pkg.bin), function (b, cb) {
    if (process.platform === 'win32') {
      chain([ [gentlyRm, path.resolve(binRoot, b) + '.cmd', true, folder],
              [gentlyRm, path.resolve(binRoot, b), true, folder] ], cb)
    } else {
      gentlyRm(path.resolve(binRoot, b), true, folder, cb)
    }
  }, gentlyRmBinRoot)

  function gentlyRmBinRoot (err) {
    if (err || top) return cb(err)
    return gentlyRm(binRoot, true, parent, cb)
  }
}
```

If `top` is true, `gentlyRmBinRoot` will not attempt to gentlyRm `binRoot`, which I believe is what *should* happen. However, it doesn't, because back in `rmStuff`, we do this:

```javascript
  var parent = pkg.name[0] === '@' ? path.dirname(path.dirname(folder)) : path.dirname(folder)
  var gnm = npm.dir
  var top = gnm === parent
```

and top ends up being false, because (when running [this new test case](https://github.com/fasterthanlime/npm/blob/gh-17781-prune/test/tap/prune-dev-dep-with-bins.js) on windows, at least), `gnm` is the absolute path `C:\Dev\npm\node_modules` whereas `parent` is the relative path `node_modules`.

So, even though they're the same folder, `top` is false, `gentlyRmBinRoot` attempts to remove `node_modules/.bin`, which `gentlyRm` forbids (by checking for 6-ish known prefixes), that throws an exception and the whole pruning is aborted.

## A fix

Here's the crux of my fix:

```javascript
 // in rmStuff:
 var top = '' === path.relative(gnm, parent)
```

There's comments above this line, because it might not be well-known that `path.relative` returns an empty string if they're the same directory. This method works if either path is relative or absolute (doesn't matter which).

I'm not aware of a better method in node.js, but feel free to use another one if it fits the codebase better.

## A test

[This test](https://github.com/fasterthanlime/npm/find/gh-17781-prune) simulates what happens when following the [repro steps](https://github.com/npm/npm/issues/17781#issuecomment-317678191) I posted in the original issue.

It's a bit lenghty, because the shims need to be something [read-cmd-shim](https://www.npmjs.com/package/read-cmd-shim) can resolve to an in-package location.

*It might be possible to make this test shorter by just including the bare minimum read-cmd-shim is looking for, but imho it's not that problematic to keep it exactly lke what an npm@5 install would produce*

### DIFF
--- a/lib/unbuild.js
+++ b/lib/unbuild.js
@@ -60,7 +60,9 @@ function rmStuff (pkg, folder, cb) {
   // otherwise, then bins are in folder/../.bin
   var parent = pkg.name[0] === '@' ? path.dirname(path.dirname(folder)) : path.dirname(folder)
   var gnm = npm.dir
-  var top = gnm === parent
+  // gnm might be an absolute path, parent might be relative
+  // this checks they're the same directory regardless
+  var top = path.relative(gnm, parent) === ''
 
   log.verbose('unbuild rmStuff', pkg._id, 'from', gnm)
   if (!top) log.verbose('unbuild rmStuff', 'in', parent)
--- a/test/tap/prune-dev-dep-with-bins.js
+++ b/test/tap/prune-dev-dep-with-bins.js
@@ -1,0 +1,105 @@
+'use strict'
+var fs = require('fs')
+var path = require('path')
+var test = require('tap').test
+var Tacks = require('tacks')
+var File = Tacks.File
+var Dir = Tacks.Dir
+var common = require('../common-tap.js')
+var testdir = path.join(__dirname, path.basename(__filename, '.js'))
+
+var fixture = new Tacks(
+  Dir({
+    node_modules: Dir({
+      'yes': Dir({
+        'package.json': File({
+          _requested: {
+            rawSpec: 'file:///mods/yes'
+          },
+          dependencies: {},
+          bin: {
+            'yes': 'yes.js'
+          },
+          name: 'yes',
+          version: '1.0.0'
+        }),
+        'yes.js': File('while (true) { console.log("y") }')
+      }),
+      '.bin': Dir({
+        // verbose, but needed for `read-cmd-shim` to properly identify which
+        // package this belongs to
+        'yes': File(
+          '#!/bin/sh\n' +
+          'basedir=$(dirname "$(echo "$0" | sed -e \'s,\\\\,/,g\')")\n' +
+          '\n' +
+          'case `uname` in\n' +
+          '    *CYGWIN*) basedir=`cygpath -w "$basedir"`;;\n' +
+          'esac\n' +
+          '\n' +
+          'if [ -x "$basedir/node" ]; then\n' +
+          '  "$basedir/node"  "$basedir/../yes/yes.js" "$@"\n' +
+          '  ret=$?\n' +
+          'else\n' +
+          '  node  "$basedir/../yes/yes.js" "$@"\n' +
+          '  ret=$?\n' +
+          'fi\n' +
+          'exit $ret\n'),
+        'yes.cmd': File(
+          '@IF EXIST "%~dp0\node.exe" (\n' +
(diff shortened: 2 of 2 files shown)


fasterthanlime:
More notes: running the tests takes 40 minutes on my machine, and 26 of them were failing when I cloned master (v5.3.0), so I'm hoping Travis catches any regressions.

Also: I'm squashing my N commits now for easier merging - there's no real timeline we need to keep track of.

Edit: all squashed :shipit:

fasterthanlime:
All TravisCI pipelines passed except for the Node 8.x one, but it seems it passed all the tests anyway? Not sure why the exit code is 1 in the end: https://travis-ci.org/npm/npm/jobs/257263030

mgol:
@fasterthanlime It looks to me this is because of a `standard` lint check failure; apparently it's only invoked on Node 8:
```
$ standard
standard: Use JavaScript Standard Style (http://standardjs.com) 
  /home/travis/build/npm/npm/lib/unbuild.js:65:13: Expected literal to be on the right side of ===.
  /home/travis/build/npm/npm/test/tap/prune-dev-dep-with-bins.js:60:34: Unexpected trailing comma.
The command "standard" exited with 1.
```

I guess it proceeds anyway but remembers to fail the run in the end.

fasterthanlime:
Ah, thanks @mgol, that's definitely it!

Pushed a (squashed) fix 🌟

fasterthanlime:
(Just want to mention that I've read the 5.2.0...5.3.0 diff a bunch of times and still haven't figured out exactly what change introduced this bug - I'm guessing it's the splitting of remove and unbuild? But not familiar enough to understand exactly how the flow changed.)

(Also: maybe there's another fix that's just passing `pkg.realpath` somewhere above `rmStuff` instead of `pkg.path` but, again, I wouldn't know exactly where to look. Stacktraces were truncated to the last promise/event callback in my debugger so I pretty much had to grep/breakpoint my way to the interesting bits!)

zkat:
Pst, just a heads up -- I'll review this in a wee bit, but the CLI time has been either conferencing or away from the office for most of this week. Thanks for the contrib!

zkat (approved):
Good sleuthing, and thanks for the fix! 🎉

mikesherov:
Just wanted to drop in and say this test is super great :heart:

uLan08:
any updates on a release?

SimenB:
5.4.0 was released as `next` a few days ago, IIRC they let it cook for a week before it's `latest`.

https://github.com/npm/npm/commit/82d71c05b0402fb21c3db39dac5c58b3b3d3c64a

PR #18022, opened 2017-08-01
doc: clarify usage for the `files` field in `package.json`

xcambar:
Following up a Twitter discussion with @zkat about the `files` field in `package.json (https://twitter.com/xcambar/status/892066155615203329), here's a documentation PR that attempts to clarify what it does and when it is used.

Mostly taken from https://github.com/npm/npm/wiki/Files-and-Ignores#files + a bit or rewording.

### DIFF
--- a/doc/files/package.json.md
+++ b/doc/files/package.json.md
@@ -168,9 +168,12 @@ npm also sets a top-level "maintainers" field with your npm user info.
 
 ## files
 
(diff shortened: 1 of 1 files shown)


iarna:
This looks good. I've landed it as 1a420c90e in release-next and it'll go out today or tomorrow.

PR #18025, opened 2017-08-01
lifecycle: remove all local dependencies other than spawn

mikesherov:
**Ignore all but the last commit. The other commits are in a separate PR: https://github.com/npm/npm/pull/18073/files**

## Abstract

This PR extracts all of npm out of lifecycle.js, and prepares it for extraction to another repo (other than spawn.js, which will be copy-pasted for now).

## The Plan

0. Get https://github.com/npm/npm/pull/18073 merged.
1. Rebase this. Get this merged.
2. yank `lifecycle.js` out, other than the `runLifecycle` function, which will remain as a local adapter for the generic `lifecycle` function. Publish it as `npm/lifecycle`... taking the tests with it.
3. Pulling in `npm/lifecycle` and using it in npm proper.
4. Use #18073 to grab config and input it into `npm/lifecycle` in `cipm`.

Let me know if you have thoughts, questions, or concerns! Thanks again

PR #18073, opened 2017-08-04
`npm config list --json` support

mikesherov:
As per a conversation with @zkat, this is to support other CLI tools like `cipm` that will need to query npm for config and avoid the hassle of ini parsing the result.

Please let me know if you have any questions, and thanks for all the great work you do on npm!

I also took the time to clean up some of the config code while I was there, and learned a lot about npm's history. Changes annotated!

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -19,7 +19,7 @@ config.usage = usage(
   'npm config set <key> <value>' +
   '\nnpm config get [<key>]' +
   '\nnpm config delete <key>' +
-  '\nnpm config list' +
+  '\nnpm config list [--json]' +
   '\nnpm config edit' +
   '\nnpm set <key> <value>' +
   '\nnpm get [<key>]'
@@ -69,7 +69,7 @@ function config (args, cb) {
       return del(args[0], cb)
     case 'list':
     case 'ls':
-      return list(cb)
+      return npm.config.get('json') ? listJson(cb) : list(cb)
     case 'edit':
       return edit(cb)
     default:
@@ -175,6 +175,23 @@ function publicVar (k) {
 
 function getKeys (data) {
   return Object.keys(data).filter(publicVar).sort(sort)
+}
+
+function listJson (cb) {
+  const publicConf = npm.config.keys.reduce((publicConf, k) => {
+    var value = npm.config.get(k)
+
+    if (publicVar(k) &&
+        // argv is not really config, it's command config
+        k !== 'argv' &&
+        // logstream is a Stream, and would otherwise produce circular refs
+        k !== 'logstream') publicConf[k] = value
+
+    return publicConf
+  }, {})
+
+  output(JSON.stringify(publicConf, null, 2))
+  return cb()
 }
 
 function listFromSource (title, conf, long) {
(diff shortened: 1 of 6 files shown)


lib/config.js
```diff
@@ -57,12 +59,21 @@ config.completion = function (opts, cb) {
...
+      return set(args[0], args[1], cb)
+    case 'get':
+      return get(args[0], cb)
+    case 'delete':
+    case 'rm':
+    case 'del':
+      return del(args[0], cb)
+    case 'list':
+    case 'ls':
+      return npm.config.get('json') ? listJson(cb) : list(cb)
```
mikesherov, line 72: this is the real meat of this change.

lib/config.js
```diff
@@ -159,15 +170,49 @@ function sort (a, b) {
 }
 
 function publicVar (k) {
-  return !(k.charAt(0) === '_' ||
-           k.indexOf(':_') !== -1 ||
-           types[k] !== types[k])
```
mikesherov, line 164: `types[k] !== types[k]` is a hold over from when the type for `argv` was `NaN`. Cute trick to exclude `argv` but was removed a long long time ago, so this has done nothing for like 4 years except confuse.

lib/config.js
```diff
@@ -185,92 +230,22 @@ function list (cb) {
...
 
   // project config file
   var project = npm.config.sources.project
-  var pconf = project.data
-  var ppath = project.path
-  var pconfKeys = getKeys(pconf)
-  if (pconfKeys.length) {
-    msg += '; project config ' + ppath + '\n'
-    pconfKeys.forEach(function (k) {
-      var val = (k.charAt(0) === '_')
```
mikesherov, line 210: this was excluded from the consolidate `listFromSource` because `getKeys` already excludes private keys! Line was doing nothing.

mikesherov:
Build fails because trusty doesn't support couchdb yet. Will fix.

lib/config.js
```diff
@@ -159,15 +170,49 @@ function sort (a, b) {
...
+    if (publicVar(k) &&
+        //argv is not really config, it's command config
+        k !== 'argv' &&
+        //logstream is a Stream, and would otherwise produce circular refs
+        k !== 'logstream') publicConf[k] = value
+
+    return publicConf
+  }, {})
+
+  output(JSON.stringify(publicConf))
```
zkat: How about using `JSON.stringify(publicConf, null, 2)` here to format the output a bit more nicely for humans?

lib/config.js
zkat: TIL

zkat (approved):
LGTM! Some small comments but otherwise 🐑 👍

lib/config.js
mikesherov: Good call. Fixing v

hzoo (approved):

mikesherov:
@zkat fixed up! Please rereview when convenient!

mikesherov:
@zkat I also had to fix the semver docs, as there was a semver release 11 days ago and the test suite ensures the docs are in sync. Should be all good now!

mikesherov:
Hooray it passes finally!

zkat:
LGTM, still :)

mikesherov:
Closing in favor of #18025

PR #18138, opened 2017-08-11
A typo in the test case

supertong:
I believe this is a typo in the test case.


supertong:
Find this one together with https://github.com/npm/npm/pull/18137

iarna:
Thanks for putting this together but I don't believe this is a typo.

iarna:
Ah, I was too hasty here and misread the change!

iarna:
This has landed as afb086230 in `release-next` and will be in today's npm release.

PR #18164, opened 2017-08-14
extract: only do multiple procs on OSX for now

zkat:


lib/install/action/extract.js
```diff
@@ -20,16 +20,34 @@ const workerFarm = require('worker-farm')
...
+// * Messes with Docker (I think)
+//
+// There are other issues that should be fixed that affect OSX too:
+//
+// * Logging is messed up right now because pacote does its own thing
+// * Global deduplication in pacote breaks due to multiple procs
+//
+// As these get fixed, we can start experimenting with re-enabling it
+// at least on some platforms.
+const ENABLE_WORKERS = process.platform !== 'darwin'
```
mikesherov: isn't this logic backwards? This says "enable workers if the platform isn't darwin", which is the opposite of the PR title.

zkat: ahahahah....


...yes

iarna:
Landed as 867df2b02

PR #18176, opened 2017-08-16
Lifecycle: use npm-lifecycle for lifecycle scripts

mikesherov:
Follow up to #18025 and uses the extracted `npm-lifecycle` package to run lifecycle scripts!


zkat (approved):
👍 🐑

PR #18229, opened 2017-08-18
Make install.sh find nodejs on debian

### LINKED ISSUE #17875: npmjs.com/install.sh fails silently on clean Debian 9 install

  - [ ] npm is crashing.

1. Spin up a clean Debian 9 ("stretch") VM. I used a DigitalOcean droplet.
2. Log in via SSH
3. Run `apt update` and `apt upgrade`
4. `apt install nodejs nodejs-legacy curl`
5. `curl -L https://www.npmjs.com/install.sh > npm-install.sh`, because I don't do `curl | sh`
6. `bash npm-install.sh`

Notice how the install script claims that "it worked", but actually `/usr/lib/node_modules/npm` is a broken symlink to a deleted folder in `/tmp` (`/usr/bin/pump` points to where it should, and would definitely work if the symlink in `/usr/lib/node_modules` wasn't broken).

### supporting information:

 - `npm -v` prints: N/A
 - `node -v` prints: v4.8.2
 - `npm config get registry` prints: N/A
 - Windows, OS X/macOS, or Linux?: Linux
 - Network issues:
   - Geographic location where npm was run: reproduced worldwide (DigitalOcean San Francisco 1 datacenter, plus a private server Germany)
 - Container:

cebe:
partially fixes https://github.com/npm/npm/issues/17875#issuecomment-323370102

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -53,6 +53,11 @@ export npm_config_loglevel
 # make sure that node exists
 node=`which node 2>&1`
 ret=$?
+# if not found, try "nodejs" as it is the case on debian
+if [ $ret -ne 0 ]; then
+  node=`which nodejs 2>&1`
+  ret=$?
+fi
 if [ $ret -eq 0 ] && [ -x "$node" ]; then
   (exit 0)
 else


cebe:
After this change the behavior is the same as in the original issue report. A broken symlink is created in /usr/bin/npm

iarna (approved):
Hokay, so the  status of this is:

We need to do a release with this.

And then figure out exactly how we get install.sh on to the website, as that was apparently neglected on our last release with an update to install.sh.

iarna:
This is merged as 5ec72ab5b in `release-next` and will be in the next npm release, it likely won't get to the website for a business day or two after that.

PR #18354, opened 2017-08-30
Paraquire - paranoidal require

nickkolok:
NPM ecosystem is [quite unresistant](https://github.com/ChALkeR/notes/blob/master/Gathering-weak-npm-credentials.md) against dependency attacks. It means that if a hacker gets control over one small abandoned package, the hacker gets control over all it's dependents, dependents of dependents, etc.
So, hijacking only one of small `npm`'s dependencies will be very crucial for whole NPM ecosystem. This PR improves `npm`'s security by jailing most dependencies into `paraquire` - secure [alternative](https://github.com/nickkolok/paraquire) for standard `require`. I jailed only two libraries just to demonstrate the mechanism.
I realise that this PR probably will not be merged, but I hope for having a productive discussion. I could not open an issue first, because only PR can show my suggestion clearly enough.

### DIFF
--- a/lib/npm.js
+++ b/lib/npm.js
@@ -16,6 +16,8 @@
   var unsupported = require('../lib/utils/unsupported.js')
   unsupported.checkForBrokenNode()
 
+  var paraquire = require('paraquire')(module)
+
   var gfs = require('graceful-fs')
   // Patch the global fs module here at the app level
   var fs = gfs.gracefulify(require('fs'))
@@ -30,7 +32,7 @@
   var which = require('which')
   var glob = require('glob')
   var rimraf = require('rimraf')
-  var lazyProperty = require('lazy-property')
+  var lazyProperty = paraquire('lazy-property')
   var parseJSON = require('./utils/parse-json.js')
   var clientConfig = require('./config/reg-client.js')
   var aliases = require('./config/cmd-list').aliases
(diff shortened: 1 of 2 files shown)


nickkolok:
UPD: I had to add `paraquire` as dependency and I don't know how to do it correctly (usual `npm install --save` is not enough?)

iarna:
Adding production dependencies to npm is done with `npm install -B` and adding dev dependencies with `npm install -D`, though in either case you need to be using the current version of npm in order to have the `package-lock.json` updated. Production dependencies are also added to `git` as they're all bundled.

Which brings us to the second thing:

As npm bundles all of its production dependencies (and locks all of its development dependencies), and we actually evaluate the content of every dependency update we are not running the risk you're seeking to mitigate.

This is rather unusual—most projects I don't actually manually inspect the code of every module I upgrade, but because npm bundles its dependencies (a copy is included with the npm tarball you download from the registry) and because of our release process, we do.

As such, while I can absolutely see the utility in your module, and I very much appreciate your bring it to our attention, I can't see us using it in the npm cli at this time.

(As an aside, trying to sandbox graceful-fs is particularly weird as it monkey-patches the global fs module. It's as far from a sandbox as you can get.)

nickkolok:
Thanks for such a detailed response!
May be ypu could advice me a project in which `paraquire` will be useful? I'll be very glad to prepare a pull request there.

PR #18372, opened 2017-09-01
Env. Variables: need to replace - with _

holyjak:


iarna (approved):
Thanks, that's a nice clarification!

iarna:
Thanks again for the docs improvement! This landed in 5.4.2.

PR #18382, opened 2017-09-02
Document browser field

mxstbr:
Based on https://github.com/defunctzombie/package-browser-field-spec and https://twitter.com/seldo/status/903997474645434368 this adds a very short note about the `browser` field to the package.json docs.

/cc @seldo @mjackson @wycats

### DIFF
--- a/doc/files/package.json.md
+++ b/doc/files/package.json.md
@@ -164,8 +164,7 @@ Or you can shorten that all into a single string, and npm will parse it for you:
 
 Both email and url are optional either way.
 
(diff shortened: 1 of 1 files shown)


mxstbr:
Note that this was shittly worded, I had to get out at my stop of the train, any suggestions very welcome!

zkat (approved):
I think this is ok to include in here. We don't do anything with it right now, but I think the field's become widespread enough to warrant this.

mxstbr:
:100:

Issue #16825, opened 2017-05-31
npm5 npm outdated leads to crash with cb() never called


dananichev:
#### I'm opening this issue because:

  - [x] npm is crashing.
  - [ ] npm is producing an incorrect install.
  - [x] npm is doing something I don't understand.
  - [ ] Other (_see below for feature requests_):

#### What's going wrong?
```js
beam-web git:ISC-1001 ✗ ✶ ✩  ❯ npm outdated
npm ERR! cb() never called!

npm ERR! This is an error with npm itself. Please report this error at:
npm ERR!     <https://github.com/npm/npm/issues>

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/dananichev/.npm/_logs/2017-05-31T12_57_45_180Z-debug.log
 ⌚︎ 7s
```

#### How can the CLI team reproduce the problem?
I'm not sure. Here is my debug-log https://gist.github.com/dananichev/86e10dcca3614cf8965fe03b621f8bdc

### supporting information:

 - `npm -v` prints: 5.0.0
 - `node -v` prints: v7.10.0
 - `npm config get registry` prints: http://builds.ic360:4873/
 - Windows, OS X/macOS, or Linux?: OS X
 - Network issues:
   - Geographic location where npm was run:
   - [ ] I use a proxy to connect to the npm registry.
   - [ ] I use a proxy to connect to the web.
   - [ ] I use a proxy when downloading Git repos.
   - [ ] I access the npm registry via a VPN
   - [ ] I don't use a proxy, but have limited or unreliable internet access.
 - Container:
   - [ ] I develop using Vagrant on Windows.
   - [ ] I develop using Vagrant on OS X or Linux.
   - [ ] I develop / deploy using Docker.
   - [ ] I deploy to a PaaS (Triton, Heroku).

Also, i tried to clean my cache (both ``npm cache clean`` and ``sudo rm -rf ~/.npm``). And i'm using ``nodenv`` to run multiple Node versions.

gizburdt:
Same error:

```
npm ERR! cb() never called!

npm ERR! This is an error with npm itself. Please report this error at:
npm ERR!     <https://github.com/npm/npm/issues>

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/Gijs/.npm/_logs/2017-05-31T13_13_59_135Z-debug.log
```

npm: 5.0.0
node: 7.10.0

gizburdt:
[...]

PR #18385, opened 2017-09-03
Change metadata retrieval for outdated

joshclow:
This addresses #16825 

Use fetchPackageMetadata to get metadata for extracting version from packages instead of cache.add (which in npm5 does not actually return package metadata anymore)


intentionally-left-nil (approved):

iarna:
Thank you for finding this oversight and putting together this PR. It's landed in 5.4.2.

PR #18397, opened 2017-09-04
doc: add "notice" loglevel and correct the default

kenany:
The "notice" loglevel was added and made the default in `npm@5`, but the
docs were not updated to this change. This patch updates the loglevel
docs to the new default and adds "notice" to the list of possible
values.

I don't know how to describe "notice" so I just left it at that.

### DIFF
--- a/doc/misc/npm-config.md
+++ b/doc/misc/npm-config.md
@@ -583,15 +583,15 @@ to the npm registry.  Must be IPv4 in versions of Node prior to 0.12.
 
 ### loglevel
 
-* Default: "warn"
+* Default: "notice"
 * Type: String
-* Values: "silent", "error", "warn", "http", "info", "verbose", "silly"
+* Values: "silent", "error", "warn", "notice", "http", "timing", "info",
+  "verbose", "silly"
 
 What level of logs to report.  On failure, *all* logs are written to
 `npm-debug.log` in the current working directory.
 
-Any logs of a higher level than the setting are shown.
-The default is "warn", which shows warn and error output.
+Any logs of a higher level than the setting are shown. The default is "notice".
 
 ### logstream
 


Butterfly:
Would trying an online thesaurus maybe help find other words for your "notice"? That's my personal go-to when I "run out of words", grin. I'm sorry I'm not more versed in `npm` yet, or I'd try to help fill that in. That word "notice" immediately makes me think of the word "advisement", but I'm not comfortable with the context here yet so as to know if that is actually what this is. :smile:

iarna:
`notice` is one of the [standard log levels](https://en.wikipedia.org/wiki/Syslog#Severity_level).

Wikipedia describes it as:

> Normal but significant conditions.
> 
> Conditions that are not error conditions, but that may require special handling.

Butterfly:
@iarna Cool, [that description is "cognitively friendly"](https://github.com/npm/npm/pull/18397#issuecomment-329573447) (is understandable). It's the kind of word that needs emphasis so users, meaning *us* newbies, know to actually... stop and "notice" when it occurs.

It now sounds like an important level. For me, it only sounded "eh, whatever, keep moving" until your description was added. That description reminds users that things are probably ok, but there's a potential that an extra step or two may be necessary to ensure success. :smile:

iarna:
This has landed in 5.4.2! =)

@Butterfly Our actual use doesn't seem to be out of line with your gut instinct here. It's important enough that we're actually showing it to you (contrast with `verbose` or `silly`), but not so important that you have to act on it. Here are all of the current use cases (there aren't many):

```
/Users/rebecca/code/npm/lib/install/save.js
101:        log.notice('save', `${pkg.name} is being moved from ${movedFrom.join(' and ')} to ${pkg.save}`)
```
```
/Users/rebecca/code/npm/lib/shrinkwrap.js
49:    log.notice('', `${PKGLOCK} has been renamed to ${SHRINKWRAP}. ${SHRINKWRAP} will be used for future installations.`)
```
```
/Users/rebecca/code/npm/lib/shrinkwrap.js
195:            log.notice('', `created a lockfile as ${path.basename(info.path)}. You should commit this file.`)
```

PR #18407, opened 2017-09-05
Clarify that "files" is an exclusive list of files

bmacnaughton:
It wasn't clear that files not included in the "files" field are excluded if the field is present.


fahrradflucht:
Just out of curiosity: I don't understand what else the text could have meant before your change. Could you elaborate?

iarna:
I think what they're trying to communicate is that if you use `files` it overrides a top level `.npmignore` or `.gitignore`?

zkat (approved):
This sounds about right, to me. Hopefully it'll help clarify things to people.

PR #18426, opened 2017-09-06
Do replace environment in config files (like .npmrc) even for keys

misak113:
Current .npmrc will keep keys with not replaced Environment variables. However values are replaced. So it may confuse somebody & has no option to solve this  problem

`.npmrc example`
```ini
registry=${NPM_REGISTRY_URL}
//${NPM_REGISTRY_HOST}/:_authToken=${NPM_AUTH_TOKEN}
always-auth=true
```
With env.vars:
```sh
export NPM_REGISTRY_URL=http://my.registry.com/
export NPM_REGISTRY_HOST=my.registry.com
export NPM_AUTH_TOKEN=xxxxxxxxxxxxxxx
```
NOW it will produce bellow options:
```ini
registry=http://my.registry.com/
//${NPM_REGISTRY_HOST}/:_authToken=xxxxxxxxxxxxxxx
always-auth=true
```

But expected options are:
```ini
registry=http://my.registry.com/
//my.registry.com/:_authToken=xxxxxxxxxxxxxxx
always-auth=true
```

Probably it is expected state & it would be very helpful when it will be merged for private registries which are using CI for publishing npm packages. For example: Then no more "`npm login | answer_input $AUTH_TOKEN ... # etc. constructions`" are necessary.

### DIFF
--- a/lib/config/core.js
+++ b/lib/config/core.js
@@ -332,7 +332,10 @@ Conf.prototype.parse = function (content, file) {
 Conf.prototype.add = function (data, marker) {
   try {
     Object.keys(data).forEach(function (k) {
-      data[k] = parseField(data[k], k)
+      const newKey = envReplace(k)
+      const newField = parseField(data[k], newKey)
+      delete data[k]
+      data[newKey] = newField
     })
   } catch (e) {
     this.emit('error', e)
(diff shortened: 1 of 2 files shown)


misak113:
Did anyone get this PR? Just asking. It's my first contribution here.

iarna:
@misak113, yup we've seen it, we just haven't had an opportunity to comment yet.

zkat (approved):
I talked with @iarna and while this is pretty darn weird, we think we should accept it because it _can_ be a nice escape valve, and I'm having a hard time figuring out how this would cause huge problems that the current replacement doesn't already.

Issue #18449, opened 2017-09-08
When used as a library, installation of packages with prepare script fails


olydis:
#### I'm opening this issue because:

  - [x] npm is crashing.
  - [x] npm is producing an incorrect install.
  - [x] npm is doing something I don't understand.

#### What's going wrong?

When using `npm` as a package (to install stuff from within a Node application), installation of packages with `prepare` script fails.

https://github.com/npm/npm/blob/latest/lib/pack.js#L174 is intended and expected to return `npm-cli.js` (essentially in order to do a dev-like `npm install` of a package with `prepare` script). However, this expectation is wrong when `npm` is used as a library. Whichever is the entry point script using `npm` will be returned by `require.main.filename`. As a result a non-sense process is spawned that obviously doesn't do what you expect. If that spawned process has non-zero exit code, the call to npm fails, otherwise it silently finishes but leaves the package in an un"prepare"d state.

#### How can the CLI team reproduce the problem?

Use the `npm` package (from code, not CLI) to install a package that has a prepare script.

### supporting information:

 - `npm@5.4.1` package
 - Windows, but shouldn't matter

iarna:
Hi there, thanks for reporting this but I'm afraid that using  npm as a package is not supported.  We can and will change internal APIs w/o major releases. Supported integration is to call `npm` with `child_process`.

The alternative in the future will be to build off of the libraries that npm is built off of (for example, `pacote` can turn any valid package specifier (eg `module@^1.2.3`, `github:foo/bar`) into a tarball.

olydis:
Thanks for the response and info about pacote! [We](https://github.com/Azure/autorest/) sort of require all `npm` features since we specifically support any `npm` packages (with some additional info) as extensions, which will be acquired and run automatically. But yes, the `child_process` way is of course a fine option 🙂

PR #18450, opened 2017-09-08
Don't rely on entry point to be npm-cli.js

olydis:
fixes https://github.com/npm/npm/issues/18449

### DIFF
--- a/lib/pack.js
+++ b/lib/pack.js
@@ -171,7 +171,7 @@ function packGitDep (manifest, dir) {
         return acc
       }, [])
       const child = cp.spawn(process.env.NODE || process.execPath, [
-        require.main.filename,
+        require.resolve('../bin/npm-cli.js'),
         'install',
         '--dev',
         '--prod',


lib/npm.js
```diff
@@ -70,8 +70,7 @@
 
   try {
     // startup, ok to do this synchronously
-    var j = parseJSON(fs.readFileSync(
-      path.join(__dirname, '../package.json')) + '')
+    var j = require(path.join(__dirname, '../package.json'))
```
iarna: We don't use require for a reason (it provides poor error messages when there are errors in the file and I don't think it supports files starting with UTF-8 BOMs.

lib/pack.js
```diff
@@ -171,7 +171,7 @@ function packGitDep (manifest, dir) {
         return acc
       }, [])
       const child = cp.spawn(process.env.NODE || process.execPath, [
-        require.main.filename,
+        require.resolve('../bin/npm-cli.js'),
```
iarna, line 174: This seems reasonable

iarna (changes requested):
So the thing is:

We don't support using npm as a package. We can and do change APIs related to that without ticking the major version number. That it works as well as it does for you is more of an accident than a fature.

That said: I'm not opposed changing `require.main.filename` to `npm-cli.js`. I'm ok with taking this PR, but only without the change to how the `package.json` is required.

lib/npm.js
olydis: > I don't think it supports files starting with UTF-8 BOMs.

it does

olydis: but this change is not at all important, the error message part makes perfect sense :) will undo

olydis:
@iarna Thanks, what you say makes perfect sense! I have reverted the `require` part. We call `npm` functions at a very "close to CLI" level, but are seriously considering the switch to child_process as soon as we get to it. Rest assured, we don't and won't expect the slightest support for what we did 😄

zkat (approved):
I'm ok with this change. Thanks!

PR #18458, opened 2017-09-08
install: fix bug with rollback of failed optional packages on Windows

marcins:
This change ensures an fsync after updating `package.json` on Windows,
which prevents an issue that intermittently occurs when rolling back a
package with many dependencies such as fsevents.

In testing it does not appear that fsync has as much of a performance
impact on Windows as it does on other platforms, so this should not affect
install times on Windows materially.

Fixes: #17671

### DIFF
--- a/lib/install/update-package-json.js
+++ b/lib/install/update-package-json.js
@@ -47,7 +47,9 @@ module.exports = function (mod, buildpath, next) {
   var data = JSON.stringify(sortedObject(pkg), null, 2) + '\n'
 
   writeFileAtomic(path.resolve(buildpath, 'package.json'), data, {
-    // We really don't need this guarantee, and fsyncing here is super slow.
-    fsync: false
+    // We really don't need this guarantee, and fsyncing here is super slow. Except on
+    // Windows where there isn't a big performance difference and it prevents errors when
+    // rolling back optional packages (#17671)
+    fsync: process.platform === 'win32'
   }, next)
 }


marcins:
Looks like Travis is failing with some sort of environment related error - I see the same error on other PRs.

marcins:
I should also probably clarify - #17671 has a lot more detail on my investigation, root cause analysis and test results of fsync on Windows vs macOS. Look for my comments. I’d link them but I’m on mobile now so it’d be a bit painful. I’ll do it tomorrow!

ansballard:
Just tested on my machine (Windows 10, Node 8.4.0). Confirmed 5.4.1 fails on `npm i`, and that `marcin/fix-windows-rollback` runs without error and runs all my tests/build scripts correctly.

1. clone pull request branch
1. `npm i -g npm@latest`
1. try `npm i` to confirm 5.4.1 is borked
1. try `node ../npm/bin/npm-cli.js i`
1. success!

```
Windows 10
Node 8.4.0
NPM 5.4.1/this branch
```

marcins:
As promised, here's a bit of a summary of my comments from #17671 with some links:

Discovering when the bug started appearing:
https://github.com/npm/npm/issues/17671#issuecomment-325666759

Identifying that something seems to still be writing to node_modules while rollback is deleting:
https://github.com/npm/npm/issues/17671#issuecomment-327653691

Discovering that it’s `update-package-json` that’s doing the writing. The root cause appears to be that while fsevents fails the platform check in `refresh-package-json` and rolls back it’s dependencies don’t fail the platform check so they continue to the refresh part. This refreshing doesn’t always complete by the time fsevents starts getting deleted, and the cause of the bug appears to be that files are being written to directories that are attempting to be deleted and this causes problems on Windows:
https://github.com/npm/npm/issues/17671#issuecomment-328083157

Identified that changing the call to `writeFileAtomic` in `refreshPackageJson` to do an fsync on Windows appears to avoid the issue:
https://github.com/npm/npm/issues/17671#issuecomment-328099129

So technically this doesn’t fix the root cause - the problem statement for that is something like “Don’t start rolling back a package until it’s children have finished their install”, however given that this is only a problem on Windows and not other platforms (it seems) my fix may be acceptable? Up to y’all!

iarna (approved):
This is some great work, thank you. I'm probably going to sneak out a release before the weekend with this fix!

iarna:
And yeah, travis changed something during the last round of updates that's broken us. =/

iarna:
Thanks again for tracking this down. This landed in 5.4.2!

PR #18500, opened 2017-09-12
Extract bin-links and gentle-fs in preparation of extraction to a separate library.

mikesherov:
As I did with `lifecycle`, the first PR simply isolates the code within npm, so it's clear to see what's happening in a single repo. A couple of things have happened here:

1. Extraction of `lib/utils/bin-links` which will be extracted to an `npm-bin-links` repo after this is merged.
2. In order to accomplish 1., `lib/utils/gently-rm` and `lib/utils/link` needed to be isolated. I combined the two into `lib/utils/gentle-fs`, which will be become yet another extracted package.
3. I deleted the unit-link test suite, and the `unit-gentlyrm` test suites which will be reintroduced in the separate extracted packages.

Please let me know if you have any questions. Thanks!


zkat (approved):
LGTM from my end. I think it's fine for this interim patch to remove the test suite, as long as we have the test suite in those external packages passing before this PR goes out in an actual npm release.

I'd also like to get a +1 from @iarna too cause she's messed around more with the gently-rm stuff than I have and might spot things.

iarna (approved):
I'm 👍 once we have the unit tests that were removed here passing over in the new library.

Issue #18356, opened 2017-08-30
NPM Docs: package-lock.json missing description of "requires".


kreig303:
Only bringing this one up because I had to do some minor hunting...

When I made a `package.json` update for a project I'm looking to contribute to I noticed that their `package-lock.json` had been updated with `"requires": true` as a property and had relevant information added which made for a large and unexpected diff.

As I was trying to avoid contention I had to research this and found it was introduced with NPM 5.1.0 which is told in the release notes here on GitHub but is not described in the standard NPM docs at https://docs.npmjs.com/files/package-lock.json.

I would of course willingly try my hand at this update and PR this but I've never contributed and would want to wait for someone else before I went ahead with my own take...

ghost:
@kreig303 can you direct me to where in this document this change was mentioned?

https://github.com/npm/npm/releases/tag/v5.1.0

kreig303:
@josephmcasey

In **Features** section, second item down... 

> 4f45ba222 a48958598 901bef0e1 #17508

kreig303:
#18103 is vaguely related to this.

PR #18555, opened 2017-09-17
#18356 Added description of requires boolean in package-lock.json to …

kreig303:
…npm docs.

Referenced in #18356.


doc/files/package-lock.json.md
```diff
@@ -46,6 +46,9 @@ The version of the package this is a package-lock for. This must match what's in
 An integer version, starting at `1` with the version number of this document
 whose semantics were used when generating this `package-lock.json`.
 
+### requires
```
kevinburke, line 49: I'm not sure that this is accurate, here's what the package-lock.json file looks like for me (generated using 5.4.2):

```patch
     "@types/supertest": {
       "version": "2.0.3",
       "resolved": "https://registry.npmjs.org/@types/supertest/-/supertest-2.0.3.tgz",
       "integrity": "sha512-QxgjuDhQEq4tPJXTOnGoN4P/BWwvftvkaCZhXMp1C1Otk9kGXb7l/wA7kwz3MwtEMnovdYikS+ZtvaQ4RbhF4g==",
-      "dev": true
+      "dev": true,
+      "requires": {
+        "@types/superagent": "3.5.5"
+      }
     },
```

In other words, it's an object, not a boolean.

kevinburke: My apologies, it looks like it's specified in the file in two different places; at the top level, it's a boolean, inside of a package descriptor, it's an object. Not sure the best way to handle that.

kreig303: So how should we handle this?  Some docs are better than none... and if this isn't to be accepted I'll just close it.

kreig303:
Closing.  Out of sync and need to properly PR.

PR #18563, opened 2017-09-18
Disable auto updater at every single start

ghost:


zkat:
@kazan1000 https://github.com/yeoman/update-notifier#user-settings is how you can disable it.

PR #18568, opened 2017-09-18
Fix a typo in the documentation for the "engine" section of package.json files

apitman:
I was checking the documentation about the "engine" section of package.json files and found a grammar problem in this sentence. I think this is what it means to say, but I don't know enough about NPM to know for sure that this is an accurate statement.


zkat (approved):
This is definitely easier to read 👍

PR #18571, opened 2017-09-18
translation typo

devmount:


kenany:
Why does _verboten_ not work here?

devmount:
Is *verboten* an actual English word? Didn't know that... It is the German word for *prohibited* or *forbidden*... I thought that was meant here.

kenany:
Seems like it is: [this page](http://www.wordnik.com/words/verboten) says it's in _The American Heritage® Dictionary of the English Language, 4th Edition_, and [this page](http://www.thefreedictionary.com/verboten) lists a couple more (admittedly unknown to me) sources.

devmount:
I see. In that case, just delete my PR. I thought it was a mistake and was confused to read a German word here.

kenany:
I wouldn't say that this need to be deleted so quickly. One could argue that "forbidden" is a more familiar word to English speakers, so perhaps this could be accepted as an improvement to the docs. I'll leave it to the rest of the team to decide.

zkat:
_verboten_ is indeed an English word, but I also like _forbidden_ much more!

zkat (approved):
LGTM! Forbidden is probably easier to understand anyway :)

PR #18577, opened 2017-09-19
Document: Non-rooted package.json are not packed

robpalme:
There was a recent breaking change to NPM 5.4 that stopped publishing/packing `package.json` files that exist below the root.


iarna:
@robpalme That's a bug and not an intended change.

zkat:
Closing this, as this was a bug, and has definitely been fixed since.

PR #18610, opened 2017-09-23
Update npm-scripts.md

Andarist:
Add information that `prepare` is ran also before `npm link`


Andarist:
It seems the behaviour is different - `npm link` causes `npm install` and `npm uninstall` in turns

PR #18618, opened 2017-09-24
update graceful-fs dependency from 4.1.9 to 4.1.11

dusura:
npm 3.x does not work reliably on Windows systems with AntiVirus when installing modules because it copies module files to a temporary location before completing installation and AntiVirus software can lock these files causing npm to fail when trying to move the module files to their permanent location. graceful-fs 4.1.11 has a fix for this.

### DIFF
--- a/node_modules/graceful-fs/package.json
+++ b/node_modules/graceful-fs/package.json
@@ -2,39 +2,39 @@
   "_args": [
     [
       {
-        "raw": "graceful-fs@4.1.9",
+        "raw": "graceful-fs@4.1.11",
         "scope": null,
         "escapedName": "graceful-fs",
         "name": "graceful-fs",
-        "rawSpec": "4.1.9",
-        "spec": "4.1.9",
+        "rawSpec": "4.1.11",
+        "spec": "4.1.11",
         "type": "version"
       },
-      "/Users/rebecca/code/npm"
+      "C:\\Users\\Dave\\git\\npm-new"
     ]
   ],
-  "_from": "graceful-fs@4.1.9",
-  "_id": "graceful-fs@4.1.9",
+  "_from": "graceful-fs@4.1.11",
+  "_id": "graceful-fs@4.1.11",
   "_inCache": true,
   "_location": "/graceful-fs",
   "_nodeVersion": "6.5.0",
   "_npmOperationalInternal": {
-    "host": "packages-12-west.internal.npmjs.com",
-    "tmp": "tmp/graceful-fs-4.1.9.tgz_1475103672016_0.7011275647673756"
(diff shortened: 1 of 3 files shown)


iarna:
npm@3 has been unsupported for almost a year now. Outside of security patches or specific requests from the Node.js project (for the purposes of new Node.js releases) we don't ordinarily accept patches or do releases to unsupported versions.

I would note that npm@4 has this version and has only limited breaking changes. Quoting from the changelog here:

> * `npm search` rewritten to stream results, and no longer supports sorting.
> * `npm scripts` no longer prepend the path of the node executable used to run
>   npm before running scripts. A `--scripts-prepend-node-path` option has been
>   added to configure this behavior.
> * `npat` has been removed.
> * `prepublish` has been deprecated, replaced by `prepare`. A `prepublishOnly`
>   script has been temporarily added, which will *only* run on `npm publish`.
> * `npm outdated` exits with exit code `1` if it finds any outdated packages.
> * `npm tag` has been removed after a deprecation cycle. Use `npm dist-tag`.
> * Partial shrinkwraps are no longer supported. `npm-shrinkwrap.json` is
>   considered a complete installation manifest except for `devDependencies`.

I'm curious which of those changes is stopping you from updating to at least npm@4? (Obviously I'd prefer you upgrade to the supported version (npm@5), but its changes _are_ more extensive.)

mekwall:
@iarna npm@5 works horrible on Windows because of similar issues this user is having. I have a PR waiting to get accepted upstream at https://github.com/isaacs/node-graceful-fs/pull/119 that should solve all the rename issues at least. Maybe you could poke @isaacs about it.

dusura:
@iarna regarding npm@3 being unsupported, why is npm@3 bundled with Nodejs 6.x LTS if it is not supported? The reason I raised this PR was I was hoping to get this fix in a future Nodejs 6.x LTS release. But perhaps this is not the correct way or this fix does not fit the criteria for an LTS release? I originally raised this issue [here ](https://github.com/nodejs/Release/issues/252) but was advised this type of change needs to come from the npm team, hence this PR.

I prefer to have an LTS release that includes this fix installed on all my developer workstations rather than also requiring the additional step of patching npm to npm@4.

iarna:
@desura The Node.js Foundation has their own support schedule and does not provide resources to npm, is why, basically.

That said, if Node.js intends to ship a new LTS release with this change, I'm prepared to do a special release. We _do_ accept patches at their request to out-of-support npm releases.

MylesBorins:
if a new version of npm 3.x was cut including this patch I would include it in a future v6.x

zkat:

PR #18625, opened 2017-09-25
Fix: Update npm-packlist to version which does not ignore test and tests directories

### LINKED ISSUE #18341: npm pack is filtering out test directory

  - [ ] `npm pack` is filtering out `test` directories
     -  https://github.com/npm/npm-packlist/blob/master/index.js#L36
  - [ ] Documentation does not match behavior
     - https://docs.npmjs.com/misc/developers
  - [ ] `npm upgrade -g npm` installs 5.4.0 even though 5.4.0 is Pre-release

#### What's going wrong?
On previous versions of `npm`, when running `npm pack`, the tar file would include all directories (including `test`). With the most recent version of `npm` (5.4.0), this is not the case. The `test` directory is filtered out which is causing issues in our dependencies.

Along with the above described issue, the documentation does not match the actual behavior. If you take a look at https://docs.npmjs.com/misc/developers , you'll notice that the documentation does not call out that `test` will be ignored. If you take a look at the behavior as well as the specific line of code that caused this (https://github.com/npm/npm-packlist/blob/master/index.js#L36), you'll see that their is a disconnect.

The other issue is that we expected `5.3.0` to be installed when running `npm upgrade -g npm` isntead of 5.4.0. According to your docs, you say to install 5.4.0 you need to run `npm upgrade -g npmc`, but it appears that you can install it with just `npm` as well.

- Setup a repo with a `test` directory
- with version `5.4.0` of npm, run `npm pack`
- run `tar zxf`
- examine the contents
  - NOTICE: `test` is not included

JakeChampion:
Fixes #18341

### DIFF
--- a/node_modules/npm-packlist/README.md
+++ b/node_modules/npm-packlist/README.md
@@ -48,12 +48,11 @@ This uses the following rules:
     1. .npmignore and .gitignore files (their effect is in the package
        already, there's no need to include them in the package)
     2. editor junk like `.*.swp`, `._*` and `.*.orig` files
-    3. A `/test/` or `/tests/` folder at the root
-    4. `.npmrc` files (these may contain private configs)
-    5. The `node_modules/.bin` folder
-    6. Waf and gyp cruft like `/build/config.gypi` and `.lock-wscript`
-    7. Darwin's `.DS_Store` files because wtf are those even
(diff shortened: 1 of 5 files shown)


imaustink:
Any chance this will be merged and released soon?

PR #18631, opened 2017-09-25
Restore ability to use search from a script

gamtiq:
It was impossible to use search programmatically from the moment when streamed search was introduced. I have restored this ability.
Also I have added config parameter `full` to supply all fields into search results.


iarna (changes requested):
Can you explain to me what you mean by "impossible to use search programmatically"?

`npm search --json` and `npm search --parseable` seem to work just fine? 
 
Given the changes it looks like you want to use npm as a library? That's not supported and we won't accept patches to support that use case. (That's why there's no documentation of the internal APIs.) The supported way to use npm from other software is to use `child_process` to spawn a copy of `npm` with `--json` or `--parseable` options.

We would be interested in a patch that moves this functionality into a library and has npm use the library. If you're interested in pursuing that then please drop by the #npm channel on the https://package.community/ Discord server and we'd be happy to chat about it.

gamtiq:
I use npm as a library in my project [seeq](https://github.com/gamtiq/seeq). Earlier (at least in version 1.x, perhaps in versions 2.x and 3.x too) it was possible to make a search from a script like the following:
```js
npm.load({loglevel: "silent"}, function(err, npmObj) {
    if (err) {
        callback(err, []);
    }
    else {
        npm.commands.search(['search term'], true, function(err, data) {
            ...
        });
    }
});
```
Now this way cannot be used without hacks because results are printed in console.

iarna:
That usage is not supported and the fact that it worked as long as it did was just luck. The supported mechanism is:

```
const exec = require('child_process').exec

exec('npm search --json abc', (err, stdout) => {
  const data = JSON.parse(stdout)
  …
}) 
```

I'm going to close this. If you have trouble using `npm` in this manner then your best bet for support is probably to drop by the [WeAllJS](https://wealljs.org/) #npm channel and ask there. (The npm devs all hang out there, along with a bunch of other folks.)

gamtiq:
It is a pity.
In that case callback parameter of [search function](https://github.com/npm/npm/pull/18631/files#diff-72d460be7ab8987d776046e325845bb2R23) is useless.

What is about [full option](https://github.com/npm/npm/pull/18631/files#diff-044c448c4d6eba953bf61072bbd0fb31R68)? Can I create another PR to save it?

PR #18646, opened 2017-09-27
fix(install): Find required package even if it's a local dependency

### LINKED ISSUE #18638: Repro repo: npm5: Maximum call stack size exceeded: linked packages

  - [ *] npm is crashing.

#### What's going wrong?

Trying to install the packages in this repo (https://github.com/briandipalma/test-npm) fails.

The README in https://github.com/briandipalma/test-npm explains how to reproduce.

Gist of output below

https://gist.github.com/briandipalma/9880f7dacb082915a8e76644fe821d10

### supporting information:

 - `npm -v` prints: v5.4.2-canary.2
 - `node -v` prints: v8.1.0
 - Windows, OS X/macOS, or Linux?: All?

\<snip\> No remote dependencies are used in the repo.

I know there are similar issues but their stack traces are different and they haven't provided reproduction repos.

briandipalma:
Filtering out local dependencies from the `findRequirement` function leads
to infinite loops if two local packages depended on each other.

closes: #18638

### DIFF
--- a/lib/install/deps.js
+++ b/lib/install/deps.js
@@ -689,7 +689,6 @@ var findRequirement = exports.findRequirement = function (tree, name, requested,
     return null
   }
   if (tree.isTop) return null
-  if (!preserveSymlinks() && /^[.][.][\\/]/.test(path.relative(tree.parent.realpath, tree.realpath))) return null
   return findRequirement(tree.parent, name, requested, requestor)
 }
 


lib/install/deps.js
```diff
@@ -689,7 +689,6 @@ var findRequirement = exports.findRequirement = function (tree, name, requested,
     return null
   }
   if (tree.isTop) return null
-  if (!preserveSymlinks() && /^[.][.][\\/]/.test(path.relative(tree.parent.realpath, tree.realpath))) return null
```
iarna, line 692: As you might imagine, this code was added for a reason and just removing it will introduce a new class of bugs.

In particular, this is would break how we resolve dependencies inside of symlinks, making npm think the dependency is resolvable even when it isn't. A solution to this bug will have to come some other way.

iarna (changes requested):
Thanks for giving this a go. I suspect that the fix for the bug will be cycle detection elsewhere in npm.

lib/install/deps.js
briandipalma: I figured that it would break something, the tests that were added didn't pick that up btw.

iarna: I saw that, which is surprising and definitely does call for a look.

zkat:

PR #18707, opened 2017-10-04
Return non zero exit code when no tests defined.

DanBUK:
When running npm test, if it's not defined in package.json currently npm returns an exit code of 0, yet displays an error.

This PR changes that to return an exit code of 1 if it's not defined.

### DIFF
--- a/lib/run-script.js
+++ b/lib/run-script.js
@@ -136,7 +136,7 @@ function run (pkg, wd, cmd, args, cb) {
   } else {
     if (!pkg.scripts[cmd]) {
       if (cmd === 'test') {
-        pkg.scripts.test = 'echo \'Error: no test specified\''
+        pkg.scripts.test = 'echo \'Error: no test specified\' && exit 1'
       } else if (cmd === 'env') {
         if (process.platform === 'win32') {
           log.verbose('run-script using default platform env: SET (Windows)')


iarna:
This would also bring things into alignment with `npm init`. This seems reasonable to me.

iarna:
TBH, I didn't even know we had this special case. I wish we didn't. =p

iarna:
This was merged to `release-next` as e71386d98 and will be in (as the branch name implies) our next release. Thanks!

iarna:
Ok, I tracked down where this was added (https://github.com/npm/npm/issues/2959) and it's sufficiently well discussed and tested that I'm not comfortable taking this in a patch/minor release so I'm gonna reopen this and tag it to be in the next major release.

iarna (changes requested):
I would also note that there are a couple of tests failing with this. Before we land this those will need to be updated to take it these changes into account.

bburhans:
The [official documentation](https://docs.npmjs.com/cli/test) says:
>[`npm test`] runs a package's "test" script, if one was provided.

That doesn't define any behavior if one was *not* provided, but I feel like it slightly implies *no* behavior in that case, because the "if one was provided" case is explicitly mentioned. If the expectation is that it break (e.g. stop a CI process) instead of do nothing, we should clarify that behavior or remove the "if none was provided" clause.

I don't know how many CI frameworks run `npm test` or `npm run test` out of the box, but if it's nonzero because of [the precedent we set in 2012](https://github.com/npm/npm/issues/2959), we might consider leaving it exiting zero. However, the modern npm init defaults are contrary:
```javascript
    "test": "echo \"Error: no test specified\" && exit 1"
```
We should make sure that the "no test script is (okay|bad)" expectation is consistent one way or the other.

Recommendation: use the word "error" and exit nonzero everywhere, or downgrade it to info/warn and exit zero everywhere. The discrepancy is misleading and sets a confusing precedent. `npm init` defaults, `npm test`, `npm run test`, and npm docs should all agree.

zkat:

PR #18709, opened 2017-10-04
fix: improve quoting for run-script arguments

### LINKED ISSUE #18670: Passing arguments to scripts using -- is inconsistent and wrong if argument ends with backslash

  - [ ] npm is crashing.

#### What's going wrong?

**Windows**

`npm run script -- foo\` gives arguments to the program as `[node, script, 'foo"']` on Windows. If you have several arguments ending with back slashes, it will put them into single argument

`npm run script -- foo\ bar\` -> `[node, script, 'foo" bar"']`

For example:

```
npm run demo -- --foo c:\bar\


> test-npm-cli-pass@1.0.0 demo C:\Users\glebb\git\test-npm-cli-pass
> node index.js "--foo" "c:\bar\"

process.argv
[ 'C:\\Program Files\\nodejs\\node.exe',
  'C:\\Users\\glebb\\git\\test-npm-cli-pass\\index.js',
  '--foo',
  'c:\\bar"' ]
```

Passing several arguments it puts them into single string

```
npm run demo -- c:\bar\ c:\baz\


> test-npm-cli-pass@1.0.0 demo C:\Users\glebb\git\test-npm-cli-pass
> node index.js "c:\bar\" "c:\baz\"

process.argv
[ 'C:\\Program Files\\nodejs\\node.exe',
  'C:\\Users\\glebb\\git\\test-npm-cli-pass\\index.js',
  'c:\\bar" c:\\baz"' ]
```

**Mac**

And if you do this on Mac, then it is very interesting. 

```
$ npm run demo -- c:\\bar\\ c:\\baz\\

> test-npm-cli-pass@1.0.0 demo /Users/irinakous/git/test-npm-cli-pass
> node index.js "c:\bar\" "c:\baz\"

process.argv
[ '/Users/irinakous/.nvm/versions/node/v6.8.1/bin/node',
  '/Users/irinakous/git/test-npm-cli-pass/index.js',
  'c:\\bar" c:baz"' ]
```

I put small repo with just example commands at https://github.com/bahmutov/test-npm-cli-pass and it has good readme showing the output I got (npm4, npm5). But here is the script file, it just prints `process.argv`

```js
// index.js
console.log('process.argv')
console.log(process.argv)
```
and here is the package.json

```json
{
  "scripts": {
    "test": "npm run demo -- --foo bar",
    "test-win": "npm run demo -- --foo c:\\bar\\",
    "test-win2": "npm run demo -- c:\\bar\\ c:\\baz\\",
    "demo": "node index.js"
  }
}
```

### supporting information:

 - `npm -v` prints: 5.4.2, 4.6.1, assume this is common to most NPM versions
 - `node -v` prints: 8, 6
 - Windows, OS X/macOS, or Linux?: Windows, Mac
 - Network issues:
 - Container:

rhendric:
Fixes #18670

Disclosure: I'm the author of [puka](https://www.npmjs.com/package/puka)—I wrote it to solve exactly this kind of problem, more comprehensively (correctly quoting shell arguments on Windows in particular is terribly easy to get almost right, until it's wrong) than any other library I could find.


iarna:
@rhendric Wow, this is lovely, thank you! (I'd love to puka replace `lib/utils/escape-arg.js` and `lib/utils/escape-exec-path.js` as well.)

iarna (changes requested):
I'm concerned that `puka` doesn't seem to be aware of things like MINGW bash? While the current solution is… bad… I also don't want to break other users to fix this. I'm pretty sure that's easy enough to fix by using `ShellString.sh` and npm's own `is-windows-shell`…

But my other concern is that it seems to have pretty different behavior on Unix as well:

Given: `npm run env -- echo '"hi $npm_config_email there"'`
Before: `env "echo" "\"hi $npm_config_email there\""`
After: `env echo '"hi $npm_config_email there"'`

Specifically, because it escapes using single quotes and previously we used double it substantially changes how commands are processed. This in particular matters when it comes to environment processing as the `npm_config_` prefixed env vars are not set in your regular shell, but they are set in the `npm run` environment.

I would add that arguably `puka`'s behavior _is_ consistent w/ what other unix tools do, eg `env` does this when execing subcommands. That being said, correct may constitute a breaking change for npm and so we'd have to wait for a major release to adopt it.

rhendric:
I don't have a MinGW system on which to test, but I'd happily incorporate support for that into puka if you or someone else can do the necessary testing. It's also true that you can do that yourself using `ShellString.sh`, if MSYS supports the same conventions as bash.

The environment variable thing is an interesting twist—I wasn't aware that npm added its own environment variables to script processes. It strikes me that taking advantage of those environment variables through script arguments passed through npm is not cross-platform, given the different environment variable conventions. It's also pretty non-obvious that, after quoting some argument that contains a dollar sign before passing it to `npm run`, the dollar sign is still going to get expanded anyway unless you escape/quote it a second time.

You don't happen to have any sense of how frequently people specifically use those `npm_` variables in arguments to `npm run` commands, do you? Maybe a good solution would be for those people to prefix their package scripts with a tiny cross-platform utility that does another level of dollar expansion. So they'd change to:

Given: `npm run expand -- env echo '"hi $npm_config_email there"'`
npm runs: `expand env echo '"hi $npm_config_email there"'`
expand runs: `env echo 'hi user@example.com there'`

And that way, the dollar syntax would work on all platforms including Windows, and it would be explicitly opt-in for people who want their quoted dollar signs to be more than just dollar signs. What do you think about something like that?

**Edit:** Maybe that utility is [cross-env](https://www.npmjs.com/package/cross-env).

rhendric:
I found out that AppVeyor build machines have all manner of bashes installed, so I ran some tests myself. Puka's Windows behavior seems to be correct when Node is being invoked from MinGW bash, Cygwin bash (both 32- and 64-bit installations), and MSYS2 bash. I ran all of the above bashes with `--login`, which sets the `MSYSTEM` environment variable that npm currently checks in [lib/utils/is-windows-bash.js](/npm/npm/blob/latest/lib/utils/is-windows-bash.js). Looking at the [relevant Node.js code](/nodejs/node/blob/master/lib/child_process.js#L450), it looks like `child_process` will always use `cmd.exe` on Windows unless the `COMSPEC` environment variable is tinkered with, which to my knowledge Cygwin/MinGW/MSYS don't do.

So, for its primary purpose (preparing strings to be `exec`'d), I believe Puka works just fine under MinGW and friends. For the purpose of outputting strings for bash completion (which is what it looks like `is-windows-bash` is ultimately used to gate), using either `ShellString.sh` or `quoteForSh` definitely sounds like the appropriate approach to me.

On the other issue—whether people are relying on npm performing a round of escaping environment variables, in order to take advantage of `npm_` variables—I want to point out (just in case you didn't realize it) that using `npm_` variables inside `package.json` should still work (as well as they ever did) after this PR. The package scripts are passed through `unquoted`, so everything in them is provided verbatim to the shell. (I say ‘as well as they ever did’ because the environment variable syntax is still not cross-platform.) We're just talking about people invoking npm with such variables in the arguments, as in your example `npm run env -- echo '"hi $npm_config_email there"'`.

iarna:
Yes, believe it or not, people actually put `npm run` commands like that in automation and in lifecycle scripts. I would prefer not to break them.

rhendric:
Fair enough. Do you think I should try to address that somehow in this PR, or just sit on this until the next major version (or drop it altogether)?

iarna:
If you can get it using the same kind of quotes we already do then I would love to land this sooner rather than later!

rhendric:
Well, it's not really a question of what kind of quotes to use; it's more a question of what characters/shell constructs should bash interpret versus pass through to the command being invoked.

In #18670, (half) the issue was caused by bash interpreting backslashes. The thing to do there seems to be to escape backslashes so that they pass through. If we do that, though, that is a potentially breaking change if anyone had already run into this issue in their automation scripts and doubled up their backslashes.

Similarly, subshell syntax (`$()`, ` `` `) is also interpreted by bash inside double quotes. Unlike environment variables, I can't think of an advantage to expanding subshell expressions in the child shell rather than in the shell invoking npm, but the fact remains that people could have depended on that behavior.

So where do we draw the line? If we say that *anything* that worked with the current quoting conventions should continue to work, then there's no point in using Puka (at least on Unix); just keep the current code and either add some special handling for terminal backslashes or close #18670 as wontfix, since the de facto semantics npm currently offers is, ‘`run` arguments get interpreted by bash as double-quoted strings before being passed to the child command,’ and singleton backslashes are illegal at the end of bash double-quoted strings.

I could see a compromise working where just `$npm_` variables (or maybe all environment variables?) get specially interpolated before passing the string to Puka, either as a temporary hack (we could print a deprecation warning too in that case, and then remove the hack next major version for the ‘correct’ approach) or as a permanent change (in which case it needs clear documentation). But again, for the above reasons, even that could potentially break existing automation.

rhendric:
I guess you aren't that enthusiastic about a semi-backwards-compatible compromise? If that's the case, is there anything I should/can do to line this up for npm@6?

zkat:
@iarna and I discussed this a bit, and I think the conclusion was that "only-kinda-breaking" is still breaking, but that the `puka` behavior _is_ what we want to eventually have. So we're not going to take a "partial" solution to this.

I had a slight concern that folks would no longer have a way to interactively inspect npm-injected env vars, but it turns out you can do this as of `npm@5.2.0` because of the inclusion of [`npx`](https://npm.im/npx) and its `-c` option: `$ npx -c 'echo "this is $npm_package_name"'`.

So... we're probably going to include this in `npm@6`, and I'm excited because npm's weird, homegrown escaping has consistently been a brittle part of the CLI and related tooling. I'm looking forward to having this bit concentrated in a single, consistent package.

Cheers!

rhendric:
Looks like I missed the npm@6 boat... just an oversight or is there some other issue (possibly something I can help with)?

zkat:

PR #18711, opened 2017-10-04
Upgrade to bearer with otp

iarna:
This adds legacy auth support to the new commands. (They thought they had it previously but weren't passing it along correctly.)

This also makes it so that enabling two-factor authentication upgrades your auth method to a bearer token. 2fa does not support basic authentication. If your registry does not support bearer authentication then we will now refuse to enable two-factor authentication.

### DIFF
--- a/lib/profile.js
+++ b/lib/profile.js
@@ -213,13 +213,28 @@ function enable2fa (args) {
       'Enabling two-factor authentication is an interactive opperation and ' +
       (conf.json ? 'JSON' : 'parseable') + 'output mode is not available'))
   }
-  log.notice('profile', 'Enabling two factor authentication for ' + mode)
+
   const info = {
     tfa: {
       mode: mode
     }
   }
-  return readUserInfo.password().then((password) => {
+
+  return Bluebird.try(() => {
+    // if they're using legacy auth currently then we have to update them to a
+    // bearer token before continuing.
+    if (conf.auth.basic) {
+      log.info('profile', 'Updating authentication to bearer token')
+      return profile.login(conf.auth.basic.username, conf.auth.basic.password, conf).then((result) => {
+        if (!result.token) throw new Error('Your registry ' + conf.registry + 'does not seem to support bearer tokens. Bearer tokens are required for two-factor authentication')
+        npm.config.setCredentialsByURI(conf.registry, {token: result.token})
+        return Bluebird.fromNode((cb) => npm.config.save('user', cb))
+      })
+    }
+  }).then(() => {
+    log.notice('profile', 'Enabling two factor authentication for ' + mode)
+    return readUserInfo.password()
+  }).then((password) => {
     info.tfa.password = password
     log.info('profile', 'Determine if tfa is pending')
     return pulseTillDone.withPromise(profile.get(conf)).then((info) => {


lib/profile.js
```diff
@@ -81,7 +81,18 @@ function config () {
...
   }
-  conf.auth = npm.config.getCredentialsByURI(conf.registry)
+  const creds = npm.config.getCredentialsByURI(conf.registry)
+  if (creds.token) {
+    conf.auth = {token: creds.token}
+  } else if (creds.username) {
+    conf.auth = {basic: {username: creds.username, password: creds.password}}
+  } else if (creds.auth) {
+    const auth = Buffer.from(creds.auth, 'base64').toString().split(':', 2)
+    conf.auth = {basic: {username: auth[0], passowrd: auth[1]}}
```
chrisdickinson: typo: `passowrd`

lib/profile.js
```diff
@@ -213,51 +224,65 @@ function enable2fa (args) {
...
+
+  return Bluebird.try(() => {
+    // if they're using legacy auth currently then we have to update them to a
+    // bearer token before continuing.
+    if (conf.auth.basic) {
+      log.info('profile', 'Updating authentication to bearer token')
+      return profile.login(conf.auth.basic.username, conf.auth.basic.password, conf).then((result) => {
+        if (!result.token) throw new Error("Your registry " + conf.registry + "does not seem to support bearer tokens. Bearer tokens are required for two-factor authentication")
+        npm.config.setCredentialsByURI(conf.registry, {token: result.token})
+        return Bluebird.fromNode((cb) => npm.config.save('user', cb))
```
chrisdickinson: TIL about `fromNode`!

lib/profile.js
```diff
@@ -213,51 +224,65 @@ function enable2fa (args) {
...
-    tfa: {
-      mode: mode
+
+  return Bluebird.try(() => {
+    // if they're using legacy auth currently then we have to update them to a
+    // bearer token before continuing.
+    if (conf.auth.basic) {
+      log.info('profile', 'Updating authentication to bearer token')
+      return profile.login(conf.auth.basic.username, conf.auth.basic.password, conf).then((result) => {
+        if (!result.token) throw new Error("Your registry " + conf.registry + "does not seem to support bearer tokens. Bearer tokens are required for two-factor authentication")
```
chrisdickinson: I note that it's possible that they enabled TFA in another session, & the reason that they cannot authenticate here is because of their already-active TFA status. Does `profile.login` handle the OTP prompt in that case?

lib/profile.js
iarna: ty!

lib/profile.js
iarna: ~~yes~~

iarna: ah, I see what you're saying… 

So this flow only happens if you run `npm profile enable-2fa`. So if you've enabled 2fa elsewhere but have a legacy auth on this account, I wouldn't expect you to be enabling 2fa here too? If you try you'll get an uncaught EOTP error. We _could_ catch it here, but it's really a more general problem because sudenly _every_ operation you do will ask for an OTP and I think it's far more likely that you'll get failures due to, for example, `npm install`.

Tweaking the error handler to be more specific, that is, to tell folks to `npm login` if they have bearer auth and just errored out due to EOTP (or E401 w/ no www-authenticate headers) would probably be good though.

zkat (approved):
🐑 LGTM

PR #18742, opened 2017-10-07
Update package.json.md

felicio:
Explicitly states 'github' as a valid shorthand option of repository reference.


iarna (approved):
Looks good! Thanks! 👍

PR #18778, opened 2017-10-10
node-gyp@3.6.2

watilde:
Let's push https://github.com/nodejs/node/pull/12479 forward. Once npm@2 is bumped, I also can take care to make a new PR to push this to node@4.


watilde:
Could I possibly be a collaborator to take care the low priority issues btw? That can be a big motivation to me :)

iarna:
@watilde If you want to participate more, head on over to https://package.community/ You can help us figure out how we assign that kind of thing! (All changes go through PR even if you are a collab so it's not particularly different. Merging is currently really a function of the release process so it's kind of. ¯\_(ツ)_/¯)

.travis.yml
```diff
@@ -14,20 +14,12 @@ matrix:
     # next LTS and master is next most important
     - node_js: "6"
       env: DEPLOY_VERSION=testing
-    # still in LTS maintenance until fall 2016 (also still in wide use)
```
iarna: We can't drop 0.12 or 0.10. Dropping support is a breaking change and 2.x can't take breaking changes.

iarna: It looks like test failures are coming from a transitive dependency of `npm-registry-couchapp`, I'm prepared to give them a pass for the time being to get this out the door.

iarna (changes requested):
Hey, if you can just remove the commit that changes our test matrix I think this is fine.

watilde:
Thx for the review comments. I've reverted the last commit that removed `v0.10` and `v0.12`.

.travis.yml
watilde: I just opened PR to `npm-registry-couchapp` to fix this: https://github.com/npm/npm-registry-couchapp/pull/273

PR #18812, opened 2017-10-12
docs(version): fix git add in version script

phra:
`git add -A dist` is equivalent to `git add -A` or `git add dist` so we should choose one of them. i suggest to change it to `git add -A` supposing that every temporary files is alredy ignored via `.gitignore`

### DIFF
--- a/doc/cli/npm-version.md
+++ b/doc/cli/npm-version.md
@@ -68,7 +68,7 @@ Take the following example:
 
     "scripts": {
       "preversion": "npm test",
-      "version": "npm run build && git add -A dist",
+      "version": "npm run build && git add -A",
       "postversion": "git push && git push --tags && rm -rf build/temp"
     }
 


doc/cli/npm-version.md
```diff
@@ -68,7 +68,7 @@ Take the following example:
 
     "scripts": {
       "preversion": "npm test",
-      "version": "npm run build && git add -A dist",
+      "version": "npm run build && git add -A",
```
iarna, line 71: The `git add` was correct. If you look at the explanation below this it talks about ONLY adding the things in the `dist` direectory.

iarna (changes requested):
Thanks for keeping a close eye on things. I'm going to close this because the existing example was correct.

doc/cli/npm-version.md
phra: i think that `git add dist` is enough, without the `-A` switch.
EDIT: at least with git v2.x.x where the `git-add(1)` also removes entries from the index.

PR #18823, opened 2017-10-12
fix: spelling of the word authenticator

tmcw:
Does what it says on the tin: the message that asks a user for an OTP has a typo, and this fixes that typo.

### DIFF
--- a/lib/auth/legacy.js
+++ b/lib/auth/legacy.js
@@ -25,7 +25,7 @@ module.exports.login = function login (creds, registry, scope, cb) {
       return profile.adduser(username, email, password, {registry: registry})
     }).catch((err) => {
       if (err.code === 'EOTP' && !auth.otp) {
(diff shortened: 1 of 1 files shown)


lib/auth/legacy.js
```diff
@@ -25,7 +25,7 @@ module.exports.login = function login (creds, registry, scope, cb) {
       return profile.adduser(username, email, password, {registry: registry})
     }).catch((err) => {
       if (err.code === 'EOTP' && !auth.otp) {
-        return read.otp('Authenicator provided OTP:').then((otp) => {
+        return read.otp('Authenticator provided OTP:').then((otp) => {
```
iarna, line 28: Ouch, I thought I'd gotten all of those typos. My fingers really don't like the `t` in that word.

iarna (approved):
Looks good, thank you! It'll be in the next release. =)

PR #18845, opened 2017-10-14
The great tfa → 2fa rename

iarna:
Plus we rename `token delete` to `token revoke`.

### DIFF
--- a/lib/token.js
+++ b/lib/token.js
@@ -15,17 +15,17 @@ module.exports = token
 
 token.usage =
   'npm token list\n' +
-  'npm token delete <tokenKey>\n' +
+  'npm token revoke <tokenKey>\n' +
   'npm token create [--read-only] [--cidr=list]\n'
 
-token.subcommands = ['list', 'delete', 'create']
+token.subcommands = ['list', 'revoke', 'create']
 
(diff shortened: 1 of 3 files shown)


lib/token.js
```diff
@@ -15,17 +15,17 @@ module.exports = token
...
-token.subcommands = ['list', 'delete', 'create']
+token.subcommands = ['list', 'revoke', 'create']
 
 token.completion = function (opts, cb) {
   var argv = opts.conf.argv.remain
 
   switch (argv[2]) {
     case 'list':
-    case 'delete':
+    case 'revoke':
```
zkat, line 28: I think you should accept both here. This would be a breaking change otherwise, and I think it's fine to have `delete` as an alias.

zkat (changes requested):
Generally LGTM but I'd rather not have this be a breaking change if we can help it?

lib/token.js
iarna: oh oops, I didn't mean to ditch it entirely!

iarna: ah, this is tab completion. I'm ok with not tab completing delete.

zkat (approved):

PR #18846, opened 2017-10-14
Correctly save transitive dependencies

iarna:
Direct deps get `package.json` and `package-lock.json`.  Transitive deps get `package-lock.json`.

This is a first step to ditching the depth=0 on `npm update`.  The other thing we need to do before we do that is fix a bug where asking to update a named transitive dependency can cause an infinite loop in `outdated`.

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -132,6 +132,7 @@ var decomposeActions = require('./install/decompose-actions.js')
 var validateTree = require('./install/validate-tree.js')
 var validateArgs = require('./install/validate-args.js')
 var saveRequested = require('./install/save.js').saveRequested
+var saveShrinkwrap = require('./install/save.js').saveShrinkwrap
 var getSaveType = require('./install/save.js').getSaveType
 var doSerialActions = require('./install/actions.js').doSerial
 var doReverseSerialActions = require('./install/actions.js').doReverseSerial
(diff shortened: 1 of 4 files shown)


lib/install.js
```diff
@@ -215,23 +217,31 @@ function Installer (where, dryrun, args) {
   this.progress = {}
   this.noPackageJsonOk = !!args.length
   this.topLevelLifecycles = !args.length
-  this.dev = npm.config.get('dev') || (!/^prod(uction)?$/.test(npm.config.get('only')) && !npm.config.get('production')) || /^dev(elopment)?$/.test(npm.config.get('only'))
-  this.prod = !/^dev(elopment)?$/.test(npm.config.get('only'))
-  this.rollback = npm.config.get('rollback')
-  this.link = npm.config.get('link')
-  this.global = this.where === path.resolve(npm.globalDir, '..')
+  this.dev = opts.dev != null ? opts.dev : npm.config.get('dev') || (!/^prod(uction)?$/.test(npm.config.get('only')) && !npm.config.get('production')) || /^dev(elopment)?$/.test(npm.config.get('only'))
```
zkat: Can I ask that we parenthesize, and maybe split between lines, all these different conditionals? I can never tell how it's actually supposed to parse when you do ternary + or + and all in on line >_<

zkat: this change otherwise looks good

zkat (approved):
LGTM -- just a small readability change that you can do if you want.

lib/install.js
iarna: er wow, yeah, that looks awful…

PR #18852, opened 2017-10-16
Check validity of version before making fake pkg

### LINKED ISSUE #17858: Cannot read property '0' of undefined

  - [✓] npm is crashing.

#### What's going wrong?

```
~/P/dreamer-api *master> npm install --save caminte
npm WARN eslint-config-airbnb@15.0.2 requires a peer of eslint-plugin-jsx-a11y@^5.1.1 but none was installed.
npm WARN eslint-config-airbnb@15.0.2 requires a peer of eslint-plugin-react@^7.1.0 but none was installed.
npm WARN dreamer-api@0.1.0 No repository field.

npm ERR! Cannot read property '0' of undefined

npm ERR!     /Users/noah/.npm/_logs/2017-07-19T19_09_33_702Z-debug.log
```

and

```
~/P/dreamer-api *master> npm remove --save sequelize
npm WARN eslint-config-airbnb@15.0.2 requires a peer of eslint-plugin-jsx-a11y@^5.1.1 but none was installed.
npm WARN eslint-config-airbnb@15.0.2 requires a peer of eslint-plugin-react@^7.1.0 but none was installed.
npm WARN dreamer-api@0.1.0 No repository field.

npm ERR! Cannot read property '0' of undefined

npm ERR!     /Users/noah/.npm/_logs/2017-07-19T19_11_35_695Z-debug.log
```

I do not know.

### supporting information:

 - `npm -v` prints:

`5.3.0`

 - `node -v` prints:

`v8.1.4`

`https://registry.npmjs.org/`

 - Windows, OS X/macOS, or Linux?:

macOS

 - Network issues:
   - [✓] None
 - Container:
   - [✓] None

**EDIT:** Seems to only happen with node 8.1.4

joshclow:
Cheking the version against semver here ensures that a fake pkg with URL/path data in its version field doesn’t get created (if that happens, then that will propagate down into the version field of any refreshed package.json, overwriting whatever version should actually be there)

This resolves #17858 without violating the spec for package-lock.json (mea culpa!)

### DIFF
--- a/lib/install/inflate-shrinkwrap.js
+++ b/lib/install/inflate-shrinkwrap.js
@@ -13,6 +13,7 @@ const npm = require('../npm.js')
 const realizeShrinkwrapSpecifier = require('./realize-shrinkwrap-specifier.js')
 const validate = require('aproba')
 const path = require('path')
+const semver = require('semver')
 
 module.exports = function (tree, sw, opts, finishInflating) {
   if (!fetchPackageMetadata) {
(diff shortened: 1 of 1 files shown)


iarna:
This would disable fake children for git urls and http tarball urls… which ok, we can do as an interum measure, though I'd like to restore it.

I'm a bit confused as to how this is related to #17858, however and maybe I just missed that discussion? The bug reported in #17858 involves unbuild being called on a package entry where `name` isn't defined, which… I don't get how a fake child in this case would have an empty name field because the version was non-semver?

iarna:
Also, I encourage you to drop by the https://package.community/ discord server and we can chat real time about this, 'cause I'd like to get this out. =)

joshclow:
I'll jump in there as soon as I can. Mostly, I've been looking at making minimally invasive changes on principle, but I'm 100% prepared to believe this one really wants something more sophisticated.

joshclow:
Proper fix was in refresh-package-json.js

PR #18881, opened 2017-10-17
install: support legacy npa output for type checking

iarna:
Current npa will pass `registry: true` for specs that will ultimately be
fetched from a registry. Older npa didn't do this.

When dealing with a `node_modules` that was created with older versions of
npm (and thus older versions of npa) we need to gracefully handle these
older spec entries. Failing to do so results in us treating those packages
as if they were http remote deps, which results in invalid lock files.

This fixes the issue where an `npm i` or `npm shrinkwrap` on a directory with a `node_modules` created by `npm@3` or `npm@4` would result in a lockfile that had http URLs for all the modules, instead of versions.

### DIFF
--- a/lib/install/action/extract.js
+++ b/lib/install/action/extract.js
@@ -16,6 +16,7 @@ let pacoteOpts
 const path = require('path')
 const localWorker = require('./extract-worker.js')
 const workerFarm = require('worker-farm')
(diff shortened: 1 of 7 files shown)


zkat (approved):
🐑 lgtm

PR #18883, opened 2017-10-18
npm: Capture logging from log events on the process global

iarna:
This allows npm to use npmlog to report logging from libraries like npm-profile.

### DIFF
--- a/lib/npm.js
+++ b/lib/npm.js
@@ -24,6 +24,16 @@
   var npm = module.exports = new EventEmitter()
   var npmconf = require('./config/core.js')
   var log = require('npmlog')
+  var inspect = require('util').inspect
+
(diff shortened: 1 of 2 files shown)


zkat (approved):
Exciting! 👍

zkat (changes requested):
Actually, could I get a small test for this? <3

iarna:
@kat what did you have in mind? Make sure the event listener is setup? I dunno, convince me there's value in a test for this?

iarna:
mm, I will say that it should probably wrap itself in a try/catch because we don't want bogus input to set npm on fire, so a test for _that_ wouldn't be amiss.

zkat (approved):

PR #18884, opened 2017-10-18
Profile UX improvements

iarna:
Two small things:

1. Stop saying "password set to undefined" when setting your password.
2. If the 2fa endpoint doesn't respond with an otpauth URL, inspect the result instead of printing it, as it may be an object. (This should never happen during normal function of the registry.)


zkat (approved):
LGTM

PR #18909, opened 2017-10-19
Adding back in bin/read-package-json.js as it is still used by https:…

### LINKED ISSUE #18524: install.sh fails because bin/read-package-json.js was removed

  - [x] npm is crashing.

#### What's going wrong?
https://github.com/npm/npm/blob/latest/scripts/install.sh fails because bin/read-package-json.js was removed. 

The file was removed in commit https://github.com/npm/npm/commit/9462e5d9cfbaa50218de6d0a630d6552e72ad0a8 . However, that file is referenced at the following line https://github.com/npm/npm/blob/e2745004628e595b5017588aae794638568ef0fc/scripts/install.sh#L211 
curl -L https://www.npmjs.org/install.sh | sh

mikehenrty:
…//npmjs.org/install.sh, fixes #18524

This reverts commit 9462e5d9cfbaa50218de6d0a630d6552e72ad0a8. #16547

### DIFF
--- a/bin/read-package-json.js
+++ b/bin/read-package-json.js
@@ -1,0 +1,24 @@
+var argv = process.argv
+if (argv.length < 3) {
+  console.error('Usage: read-package.json <file> [<fields> ...]')
+  process.exit(1)
+}
+
+var file = argv[2]
+var readJson = require('read-package-json')
+
+readJson(file, function (er, data) {
+  if (er) throw er
+  if (argv.length === 3) {
+    console.log(data)
+  } else {
+    argv.slice(3).forEach(function (field) {
+      field = field.split('.')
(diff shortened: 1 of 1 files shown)


zkat:
https://github.com/npm/npm/commit/5ee55dc71b8b74b8418c3d5ec17483a07b3b6777#diff-80423f13d45b53af35790ffc6009be4d already fixed this separately, and it looks like it's just an issue with the new `install.sh` not having been deployed to production properly. Stand by, as I've pinged the registry folks.

andelmad (approved):

zkat:
I poked the registry folks again. This PR isn't really necessary 'cause the issue is already fixed.

PR #18921, opened 2017-10-22
put auto-pruning feature behind --no-auto-prune or auto-prune config setting

ghost:
This patch adds an `auto-prune` configuration setting that when set to `false` (either through `--no-auto-prune` or `npm config set auto-prune false`) disables the removal of unused packages.

The motivating use-cases for this patch are documented in this issue: https://github.com/npm/npm/issues/16853

Related no-delete-symlinks feature: https://github.com/npm/npm/pull/18922

Issue #17257, opened 2017-06-16
npm i deletes all the symlinked repos


yawhide:
#### I'm opening this issue because:

  - [ ] npm is crashing.
  - [x] npm is producing an incorrect install.
  - [x] npm is doing something I don't understand.
  - [ ] Other (_see below for feature requests_):

#### What's going wrong?
When I do npm link B inside repo A, npm deletes all the node_modules inside B.

#### How can the CLI team reproduce the problem?
- git clone repo A and B
- call `npm i` inside repo A
- call `npm i` inside repo B
- repo A has a dependency on repo B (but nothing is symlinked yet)
- call `npm link` in repo B.
- go to repo A, call `npm link B`.
- call `npm i` in repo A. <---- this deletes a/node_modules AND b/node_modules but it shouldn't delete b/node_modules.
- go back to repo B and observe that the node-modules folder is empty.
- the symlink to repo B from repo A is still preserved (thankfully) so all you have to do is go back to repo B and call `npm i`.

this is especially annoying if I add a dependency in repo A, call `npm i lodash` for example, which then deletes all the symlinked repos...

### supporting information:

 - `npm -v` prints: 5.0.3
 - `node -v` prints: v8.1.2
 - `npm config get registry` prints: https://registry.npmjs.org/
 - Windows, OS X/macOS, or Linux?: Linux/Ubuntu
 - Network issues:
   - Geographic location where npm was run:
   - [ ] I use a proxy to connect to the npm registry.
   - [ ] I use a proxy to connect to the web.
   - [ ] I use a proxy when downloading Git repos.
   - [ ] I access the npm registry via a VPN
   - [ ] I don't use a proxy, but have limited or unreliable internet access.
 - Container:
   - [ ] I develop using Vagrant on Windows.
   - [ ] I develop using Vagrant on OS X or Linux.
   - [ ] I develop / deploy using Docker.
   - [ ] I deploy to a PaaS (Triton, Heroku).

yawhide:
basically if you have a symlink in node_modules and you call npm i, npm i deletes the current node_modules AND goes into the symlink'ed node_modules and deletes it. VERY annoying.

onury:
Does this happen on older versions? I'm tired of re-installing removed packages and re-linking..

yawhide:
i cant repro this in npm v4.2.0

ile:
It seems that npm 5.0.3 just deleted all the git repositories in node_modules, symlinked or not. I did `npm i some-module`.

[...]

Issue #17287, opened 2017-06-18
npm install after npm link deletes linked module dependencies in it's node_modules folder


HosseinAgha:
#### I'm opening this issue because:

  - [ ] npm is crashing.
  - [x] npm is producing an incorrect install.
  - [ ] npm is doing something I don't understand.
  
- [ ] Other (_see below for feature requests_):
#### What's going wrong?

npm installing should not do anything with linked package modules but it completely messes up linked package dependency tree and installed modules.  
In other words you cannot do anything in npm 5 if you rely on linked packages.

#### How can the CLI team reproduce the problem?

npm link to package that you already `npm installed 

### supporting information:

 - `npm -v` prints: 5.0.3
 - `node -v` prints: v6.10.3
 - `npm config get registry` prints:
 - Windows, OS X/macOS, or Linux?: macOS

This is probably a duplicate of https://github.com/npm/npm/issues/10343 but as it is an old issue and none of the project maintainers noticed that this issue is popped up again in version 5 I'm opening a new issue here. I think some people have not faced this issue, because they have not `npm installed` in their main package yet.

lejeunerenard:
This is probably a duplicate of #16788.

HyperBrain:
@lejeunerenard I think this one is a superset of #16788 . The major issue here is, that you cannot do any module development and local testing with npm 5 because it just removes the linked modules as soon as you do a "npm install".

marijnh:
(Still happening with 5.3. My multi-module workflow which worked fine with 3.x and 4.x has become very painful since 5.x, due to unpredictable changes in sub-module's `node_modules` directory. The most obvious/consistent one is that `node_modules/.bin` is deleted in the linked modules every time I run `npm install` on the outer module.)

HosseinAgha:
I don't understand the existence of this bug after such a long time. maybe link is not a priority for npm community?!

bertho-zero:
Same here

Legogris:
Still an issue with npm 5.4.2. This really breaks the workflow.

jimsugg:
Our team relies on linking to test a set of interacting packages. In this regard, npm 5 is a nightmare. We have no choice but to revert to npm 4 until we have the time to figure out how to adjust our workflow, or find an acceptable workaround. Note this is one of many issues open related to problems related to linking introduced in npm 5.

HosseinAgha:
[...]

PR #18922, opened 2017-10-22
configuration setting to never delete symlinks

ghost:
This patch turns off pruning for symlinks and packages inside of symlinked directories.

This patch would mitigate the effects of https://github.com/npm/npm/issues/17257
and may still be wanted behavior once that bug is fixed.

Related no-auto-prune feature: https://github.com/npm/npm/pull/18921

### DIFF
--- a/doc/misc/npm-config.md
+++ b/doc/misc/npm-config.md
@@ -284,6 +284,14 @@ This is a list of CIDR address to be used when configuring limited access tokens
 If false, never shows colors.  If `"always"` then always shows colors.
 If true, then only prints color codes for tty file descriptors.
 
+### delete-symlinks
+
+* Default: false
+* Type: Boolean
+
+If true, do not delete symlinks in `node_modules` nor any `node_modules`
+directories inside a symlinked directory during tree pruning.
+
 ### depth
 
 * Default: Infinity
--- a/lib/config/defaults.js
+++ b/lib/config/defaults.js
@@ -131,6 +131,7 @@ Object.defineProperty(exports, 'defaults', {get: function () {
     cidr: null,
 
     color: true,
+    'delete-symlinks': true,
     depth: Infinity,
     description: true,
     dev: false,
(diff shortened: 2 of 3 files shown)


fstoerkle:
@substack looking at the Travis CI build, I guess you need to add a default for this new option in https://github.com/npm/npm/blob/latest/lib/config/defaults.js

zkat:
Gonna hold off on this PR until @iarna is back from conference stuff so we can actually talk about it.

jimsugg:
See also #17287. Please expedite this if possible.

billiegoose:
OH MY GOD - this is why I got so frustrated and blew away all of my repos and reinstalled them all from scratch on Friday... I was going crazy. Here is what was going on:
1. git clone foo; cd foo
2. npm i
3. npm link foo-common # our development copy of foo-common is developed in tandem with foo
4. npm i --save new-dependency

and now the "npm link" to foo-common is messed up.

I'm not 100% sure this is the exact steps and the cause of the problem... but I will test on Monday when I get into work.

jbcpollak:
Steps 2, 3, 4 are repeatable and consistent no matter what you do prior to
them. In fact step 4 could just be `npm i` - any `npm install` after
linking will blow away your locally linked repos.

>

AndreiShostik:
@zkat, @iarna was back from a conference?
guys, please, review.
it turned into a very annoying thing to come up with workarounds to return back symlinks

zkat:
@substack the latest release added a mode where autoprune only happens if you're running with package locks/shrinkwrap enabled. I believe this solves your specific use-case without adding this flag.

We're going to have a better workflow around links in general when `npm@6` comes out, and overhauling link is literally the reason for that breaking change.

Are you ok enough with having this closed for now?

ljharb:
@zkat even with a lockfile i don’t want the pruning behavior; it breaks anything i set up in a postinstall script, links or not. It’d be really great to get this merged.

bluelovers:
need this lol

zkat:

PR #18943, opened 2017-10-24
fix: Output package.json with original line endings

### LINKED ISSUE #17161: Bug: npm is platform opinionated concerning line endings

  - [ ] npm is crashing.

#### What's going wrong?
When having a checkout of a package.json + package-lock.json on a Windows machine with git "core.autocrlf=true", then all files will have CRLF line endings. Now, when running "npm install", then npm thinks that the line endings of package.json + package-lock.json should be replaced with pure LF. This is bad, as git now thinks there are changes to the files, while in fact there are none (only line endings):

image

image

And the same is true for IntelliJ:
image

### supporting information:

 - `npm -v` prints: 5.0.3
 - `node -v` prints: v6.10.3
 - Windows, OS X/macOS, or Linux?: Windows

nwoltman:
Same for `package-lock.json` and `npm-shrinkwrap.json`.

Detect the line endings used in the original file and use the same line endings when updating the file.

Fixes #17161

### DIFF
--- a/lib/install/save.js
+++ b/lib/install/save.js
@@ -3,6 +3,7 @@
 const createShrinkwrap = require('../shrinkwrap.js').createShrinkwrap
 const deepSortObject = require('../utils/deep-sort-object.js')
 const detectIndent = require('detect-indent')
+const detectLineEnding = require('../utils/detect-line-ending')
 const fs = require('graceful-fs')
 const iferr = require('iferr')
 const log = require('npmlog')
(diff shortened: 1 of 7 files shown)


vanjan:
Is this going to get merged this century?

nwoltman:
@zkat @iarna This should be pretty easy to review. The code is very similar to how indentation is handled.

zkat:
@nwoltman oh no! I'm sorry I didn't see this before I saw the other one. We were gonna go through all our PRs for the next release and stacks work different than queues :<

I'm really happy that you wrote a test for this, and that's awesome -- would you be willing to rebase the test onto the patch for the other detection PR? I'll definitely take that and I'm really grateful you took the time to do this. <3

nwoltman:
Thank you for your kind response @zkat. I understand why the other PR showed up on your radar before this one. I'll rebase these changes using the `release-next` branch where the other PR landed and target that branch.

PR #18953, opened 2017-10-25
Remove ES6 from unsupported error logging

deployable:
NPM throws an error on old node versions that don't support template strings, rather than reporting the unsupported information to the user

### DIFF
--- a/lib/utils/unsupported.js
+++ b/lib/utils/unsupported.js
@@ -24,7 +24,7 @@ exports.checkForBrokenNode = function () {
     supportedNode.forEach(function (rel) {
       if (semver.satisfies(nodejs.version, rel.ver)) {
         console.error('Node.js ' + rel.ver + " is supported but the specific version you're running has")
-        console.error(`a bug known to break npm. Please update to at least ${rel.min} to use this`)
+        console.error('a bug known to break npm. Please update to at least ' + rel.min + ' to use this')
         console.error('version of npm. You can find the latest release of Node.js at https://nodejs.org/')
         process.exit(1)
       }


deployable:
[Currently throws ](https://stackoverflow.com/q/46922146/1318694)

```
/home/vagrant/.nvm/v0.10.32/lib/node_modules/npm/lib/utils/unsupported.js:27
        console.error(`a bug known to break npm. Please update to at least ${r
                      ^
SyntaxError: Unexpected token ILLEGAL
    at Module._compile (module.js:439:25)
    at Object.Module._extensions..js (module.js:474:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:312:12)
    at Module.require (module.js:364:17)
    at require (module.js:380:17)
    at /home/vagrant/.nvm/v0.10.32/lib/node_modules/npm/bin/npm-cli.js:19:21
    at Object.<anonymous> (/home/vagrant/.nvm/v0.10.32/lib/node_modules/npm/bin/npm-cli.js:92:3)
    at Module._compile (module.js:456:26)
    at Object.Module._extensions..js (module.js:474:10)
```

deployable:
@KenanY @iarna Is there any reason why this isn't being merged? There's been a number of dupes to it and it seems to meet the policy from: https://github.com/npm/npm/blob/latest/changelogs/CHANGELOG-4.md

>     On the flip side, we'll continue to (happily!) accept patches that address regressions seen when running the CLI with Node.js 0.12.
> 
> What this doesn't mean:
> 
>     The CLI is going to start depending on ES2015+ features. npm continues to work, in almost all cases, all the way back to Node.js 0.8, and our long history of backwards compatibility is a source of pride for the team.
>     We aren't concerned about the problems of users who, for whatever reason, can't update to newer versions of npm. As mentioned above, we're happy to take community patches intended to address regressions.

zkat (approved):
Oh hey, thanks. Yeah, we can totally use modern Node features but we should probably avoid that for `lib/unsupported.js` which is what reports that this is even an issue.

PR #18964, opened 2017-10-26
chore: add support for node 9.x

MylesBorins:
Seems like something that needs to be done 😄


depoulo:
Also, the error message should say "to a newer version of **npm**". Maybe you can just sneak this in here?
grafik

**Update**: Sorry my bad. In many cases, the error message is right. It's still wrong in this case though, but that should never happen in the first place.

graingert:
you'll need to update the ci configs too...

graingert:
@MylesBorins https://github.com/npm/npm/pull/19025

MylesBorins:
@depoulo this patch should fix the exact problem

@graingert I've re targeted this to canary so no need to worry about the zlib

I've updated the travis yaml in another commit so you might be able to close #19025

ilovezfs:
Does anyone have any idea what the ETA is for an npm 5.5.2 with support for node 9?

ilovezfs:
@iarna @zkat any idea what the impediments are to a 5.5.2 release with the node 9 patches? That would be _very_ helpful for the Homebrew formula, which is stuck on node 8 since it needs to work with the stock upstream npm.

MylesBorins:
this is fixed by 6caf23096

mightyiam:
Hurrah! :fireworks: :tada: and thanks!

PR #19025, opened 2017-11-01
support node 9, also update CI

graingert:


graingert:
@MylesBorins looks like node 9 doesn't pass the build

chrmoritz:
That's because https://github.com/npm/npm/commit/4ca6958196ae41cef179473e3f7dbed9df9a32f1 is required for node 9 support too (and this commit is only available in npmc canary atm, targeting the release-next branch instead of latest should work too).

graingert:
Ok will try that.

ilovezfs:
What's the status here?

graingert:
@ilovezfs see https://github.com/npm/npm/pull/18964

ilovezfs:
:tada:

PR #19030, opened 2017-11-01
install: fix the .hooks script does not running.

zhrln:
in npm@5.0.3 , `node_modules/.hook/preinstall` It works, but in npm@5.5.1 `node_modules/.hook/preinstall` It doesn't work.


zhrln:
Hook scripts can run normally in the earlier version **5.0.4**, but `npm i` can’t execute hook scripts from version **5.1.0**. 
The major cause is that, if currently running is `node_modules/.hooks/preinstall` and the `preinstall` scripts are not configured in `package.json`, the judgment logic result is `pkg.scripts` don’t run `preinstall` and return to `resolve`(or `cb`).

Issue #18658, opened 2017-09-28
npm 5.4.2 - removes '@' from scoped package subfolder names


jdbranham:
#### I'm opening this issue because:

  - [ X] npm is crashing.
  - [ ] npm is producing an incorrect install.
  - [ ] npm is doing something I don't understand.
  - [ ] Other (_see below for feature requests_):

#### What's going wrong?  
After updating to the latest version of npm [5.4.2], I've began getting this error -
npm ERR! cb() never called!

I know it is because of my scoped package name - it is removing the '@' symbol from the folder names when trying to do a npm publish or npm pack.

This path -
'C:\ProgramData\Git\ngx-menu\dist\savantly\ngx-menu.es5.js'

Should actually be this path -
'C:\ProgramData\Git\ngx-menu\dist\@savantly\ngx-menu.es5.js'

Using node version - v6.11.3

#### How can the CLI team reproduce the problem?  

Create a scoped package with a subfolder starting with '@'.  
do `npm publish`  

example package.json -   
```
{
  "name": "@savantly/ngx-menu",
  "version": "1.1.12",
  "description": "Menu service and component for an Angular app",
  "main": "bundles/ngx-menu.umd.js",
  "homepage": "https://github.com/savantly-net/ngx-menu",
  "repository": {
    "type": "git",
    "url": "https://github.com/savantly-net/ngx-menu.git"
  },
  "scripts": {
    "ng": "ng",
    "start": "ng serve",
    "build": "ng build",
    "build:dev": "ng build --app dev",
    "build:lib": "ng-packagr",
    "test-watch": "ng test",
    "test": "karma start --single-run",
    "lint": "ng lint",
    "e2e": "ng e2e",
    "clean": "rimraf .ng_build dist"
  },
  "keywords": [
    "Sprout",
    "Menu",
    "Angular"
  ],
  "author": "Jeremy@Savantly.net",
  "license": "MIT",
  "peerDependencies": {
    "@angular/animations": "^4.4.3",
    "@angular/cdk": "^2.0.0-beta.11",
    "@angular/common": "^4.4.3",
    "@angular/core": "^4.4.3",
    "@angular/flex-layout": "^2.0.0-beta.9",
    "@angular/forms": "^4.4.3",
    "@angular/http": "^4.4.3",
    "@angular/material": "^2.0.0-beta.11",
    "@angular/router": "^4.4.3",
    "@savantly/ngx-security": "^1.1.12",
    "rxjs": "^5.4.3",
[...]

PR #19033, opened 2017-11-01
pack: fix inclusion of files with @ signs

zkat:
Fixes: #18658
Credit: @zkat


isaacs:
LGTM!  I'll fix that bug in node-tar where it tries to parse a directory as if it's a tarball if it has an `@` sign in it, but in the meantime, this is a 💯 workaround. 🐑

PR #19057, opened 2017-11-04
install: fix: can’t execute .hooks scripts

zhrln:
Hook scripts can run normally in the earlier version **5.0.4**, but `npm i` can’t execute hook scripts from version **5.1.0**. 
The major cause is that, if currently running is `node_modules/.hooks/preinstall` and the `preinstall` scripts are not configured in `package.json`, the judgment logic result is `pkg.scripts` don’t run `preinstall` and return to `resolve`(or `cb`).

### DIFF
--- a/node_modules/npm-lifecycle/index.js
+++ b/node_modules/npm-lifecycle/index.js
@@ -42,8 +42,6 @@ function lifecycle (pkg, stage, wd, opts) {
       delete pkg.scripts.prepublish
     }
 
-    if (!pkg.scripts[stage]) return resolve()
-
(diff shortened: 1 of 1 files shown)


ties-s:
Wouldn't it be better to create a pull request in the lifecycle package? The problem lies in there.

zhrln:
@ties-s You are right, thanks very much.

ties-s:
Already made a new pull request over there since I also need this feature: https://github.com/npm/lifecycle/pull/9

zhrln:
@ties-s 
I moved this PR to npm/lifecycle repository. Thanks for helping me make a PR, please close your PR.

PR #19090, opened 2017-11-07
Fix npm command typo

orbxball:
There's a typo in the command list file `lib/config/cmd-list.js`.
For now, I've just found that it will be used in the help message.
I don't know whether it will be used in another place or not.
Thus, I revise it in case that `cmd-list.js` is required in another place and get unexpected behavior.

### DIFF
--- a/lib/config/cmd-list.js
+++ b/lib/config/cmd-list.js
@@ -22,7 +22,7 @@ var affordances = {
   'la': 'ls',
   'll': 'ls',
   'verison': 'version',
-  'isntall': 'install',
+  'install': 'install',
   'dist-tags': 'dist-tag',
   'apihelp': 'help',
   'find-dupes': 'dedupe',


kenany:
This is an intentional typo, to serve as an alias to `install`. See https://github.com/npm/npm/issues/2933.

orbxball:
But just as most of the people said in #2933, it will misguide when developers first see this.
Thus, I'm not the only person that is confused by this and dig into deeply finding the reason.
(It cost me about 2 hr on it to find which is the correct file that matters...)

`npm` should be focused on correctness, not the sloppiness or carelessness of developers.
As you say if `npm` should take care of this typo, why other commands don't have intentional typo?

Otherwise, we can merge this and then and add `isntall` back so that it will quickly be recognized as the intentional typo of `install` which is next to `isntall`.

kenany:
> why other commands don't have intentional typo?

See `npm verison` (`version`) and `npm udpate` (`update`) for two other examples.

a5sk4s:
In lib/npm.js is the following code `var littleGuys = [ 'isntall', 'verison' ]`. I do not fully understand the reason for it, but if `udpate` is another of those 'intentional typos', why is it not listed there?

edmorley:
Perhaps it would help to have a concept of "hidden aliases", which work as normal but just don't show up in the aliases list printed in the help text?

zkat:
This is totally on purpose, and is a pretty known-in-some-circles alias that occasionally saves someone some trouble. Good eye, though ;)

Generally, this section of `affordances` is used for exactly this sort of thing!

Because it's not an "official" spelling of the command, and more of a safety net that is part of our UX, I'm going to close this, but thanks for taking a look and taking the time to submit this patch!

PR #19099, opened 2017-11-08
Switch to using stand alone module for finding the npm prefix/project directory

iarna:
This is a part of the project to extern the configuration. It turns out there are a distressing number of modules out there that mostly implement this, but don't actually implement all of our quirks, thus the new module.


zkat (approved):
ligtum! 🐑

iarna:
Merged as 829893d61

PR #19104, opened 2017-11-08
config: expose "node-options" passthrough for lifecycle scripts

bmeck:
intent here is to open a path to use `--inspect` with `npm run`

requires https://github.com/npm/lifecycle/pull/7


iarna:
I'll be bringing in the updated `npm-lifecycle` soon and I'll land this then.

iarna:
Merged into `release-next` as a809509b5

PR #19118, opened 2017-11-09
doc: use valid JSON in example for bundledDependencies

charmander:


Saturate:
Looks fine, and it's a nice fix.

Issue #18789, opened 2017-10-11
command to only generate package-lock.json and not install any dependencies


rarkins:
#### What's the feature?

A CLI option (e.g. for `npm install`) or new command whose purpose is to simply generate the lock file and *not* check or install `node_modules`.

#### What problem is the feature intended to solve?

There are use cases where we need to simply generate a lock file, such as to validate if an existing lock file is up-to-date, and do not wish to download the actual packages.

#### Is the absence of this feature blocking you or your team? If so, how?

No but it's inefficient to install a lot of packages when you don't need to.

#### Is this feature similar to an existing feature in another tool?

No

#### Is this a feature you're prepared to implement, with support from the npm CLI team?

I could give it a shot if someone can give a pointer of where to modify.

alopezsanchez:
Hello! Is this feature viable to be merged?

I'd like to contribute to the project and I could do It if you don't have any inconvenience.

Cheers!

iarna:
@alopezsanchez This has been on our roadmap for a while, so I'd be happy to see it done. Looking at this now, I think it's actually a pretty good intro project.

I think all of your changes will be in `lib/install.js` to `Installer.prototype.run`. You'll be doing something very similar to what `dryrun` does, except that you'll be running the `saveToDependecies` step to update the package-lock.

If you take a crack at this, please drop by #npm on https://package.community/ and chat with us when you've got something (or if you run into any problems—we'd be happy to hel‼)

alopezsanchez:
@iarna thank you so much!

I take this then, I'm pretty sure I will have have doubts, so see you at Discord!

rarkins:
Great work @alopezsanchez

PR #19141, opened 2017-11-12
Feature/package lock only

alopezsanchez:
Added `package-lock-only` config option to allow only generate the `package-lock.json` when install, instead of checking `node_modules` and downloading dependencies.

I added a new argument to `Installer`, so some files needed to be updated to make tests work.

Please, review it! :smiley: 

Ref: https://github.com/npm/npm/issues/18789


alopezsanchez:
Closed because I prefer to make another one with branch `release-next` rebased.

PR #19142, opened 2017-11-13
ci: Update node versions

alberto:
Set Node 8 as most important, add Node 9 as master and remove Node 7.


alberto:
Sorry, I didn't see https://github.com/npm/npm/pull/18964 already existed.

PR #19157, opened 2017-11-14
Support for noProxy configuration

### LINKED ISSUE #18350: Does npm support “no_proxy” option?
My team need to set the proxy for npm while ignoring some Urls(internal websites), but I can't find "no_proxy" option in npm-config(https://docs.npmjs.com/misc/config). 

Does npm support "no_proxy" now?

Saturate:
This minor commit adds support for the noProxy configuration.

All whats really needed is passing the option to the `pacote` package,
this will pass it to `make-fetch-happen`.

pacote: https://github.com/zkat/pacote/blob/latest/lib/util/opt-check.js#L26
make-fetch-happen: https://github.com/zkat/make-fetch-happen#opts-no-proxy

Fixes #18350, #7168

### DIFF
--- a/doc/misc/npm-config.md
+++ b/doc/misc/npm-config.md
@@ -663,6 +663,13 @@ The registry you want to send cli metrics to if `send-metrics` is true.
 
 The node version to use when checking a package's `engines` map.
 
+### no-proxy
+
+* Default: null
+* Type: String or Array
+
+A comma-separated string or an array of domain extensions that a proxy should not be used for.
+
 ### offline
 
 * Default: false
--- a/lib/config/defaults.js
+++ b/lib/config/defaults.js
@@ -310,7 +310,7 @@ exports.types = {
   message: String,
   'metrics-registry': [null, String],
   'node-version': [null, semver],
-  'no-proxy': [null, String, Array],  
+  'no-proxy': [null, String, Array],
   offline: Boolean,
   'onload-script': [null, String],
   only: [null, 'dev', 'development', 'prod', 'production'],
--- a/lib/config/defaults.js
+++ b/lib/config/defaults.js
@@ -310,6 +310,7 @@ exports.types = {
   message: String,
   'metrics-registry': [null, String],
   'node-version': [null, semver],
+  'no-proxy': [null, String, Array],  
   offline: Boolean,
   'onload-script': [null, String],
   only: [null, 'dev', 'development', 'prod', 'production'],
--- a/lib/config/defaults.js
+++ b/lib/config/defaults.js
@@ -191,6 +191,7 @@ Object.defineProperty(exports, 'defaults', {get: function () {
     'progress': !process.env.TRAVIS && !process.env.CI,
     proxy: null,
     'https-proxy': null,
+    'no-proxy': null,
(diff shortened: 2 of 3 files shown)


Saturate:
I am aware of the issues and will fix them.

jacobjuul:
+1 👍

iarna:
`make-fetch-happen` doesn't use `request`, it uses `http-proxy-agent` and `https-proxy-agent`  in conjunction with `node-fetch-npm`. I don't see mention of this option in the docs for any of those… have you verified that this patch actually works?

Saturate:
No idea where I got that from, maybe it was used in an old `npm` version? I tested it and it worked, but with this I'll just reconfirm it :-) However `make-fetch-happen` has documentation on this feature so I'm pretty sure it works :-) also if `request` is out of the picture.

Saturate:
@iarna Tested again and it works.

Saturate:
Any feedback on this @iarna ?

sebinsua:
Is anybody going to look at this at some point? I'd prefer not to have to switch to `yarn`.

Saturate:
I'm still waiting for this pretty important feature, but there seems to be
no progress or feedback yet.

FabienDehopre:
I just test this PR in my environment where we use a private repo for our private dependencies.
This private repo in inside our network and NPM has to bypass the proxy to access it.
Everything work fine with this PR.
We used to use environment variables to set the proxy configuration (HTTP_PROXY, HTTPS_PROXY and NO_PROXY) but since a few updates, this does not work anymore so this PR is very important for us.

thecontrarycat:
+1 from me.  I need NPM to work with an external proxy and a local registry and we can't use the environment variables because that breaks git (which needs to talk to a local git server) and other things.

jacobjuul:
Eagerly awaiting this as well

Saturate:
@sebinsua as far a I know Yarn does not support this? But maybe I should create a PR for them as well.

sebinsua:
> as far a I know Yarn does not support this? But maybe I should create a PR for them as well.

@Saturate They only support the `NO_PROXY` environment variable in the pre-release of `1.4.1`. They don't yet support a configuration option of `noProxy` and, in fact, setting any of the proxy configuration via config stops the fallback to the environment variables. However, I've been able to achieve what I need by just using the environment variables.

jacobjuul:
@iarna did you have a chance to review this?

Saturate:
@sebinsua thanks, I'll take a look at it soon, if this goes no further as it would help me with a lot of the big corporate environments where they lock most things down, and everything has to be proxied.

wvandeneede:
+1 👍

zkat (approved):
Let's do it 😎 

Thanks for the patch! I added this a whiiiile back to `make-fetch-happen`/`pacote`, but obviously never got around to integrating it with npm itself. Thanks for your patience while we managed to get around to going through all these PRs!

zkat:
For the rest of y'all: npm has supported the _environment variable_ `NO_PROXY` for a while, so if you don't have this patch, you can still set your no-proxy config by using that env var instead of an npm arg (`NO_PROXY="..." npm install ...`)

PR #19160, opened 2017-11-14
feature: `package-lock-only` config option

alopezsanchez:
Added `package-lock-only` config option to allow only generate the `package-lock.json` when install, instead of checking `node_modules` and downloading dependencies.

Ref: https://github.com/npm/npm/issues/18789


iarna:
Thank you, this is awesome!

This has been merged as f94fcbc50 into `release-next` and will be in the next release of npm.

Issue #19151, opened 2017-11-13
npm install corrupts binary executable shebang files


marcosscriven:
#### I'm opening this issue because:

  - [*] npm is producing an incorrect install.

#### What's going wrong?

A pull request was merged to convert any shebang scripts to unix line endings:  https://github.com/npm/npm/pull/16634 

Unfortunately this breaks binary shebang files, such as [Python Pex](https://pex.readthedocs.io/en/stable/whatispex.html) files, which look like this:

```
#!/usr/bin/env python2.7
<binary>
```

Processing such files with `dos2unix` completely corrupts them.

I would be happy to PR a fix, but I was unclear why the original PR made this change happen by default? 

Even if there's an argument for making it the default for publishing, I really don't think npm should be modifying downloaded artifacts during the `build` phase of `install`.

marcosscriven:
Ping @ScottFreeCode

robbieg:
It could also break executable jars for example, if they're made executable by prepending a shebang and exploiting the fact that the zip format is robust to prepended content.

ScottFreeCode:
>  I would be happy to PR a fix, but I was unclear why the original PR made this change happen by default?

Per discussion of #2097 and #12371 I was under the impression that beginning with a hashbang would be adequate to identify a non-binary file; I don't recall any discussion of making it require opt-in.

marcosscriven:
We would like to propose two alternatives to resolve this issue, in order of preference:

1. Add an attribute to `package.json` that indicates a file (or files) should not be processed by npm upon install. 

I intentionally use the word 'processed', because I could envision some other processing being introduced in the future, other than simple line ending changes, especially as the original intention here was to fix already published packages without requiring them to be re-published.

Quite what the attribute should be called is open for discussion - there is already a `files` attribute that takes the form of an array of paths expressed as strings. We could perhaps make the contents of that array heterogenous, with either a string for simplicity/compatibility, or an object with `path` and `noProcess` fields? I struggle to find any packages actually using the `files` attribute anyway, as it seems to be idiomatic to simply use `.npmignore`.

```
...
  "files": [
    "bin/script.sh",
   {
      "path": "bin/binaryHashbangFile",
      "noProcess": true
   } 
  ],
...
```

Clearly this would be a larger change, not least because this would require updating public documentation. However, I do feel it's the most robust and comprehensive solution, whilst also ensuring the changes does not affect current behaviour by default.

[...]

PR #19172, opened 2017-11-16
19151: Log the fact line endings are being changed upon install.

marcosscriven:
Per https://github.com/npm/npm/issues/19151, as an interim measure I'm just adding a log line on the `silly` level to make it clearer some files are having their line endings changed.

This is to save other developers time if they hit the same issue of having their binary hashbang files corrupted during install.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -526,3 +526,4 @@ Carsten Brandt <mail@cebe.cc>
 Marcin Szczepanski <marcin@imagichine.com.au>
 Josh Clow <josh@textio.com>
 Jakub Holy <jakubholy@jakubholy.net>
+Marcos Scriven <marcos@scriven.org>
--- a/AUTHORS
+++ b/AUTHORS
@@ -526,4 +526,3 @@ Carsten Brandt <mail@cebe.cc>
 Marcin Szczepanski <marcin@imagichine.com.au>
 Josh Clow <josh@textio.com>
 Jakub Holy <jakubholy@jakubholy.net>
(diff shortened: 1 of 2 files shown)


AUTHORS
```diff
@@ -526,3 +526,4 @@ Carsten Brandt <mail@cebe.cc>
 Marcin Szczepanski <marcin@imagichine.com.au>
 Josh Clow <josh@textio.com>
 Jakub Holy <jakubholy@jakubholy.net>
+Marcos Scriven <marcos@scriven.org>
```
iarna: Please don't add yourself to AUTHORS (that happens automatically during releases)

iarna (approved):
Looks good

AUTHORS
marcosscriven: Removed.

marcosscriven:
@iarna - how do I get this merged please?

iarna:
@marcosscriven Kat or I merge it when we're building a release. (We go through PRs methodically when doing so.)

Speaking of, I was just here for that purpose and something that's happened in our release branch that had slipped my mind or I'd've mentioned it earlier is that this code was externed to a new module `bin-links` that exists over here: https://github.com/npm/bin-links/blob/latest/index.js#L119

I'm gonna see if I can't port your commit over there and do a new release from there though…

iarna:
This landed in `bin-links` as https://github.com/npm/bin-links/commit/e9f8a6f705b3c04a2e76258380d9025ea0c5cb2d and in npm in the `release-next` branch as 5645a9597.

PR #19173, opened 2017-11-16
Update npm-dedupe.md

ghost:
Update for "Modules" line.

**Note:** If it is not correct, please update is as needed.


zkat:
I'd rather just remove that section altogether, tbh

PR #19176, opened 2017-11-16
feature: add --package-json-file to set a custom name for package.json

### LINKED ISSUE #6661: Request: allow npm install to accept path to json file describing dependencies
It would be very useful if `npm install` accepted a path to a file that included json formatted the same as package.json's `dependencies` value.  My use-case is to programmatically configure a list of global packages on, say, a new box.  I haven't found a good solution to arbitrarily translate
`{"package-name": "version/uri"}` (a la package.json) into something like `npm install package-name@version`.

### LINKED ISSUE #8263: dev and prod install script
Right now, there is only the `install`/`postinstall` lifecycle you can use to run scripts on install. I am aware that using install script are not recommended. However, I wanted to have one mostly to setup some thing for development only that would use `devDependencies`.

The most obvious problem I ran into, is that if I have an `install` script, it will always run regardless if you are working in development, if someone else installed your module or if you are in production.

I tried looking into npm source to find a way to figure out which scenario we're in when an install script is ran and unfortunately, the only thing I could find was the path (ie: if your parent directory is `node_modules` you are not in dev). 
It kinda works, see my attempt to use this [here](https://github.com/Cellule/npm-lifecycle-runner), but it feels very hacky.

I was wondering if there could be any support directly from npm by adding lifecycle `devinstall` and `prodinstall`. 

`devinstall` would run anytime you run `npm install` or `npm install x` directly in your project.
`prodinstall` would run anytime you either have `NODE_ENV=production` or your module is installed as a child (ie: no `devDependencies` are installed and your parent folder is `node_modules`).

### LINKED ISSUE #8836: Speeding up `npm install` in docker
I'm using a `Dockerfile.dev` file for Node.js development and am using the following "trick" to only `npm install` when `package.json` is modified:

``` dockerfile
COPY package.json /src/
RUN npm install --unsafe-perm
COPY . /src
```

However, as I frequently install new modules and the `npm install` is quite slow, this is not good enough. I've tried mounting my local `node_modules` as a mounted volume but it seems `npm install` installs everything from scratch regardless of whether there are already installed modules in `node_modules/`. Is this correct? in this case, are there any flags or commands I can use to only `npm install` modules which aren't already present in `node_modules/`? Would mounting `./docker_npm_cache:/root/.npm` in the Docker container be another suitable option?

### LINKED ISSUE #11446: Think about splitting package.json into parts in order to improve docker experience
Hi there,

The problem is very simple. In order to speed up docker builds, one needs to leverage layers caching and usual approach is:

```
COPY package.json /usr/src/app/
RUN npm install

COPY . /usr/src/app/
```

It works well in cases when you don't touch package.json too often, but in case of active development you start messing around with everything in that file, but dependencies. You adjust scripts sections, change main property, etc., but dependencies are unchanged. What happens next is docker recognizes files as modified and so bypasses cached `npm install` results.

It would be really helpful to have a composable package.json in order to extract dependencies into external files so further changes to package.json doesn't affect docker caching.

Thanks!

### LINKED ISSUE #13869: extendingDependencies - another type of dependency

I am trying to figure out how to best write one particular kind package and I struggle find the best way with the current limitations of NPM. I even wrote a [medium post](https://medium.com/@leichtgewicht/good-dev-offers-dependency-234ce6e1fbf2#.bz34glk2w) on it.

#### What's the feature?

Add a new way to specify dependencies that are optional yet not installed by default.

If you want to write a package like https://github.com/tj/consolidate.js you need to exclude the dependencies from your `package.json` else all the dependencies are installed by default, which would lead to signficant higher install times and bandwidth use. However if you do that all the goodness of semver are lost on the package.

#### How would you solve this?

I would add a new property to the `package.json` called `extendingDependencies` _(not quite happy with the naming yet)_. I will reference to the property as `extDep` from now on.  The notation of the properties are exactly same as with the other `*Dependencies` properties:
```json
{
  "name": "sample-pkg",
  "extendingDependencies": { "js-yaml": "3.6.x"}
}
```

At installation-time of `sample-pkg` if ...

1. ... a matching `js-yaml` **has been** installed before:
     NPM should move/dedupe the `js-yaml` package from the other peer dependency.
2. ... a matching `js-yaml` has **never been** installed:
     NPM should Install a placeholder package in the `sample-pkg/node_modules/js-yaml` folder.  
3. ... a **non-matching** `js-yaml` has been installed in the hierarchy above `sample-pkg`:
     (same like 2.)

##### Placeholder
The Placeholder is a usability feature: If some code in `sample-pkg` were to call `require('js-yaml)` then the placeholder would show a [helpful message](https://www.npmjs.com/package/package-missing) instructing the user on how to install an extending dependency. This should reduce friction between developer and user.

If NPM would simply not install this placeholder then node would bubble up folder hierarchy to look for `js-yaml` and if a non-matching version is found there the code might break. _(in other words: the placeholder is important!)_

Another consideration here is that older versions of `npm` would simply not install those `extDeps` which means would lead to problems but in many cases it would also still work.

##### Specifying the installation of an `extDep`.
For the installation of an `extDep` the user needs to specify possibly the whole hierarchy of packages above, my thought is to add a new "semver-extension" for this case:

`npm i sample-pkg:js-yaml` which would then install `js-yaml` with the highest version matching the specification in `extDeps`.

If `sample-pkg` were to be in a deeper hierarchy the installation could look like:
`npm i other-pkg:sample-pkg:js-yaml`.

##### Saving the `extDep`

If a package depends on `sample-pkg` having `extDep` installed then it would need a way to reference those. I initially thought it might be good to have a semver extension for this: like `"sample-pkg": "1.1.0+js-yaml"` but this would be bad for backwards compatibility. Instead I think it would need to be a separate field `extendedDependencies` _(again, the naming is problematic: extending/extended is too similar)_:

```json
{
  "name": "different-pkg",
  "dependencies": {
    "sample-pkg": "1.1.0"
  },
  "extendedDependencies": {
     "sample-pkg": {
        "js-yaml": "3.6.0"
     }
  }
}
```

It could even support deep-extending like: `"other-pkg:sample-pkg": { "sample-pkg": "3.6.0" }`

### LINKED ISSUE #15135: Feature request/question - Optional dependencies on a case-by-case basis

  - [ ] npm is crashing.

#### What's going wrong?

there is nothing really going wrong  - this is a question to see if this feature exists, and if the feature does not exist, this is a feature request! :)

so NPM allows for optional dependencies -   the syntax is like so:

```json
"dependencies" : {
   "foo": "1.2.3",
   "zoom": "4.5.3"
},
"devDependencies" : {
    "bar": "2.3.2"
},
"optionalDependencies" :{
    "baz": "1.4.22"
}
```

what I am looking to do is conditionally make dependencies optional. Such that I could make `foo` or `zoom` optional. A command line argument, let's call it "omit" would be sufficient in my use case, e.g.:

`npm install <my-package> --omit foo --omit zoom`

That's the functionality I am looking for. My use case is not a common one, but I can explain it if you want!

also, as an aside, in the docs it says:

> "Entries in optionalDependencies will override entries of the same name in dependencies, so it's usually best to only put in one place."

I actually don't know what that means

### LINKED ISSUE #16342: Feature request =>  npm install --except X

  - [ ] npm is crashing.

#### What's going wrong?

I am running npm install on a couple of local repos (let's say 5 of them)

each repo has a fairly large devDependency X; I own X, and it's undergoing local development.

I am going to end up symlinking X into each of these 5 repos, so it actually doesn't make sense to install it at all, it's just a time suck as far as I can tell.

Is there a way to do

`npm install --except X`

? 

that would be really useful!

In other words, X would be an @latest entry for all 5 of these local repos. Why both install @latest from NPM, when in fact I want the absolute latest: the one on my filesystem!

BelfordZ:
First, thank you for looking at this pr.

The motivation behind this PR is largely to improve docker support so that we can leverage docker caching to the best of docker's ability! 

Currently there are two major issue when using node/npm with docker. 

A) there is no way to split up installing of node modules such that docker can efficiently cache packages that are infrequently updated. 

B) when you have dependencies that point to private git repos, node and docker are less than optimal.
  1. manually writing a script to seperate package.json's into private and public deps, installing the public ones at docker build time to leverage cache while not leaking secrets into docker images. You then npm install the private ones at run time.
  2. adding your secrets to the image in order to pull the private repo (security concern)
  3. adding your ssh keys to the image and squashing the intermediate layers (bye bye docker caching)
  4. including tokens in your package.json (again ignoring security best practices)

Further, many people have reported needs/desires to create their own separation of deps, scripts, names, all sorts of things. 

I have started by implementing this for npm install to start with. My general approach was to 
0. clone things, run the tests, figure out how to get stdout so I could see the errors instead of just the failing test (I must be doing something wrong here)
1. get the new config value in there and make sure its correctly defaulted to `package.json` in the  install tests
2. write a new test that uses the param and changes the value, fix everything that breaks.

I came across a couple interesting issues with this.
1. [Some of the code used to pull out the prefix](https://github.com/npm/npm/pull/19176/files#diff-cb61c79dd0154389870f5c13853305fcR46) has `package.json` hard coded, and is used before the command line arguments are parsed. I admittedly handled this in a hacky way, and I think the correct thing to do here would be to ensure that the command line arguments are parsed and stored in the configs pretty early on the in the chain.
2. [The read-package-tree](https://github.com/npm/npm/pull/19176/files#diff-7d67c5bca38f2cd29f790b16637a7819R97) also hard codes the `package.json` string. I monkey patched it to make the tests pass and move on.

I welcome any and all feedback, please don't be shy! Thanks again for your time in taking a look at this pr.

fixes #16342
fixes #15135
fixes #13869
fixes #8263
fixes #6661
fixes #11446 
fixes #8836 

im sure it fixes many, many other issues/feature requests.


lib/config/find-prefix.js
```diff
@@ -43,8 +43,14 @@ function findPrefix_ (p, original, cb) {
     // walked up too high or something.
     if (er) return cb(null, original)
 
+    var packageJsonName = 'package.json'
```
BelfordZ, line 46: `npm.config.get` is not available in this context -- is there a better way to determine cli params used when this method is called called?

node_modules/read-package-tree/rpt.js
```diff
@@ -94,7 +94,7 @@ Link.prototype = Object.create(Node.prototype, {
 Link.prototype.target = null
 Link.prototype.realpath = ''
 
-function loadNode (logical, physical, cache, cb) {
+function loadNode (logical, physical, cache, packageFilename, cb) {
```
BelfordZ, line 97: Very unfortunate that changes to this package also need to be made. Not sure how you folks have handled this type of dep change in the past -- but since they are checked in, monkey patching seemed like the way to go at the time. 

Any suggestions for this?

iarna: Monkey patching things in node_modules is never something we do. We patch the other module, do a release there, then update as a part of the commit.

While figuring out what changes to make it's fine though, it just can't be merged like that.

iarna:
As a rule, I'm opposed to making the name of `package.json` configurable, but I would like to see the underlying needs met.

I'm gonna be away from the computer for a bit, so I don't have time to write up more, but I will be back. =)

In the meantime, I'd encourage you to drop by #npm on https://package.community/ — it's where the core devs (and the more active contributors) hang out and it's also where we have discussions about this sort of thing.

BelfordZ:
The only way the underlying needs are met is by being able to have arbitrary partitioning of dependencies in your package json, which IMO is much more likely to cause breaking changes.

I wonder why every other package manager under the sun has this functionality and yet its so quickly dismissed?

BelfordZ:
@iarna RIP npm cli. I'm moving to yarn and I'll evangelize how great of an open source community you are fostering over here ;)

iarna:
Ok, this landed as https://github.com/npm/bin-links/commit/e9f8a6f705b3c04a2e76258380d9025ea0c5cb2d over there…

iarna:
@BelfordZ Well, I'm glad you found something that meets your needs!

iarna:
Final bit for any future readers of this PR:

Some discussion went on in the package.community discord. The PR never actually got far enough to be rejected. (I hadn't closed it because I wanted to see if we could address its needs without taking the changes here. I was in fact still prepared to be convinced that this was the best approach though.)

Soo, I hacked together this proof-of-concept of an alternative take on this: https://github.com/iarna/npm-install-with-only#npm-install-with-only

Issue #17722, opened 2017-07-11
package-lock.json and optional packages


Strate:
#### I'm opening this issue because:

  - [x] npm is doing something I don't understand.

#### What's going wrong?

Npm accidently make changes to package-lock.json

#### How can the CLI team reproduce the problem?

Assume you have a 2 developers, one on mac, and one on linux. You use npm@5.1 and your project depends on `chokidar` package. That package has optional dependency of `fsevents`, which is useful only for mac. So, you are on linux, and do `npm i chokidar`. npm generates package-lock.json withous fsevents, because it is useless on linux. You commit that generated file.

Your teammate pulls your changes, and do `npm i`, to get `node_modules` in sync with `package-lock.json`. Npm installs `fsevents`, and write it to `package-lock.json`. What should mac user to do? commit that file?

Assume that mac user commits file. Linux user pulls it, and make `npm install`. Npm does not install `fsevents`, and remove it from `package-lock.json`. And this become annoying very quickly.

How lock files could be used in this case?

### supporting information:

 - `npm -v` prints: 5.1.0
 - `node -v` prints: v6.10.0
 - `npm config get registry` prints: undefined
 - Windows, OS X/macOS, or Linux?: max + linux
 - Network issues:
   - Geographic location where npm was run:
   - [ ] I use a proxy to connect to the npm registry.
   - [ ] I use a proxy to connect to the web.
   - [ ] I use a proxy when downloading Git repos.
   - [ ] I access the npm registry via a VPN
   - [ ] I don't use a proxy, but have limited or unreliable internet access.
 - Container:
   - [ ] I develop using Vagrant on Windows.
   - [ ] I develop using Vagrant on OS X or Linux.
   - [ ] I develop / deploy using Docker.
   - [ ] I deploy to a PaaS (Triton, Heroku).

mborisv:
👍

topheman:
We have the exact same problem. Developers use MacOS and Linux across our team and a simple `npm install` might update the `package-lock.json` (adding or removing `fsevents`) (according on the plaform this file was created first) - even though there wasn't any package added.

Is there a flag to do the opposite of `npm install --no-optional` ? (though, this might not work, since the platform will override anyway)

Strate:
We also can not force to use `fsevents` package, because it has strict requirement of os.

topheman:
[...]

PR #19054, opened 2017-11-03
Stop dropping failed optional deps

iarna:
Makes full package-locks even the package.json has optional deps that are currently uninstallable. Also stops deleting optionals from package-locks when we can't install them.

PR #19233, opened 2017-11-23
Stabilize optional shrinkwrapped dependencies.

jacksonrayhamilton:
Fixes https://github.com/npm/npm/issues/17722, ensuring that optional dependencies from a lockfile which fail to install are not removed from the lockfile.  Example: When `fsevents` is installed on macOS, but later fails to install on GNU/Linux, `fsevents` will not be removed from the lockfile.

Besides the unit test, I also confirmed that lockfiles remain unchanged on GNU/Linux by running `npm i` in this repo: https://github.com/jacksonrayhamilton/chokidar-test

### DIFF
--- a/lib/install/deps.js
+++ b/lib/install/deps.js
@@ -81,6 +81,12 @@ function childDependencySpecifier (tree, name, spec) {
   return npa.resolve(name, spec, packageRelativePath(tree))
 }
 
+function shouldRetainChild (child) {
+  // Forgive shrinkwrapped dependencies for failing.  They should be retained
+  // for the sake of whatever environment they didn't fail in.
+  return !child.removed && (!child.failed || child.fromShrinkwrap)
+}
+
 exports.computeMetadata = computeMetadata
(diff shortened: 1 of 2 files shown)


jacksonrayhamilton:
And now I found https://github.com/npm/npm/pull/19054.  Is that a solution for https://github.com/npm/npm/issues/17722 and we should close this?

Karasuni:
Regardless we might still need this PR to ensure consistency on lockfiles across OS.

jacksonrayhamilton:
Maybe.  I bring it up because the three changes I made in deps.js are more or less mirrored in the other PR.

zkat:
This should indeed have been overridden by #19054! 👍

jacksonrayhamilton:
@zkat!!  Thank you for responding.  I implore you to also take a look at my other PR which has been open for a while and fixes a long-outstanding NPM issue: https://github.com/npm/npm/pull/19198

PR #19307, opened 2017-12-04
docs:sync package-lock spec and help page

jcrben:
Per https://twitter.com/ReBeccaOrg/status/935307487707774976 at least this field `requires` didn't make it over into the main help file.

I didn't change the others as I'm not sure whether they are intended for the main helpfile - for example, `from *(deprecated)*` in spec, `packageIntegrity` in helpfile and not spec, and `link` (to be implemented post npm@5).

Note: folks at https://stackoverflow.com/questions/45117589/what-does-requires-true-do-in-package-lock-json noticed that there was no docs on this - I will update that when this is merged.


zkat (approved):
Thanks for the sync! 👍

PR #19418, opened 2017-12-18
Do not redirect the `env` command to `set`, for windows bash shell

gucong3000:
- Redirect the `env` only for window cmd shell
- Add support `%ProgramFiles%\Git\usr\bin\bash.exe`  for `lib/utils/is-windows-bash.js`

https://github.com/npm/npm/issues/19419

### DIFF
--- a/lib/run-script.js
+++ b/lib/run-script.js
@@ -8,6 +8,7 @@ var log = require('npmlog')
 var chain = require('slide').chain
 var usage = require('./utils/usage')
 var output = require('./utils/output.js')
+var isWindowsShell = require('./utils/is-windows-shell.js')
 
 runScript.usage = usage(
   'run-script',
@@ -138,7 +139,7 @@ function run (pkg, wd, cmd, args, cb) {
       if (cmd === 'test') {
         pkg.scripts.test = 'echo \'Error: no test specified\''
       } else if (cmd === 'env') {
-        if (process.platform === 'win32') {
+        if (isWindowsShell) {
           log.verbose('run-script using default platform env: SET (Windows)')
           pkg.scripts[cmd] = 'SET'
         } else {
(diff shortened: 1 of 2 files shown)


gucong3000:
Is there anything I can do to help this PR?

andelmad (approved):

iarna:
Hi @gucong3000, thank for your patience! While we don't officially support cygwin, I'm ok with taking patches for it that we can be confident won't cause other issues. In this case, I would suggest you add your Cygwin detection to `lib/utils/is-windows-bash.js` and then use that in the if statement here.

lib/run-script.js
```diff
@@ -138,7 +138,7 @@ function run (pkg, wd, cmd, args, cb) {
       if (cmd === 'test') {
         pkg.scripts.test = 'echo \'Error: no test specified\''
       } else if (cmd === 'env') {
-        if (process.platform === 'win32') {
+        if (process.platform === 'win32' && !/^(msys|cygwin)$/.test(process.env.OSTYPE)) {
```
iarna: Use `lib/utils/is-windows-bash.js` and update that to include cygwin.

iarna (changes requested):

gucong3000:
@iarna Thanks for your patience. Modified.

lib/utils/is-windows-bash.js
```diff
@@ -1,3 +1,4 @@
 'use strict'
 var isWindows = require('./is-windows.js')
-module.exports = isWindows && /^MINGW(32|64)$/.test(process.env.MSYSTEM)
+var env = process.env
+module.exports = isWindows && (env.MSYSTEM ? /^MINGW(32|64)$/.test(env.MSYSTEM) : env.TERM === "cygwin")
```
iarna, line 4: The env.MSYSTEM guard here is unnecessary… but ah, I'll take and make a small change. =)

iarna:
Thanks! This is landing as 4c32413a5b!

Issue #9111, opened 2015-07-30
npm version not creating commit


JaKXz:
From https://github.com/npm/npm/issues/8620#issuecomment-126098728

I have also tried using the `--force` flag with no luck.

JaKXz:
#9099 looks like it's related, because my directory structure is a bit weird for this project:

```
my-project-root
|----.git/
|----other-files.txt
|----subdir/
|--------package.json
|--------index.html
```

and I'm running `npm version minor -m "upgrade to %s for reasons"` from `subdir`.

aredridel:
Oof. that layout presents a problem since npm is definitely not in charge of the whole tree, and you can't commit just a subtree in git.

dreyescat:
Definitely the layout presents a problem. `npm version` is assuming that both `.git` directory and `package.json` are found in the same directory. As you have your `package.json` in a subdirectory, when you run `npm version` it [will fail to create a tag because it is not able to find the `.git` directory](https://github.com/npm/npm/blob/master/lib/version.js#L144-L150).

As a *dirty* workaround... you could just bypass the `.git` checking just creating a fake empty `.git` folder along with your `package.json`. And when I say a fake empty folder I mean `mkdir .git` in the same folder the `package.json` file resides.

JaKXz:
@aredridel @dreyescat thanks for the responses - making a fake `.git` directory worked and is an acceptable workaround for now.

andrewbranch:
Yeah, I think it would be nice to enable this through an option or something. We have a single repo that contains many npm packages (for reasons that we just have to deal with for now), and I thought to myself, "I can still make `npm version` work pretty nicely by making project-specific values for `tag-version-prefix` and `messag` in each package's .npmrc. Not so. 😢

adambuczynski:
I will +1 on this, and I think I have a valid use case. My folder structure looks approximately like this:

```
project
|_ client
|_ server
```

The `server` folder contains a Node Express app, with it's own `package.json` in the sub directory. The `client` folder contains an Angular application, again with it's own `package.json` in the sub directory. 

Because the Angular and Node apps have vastly different requirements in terms of dependencies, dev dependencies, run scripts, etc. it's not an option to use one shared `package.json`. However, I do want to keep both the client and server apps in one git repository, because it's one project and splitting it up in two repositories will create too much unnecessary overhead.

[...]

Issue #9099, opened 2015-07-29
npm version while in a subdirectory of the project increments the version but fails to commit to git


aredridel:
```
ariastewart@LM-BOS-00872654:cases (master:8dbeb8c)
:; npm version patch
v3.0.1
npm ERR! Darwin 14.5.0
npm ERR! argv "node" "/usr/local/bin/npm" "version" "patch"
npm ERR! node v0.12.7
npm ERR! npm  v3.1.3
npm ERR! code 128

npm ERR! Command failed: git add package.json
npm ERR! fatal: pathspec 'package.json' did not match any files
npm ERR!
npm ERR!
npm ERR! If you need help, you may report this error at:
npm ERR!     <https://github.com/npm/npm/issues>

npm ERR! Please include the following file with any support request:
npm ERR!     /Users/ariastewart/Projects/json-ometajs/test/cases/npm-debug.log
ariastewart@LM-BOS-00872654:cases (master:8dbeb8c Δ )
```

othiym23:
Good catch! Thanks! It would be nice to have this just work, but it also would be nice to document how npm works here, because there's some conflusion (and differences of opinion) about what should happen when `npm version` is run somewhere other than the root of a package and / or git repository.

It also might be a little while until the team has cycles to address this, so I stuck `patch-welcome` on this issue if somebody wants to take a whack at it.

simonvanderveldt:
This does not work for me on 2.15.1 which does include this fix. Looking at the commit I don't really see how that change would solve this issue to be honest.
@aredridel did this actually work for you?

dcolens:
npm: 3.9.5, it fails to commit when running from a subfolder (package.json is also in the subfolder)

othiym23:
@simonvanderveldt, @dcolens: are you saying

1. `package.json` doesn't get incremented and checked in when `npm version` is run from a subdirectory, or
2. dirty files in the current directory don't get checked in when `npm version` is run from a subdirectory?

1 is the intent of this patch, but 2 is not.

simonvanderveldt:
@othiym23 To be more precise this is what I'm doing and what happens:
- I `cd` in to a project with a clean working directory that resides in a subdir of the repository
- Then run `npm version minor` (but the same happens for other ones obviously)
- `package.json` gets updated with the new version
- `package.json` is not staged/added nor committed
- No tag is created

drew-r:
Get anywhere with this @simonvanderveldt ?

[...]

PR #19468, opened 2017-12-27
version doesn't commit when run in subdir

ryanhos:
The version command doesn't currently detect that it should create a commit unless the current working directory is the root of a git working tree.  See #9099 and #9111.

### DIFF
--- a/doc/cli/npm-version.md
+++ b/doc/cli/npm-version.md
@@ -20,7 +20,7 @@ valid second argument to [semver.inc](https://github.com/npm/node-semver#functio
 the existing version will be incremented by 1 in the specified field.
 `from-git` will try to read the latest git tag, and use that as the new npm version.
 
-If run in a git repo, it will also create a version commit and tag.
+If run in the root directory of a git working tree, it will also create a version commit and tag.
 This behavior is controlled by `git-tag-version` (see below), and can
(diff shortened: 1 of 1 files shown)


zkat:
This seems more like a bug worth fixing, rather than documenting?

pierre-charpentier:
@zkat The problem only occurs if you have a subdir that also has a package.json like so:

project-root
|----.git/
|----subdir/
|--------package.json

The git repository tag should not be updated by a version command in the subdirectory in my opinion.

Everything works fine if the subdirectory has no package.json file.

zkat:
@Qayou then I guess that specific bit should be clarified. :)

luisdavim:
I think this a bug that should be fixed.

markcellus:
I agree with @luisdavim. Will this be fixed eventually?

legodude17:
@luisdavim as @Qayou said, the current behavior seems correct.

luisdavim:
Can you explain the logic?

legodude17:
If you have a `package.json` in the subdir, then you want to dump the version of the subdir, and it wouldn't make sense to commit that to it's parent, as you didn't change the parent's version.

luisdavim:
What if my repo structure is different, I have a folder with some related scripts, another with Docker files and another with an helm chart and what should go into my actual npn package is in a subfolder of my git repo...

Issue #7990, opened 2015-04-17
Implement silent param for install/uninstall commands


marcominetti:
When using npm programmatically as module, install and uninstall does not accept any silent param. Just for example list command supports it.

smikes:
If you want npm to write minimal output when installing, you can set the `loglevel` parameter to `silent`:

```
$ npm --loglevel=silent install
```

However, npm will still write to standard output and standard error.  On a unix-like system such as Linux or OSX, you can suppress this output as well:

```
$ npm --loglevel=silent install 2>&1 >/dev/null
```

marcominetti:
Thanks, but I'm using it programmatically through
```
var npm = require('npm');
```
Do you have a quick hint to initialize with log level silent in npm.load?

smikes:
Sure

npm.load( {loglevel: "silent"}, function … )

But you should be careful, as the npm api is intended for internal use, and project policy allows npms api to change without affecting semver; semver applies only to the command line interface and behavior.

marcominetti:
Yep, I know... I will probably fork it for internal use and manage lifecycle directly... ;) Thanks!

marcominetti:
Perfect,
```
npm.load( {loglevel: "silent"}, function … )
```

does exactly fit my needs.

Thanks...

;)

mattdesl:
Testing with npm@2.8.4 and `npm.commands.install` does not seem to respect `loglevel` or `silent` options on `npm.load`. Ideas?

marcominetti:
Hi @mattdesl, just to confirm we're talking about the same thing...

With the snippet above we define log level silent that works only for console log from npm... All the logs you see at stdout when loglevel is silent should come directly from nodegyp or any other process (npm itself too) spawned with child stdio piped to your process.stdio...

Did I understand correctly @smikes?

KoryNunn:
Hi, please pull and give it ago again. new API:

```
globalify module -o module.js
```

wilk:
Like @mattdesl said, I've the same problem with `npm@5.6.0` and `npm.commands.install`

<img width="912" alt="screen shot 2017-12-26 at 11 55 36" src="https://user-images.githubusercontent.com/647020/34355181-b8e13a4c-ea33-11e7-94c0-1e717eba621d.png">

Between "done!" and "done!" there's another debug line printed with `process.stdout.write` (yes, you cannot see it): it seems that only `console.log` is working after the `npm.commands.install` gets executed.
[...]

PR #19473, opened 2017-12-28
deactivate progress bar when loglevel is set to silent

wilk:
deactivate progress bar when loglevel is set to silent, due to this bug: https://github.com/npm/npm/issues/7990


wilk:
Ok, maybe the `progress` config is enough.

PR #19502, opened 2018-01-03
Make outdated handle https URLs

### LINKED ISSUE #17835: NPM update fails with a non-registry tarball URL dependency

  - [X] npm is doing something I don't understand.

#### What's going wrong?

NPM update fails when a non-registry tarball URL dependency exists for a module. It errors out of the process with this:

```
npm ERR! code E404
npm ERR! 404 Not found : <module_name>
npm ERR! 404
npm ERR! 404  '<module_name>' is not in the npm registry.
npm ERR! 404 You should bug the author to publish it (or use the name yourself!)
npm ERR! 404
npm ERR! 404 Note that you can also install from a
npm ERR! 404 tarball, folder, http url, or git url.

npm ERR!     /Users/michael.stone/.npm/_logs/2017-07-18T15_28_47_274Z-debug.log
```

It looks like this issue was raised previously as https://github.com/npm/npm/issues/9531

A summary of that thread:
- This previously worked, but started throwing the error as the result of this commit:
https://github.com/npm/npm/commit/765fd4bfca8ea3e2a4a399765b17eec40a3d893d
- No problems reported with npm install
- Issue also affects npm outdated because that's used by npm update under the hood
- Setting "private: true" in the tarball dependency package.json should be a workaround as private packages are ignored by npm outdated as the result of a previous patch. This has been reported as unsuccessful by a user more recently, and may not always be an option.

The actual behaviour of the npm update relating to tarball URL dependencies is probably debatable given the possible scenarios, but at a minimum it seems like this should warn rather than fail the update for all dependencies.

It looks like the block of code [here](https://github.com/npm/npm/blob/latest/lib/outdated.js#L327) needs to prevent a dependency with a type of 'remote' from being passed to the mapToRegistry function, similar to the exclusions for linked and git dependencies.

I'm happy to take a swing at a PR for this if there's an agreement about the approach.

1. Add a dependency with a non-registry hosted tarball URL
```
{
  "dependencies": {
    "module-name": "http://myserver.com/module-name-v1.tgz"
  }
}
```

2. npm install (which successfully downloads the dependency from the url)
3. npm update

### supporting information:

 - `npm -v` prints: 5.3.0
 - `node -v` prints: 7.10.0
 - `npm config get registry` prints: <private nexus repository>
 - Windows, OS X/macOS, or Linux?: macOS

maxthoursie:
Fixes #17835

A dependency of type remote, i.e. a http or https URL, should not be considered outdated. This change will return `remote` for both _wanted_ and _latest_, similar to how git urls are handled, which will fix both `npm outdated` and `npm update`

### DIFF
--- a/lib/outdated.js
+++ b/lib/outdated.js
@@ -331,6 +331,10 @@ function shouldUpdate (args, tree, dep, has, req, depth, pkgpath, cb, type) {
   if (parsed.type === 'git' || parsed.type === 'hosted') {
     return doIt('git', 'git')
   }
+  // If this is a http or https URL, don't consider it outdated.
+  if (parsed.type === 'remote') {
+    return doIt('remote', 'remote')
+  }
 
   // search for the latest package
   mapToRegistry(dep, npm.config, function (er, uri, auth) {


andelmad (approved):

andelmad (approved):

iarna:
Thanks for submitting this! 

As of 5.4.2 remote tarball deps no longer show up in outdated (solving the crash). 

Whatever solution, it should probably apply equally to git dependencies, remote tarballs and local tarballs.

I think arguably none of these should show up in outdated if we don't have a facility for checking for their freshness. (It used to be that git deps were reinstalled every time you ran `npm install`, so having them show as eligible for updates made a sort of sense. But it's been a very long time since that was true.)

Long term, I'd like to have outdated be able to check for the freshness of all of these kinds of things, by checking hashes on local and remote tarballs and comparing the commit id we cloned the local git project from to the commit id that it resolves to now.

If you (the author of this patch or a reader of this comment) would be interested in exploring this further, please drop by #js-npm on the https://package.community/ Discord.

PR #19506, opened 2018-01-03
Fixes #19505 Use a safer method of escaping characters in run-script to allow $

### LINKED ISSUE #19505: Cannot pass an argument through run script args with dollar sign

  - [X] npm is producing behaving incorrectly.

#### What's going wrong?

npm run (run-script) fails to correctly pass through arguments with dollar signs in them because it passes through arguments using double quotes instead of single quotes. 

`npm run example-script -- -a '$500'` or:

Failing test case for `run-script`: 

```javascript
test('npm run-script with args that contain dollar signs', function (t) {
  common.npm(['run-script', 'start', '--', 'what$ up$?'], opts, testOutput.bind(null, t, 'what$ up$?'))
})
```

### supporting information:

 - `npm -v` prints: 5.5.1
 - `node -v` prints: v8.9.2

xdumaine:
as well as all existing chars.

### DIFF
--- a/lib/run-script.js
+++ b/lib/run-script.js
@@ -175,7 +175,7 @@ function run (pkg, wd, cmd, args, cb) {
 function joinArgs (args) {
   var joinedArgs = ''
   args.forEach(function (arg) {
-    joinedArgs += ' "' + arg.replace(/"/g, '\\"') + '"'
+    joinedArgs += ' \'' + arg.replace(/'/g, '\'\\\'\'') + '\''
   })
   return joinedArgs
 }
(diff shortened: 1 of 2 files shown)


xdumaine:
Bump. Only tests failing are ones marked as failing in travis.

zkat:
I'm pretty sure the existing behavior is intentional -- remember that run-scripts have their own magical internal environment with all sorts of special variables that npm injects.

iiuc, with this patch, the following would not work:

```js
// package.json
{
  "name": "world-pkg",
  "version": "1.2.3",
  "scripts": {
    "hello": "echo hello,"
  }
}
```

and then:

```
$ npm run hello -- '$npm_package_name'
hello, world-pkg
```

Is there a reason for intentionally removing this feature? (which I'm very much inclined to keep)? This would at least be a breaking change to our API, so I'd like to hear more about your motivation here.

zkat:

PR #19542, opened 2018-01-07
add npm ci command

zkat:
This adds a brand new npm command!

This command is similar to `npm-install(1)`, except it's meant to be used in automated environments such as automated testing, continuous integration, and deployment. It can be significantly faster than   a regular npm install by skipping certain user-oriented features. It is also more strict than a regular  install, which can help catch errors or inconsistencies caused by the incrementally-installed local environments of most npm users.


agirorn:
Can we please not make a completely separate command for this? Just add extra param to npm install `--ci` for this behavior #19947

drewfish:
The doc doesn't mention which arguments are handled differently than `npm install`. For example I just did a test with npm@5.7.1 and `npm ci --registry X` seems to have worked the same but `npm ci --production` didn't. (I also added `--ignore-scripts` (for safety) but am unsure if that took effect as well.)

zkat:
@agirorn It's a new command because it's a separate installer with different behavior. Those code is very very different.

@drewfish the arguments should be the same. The thing you pointed out was a bug.

Anyway, this is already merged and released so I'm closing this

PR #19552, opened 2018-01-08
update now saves by default

### LINKED ISSUE #18386: `npm update` always updates package.json, ignoring --save flag

  - [ ] npm is crashing.

#### What's going wrong?

`npm update` is always updating the package.json file, regardless of whether the **--save** flag is specified or not. According to the docs (https://docs.npmjs.com/cli/update), updates should only be recorded when the **--save** flag has been specified.

1. Setup an example project with a minimal **package.json** file:

```json
{
  "name": "test",
  "version": "1.0.0",
  "dependencies": {
    "lodash": "4.17.0"
  }
}
```

2. Run `npm i --no-package-lock`. **lodash** should now be installed with version **4.17.0**.

3. Modify **package.json** to include a tilde-range constraint:

```json
{
  "name": "test",
  "version": "1.0.0",
  "dependencies": {
    "lodash": "~4.17.0"
  }
}
```

4. Run `npm update --no-package-lock`. **lodash** should now be installed with the latest patch version, **4.17.4** at the time of writing. However, **package.json** is modified too:

```json
{
  "name": "test",
  "version": "1.0.0",
  "dependencies": {
    "lodash": "^4.17.4"
  }
}
```

If this is the intended behavior, the docs should be updated and the **--save** flag removed. Otherwise, this might be a bug in **npm-update**.

### supporting information:

 - `npm -v` prints: 5.4.0
 - `node -v` prints: 6.11.2
 - Windows, OS X/macOS, or Linux?: Linux, Debian
 - Network issues:
 - Container:

selbekk:
This commit makes a note that `npm update` updates `package.json` with the new version as
the minimum required dependency from `@5.0.0`.

Outdated documentation about how to update `package.json` in older versions of npm has been removed.

Fixes #18386

### DIFF
--- a/doc/cli/npm-update.md
+++ b/doc/cli/npm-update.md
@@ -26,6 +26,10 @@ As of `npm@2.6.1`, the `npm update` will only inspect top-level packages.
 Prior versions of `npm` would also recursively inspect all dependencies.
 To get the old behavior, use `npm --depth 9999 update`.
 
+As of `npm@5.0.0`, the `npm update` will change `package.json` to save the 
+new version as the minimum required dependency. To get the old behavior, 
+use `npm update --no-save`.
+
 ## EXAMPLES
 
 IMPORTANT VERSION NOTE: these examples assume `npm@2.6.1` or later.  For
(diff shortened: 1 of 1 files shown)


selbekk:
Not sure why the build fails on a documentation change - should be unrelated.

zkat (approved):
Thanks for the clarification! 👍

ahmehri:
This was merged since March 8 but still not [published](https://docs.npmjs.com/cli/update).

PR #19580, opened 2018-01-11
Better message when an "OTP" error occurs

jdeniau:
When using 2FA, there is a weird `Enter OTP: ` without further explaination.

This PR adds more logging and a small explaination on OTP

### DIFF
--- a/lib/profile.js
+++ b/lib/profile.js
@@ -180,7 +180,7 @@ function set (args) {
       newUser[prop] = value
       return profile.set(newUser, conf).catch((err) => {
         if (err.code !== 'EOTP') throw err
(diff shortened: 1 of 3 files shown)


jdeniau:
I do not know why the tests are failing on `COVERALLS_REPO_TOKEN="$COVERALLS_OPTIONAL_TOKEN"`. 

It does not seems to be related to my changes but to the path of some binary files

iarna (approved):
This seems like a pretty reasonable user experience improvement. Thanks!

iarna:
Thank you for this! It shipped in 6.0.0! =)

PR #19590, opened 2018-01-13
doc(npmInstall): clarify --ignore-dependencies is recursive

xml:
The fact that `--ignore-dependencies` prevents execution of scripts not only in your main project, but in all sub-dependencies as well, feels to me like it violates the principle of least-surprise, even though I admit it's logically consistent. It would be very cool if someday, you could suppress only your own scripts, and not those of dependencies you didn't realize you were messing with. 

But anyway: this PR at least documents the current behavior clearly.


zkat (changes requested):
👍 this is probably a good clarification, if you think it can confuse folks!

Could you do me a favor and:

1. Wrap this to the column width used in that file.

2. Add a similar comment to the `ignore-scripts` section in `doc/misc/npm-config.md` for consistency?

PR #19629, opened 2018-01-17
Added a warning for invalid shrinkwrap JSON

### LINKED ISSUE #19628: Context-less error thrown trying to parse an invalid package-lock.json

  - [x] npm is crashing.

#### What's going wrong?

If `package-lock.json` is empty, npm crashes with: ` Unexpected end of JSON input`

image

1. Install in a project normally
2. Delete all content of `package.json`
3. `npm i`

https://gist.github.com/JoshuaKGoldberg/74c591fa1d55462a2d290dd5dcae3051

### supporting information:

 - `npm -v` prints: 5.5.1
 - `node -v` prints: 8.8.1
 - Windows, OS X/macOS, or Linux?: Windows 10 with cmder

Extra relevant because of #18131

JoshuaKGoldberg:
Fixes #19628 

Instead of the direct `JSON.parse` error, this gives a warning mentioning the file name.

image

### DIFF
--- a/lib/utils/parse-json.js
+++ b/lib/utils/parse-json.js
@@ -1,6 +1,7 @@
 'use strict'
+var parseJsonWithErrors = require('json-parse-better-errors')
 var parseJSON = module.exports = function (content) {
-  return JSON.parse(stripBOM(content))
+  return parseJsonWithErrors(stripBOM(content))
 }
 
 parseJSON.noExceptions = function (content) {
--- a/node_modules/json-parse-better-errors/CHANGELOG.md
+++ b/node_modules/json-parse-better-errors/CHANGELOG.md
@@ -1,0 +1,36 @@
+# Change Log
+
+All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
+
+<a name="1.0.1"></a>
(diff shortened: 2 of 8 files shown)


zkat:
We should not be skipping pkglocks if they fail to parse -- I think a better patch here is to go into `lib/utils/parse-json.js` and make it use `json-parse-better-errors`, which is what we use in other parts of the CLI to output more informative parse errors.

Could you send that patch in, instead? You'll have to rewrite this anyway, because `npm@5.7.0` added support for automatic git conflict resolution for pkglocks, so this code is a bit different now.

JoshuaKGoldberg:
> `json-parse-better-errors`

Adding a new package dependency wasn't something I was _planning_ on doing for my first npm contribution, but cool!

zkat:
@JoshuaKGoldberg it's not -quite- a new dependency. It's used a lot by npm's deps and I guess I'm actually surprised we don't have it as a toplevel dependency. Just make sure it's in `bundleDependencies`, too!

zkat (approved):
Looks great! Thanks a bunch! 🎉

zkat:
```
➜ npm i
npm ERR! Unexpected token { in JSON at position 104 while parsing near '...e,
npm ERR!   "dependencies" {
npm ERR!     "eggplant": {
npm ERR! ...'
```

I think this is a much nicer error (could be better, but better than "uh idk")

PR #19670, opened 2018-01-22
Check for package name before removing.

richardpringle:
Hello friends, not trying to waste anyone's time here but I haven't actually read the "contributing" docs. 

I keep running into an error:
```
cannot read '0' of undefined
```

I've run into on npm 5.4.0 as well as 5.6.0.

This PR fixes the problem. Although... if you log out all the `pkg` contents, I'm not getting any package objects with `name`.

I'm sure there's an issue in here that's already open, if someone can just point me there for the discussion that would be much appreciated.

### DIFF
--- a/lib/unbuild.js
+++ b/lib/unbuild.js
@@ -58,7 +58,9 @@ function rmStuff (pkg, folder, cb) {
   // if it's global, and folder is in {prefix}/node_modules,
   // then bins are in {prefix}/bin
   // otherwise, then bins are in folder/../.bin
-  var parent = pkg.name[0] === '@' ? path.dirname(path.dirname(folder)) : path.dirname(folder)
+  var parent = pkg.name && pkg.name[0] === '@'
+    ? path.dirname(path.dirname(folder))
+    : path.dirname(folder)
   var gnm = npm.dir
   // gnm might be an absolute path, parent might be relative
   // this checks they're the same directory regardless


zkat:
@richardpringle I'd really love to see a repro for this issue -- this seems like the wrong place to fix it. Can you reproduce it with `npm@latest`? This seems like putting a bandaid on a corner of a bigger wound, and would probably cause further, harder-to-find errors if it doesn't (conveniently) crash this early.

richardpringle:
Hey @zkat, I actually no longer have access to the codebase where I was getting the error.

>This seems like putting a bandaid...

Haha, you're absolutely right. I would never have expected to land the code as is, but instead work together to find the deeper issue, this was just a starting point. 

Have you seen any similar issues pop up anywhere? I think you can close this PR since this is not an easy error to reproduce (without that original repo).

Thanks for the help, hopefully all is fixed in the latest release.

zkat:
@richardpringle I think I saw something like this some months ago, which is why I have a feeling we fixed it. /cc @iarna in case she remembers what it was.

andelmad (approved):

richardpringle:
Hey friends, I'm going to close this one. I can't reproduce the issue anymore.

(Great job with npm 6 by the way, 5 was a struggle but so far 6 had been smooth sailing)

PR #19676, opened 2018-01-23
Fix cache location on Windows

VladRassokhin:
Make consistent with `cache` cli command documentation


zkat (approved):
Cool! Thanks for the patch :)

PR #19701, opened 2018-01-27
Fix `npm version` not working in cygwin

ForsakenHarmony:


ForsakenHarmony:
ok this has to be done differently, but has to detect cygwin

ForsakenHarmony:
Where does `localPrefix` come from?

iarna:
@ForsakenHarmony `localPrefix` is the current module directory.. so if I'm in `/home/me/my-project/a/b/c` and there's a `package.json` in `/home/me/my-project/my-project` then that's what `localPrefix` will be set to. It's the top level of the current project and where `node_modules` will be created. It contrasts with `globalPrefix` which is where things go when you use `npm i -g`.

It's set in `lib/config/load-prefix.js` and the value is computed by [`find-npm-prefix`](https://www.npmjs.com/package/find-npm-prefix).

If you've got more questions or would like to talk about this and get responses in a somewhat more timely fashion, I encourage you to drop by #js-npm on the https://package.community Discord!

zkat:

PR #19725, opened 2018-02-01
Fixed bug in unsupported.js

jayluxferro:
Changed ``  ` `` to  ``'``.


kenany:
Can you please explain what bug this fixes?

stephnr:
@jayluxferro that string is a templated string. Switching it to single quotes will prevent the variable `rel.min` from being added to the string.

nwoltman:
Duplicate of #18953.

zkat:
Closing as duplicate of #18953 -- also, this patch as-is would... not do what it's supposed to. Thanks for sending it in, though!

PR #19726, opened 2018-02-01
Fix #18781 - name and version optional

### LINKED ISSUE #18781: Document name and version as optional for private packages
(This is a documentation change request)

I generally use npm for private projects. Even though I download public dependencies via npm, I have no intention of ever publishing most of my projects on npm, since they're too specific to be of use to others.

The [documentation for package.json](https://docs.npmjs.com/files/package.json) currently states that "The most important things in your package.json are the name and version fields. **Those are actually required**, and your package won't install without them".

This makes sense if I'm developing a package meant to be published and installed as a dependency on other projects.
However, on private projects, specifying the name and version values seems unnecessary. In fact, if I set `"private": true` and omit the name and version fields, nearly all tools work and show no warnings.

So, first, a question: are `name` and `version` officially considered to be required fields even for private projects?
And if not, could this exception be clarified in the documentation?

ngarnier:
This PR fixes #18781 by explaining that `name` and `version` are required only to publish your package.

### DIFF
--- a/doc/files/package.json.md
+++ b/doc/files/package.json.md
@@ -11,11 +11,11 @@ settings described in `npm-config(7)`.
 
 ## name
 
-The *most* important things in your package.json are the name and version fields.
-Those are actually required, and your package won't install without
-them.  The name and version together form an identifier that is assumed
-to be completely unique.  Changes to the package should come along with
-changes to the version.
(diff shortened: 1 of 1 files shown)


zkat (approved):
This seems fine to include 👍  Thanks for the patch!

PR #19807, opened 2018-02-12
Update npm-test.md

mattce:
- add `tes` as one more possible item to the list of `aliases`


iarna:
This s a bit of a weird and fine line but `tst` and `t` are official shorthands and are documented throughout the product as such. (They also show up in the commands in the usage info when you run `npm` with no arguments.) 

By contrast `tes` is a programmatically generated short hand (produced by passing the command list to `abbrev`) and none of those are documented. 

As such I think I'm going to decline to take this, but thank you for putting it together just the same.

PR #19817, opened 2018-02-13
install: count contributors and mention npx thanks

kemitchell:
This small pull request affects console reporting on `npm install` and friends.

In addition to reporting the number of packages added, the script now also reports the number of unique contributors to those added packages, based on `author` and `contributors` data. When the script reports a contributors count, it also prints a message encouraging users to run @feross' `npx thanks` for more information on how to support contributors to the code they use.

I've also added a TAP test, based on the test for `npm install --report`.

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -774,6 +774,9 @@ Installer.prototype.printInstalledForHuman = function (diffs, cb) {
   var added = 0
   var updated = 0
   var moved = 0
+  // Count the number of contributors to packages added, tracking
+  // contributors we've seen, so we can produce a running unique count.
+  var contributorsSet = []
   diffs.forEach(function (action) {
     var mutation = action[0]
     var pkg = action[1]
@@ -784,6 +787,28 @@ Installer.prototype.printInstalledForHuman = function (diffs, cb) {
       ++moved
     } else if (mutation === 'add') {
       ++added
+      // Count contributors to added packages. Start by combining `author`
+      // and `contributors` data into a single array of contributor-people
+      // for this package.
+      var people = []
+      var meta = pkg.package
+      if (meta.author) people.push(meta.author)
+      if (meta.contributors && Array.isArray(meta.contributors)) {
+        people = people.concat(meta.contributors)
+      }
+      // Make sure a normalized string for every person behind this
+      // package is in `contributorsSet`.
+      people.forEach(function (person) {
+        // Ignore errors from malformed `author` and `contributors`.
+        try {
+          var normalized = normalizePerson(person)
+        } catch (error) {
+          return
+        }
+        if (contributorsSet.indexOf(normalized) === -1) {
+          contributorsSet.push(normalized)
+        }
+      })
     } else if (mutation === 'update' || mutation === 'update-linked') {
       ++updated
     }
@@ -795,7 +820,11 @@ Installer.prototype.printInstalledForHuman = function (diffs, cb) {
     }).join('\n') + '\n'
   }
   var actions = []
-  if (added) actions.push('added ' + packages(added))
+  if (added) {
+    var action = 'added ' + packages(added)
+    if (contributorsSet.length) action += from(contributorsSet.length)
+    actions.push(action)
+  }
   if (removed) actions.push('removed ' + packages(removed))
   if (updated) actions.push('updated ' + packages(updated))
   if (moved) actions.push('moved ' + packages(moved))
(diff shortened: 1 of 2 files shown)


kemitchell:
Two lingering thoughts:

1. ~Do I need to worry more about memory usage in compiling the list of unique contributors?~

    Resolved https://github.com/npm/npm/pull/19817#issuecomment-365444384
￼￼￼￼￼
2. ~I couldn't find any existing tests of the CLI install summary output. Did I miss something ... and possibly break it?~

    Resolved https://github.com/npm/npm/pull/19817#issuecomment-365444384

lib/install.js
```diff
@@ -774,6 +774,9 @@ Installer.prototype.printInstalledForHuman = function (diffs, cb) {
   var added = 0
   var updated = 0
   var moved = 0
+  // Count the number of contributors to packages added, tracking
+  // contributors we've seen, so we can produce a running unique count.
+  var contributorsSet = []
```
iarna: You can use a `Set` here and get uniqueness for free

kemitchell: No compat. issue? Then will do.

kemitchell:
@iarna, 476e38fc66b2fa550188c7922e3f4133aedfe005 refactors to use ES2015 `Set`.

iarna:
Given that we already have all of the package info in memory (far more than just the contrib list) I'm not overly concerned about it.

There aren't tests of ONLY output, but there are a bunch of tests that include output in their assertions. It looks like your change doesn't impact them, however.

The test failure you're seeing is fixed in 30dc1b1d8 in `release-next`, so don't worry about it.

lib/install.js
iarna: Yup, we're Node 4+, which includes Set and Map and Symbol!

lib/install.js
```diff
@@ -784,6 +787,26 @@ Installer.prototype.printInstalledForHuman = function (diffs, cb) {
...
+      var people = []
+      var meta = pkg.package
+      if (meta.author) people.push(meta.author)
+      if (meta.contributors && Array.isArray(meta.contributors)) {
+        people = people.concat(meta.contributors)
+      }
+      // Make sure a normalized string for every person behind this
+      // package is in `contributors`.
+      people.forEach(function (person) {
+        // Ignore errors from malformed `author` and `contributors`.
```
iarna, line 802: Under what circumstances could `normalizePerson` throw? To my reading it would only throw if `person` were `undefined` or `null`? If that's the case, then I think it would be better to just skip those cases…

kemitchell: I didn't have a case in mind. Just `try`'d it out of prudence.

Perhaps a `TypeError` on String coercion, if `person.name` or similar is an Object?

kemitchell:
Amended to fix a typo in a comment.

iarna (approved):

lib/install.js
```diff
@@ -808,13 +835,35 @@ Installer.prototype.printInstalledForHuman = function (diffs, cb) {
...
 
   output(report)
   return cb()
 
   function packages (num) {
     return num + ' package' + (num > 1 ? 's' : '')
   }
+
+  function from (num) {
+    return ' from ' + num + ' contributor' + (num > 1 ? 's' : '')
```
billyjanitsch, line 852: A tiny suggestion:

In the message "installed N packages from M contributors", it's ambiguous whether "from" applies to "packages" or "installed". Read the latter way, it sounds like packages are being installed from multiple places (like multiple contributing registries or something?) which might be confusing for less npm-savvy folks, esp those unfamiliar with the recent Twitter etc. conversations motivating this change.

Is there a way to rephrase? s/from/created by/?

kemitchell: "by"?

billyjanitsch:
Oo also: maybe this could wait until the [issue of generality](https://github.com/feross/thanks/issues/2) is sorted out? I'm extremely excited about thanks, but if it's gonna be recommended by the npm CLI on _every_ additive install, it's a bummer for the donation information to be centralized -- especially when the tool's wording ("You depend on X authors and Y teams who are seeking donations!") suggests that _only_ authors who have explicitly opted in to this userland project require support.

iarna:
I don't see any value in waiting, especially seeing as we're not bundling any particular version of `thanks` and instead encouraging folks to use the evergreen `npx thanks`, which means the moment that lands it will become available to users.

iarna:
This PR has been merged to release-next as 737dc85e1, thanks again! =)

billyjanitsch:
@iarna would you take a PR with the [wording tweak](https://github.com/npm/npm/pull/19817#discussion_r168048436)?

iarna:
@billyjanitsch and I chatted a bunch more over on Discord:

1. That wording tweak sounds good (and I'll just make it directly, don't worry about a PR).

2. There are reasonable concerns about endorsing tools this way. No conclusions as yet.

3. @iarna would like `npx thanks` to have _some_ pointer to how you participate in the "please contribute" economy. Also maybe some credit to folks who _aren't_ looking for donations?

feross:
Wow! Just saw this. You all move quickly! 

Open to working on the mentioned issues. Is there a log of the Discord discussion somewhere so I can get up-to-speed?

**Edit:** Found it

kemitchell:
For the record, I work on a different set of tools for contributor support, have criticized Patreon, GratiPay, Liberapay, and their ilk as approaches to the problem, and am still in favor of having `npm i` mention `npx thanks`.

PR #19836, opened 2018-02-16
Fix optional dependencies with npm outdated

robscotts4rb:
Previously the only way to check for updates on optional packages was by running:

`npm outdated --save false --save-optional`

This is cumbersome, but also means you have to do two specific operations to update your normal  and optional dependencies. (e.g. One operation to check `dependencies` and one to check `optionalDependencies`)

This allows some cross over between the commands so you can update everything in one shot.

`npm outdated --save-optional` will now list all packages from both `dependencies` and `optionalDependencies`

The original behaviour is still in place when you run the `npm outdated --save false --save-optional` command.

### DIFF
--- a/lib/outdated.js
+++ b/lib/outdated.js
@@ -247,10 +247,10 @@ function outdated_ (args, path, tree, parentHas, depth, cb) {
     deps.forEach(function (dep) {
       types[moduleName(dep)] = 'devDependencies'
     })
-  } else if (npm.config.get('save')) {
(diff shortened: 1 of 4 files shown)


iarna:
I uh, wow, that original behavior is almost too strange for words. 

TBH, the fact that it doesn't do optional dependencies by default seems pretty weird to me as it is.

zkat:

PR #19846, opened 2018-02-17
update to latest version of pacote (Security Bug)

knoxcard:
✗ Medium severity vulnerability found on ssri@5.0.0
- desc: Regular Expression Denial of Service (ReDoS)
- info: https://snyk.io/vuln/npm:ssri:20180214
- from: node_services@1.0.0 > npm@5.6.0 > pacote@7.0.2 > ssri@5.0.0
Your dependencies are out of date, otherwise you would be using a newer ssri than ssri@5.0.0.
Try deleting node_modules, reinstalling and running `snyk test` again.
If the problem persists, one of your dependencies may be bundling outdated modules.

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,7 +93,7 @@
     "once": "~1.4.0",
     "opener": "~1.4.3",
     "osenv": "~0.1.4",
-    "pacote": "^7.0.2",
+    "pacote": "^7.4.0",
     "path-is-inside": "~1.0.2",
     "promise-inflight": "~1.0.1",
     "qrcode-terminal": "~0.11.0",


zkat:
Hey! Thanks for the contribution -- I'm going to skip this particular PR for the simple reason that we'll be updating `pacote` to its latest version, which is already past `7.4.0`, with the upcoming release.

For those worrying about the security implications of this: npm is not at all affected by the `ssri` issue -- in fact, no tools using `ssri` for npm compatibility are vulnerable, because the issue only affects strict-mode, which npm explicitly does not use.

Cheers!

knoxcard:
Thanks for your contribution and clarification on the issue!

PR #19857, opened 2018-02-19
docs: Fix default value for save

nalinbhardwaj:
https://docs.npmjs.com/misc/config#save mentions that the default config value is `false`, however, in actuality, it is `true`, which can be verified from the code [here](https://github.com/npm/npm/blob/0cc9d89ed2d46745f91d746fda9d205fd39d3daa/lib/config/defaults.js#L204).

I haven't read/contributed to npm before, so let me know if there's something wrong with it. I was just very confused by this information in the docs, so made a PR!

### DIFF
--- a/doc/misc/npm-config.md
+++ b/doc/misc/npm-config.md
@@ -730,7 +730,7 @@ Remove failed installs.
 
 ### save
 
-* Default: false
+* Default: true
 * Type: Boolean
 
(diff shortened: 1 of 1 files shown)


nalinbhardwaj:
Closing for separate PR to `npm:latest`(not `master`, like I did). Sorry for the error.

PR #19879, opened 2018-02-21
Get rid of nock from .gitignore

watilde:
Remove `nock` from `.gitignore` which is already removed in package.json


zkat (approved):
Thanks! nock was removed back in Dec 2015, so I think this is a pretty safe change here.

PR #19892, opened 2018-02-23
Update npm outdated, and thus npm update, to not install > the latest tag

iarna:
This makes it actually use `npm-pick-manifest` so we _know_ it's the same logic as the installer.


lib/install/deps.js
```diff
@@ -283,10 +283,12 @@ function computeVersionSpec (tree, child) {
...
   var requested
   var childReq = child.package._requested
-  if (childReq && (isNotEmpty(childReq.saveSpec) || (isNotEmpty(childReq.rawSpec) && isNotEmpty(childReq.fetchSpec)))) {
+  if (child.isLink) {
+    requested = npa.resolve(child.package.name, 'file:' + child.realpath, getTop(tree).path)
+  } else if (childReq && (isNotEmpty(childReq.saveSpec) || (isNotEmpty(childReq.rawSpec) && isNotEmpty(childReq.fetchSpec)))) {
     requested = child.package._requested
   } else if (child.package._from) {
-    requested = npa(child.package._from)
+    requested = npa(child.package._from, tree.path)
```
zkat, line 291: Aren't all `_from` supposed to use `getTop(tree)`?

zkat (approved):
Yes please

test/tap/outdated-latest.js
```diff
@@ -0,0 +1,111 @@
...
+  setup()
+  mr({port: common.port, throwOnUnmatched: true}, function (err, s) {
+    if (err) throw err
+    server = s
+    server.get('/example').reply(200, examplePackument)
+    t.done()
+  })
+})
+
+test('example', function (t) {
```
dougmoscrop, line 96: what would be the expected behavior when a user runs `npm upgrade` if `beta` stayed at `1.2.6`, installed/wanted was `^1.2.1` and then `2.0.0` had been released as `latest`?

zkat: This is the algorithm npm uses to pick versions: https://github.com/zkat/npm-pick-manifest/blob/latest/index.js

The "tag" is a special value that happens to be a cutoff if the algorithm happens to run into it.

My expectation in this case is that `1.2.6` is what would be installed if your `package.json` says `^1.2.1` for that dependency, since `latest` wouldn't be among those. But if you do `npm i pkg@latest`, then you'd get `2.0.0`.

dougmoscrop: thanks - not trying to be a pain, but think this means that a user can still get sort of accidentally steered in to a prerelease/alternative dist-tag using `npm upgrade` unless the package maintainer takes some kind of preventative measure*

it boils down to semver being assumed to be linear, but dist-tags are trying to create branches, I think?

* (example that is obviously too complicated to actually adopt: always publish a semver-patch non prerelease version after pre-releases -- so, if npm wanted to release 6 as the next thing after 5.7.1, they should actually publish 5.7.2 from their non-prelease branch to add a guard for users at ^5 that are doing upgrade)

^ this example is silly, and the docs do say `npm i -g npm@latest` is the right upgrade path, I'm just not sure how one would document `update` in such a way that people will know what to expect

iarna: The only tag that has meaning to npm is `latest` so all logic must be expressed in terms of that. All other tags are arbitrary.

lib/install/deps.js
iarna: `_from` is taken from what was in the `package.json` of the requiring package and is relative to that package, not to the top of the tree. When we save froms to the package-lock they're relative to the tree root.

test/tap/outdated-latest.js
dougmoscrop: I see. Would a `--safe` flag that only updates if `latest` semver matches wanted be useful? Should it be the default? This would bail in the above scenario rather than installing a pre-release/beta/rc version on a users system that had only ever done `npm i foo@latest` and then (some time later) `npm upgrade foo`

Maybe I am obsessing over a corner case here, and the takeaway should be to only install `latest` or a specifically known version that has been released.

andelmad (approved):

test/tap/outdated-latest.js
iarna: The entire point of this patch is to make it so that the user who does `npm upgrade foo` doesn't get a pre-release version, so I think we're pretty in alignment here. =)

iarna:
This landed in v6.0.0-next.0

PR #19893, opened 2018-02-23
Add a common open reason to the issue template

MrsTonedOne:
`npm is doing something I don't understand.` is often inappropriately used when the author means `npm is producing incorrect or undesirable behavior.`. I only had to look thru the first page of open issues to find 5 examples of this.


zkat (approved):
Hey!

We want to eventually overhaul this whole template, but I see no problem pulling this in for the benefit of folks having a somewhat better option on this.

Thanks for the contribution!

Issue #19832, opened 2018-02-16
NPM 4.6.1 depends on vulnerable request@2.81


khitrenovich:
There is a fresh NSP advisory published today - https://nodesecurity.io/advisories/566.

NPM 4.6.1 is affected via `request` package - 
```
npm@4.6.1 > request@2.81.0 > hawk@3.1.3 > hoek@2.16.3
```

NPM 5 brings `request@2.83`, which is not vulnerable, However, it is a major release that brings breaking changes, and quick upgrade is not always possible/desirable. Can we expect another v4.6.x patch that will bring non-vulnerable `request` package?

Thanks!

PR #19900, opened 2018-02-23
Fix issue #19832: NPM 4.6.1 depends on vulnerable request@2.81

yvan-sraka:
Should be merged in `4.6.1` brach version!


zkat:
We're not taking this: we aren't planning on doing a new release of npm@4 any time soon, and this particular vulnerability does not actually affect npm, so audit is being a bit overzealous here.

Thanks for taking the time to put this together, though! Security's important and stuff :)

PR #19902, opened 2018-02-23
Add emacs and vim backup files to gitignore

Mstrodl:
Vim and Emacs leave backup files in the file directory which we don't want commited

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
+\#*\#
+*~
 *.swp
 npm-debug.log
 .nyc_output/


Alexsey:
You should use `.git/info/exclude` for local environment specific ignores. It's your personal `.gitignore` ;)

Mstrodl:
@Alexsey Someone added `.swp` there though?

Alexsey:
I don't have experience with `.swp` and don't know when it could be created and why. Probably someone else should comment on that

Mstrodl:
@Alexsey `.swp` is vim's swapfile, it allows vim to recover work if it crashes

zkat:
That `.swp` was added back in 2010, by @isaacs. I kinda feel like turning down this PR because I think editor-specific utility files should be ignored the way @Alexsey says, and I don't want to keep piling stuff into that file over time.

I don't wanna just shut down your PR, actually, so I'm wondering if you have a strong argument in favor of keeping this (as opposed to just removing `.swp`, if you want consistency).

Mstrodl:
@zkat I mean, I wouldn't have made the PR if I saw that `*.swp` wasn't there and I thought it was a bit silly for there to be one but not the other. I honestly don't have much of an opinion either way so long as it stays consistent.

zkat:
@Mstrodl wanna send me a separate PR that removes `.swp`? :)

Mstrodl:
@zkat Sure! Should I close this one then? I've opened #20000

zkat:
yep. Thanks!

PR #19904, opened 2018-02-24
fix: use detect-newline to detect newline instead of default \n

tuananh:
use `detect-newline` to detect newline instead of using the default `\n` which may cause npm to find that `packagejson` string to be different even though there is no change. (causing by the `\n` newline)

### DIFF
--- a/lib/install/save.js
+++ b/lib/install/save.js
@@ -124,7 +124,7 @@ function savePackageJson (tree, next) {
       tree.package.bundleDependencies = deepSortObject(bundle)
     }
 
-    var json = JSON.stringify(tree.package, null, indent) + newline
+    var json = JSON.stringify(tree.package, null, indent).replace(/\n/g, newline) + newline
     if (json === packagejson) {
       log.verbose('shrinkwrap', 'skipping write for package.json because there were no changes.')
       next()
--- a/lib/install/save.js
+++ b/lib/install/save.js
@@ -3,6 +3,7 @@
 const createShrinkwrap = require('../shrinkwrap.js').createShrinkwrap
 const deepSortObject = require('../utils/deep-sort-object.js')
 const detectIndent = require('detect-indent')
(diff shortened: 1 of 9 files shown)


lib/install/save.js
```diff
@@ -122,7 +124,7 @@ function savePackageJson (tree, next) {
       tree.package.bundleDependencies = deepSortObject(bundle)
     }
 
-    var json = JSON.stringify(tree.package, null, indent) + '\n'
+    var json = JSON.stringify(tree.package, null, indent) + newline
```
zkat: This does not replace all the newlines in the JSON file. I'm pretty sure you're gonna have to `JSON.stringify(...).replace(/\n/g, newline)` here, not just append the final one.

lib/install/save.js
```diff
@@ -3,6 +3,7 @@
 const createShrinkwrap = require('../shrinkwrap.js').createShrinkwrap
 const deepSortObject = require('../utils/deep-sort-object.js')
 const detectIndent = require('detect-indent')
+const detectNewline = require('detect-newline')
```
zkat, line 6: This is not the only file that needs you to do this. You'll also have to add this to `lib/shrinkwrap.js` and `lib/version.js`, which both write out json files.

zkat (changes requested):
Thanks for taking the time and initiative to do this!! There's a bit more that needs to go into this patch for it to be ready, and I'm honestly fine not having test for a thing like this, provided the regular test suite continues passing.

lib/install/save.js
tuananh: i updated the PR to cover `lib/shrinkwrap.js` and `lib/version.js` as well.

zkat (approved):
LGTM! Thanks a bunch for this patch!

nwoltman:
#18943 from 4 months ago solved this in IMO a cleaner way and included tests. Guess you didn't see it...

PR #19905, opened 2018-02-24
Match node's resolution when validating peer dependencies

### LINKED ISSUE #15708: UNMET PEER DEPENDENCY
Hi,
I'm opening this issue because npm is doing something I don't understand.

I get the error message UNMET PEER DEPENDENCY even though I have the said dependency installed.

I'm tying to install eslint-config-airbnb globally and it gives me:

npm install -g eslint-config-airbnb
/usr/local/lib
├── UNMET PEER DEPENDENCY eslint@^3.15.0
├─┬ eslint-config-airbnb@14.1.0
│ ├── UNMET PEER DEPENDENCY eslint@^3.15.0
│ └── UNMET PEER DEPENDENCY eslint-plugin-import@^2.2.0
├── UNMET PEER DEPENDENCY eslint-plugin-import@^2.2.0
├── UNMET PEER DEPENDENCY eslint-plugin-jsx-a11y@^3.0.2 || ^4.0.0
└── UNMET PEER DEPENDENCY eslint-plugin-react@^6.9.0

npm WARN eslint-config-airbnb@14.1.0 requires a peer of eslint@^3.15.0 but none was installed.
npm WARN eslint-config-airbnb@14.1.0 requires a peer of eslint-plugin-jsx-a11y@^3.0.2 || ^4.0.0 but none was installed.
npm WARN eslint-config-airbnb@14.1.0 requires a peer of eslint-plugin-import@^2.2.0 but none was installed.
npm WARN eslint-config-airbnb@14.1.0 requires a peer of eslint-plugin-react@^6.9.0 but none was installed.
npm WARN eslint-config-airbnb-base@11.1.0 requires a peer of eslint@^3.15.0 but none was installed.
npm WARN eslint-config-airbnb-base@11.1.0 requires a peer of eslint-plugin-import@^2.2.0 but none was installed.

I have the all installed, latest, and even tried to uninstall and the reinstall the exact required version, but still get the error.

And this is a quit common problem for me, getting UNMET PEER DEPENDENCY even though I have the said dependency installed.

 npm v: 4.1.2
node v: 7.5.0
npm config get registry: https://registry.npmjs.org/
macOS 10.12.3

bjornstar:
Fixes #15708

I've created an organization to reproduce the problem: `peer-deps-repro`

To reproduce:
``` bash
git clone git@github.com:peer-deps-repro/main.git
git checkout v1.2.0
npm install
```

There should be no warnings when you install.

The problem is that the validator was not using the same module resolution strategy as node and matching the wrong version. We need to go up the tree and use the nearest peer dependency.

### DIFF
--- a/lib/install/deps.js
+++ b/lib/install/deps.js
@@ -683,8 +683,10 @@ var validatePeerDeps = exports.validatePeerDeps = function (tree, onInvalid) {
     try {
       var spec = npa.resolve(pkgname, version)
     } catch (e) {}
-    var match = spec && findRequirement(tree.parent || tree, pkgname, spec)
-    if (!match) onInvalid(tree, pkgname, version)
+    tree.requiredBy.forEach(function (requiredBy) {
+      var match = spec && findRequirement(requiredBy, pkgname, spec)
+      if (!match) onInvalid(tree, pkgname, version)
+    })
   })
 }
 


bjornstar:
It turns out that this is not the same resolution strategy and the deduping method used since v3.0.0 is flawed. Child dependencies are getting the wrong version.

PR #19907, opened 2018-02-24
Consolidate code for stringifying package data and add tests

nwoltman:
This reduces code duplication for detecting indentation and new lines in `package[-lock].json` and stringifying the data. This also improves performance on Unix systems since it avoids replacing newlines when it's not necessary.

Tests have been added to test that `package[-lock].json` files are written to disk with their original line endings.

Mostly based on #18943 rebased after #19904.


zkat (approved):
This is awesome! Thanks so much for the refactor and the well-written tests! 👍 

Sorry again for clobbering your old PR like that! 🙊

nwoltman:
@zkat That's okay! Thanks for all the work you do :)

PR #19908, opened 2018-02-25
pack/publish: --dry-run, --json, and tarball content previews

zkat:
screen shot 2018-02-24 at 16 38 27

This patch adds three features:

1. a `notice` on both `npm pack` and `npm publish` that always outputs a chunk of details about the tarball(s) involved.
2. support for the `--dry-run` flag for both `pack` and `publish` -- this "dry run" _will_ run npm scripts normally, but just won't create or upload the tarball.
3. support for the `--json` flag, to programmatically get the data about packages in JSON format.


andelmad (approved):

aleclarson:
This should be documented :)

PR #19910, opened 2018-02-25
view: humanize and prettify default output

zkat:
This is an overhaul of the basic version of the `npm view` command such that the default non-json view displays package information in a more compact and useful way.

Here's some examples of what it looks like now:

`npm view lodash`:
screen shot 2018-03-15 at 14 38 53

`npm view ssri`:
screen shot 2018-03-15 at 14 39 09

`npm view npm`:
screen shot 2018-03-15 at 14 39 28


zkat:
two notes:
* get rid of stars
* blue on black shitsux

PR #19929, opened 2018-02-27
config: disable color by default when $NO_COLOR is set

leahneukirchen:
This makes npm follow the NO_COLOR standard, as explained in
http://no-color.org/

> All command-line software which outputs text with ANSI color added
> should check for the presence of a NO_COLOR environment variable that,
> when present (regardless of its value), prevents the addition of ANSI
> color.

npm already provides environment variables (npm_config_color=false) and
options (--no-color) to disable coloring by default, but NO_COLOR is an
comprehensive approach to disable colors for all tools.

### DIFF
--- a/doc/misc/npm-config.md
+++ b/doc/misc/npm-config.md
@@ -284,6 +284,9 @@ This is a list of CIDR address to be used when configuring limited access tokens
 If false, never shows colors.  If `"always"` then always shows colors.
 If true, then only prints color codes for tty file descriptors.
 
+This option can also be changed using the environment: colors are
+disabled when the environment variable `NO_COLOR` is set to any value.
+
 ### depth
 
 * Default: Infinity
--- a/lib/config/defaults.js
+++ b/lib/config/defaults.js
@@ -130,7 +130,7 @@ Object.defineProperty(exports, 'defaults', {get: function () {
 
     cidr: null,
 
-    color: true,
+    color: process.env.NO_COLOR == null,
     depth: Infinity,
     description: true,
     dev: false,


lib/config/defaults.js
```diff
@@ -130,7 +130,7 @@ Object.defineProperty(exports, 'defaults', {get: function () {
 
     cidr: null,
 
-    color: true,
+    color: typeof(process.env.NO_COLOR) === 'undefined',
```
watilde: According to the error on CI, we might need to replace here with this: `typeof process.env.NO_COLOR`.

Refs: https://github.com/eslint/eslint/blob/master/docs/rules/no-extra-parens.md

zkat: This is better off as `!process.env.NO_COLOR` or, if you want `NO_COLOR=""` to be truthy, `process.env.NO_COLOR != null` (with one `=`)

doc/misc/npm-config.md
```diff
@@ -278,7 +278,7 @@ This is a list of CIDR address to be used when configuring limited access tokens
 
 ### color
 
-* Default: true
+* Default: true, unless the environment variable `NO_COLOR` is set
```
zkat: You can remove this. Please mention `NO_COLOR` in the description below, instead, and how to use it. I'd rather keep this describing what the "no action" default is.

zkat (changes requested):
I'm 👍 on this after a couple of review edits. Thanks for proposing this!

zkat (approved):
Thanks for the changes! This looks great, and I'll be including it in this week's release! 🎉

PR #19950, opened 2018-03-01
Add a note to install --production.

kyranet:
The `npm install --production` CLI command was quite confusing when it comes to the difference between installing the dependencies for a project and for adding a dependency to a project. This PR adds a note about it specifying the flag is only meaningful in the former case.


zkat (approved):
LGTM! Thanks for this clarification! I'm sure it'll help clarify things for folks trying to make sense of all our flags :)

PR #19993, opened 2018-03-07
A trio of fixes to optional dependency handling

iarna:
1. When running `npm i -O module` we weren't saving `module`'s optional status to the `package-lock.json`. This was due to our adding the module to the in memory `optionalDependencies` but not also adding it to `dependencies`, which resulted in it not having any `requiredBy`, which in turn meant that the optional check on the module object run when saving the package-lock failed, resulting it being saved as an ordinary dependency.
2. We were filtering optionals during `loadDeps`. This moves the filtering to `diff-trees`, in the same place we filter dev and production dependencies. The `loadDeps` filter had two problems:
    1. First, if you didn't have a package-lock, but did have an optional dependency, but couldn't install it on your current platform then it wouldn't be saved to the new package-lock (as it would never be added to the tree.
    2. Second, if you HAD a package-lock that already had an optional dependency then `loadDeps` wouldn't have anything to do and so it's skipping optional dependencies wouldn't matter. This meant that `--no-optional` with a package-lock had no effect.
3. Previously our ability to detect optionals was predicated on having an `optionalDependencies` field in the package manifest. Ordinarily this is fine, as indeed, that's where it's stored. But when installing from a lock file we now generate synthetic manifests (_fake children_ in the jargon of cli internals) and our lock file doesn't track which _requirements_ are optional so we can't regenerate this. The fix to this was to look for the lock-file's optional annotation on the module itself when any of the things that require it  are synthetic (and thus can't express this themselves).

* [x] A brief regression test, demonstrating proper package-lock updating with `npm i -O` and that it obeys `--no-optional` with an extant package-lock.
* [x] Check if `devDependencies` suffer from the same problem as #3 (which would call for the same solution). (edit: it does not)
* [x] Ensure that _replacing_ the optional dependency when it's `requiredBy` are synthetic still marks the dependency as optional

### DIFF
--- a/lib/install/deps.js
+++ b/lib/install/deps.js
@@ -260,6 +260,7 @@ exports.loadRequestedDeps = function (args, tree, saveToDependencies, log, next)
             delete tree.package[saveType][childName]
           }
         }
+        if (child.save === 'optionalDependencies') tree.package.dependencies[childName] = child.saveSpec
       }
 
       // For things the user asked to install, that aren't a dependency (or
--- a/lib/install/deps.js
+++ b/lib/install/deps.js
@@ -489,12 +489,6 @@ function loadDeps (tree, log, next) {
   if (!tree.package.dependencies) tree.package.dependencies = {}
   asyncMap(Object.keys(tree.package.dependencies), function (dep, done) {
     var version = tree.package.dependencies[dep]
-    if (tree.package.optionalDependencies &&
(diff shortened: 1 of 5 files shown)


andelmad (approved):

zkat (approved):
🐑

iarna:
merged

PR #20000, opened 2018-03-08
Remove *.swp files from .gitignore

Mstrodl:
Remove *.swp files from .gitignore because `.git/info/exclude` should be used for editor-specific files instead

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
-*.swp
 npm-debug.log
 .nyc_output/
 /test/bin


andelmad (approved):

iarna:
While that's technically correct I don't see any harm in having `.swp` in our gitignore nor do I seen any value in removing it.

Mstrodl:
@iarna Yeah, but then shouldn't `*~` be added then? I just don't see the value in having but not the other

PR #20027, opened 2018-03-12
Use a specific mtime when packing, rather than none at all

isaacs:
Thank god I found you. Listen, can you meet me at Twin Pines Mall tonight at 1:15? I've made a major breakthrough, I'll need your assistance.

### DIFF
--- a/lib/pack.js
+++ b/lib/pack.js
@@ -120,7 +120,9 @@ function packDirectory (mani, dir, target) {
         cwd: dir,
         prefix: 'package/',
         portable: true,
-        noMtime: true,
+        // Provide a specific date in the 1980s for the benefit of zip,
+        // which is confounded by files dated at the Unix epoch 0.
+        mtime: new Date('1985-10-26T08:15:00.000Z'),
         gzip: true
       }
 


zkat (approved):
💯

zkat:
cc @simonua

lib/pack.js
```diff
@@ -120,7 +120,9 @@ function packDirectory (mani, dir, target) {
         cwd: dir,
         prefix: 'package/',
         portable: true,
-        noMtime: true,
+        // Provide a specific date in the 1980s for the benefit of zip,
+        // which is confounded by files dated at the Unix epoch 0.
+        mtime: new Date('1985-10-26T08:15:00.000Z'),
```
MylesBorins, line 125: when this tarball hits 88... we are going to see some serious 💩

MylesBorins: Jokes aside, is there a reason to set a date in the past as opposed to using the original mtime of the file?

If I'm not mistaken older versions would maintain the original mtime, see https://github.com/google/google-p12-pem/issues/27#issuecomment-371653125

zkat: The reason for this is that, by setting mtimes like this, we're able to make it so two computers with two checkouts of the same project at the same commit, barring extra noise in the signal, will both generate tarballs with the exact same hash. It means npm packages can be made compliant with [reproducible builds](https://reproducible-builds.org/). Obvs, npm isn't the only part of this story, and this also depends on people's own run-scripts modifying things, but at least npm's packer no longer gets in the way of this.

For a while now, we were using the `noMtime` option because it allowed those stable times, but it turns out there's some implementations that can't handle beginning-of-epoch timestamps (or timestamps before 1980), so this is the compromise.

simonua:
Thanks, Doc!

lib/pack.js
MylesBorins: A++ explanation thanks.

I know that Node is interested in exploring reproducible builds, and timestamps are definitely an issue for this.

PR #20029, opened 2018-03-12
inflate-shrinkwrap: allow fakeChildren for non-integrity types

zkat:
Fixes the case helpfully described in https://gist.github.com/jmorrell/057caf31563a95c7c9c880867afcfea3

/cc @jmorrell

### DIFF
--- a/lib/install/inflate-shrinkwrap.js
+++ b/lib/install/inflate-shrinkwrap.js
@@ -70,6 +70,11 @@ function normalizePackageDataNoErrors (pkg) {
 
 function inflatableChild (onDiskChild, name, topPath, tree, sw, requested, opts) {
   validate('OSSOOOO|ZSSOOOO', arguments)
+  const usesIntegrity = (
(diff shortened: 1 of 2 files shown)


jmorrell:
Oh yay! Thank you for taking a look at this :)

zkat:
@jmorrell thanks so much for the excellent repro. It was really easy to figure out what was going on (and whether I'd fixed it) because I was able to check what exactly you were running into, in context. Seriously, thanks ^_^

iarna (approved):

iarna:
Looks good—we definitely should be constructing fake children for any entry that matches what we produce, at the very least.

zkat:
Note: I'm gonna hold off on merging this. It's making some of our tests fail (probably because they weren't written expecting to take the fakeChild code path), so it'll have to go in next release.

iarna:
Landed in 6.0.0!

PR #20082, opened 2018-03-16
feat: allow optional @ prefix on scope

bcoe:
when providing a scope, e.g., during login, we encourage our users in docs to provide an `@` prefix on scope:

`npm login --scope=@foo`

we don't currently allow a `@` prefix for `team` commands, this is a slightly confusing disconnect that I felt was worth fixing.

### DIFF
--- a/lib/team.js
+++ b/lib/team.js
@@ -41,7 +41,7 @@ function team (args, cb) {
     try {
       return npm.registry.team(cmd, uri, {
         auth: auth,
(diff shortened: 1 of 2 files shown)


nzec (approved):
Looks Good RN!

zkat (approved):

iarna:
Thank you for this! It shipped in 6.0.0! =)

PR #20089, opened 2018-03-17
Update README.md

Seliniux777:
`fix marged`
@Teamux-Packages


Seliniux777 (commented):
`usage`

iarna:
Not clear to me what you were trying to do here. All this seems to do is rename our readme?

PR #20122, opened 2018-03-21
cli: shinier update-notifier

zkat:
This message is based on the one pnpm uses, which
I think takes care of all the false update messages and
missing -g that happens with the default notifier message.

Preview:
screen shot 2018-03-21 at 12 21 46

I did 4 things:
1. Moved the update check to post-`npm.load()` so we have access to config. I think this is fine -- if your npm isn't even loading, you can't even update, so.
2. Added a colorful message about what "level" the update is (major/minor/patch)
3. Added a link to the changelog right in the message
4. Forced a `-g` in the message because `update-notifier`'s detection of globals keeps messing up and if people have npm installed locally, they probably know well enough that they need to omit the -g -- vs folks getting confused because the `-g` is missing.

Cheers to @zkochan for the [corresponding version of this in `pnpm`](https://github.com/pnpm/pnpm/blob/master/src/checkForUpdates.ts), which I think is *way better* than the default, and I hope he's ok with us cribbing it like this. 💚

### DIFF
--- a/bin/npm-cli.js
+++ b/bin/npm-cli.js
@@ -25,12 +25,6 @@
 
   unsupported.checkForUnsupportedNode()
 
-  if (!unsupported.checkVersion(process.version).unsupported) {
-    var updater = require('update-notifier')
-    var pkg = require('../package.json')
-    updater({pkg: pkg}).notify({defer: true})
-  }
-
   var path = require('path')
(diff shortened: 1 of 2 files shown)


iarna (approved):

iarna:
Landed in 6.0.0!

PR #20126, opened 2018-03-21
feat: add npm cit command

SimenB:
I think that, similar to `npm it` instead of `npm i && npm t`, an `npm cit` instead of `npm ci && npm t` would be nice to have. This was pretty quick to throw together, so no hard feelings if it's rejected


iarna:
_quietly contemplates adding an `npm sit` easter egg to this_

iarna:
Thank you again, this landed in 6.0.0 =)

Issue #20069, opened 2018-03-15
npm hangs on any command - Windows


dougllima:
#### I'm opening this issue because:

  - [X] npm is crashing.

#### What's going wrong?

After a random time (even without doing anything on the computer), any NPM command that I execute will hangs forever, and Ctrl+C won't stop the process.

If any command hang, any other command will hang too, npm only will work again after I restart my PC (I have tried finish all proccess related to node, not related, log out from user... nothing works, only restart)

With silly loglevel i have this log on any command that hangs: 

```
npm info it worked if it ends with ok
npm verb cli [ 'C:\\Program Files\\nodejs\\node.exe',
npm verb cli   'C:\\Program Files\\nodejs\\node_modules\\npm\\bin\\npm-cli.js',
npm verb cli   'prefix',
npm verb cli   '-g' ]
npm info using npm@5.6.0
npm info using node@v8.10.0
npm verb exit [ 0, true ]
npm info ok
```

I have searched about it everywhere, tried all soluctions and nothing works!

This problem happens with another 2 co-workers, and with at least 5 friends, with different computers, settings and networks.

#### How can the CLI team reproduce the problem?

Well... i really don't know what is causing the problem, but if i just run any npm command (start, -v, install...) it hangs.

### supporting information:

 - `npm -v` prints:
 - `node -v` prints: v8.10.0
 - `npm config get registry` prints:
 - Windows, OS X/macOS, or Linux?: Windows 10
 - Network issues: None

rtwfroody:
I believe I have the same issue. When I run any `npm` command at all (including `npm -v`, `npm help`, `npm asdfasdf`), there is never any output. I can watch the node.js server process which grows until it uses about 1.4GB of RAM, and chews about 35% of my CPU continuously.

![image](https://user-images.githubusercontent.com/3102687/37748919-c8eadc7a-2d42-11e8-98e4-611fb7a3e44d.png)

`node -v` prints: `v8.10.0`

I tried before with node 9.9 and the behavior was the same.

My OS is Windows 10 Pro.

rtwfroody:
It appears that npm is stuck in a loop trying to mkdir(D:\) which is not a drive that exists on my system. (However, it did recently.) I'm going to see if I can figure out where it's getting D:\ from. Presumably there's some stale configuration somewhere.

Here's the stack trace:
```
fs.mkdir (fs.js:877)
[...]

PR #20152, opened 2018-03-24
Don't attempt to mkdir() a path that just failed

rtwfroody:
This fixes an infinite recursion on Windows when a drive that doesn't exist is
not specified. Eg. `mkdir D:\` would get ENOENT, find the "parent" which is
also D:\, and repeat infinitely.

See also #20069

### DIFF
--- a/node_modules/mkdirp/index.js
+++ b/node_modules/mkdirp/index.js
@@ -31,11 +31,15 @@ function mkdirP (p, opts, f, made) {
         }
         switch (er.code) {
             case 'ENOENT':
-                mkdirP(path.dirname(p), opts, function (er, made) {
-                    if (er) cb(er, made);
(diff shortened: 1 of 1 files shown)


iarna:
This seems like a great patch, but it's being made to an npm dependency. (You can tell because the file you changed is in `node_modules`.) That means the PR needs to be against that library, not against npm itself. If this was an npm maintained library I would be inclined to move your patch over to it, but in this case it is not. As such, changing this library is outside of our direct control.

If you could please PR this change to the mkdirp repository, it would be appreciated! [mkdirp](https://github.com/substack/node-mkdirp)

rtwfroody:
Done: https://github.com/substack/node-mkdirp/pull/144

Thanks for the pointer.

PR #20177, opened 2018-03-27
Add exception to environment variable substitution for :_ and :__ 

wilvk:
Fixes https://github.com/npm/npm/issues/20159

Currently server-specific environment variables are being converted to hyphens with the regex replace, thus causing them to be inaccessible for things like credentials.

By running the command (in Linux):

```
env npm_config_//nexus:8081/nexus/content/repositories/npm-test/:_password=YWRtaW4xMjM= bash
```

An environment variable is present such as:

```
npm_config_//nexus:8081/nexus/content/repositories/npm-test/:_password=YWRtaW4xMjM=
```

this is interpreted as the config variable:

```
{ '//nexus:8081/nexus/content/repositories/npm-test/:-password': 'YWRtaW4xMjM=' }
```

Where the password component is incorrect (`-password` instead or `_password`). 
This prevents the password from being set correctly for use later on in `/lib/config/get-credentials-by-uri.js`.

The same applies for other environment variables such as `__auth`.

The replacement I've used is a bit 'clunky' so if anyone wants to clean it up a bit, feel free.


wilvk:
bump - any updates/feedback on this PR?

lib/config/core.js
```diff
@@ -353,9 +353,15 @@ Conf.prototype.addEnv = function (env) {
...
       // leave first char untouched, even if
       // it is a '_' - convert all other to '-'
+      // except for if there is :_ or :__ as
+      // these are used for _password and __auth
       var p = k.toLowerCase()
-               .replace(/^npm_config_/, '')
-               .replace(/(?!^)_/g, '-')
+              .replace(/^npm_config_/, '')
+              .replace(/:_/, 'PLACEHOLDERONE')
+              .replace(/:__/, 'PLACEHOLDERTWO')
```
iarna: this second regexp will never match because `:__ABC` will be replaced with `PLACEHOLDERONE_ABC`

lib/config/core.js
```diff
@@ -353,9 +353,15 @@ Conf.prototype.addEnv = function (env) {
...
+              .replace(/(?!^)_/g, '-')
```
iarna: I wonder if the placeholder business could be ditched in preference for:

`.replace(/(?!^)(⁈[_:])_/g, '-')`

That is, only change `_` to `-` if it's not preceded by the start of the string, a `:` or a `_` . It's not quite the same, granted…

iarna (changes requested):
Ok, so beyond the implementation nits, a change like this would need docs and tests too.

But beyond that I'm concerned that making this kind of change to processing will mess up someone else's weird edge case. As there is a work around for your use case that makes me inclined to play things safe here and not take this.

The work around, and in fact, how we encourage people to configure systems like this is:

Add the following to your `npmrc`:

```
 //nexus:8081/nexus/content/repositories/npm-test/:_password=${NEXUS_PASSWORD}
```

Then set the `NEXUS_PASSWORD` env var, eg

```
env NEXUS_PASWORD=YWRtaW4xMjM= bash
```

iarna:
Oh, and if you'd like to chat more about this in real time I encourage you to drop by #js-npm on the https://package.community/ Discord.

wilvk:
Thanks for the reply @iarna , I wasn't aware that variable substitution worked in-line for `.npmrc`. The `$NEXUS_PASSWORD` approach seems like the best solution so will go with that for my stuff.

Will implement the docs, tests and suggested changes for if/when it is decided to go ahead with this PR.

Cheers,

zkat:

PR #20198, opened 2018-03-29
Load dependencies into ideal tree if installing with arguments

### LINKED ISSUE #16839: Missing dependencies after running `npm install` a second time

greysteil:
Previously, dependencies were only loaded into the tree during install if no
arguments were passed. That caused git dependencies to be deleted from the
`package-lock.json`.

Fixes #16839 (again)

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -456,23 +456,23 @@ Installer.prototype.loadAllDepsIntoIdealTree = function (cb) {
   if (installNewModules) {
     steps.push([validateArgs, this.idealTree, this.args])
     steps.push([loadRequestedDeps, this.args, this.idealTree, saveDeps, cg.newGroup('loadRequestedDeps')])
-  } else {
-    const depsToPreload = Object.assign({},
-      this.dev ? this.idealTree.package.devDependencies : {},
-      this.prod ? this.idealTree.package.dependencies : {}
-    )
-    if (this.prod || this.dev) {
-      steps.push(
-        [prefetchDeps, this.idealTree, depsToPreload, cg.newGroup('prefetchDeps')])
-    }
-    if (this.prod) {
-      steps.push(
-        [loadDeps, this.idealTree, cg.newGroup('loadDeps')])
-    }
-    if (this.dev) {
-      steps.push(
-        [loadDevDeps, this.idealTree, cg.newGroup('loadDevDeps')])
-    }
+  }
+
+  const depsToPreload = Object.assign({},
+    this.dev ? this.idealTree.package.devDependencies : {},
+    this.prod ? this.idealTree.package.dependencies : {}
+  )
+  if (this.prod || this.dev) {
+    steps.push(
+      [prefetchDeps, this.idealTree, depsToPreload, cg.newGroup('prefetchDeps')])
+  }
+  if (this.prod) {
+    steps.push(
+      [loadDeps, this.idealTree, cg.newGroup('loadDeps')])
+  }
+  if (this.dev) {
+    steps.push(
+      [loadDevDeps, this.idealTree, cg.newGroup('loadDevDeps')])
   }
   steps.push(
     [loadExtraneous.andResolveDeps, this.idealTree, cg.newGroup('loadExtraneous')])


greysteil:
I've added all of my context for this PR [here](https://github.com/npm/npm/issues/16839#issuecomment-377306641). I'm not a JavaScript developer, but am pretty sure that this is the right fix. I'd love help from a maintainer to write a test for this and get it reviewed. :octocat:

iarna (changes requested):
This is currently causing a LOT of test failures, there are reasons for the current behavior, vis-a-vis which modules are loaded when.

That being said, git modules being removed when they shouldn't be is something that should be resolved in the most recent releases (try `npm@next`). If it _isn't_ and you have a repro, I'd be happy to dig back into that. 

I'd be happy to chat more if you want to join the https://package.community/ Discord and drop by the #js-npm channel.

greysteil:
Thanks @iarna. The reproduction steps in https://github.com/npm/npm/issues/16839#issuecomment-377306641 still reproduce the bug on `5.9.0-next.0`. I'll have another dig into this now, but would love your help.

greysteil:
Dsicussed with @iarna on #js-npm channel and she has a much better fix than I could ever come up with. I'll leave this open so she can close it when she pushes it.

Thanks @iarna! ❤️

ezze:
@iarna, @greysteil, what is a fix? When will it be released? This is a blocking bug for our npm usage since 5.0.0.

greysteil:
@iarna's fix is to add the following to `doesChildVersionMatch`:

```
  if (requested.type === 'git' && child.fromShrinkwrap) {
    const fromSw = child.fromShrinkwrap
    // FIXME: if we start storing the install specifier for git repos in the lockfile

    // might not be true, but we can't tell currently
    if (!fromSw.hosted && !requested.hosted) return true

    return fromSw.hosted.toString() === requested.hosted.toString()
  }
```

Should drop in npm 6, which will be released this week.

iarna:
It is indeed out in npm@6.0.0, try it out today! =)

Issue #17346, opened 2017-06-20
npm5 prepare scripts not executing properly


eblin:
#### I'm opening this issue because:

  - [ ] npm is crashing.
  - [x] npm is producing an incorrect install.
  - [ ] npm is doing something I don't understand.
  - [ ] Other (_see below for feature requests_):

#### What's going wrong?
`prepare` script fails to execute properly when npm is running as user `root (uid:0)`

The lifecycle error is: `"cannot run in wd... etc"` (visible only with `-ddd`) since it's part `child_process` output

I'm guessing this is happening because of the permissions on child_process, since npm it's changing user (default to nobody) when running as root.

I fixed it by adding `--unsafe-perm` in here: https://github.com/npm/npm/blob/v5.0.3/lib/pack.js#L168

But I'm not sure if this is the right fix because of the obvious security risk around it, which might not matter since you are explicitly running as root anyways.

Also the install continues fine, I think it should halt and not let it finish as most likely the `prepare` script is a build process. As a user this is confusing because you think everything went well when in fact the `prepare` script never ran and now your module is "incomplete"

Let me know if you need more information or you'd like me to submit a PR. 

Thanks!

#### How can the CLI team reproduce the problem?

Just try to install any package that has a prepare script running and make sure you are running npm as root. As far as I can tell only happens when running as root, works fine when npm uses another user (which is why it works fine locally in my machine but not on the server).

### supporting information:

 - `npm -v` prints: 5.0.3
 - `node -v` prints: 6.11.0
 - `npm config get registry` prints: https://registry.npmjs.org/
 - Windows, OS X/macOS, or Linux?: All
 - Network issues:
   - Geographic location where npm was run:
   - [ ] I use a proxy to connect to the npm registry.
   - [ ] I use a proxy to connect to the web.
   - [ ] I use a proxy when downloading Git repos.
   - [ ] I access the npm registry via a VPN
   - [ ] I don't use a proxy, but have limited or unreliable internet access.
 - Container:
   - [ ] I develop using Vagrant on Windows.
   - [ ] I develop using Vagrant on OS X or Linux.
   - [ ] I develop / deploy using Docker.
   - [ ] I deploy to a PaaS (Triton, Heroku).

shawnmcknight:
I've been encountering this issue for a while now in my CI containers.  To get around it, I haven't been executing `npm install` as root, instead executing `su node -c "npm install"` where _node_ is a non-root user that is created by the node:8.1.2 image.

[...]

PR #20201, opened 2018-03-29
Fix root user node-gyp permission issues

lucyllewy:
Per issue #17346, when running npm in docker builds some packages are failing to run installation scripts when installing globally due to the global node_modules directory being owned by `root` but npm is running the scripts as `nobody`. This commit attempts to massage the default user configuration to allow executing as root.

Please discuss whether I have approached this issue correctly, or if more thought needs to be applied....

### DIFF
--- a/lib/config/defaults.js
+++ b/lib/config/defaults.js
@@ -234,7 +234,7 @@ Object.defineProperty(exports, 'defaults', {get: function () {
                        process.getgid && process.setgid) ||
                      process.getuid() !== 0,
     usage: false,
-    user: process.platform === 'win32' ? 0 : 'nobody',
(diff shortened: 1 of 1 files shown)


iarna (commented):
The main issue with this is that breaking is "working as intended" behavior around root installs. I'm not terribly keen on that decision however and have tentatively been planning patching it out more comprehensively w/ npm@7 (scheduled for later this Summer).

That would involve making npm under sudo be the same as npm as root and the removal of the whole "run scripts as `nobody`" thing, which really solves no problems.

iarna:
I have an RFC written up on changing npm's behavior when running as root: https://github.com/npm/rfcs/pull/5

zkat:

Issue #17322, opened 2017-06-19
npm unconditionally creates $prefix/etc (reopen #11486)


Zombaya:
#### I'm opening this issue because:

  - [ ] npm is crashing.
  - [x] npm is producing an incorrect install.
  - [ ] npm is doing something I don't understand.
  - [ ] Other (_see below for feature requests_):
  - [x] the previous bugrepport was automatically closed: #11486 

#### What's going wrong?

If you use a prefix to install, npm automatically creates a subdirectory `etc` in the prefixed folder.

#### How can the CLI team reproduce the problem?

```
cd /tmp/
mkdir reproduce
cd reproduce
echo "{}" > package.json
npm install --save react
cd /tmp
npm install --prefix reproduce 
ls -l reproduce
```

### supporting information:

 - `npm -v` prints: 5.0.3
 - `node -v` prints: v7.10.0
 - `npm config get registry` prints:https://registry.npmjs.org/
 - Windows, OS X/macOS, or Linux?: Linux
 - Network issues:
   - Geographic location where npm was run:
   - [ ] I use a proxy to connect to the npm registry.
   - [ ] I use a proxy to connect to the web.
   - [ ] I use a proxy when downloading Git repos.
   - [ ] I access the npm registry via a VPN
   - [ ] I don't use a proxy, but have limited or unreliable internet access.
 - Container:
   - [ ] I develop using Vagrant on Windows.
   - [ ] I develop using Vagrant on OS X or Linux.
   - [ ] I develop / deploy using Docker.
   - [ ] I deploy to a PaaS (Triton, Heroku).

### Related issues
* #11802 npm start --prefix ./folder-that-does-not-exist creates the folder and a subfolder called etc
* #11486 npm unconditionally creates $prefix/etc

romainwurtz:
Still happening with v5.4.2

matthispaita:
I have also this issue. npm create "/etc" folder in the target directory. npm: 5.5.1, node: 8.2.1.

shine-on:
same here

$ node -v && npm -v
v6.11.4
3.10.10

peschee:
@iarna @zkat  What can we do here so that @npm-robot does not automatically close this issue (which is what happened in #11802 and #11486)?

iarna:
[...]

Issue #11802, opened 2016-03-07
npm start --prefix ./folder-that-does-not-exist creates the folder and a subfolder called etc


tkellen:
```
~/code/npm
❯ ls -al
total 0
drwxr-xr-x   2 tkellen  staff   68 Mar  7 10:15 .
drwxr-xr-x  15 tkellen  staff  510 Mar  7 10:13 ..

~/code/npm
❯ npm start --prefix ./folder-that-does-not-exist
npm ERR! Darwin 14.1.0
npm ERR! argv "/usr/local/bin/node" "/usr/local/bin/npm" "start" "--prefix" "./folder-that-does-not-exist"
npm ERR! node v5.7.0
npm ERR! npm  v3.6.0
npm ERR! path /Users/tkellen/code/npm/folder-that-does-not-exist/package.json
npm ERR! code ENOENT
npm ERR! errno -2
npm ERR! syscall open

npm ERR! enoent ENOENT: no such file or directory, open '/Users/tkellen/code/npm/folder-that-does-not-exist/package.json'
npm ERR! enoent ENOENT: no such file or directory, open '/Users/tkellen/code/npm/folder-that-does-not-exist/package.json'
npm ERR! enoent This is most likely not a problem with npm itself
npm ERR! enoent and is related to npm not being able to find a file.
npm ERR! enoent

npm ERR! Please include the following file with any support request:
npm ERR!     /Users/tkellen/code/npm/npm-debug.log

~/code/npm
❯ find .
.
./folder-that-does-not-exist
./folder-that-does-not-exist/etc
./npm-debug.log
```

jbergstroem:
I've also run into this. It even does so on a global level: #11486.

cowboy:
I also see this behavior if the folder exists but doesn't contain a `package.json` file:

```
$ mkdir foo

$ npm start --prefix ./foo
npm ERR! Darwin 14.5.0
npm ERR! argv "/Users/cowboy/.nave/installed/4.3.0/bin/node" "/Users/cowboy/.nave/installed/default/bin/npm" "start" "--prefix" "./foo"
npm ERR! node v4.3.0
npm ERR! npm  v3.7.3
npm ERR! path /Users/cowboy/tmp/npm-start-wtf/foo/package.json
npm ERR! code ENOENT
npm ERR! errno -2
npm ERR! syscall open

npm ERR! enoent ENOENT: no such file or directory, open '/Users/cowboy/tmp/npm-start-wtf/foo/package.json'
[...]

PR #20212, opened 2018-03-31
config: only create config dir before write

buddydvd:
This change fixes the bug that creates unnecessary `./etc` folder when calling `npm` with the `--prefix` flag.

See: https://github.com/npm/npm/pull/7249#issuecomment-96854630

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -9,6 +9,8 @@ var types = npmconf.defs.types
 var ini = require('ini')
 var editor = require('editor')
 var os = require('os')
+var path = require('path')
+var mkdirp = require('mkdirp')
 var umask = require('./utils/umask')
 var usage = require('./utils/usage')
 var output = require('./utils/output')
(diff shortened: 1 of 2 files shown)


buddydvd:
PR #7249 (meant to fix #7233) introduced a regression that causes creation of extraneous `etc` folders when calling `npm` with the `--prefix` flag. The effect of the regression is documented in #7249 (closed), #11486 (closed), #11802 (closed), #17322.

This PR fixes the issue by reverting #7249 (while retaining the associated tests) and implementing the suggested fix as described here: https://github.com/npm/npm/pull/7249#issuecomment-96854630.

The essence of the fix is to only create the possibly missing `etc` folder _just before_ `npm` is ready write to the global config file located at `$prefix/etc/npmrc`.

iarna (approved):

zkat:
This was already merged and is part of `npm@6`. Cheers!

PR #20256, opened 2018-04-04
hook: add command to manage npm hooks

zkat:
This is a port of [`wombat-cli`'s hook manager](https://github.com/npm/wombat-cli/blob/master/commands/hook.js) to npm itself, using [`npm-registry-fetch`](https://npm.im/npm-registry-fetch). It's the first official consumer of that library, so we're using this command as a proving ground for the thing that's meant to replace `npm-registry-client`.


iarna (approved):

zkat:
This was already merged and included as part of `npm@6`. Cheers!

PR #20270, opened 2018-04-06
start replacing remaining npm-registry-client uses

zkat:
This is the beginning of the work needed to finally obsolete `npm-registry-client` in favor of `npm-registry-fetch`. I'm starting with more trivial commands, making sure they work, and then working my way up to more involved commands.


zkat:
Moved to https://github.com/npm/cli/pull/2

PR #20276, opened 2018-04-06
node-gyp@3.6.2

zkat:
Fixes: #20163

This PR also adds a test to make doubleplus sure we don't accidentally move around node-g again _and_ that the version that `npm-lifecycle` uses matches the one we use in the toplevel, so we don't get any accidental inconsistencies even if they happen to end up duplicated (it doesn't need to be a singleton)


iarna (approved):

MylesBorins:
@zkat @iarna is there a timeline on this landing? I'd like to float this in https://github.com/nodejs/node/pull/19840 and try and get it pulled into the next 8.x RC which is currently being prepped.

zkat:
@MylesBorins Rebecca's working on the release right now. So, "soon".

zkat:
This was already merged, so closing it.

PR #20303, opened 2018-04-11
feat: add npm init <create-pkg-name> command

jdalton:
The `npm init <create-pkg-name>` command will

 * detect the argument after `npm init`
 * append the appropriate prefix to it
   _(e.g. `npm init esm` [will translate](https://github.com/yarnpkg/yarn/blob/28c89216e63ca267e2b5fdcf3662860ca8b81282/src/cli/commands/create.js#L26-L27) the `esm` arg to `create-esm`)_
 * invoke the equiv of `npx create-esm` via [`libnpx`](https://www.npmjs.com/package/libnpx)

Related to https://github.com/zkat/npx/issues/167.

**Update**:

Okay! 

Basic functionality exists in the PR now _(thanks to @zkat for getting me unstuck)_.

At the moment I defer to `npx` for parsing args and handling installs, which means it follows its rules on fresh installing packages and all that. `npx` does all the heavy lifting here so this tie-in is pretty minimal.

I based `npx` invocation on `bin/npx-cli.js`. The first arg beyond `npm init __` is used as the initer to resolve. Additional arguments are forwarded to the initer.

I used `args` passed to `init()` as the initial feature gate because it was provided info. The `args` array lacks extra flags though. So, for say `npm init esm -y` the `args` array to `init()` is `['esm']`. This is why `process.argv` is used to forward additional args.

Erroring is handled by `npx`. So for invalid packages the result looks like
```
npm init b -y
npm ERR! code E404
npm ERR! 404 Not Found: create-b@latest

npm ERR!     /Users/jdalton/.npm/_logs/2018-04-11T23_57_32_908Z-debug.log
Install for create-b@latest failed with code 1
```

**Update:**

Added `create` as an alias of `init` and added `innit` as an affordance of `init`.


lib/init.js
```diff
@@ -2,15 +2,30 @@
...
+    var packageName = initerName.replace(/^(@[^/]+\/)?/, '$1create-')
+
+    var npxArgs = [process.argv0, 'npm-init', packageName, ...process.argv.slice(4)]
+    var parsed = npx.parseArgs(npxArgs, NPM_PATH)
+
+    return npx(parsed)
+      .then(cb)
+      .catch(cb)
+  }
+
```
jdalton: The promise returned by `npx` is a vanilla promise _(not a bluebird one)_ so lacks the `nodeify` helper.

The promise `then(cb)` of `npx` is executed before the `npm init` command finishes which causes the process to exit. There is a PR open on `npx` (https://github.com/zkat/npx/pull/172) to return a promise so the `cb` will be executed after `npx` runs its command.

lib/init.js
```diff
@@ -2,15 +2,30 @@
...
 var noProgressTillDone = require('./utils/no-progress-while-running').tillDone
 
 init.usage = 'npm init [--force|-f|--yes|-y]'
 
 function init (args, cb) {
+  if (args.length) {
+    var NPM_PATH = path.resolve(__dirname, '../bin/npm-cli.js')
+    var initerName = args[0]
+    var packageName = initerName.replace(/^(@[^/]+\/)?/, '$1create-')
+
```
jdalton: ☝️ That regexp is the key yarn/npm interoperable bit.

lib/init.js
jdalton: The promise `then(cb)` is executed before the `npm init` command finishes which causes the process to exit.
There is a PR open on `npx` to return a promise to execute the `cb` after `npx` runs its command.
https://github.com/zkat/npx/pull/172

jdalton:
Any guidance on unit tests would be appreciated.

lib/init.js
```diff
@@ -2,15 +2,31 @@
...
 init.usage = 'npm init [--force|-f|--yes|-y]'
 
 function init (args, cb) {
+  if (args.length) {
+    var NPM_PATH = path.resolve(__dirname, '../bin/npm-cli.js')
+    var initerName = args[0]
+    var packageName = /^@[^/]+$/.test(initerName)
+      ? initerName + '/create'
+      : initerName.replace(/^(@[^/]+\/)?/, '$1create-')
+
```
jdalton, line 28: 👆 In addition to following yarn's rules, 
name expansion also supports transforming `@scope` into `@scope/create`.

zkat:
@jdalton as far as unit tests go, I would mock `bin/npx-cli.js` and just make sure it's being called with the right arguments.

jdalton:
@zkat
> as far as unit tests go, I would mock `bin/npx-cli.js` and just make sure it's being called with the right arguments.

Cool!

jdalton:
~~Note: This PR [depends](https://github.com/npm/npm/pull/20303#discussion_r180927254) on https://github.com/zkat/npx/pull/172 _(however it is resolved)_.~~

The `npx` issue is resolved 😋

jdalton:
Added tests!

lib/init.js
```diff
@@ -2,15 +2,31 @@
...
 
+var path = require('path')
 var log = require('npmlog')
 var npm = require('./npm.js')
+var npx = require('libnpx')
 var initJson = require('init-package-json')
 var output = require('./utils/output.js')
 var noProgressTillDone = require('./utils/no-progress-while-running').tillDone
 
 init.usage = 'npm init [--force|-f|--yes|-y]'
```
zkat: This should be updated. You should document this feature in `doc/cli/npm-init.md` as well.

package.json
```diff
@@ -74,7 +74,7 @@
     "lazy-property": "~1.0.0",
     "libcipm": "^1.6.2",
     "libnpmhook": "^4.0.1",
-    "libnpx": "^10.1.1",
+    "libnpx": "^10.2.0",
```
zkat: This can be removed -- I've updated libnpx

zkat (changes requested):

jdalton:
Added docs and removed the `libnpx` bump in the package.json.

zkat (approved):

jdalton:
Thank you 🎉

lib/config/cmd-list.js
```diff
@@ -4,6 +4,7 @@ var shorthands = {
   'rb': 'rebuild',
   'list': 'ls',
   'ln': 'link',
+  'create': 'init',
```
statianzo, line 7: Thanks for this alias 👍

jdalton: Thanks goes to @zkat :clap:

PR #20322, opened 2018-04-12
install: add support for package aliases

zkat:
note: This requires yet-to-land changes in `pacote` in order to work. This is also a work in progress and barely even works right now, so.


zkat:
Moved to https://github.com/npm/cli/pull/3

PR #20331, opened 2018-04-13
Fix broken link to 'private-modules' page.

vipranarayan14:
Fix the broken link to 'private-modules' page by changing the link from 'https://www.npmjs.com/private-modules' to 'https://docs.npmjs.com/private-modules/intro'.


zkat (approved):
👍

PR #20372, opened 2018-04-18
test: ensure npm init forwards arguments

jdalton:
A follow-up PR to #20303 to add a test to ensure `npm init` forwards arguments.


zkat (approved):
tesssstttttsssssss 🤩

PR #20384, opened 2018-04-20
pkglock: add from field back into git dependencies

zkat:


iarna (approved):
Looks good, thank you for putting this together!

PR #20390, opened 2018-04-20
install/deps: Let git deps w/ lock only match package.json

iarna:
Previously they would only match if the complete commitid was used.  This allows partial matching when no on-disk component is available.

This will need a minor patch to coordinate with https://github.com/npm/npm/pull/20384


iarna:
This landed starting at https://github.com/npm/npm/commit/e4ed976e20b7d1114c920a9dc9faf351f89a31c9 in 6.0.0

PR #20398, opened 2018-04-20
makefile: call cache clean with --force

kapouer:


iarna (approved):
Ah yeah given the new `cache clean` behavior this looks good.

iarna:
Merged as be51b382b. Thank you for this!

Issue #20399, opened 2018-04-21
[BETA issue] npm init is not working as expected with github dependencies


mansona:
I don't know if this is the right channel to give some BETA channel feedback, please let me know if there is a better place to put these sorts of comments.

I just tested something with the new `npm init` command that I expected to work based on previous conventions but maybe it's not on the spec for this new feature.

I'm trying to get the following command to work: 

```
npx npm@v6.0.0-next.2 init ember-app my-app
```

(real talk for a second: this is amazing that I can do this!! npx is so cool!!)

as it turns out there is some specific behaviour in `create-ember-app` that expects yarn to be installed. As this is open source we're talking about I forked it and removed the yarn specific command and then wanted to run this to test the new fork: 

```
npx npm@v6.0.0-next.2 init stonecircle/create-ember-app my-app
```

It seems to be *trying* to do the right thing but misses the mark by a tiny amount 😂 here is the output: 

```
npx: installed 680 in 10.394s
npm ERR! Error while executing:
npm ERR! /usr/bin/git ls-remote -h -t ssh://git@github.com/create-stonecircle/create-ember-app.git
npm ERR!
npm ERR! ERROR: Repository not found.
npm ERR! fatal: Could not read from remote repository.
npm ERR!
npm ERR! Please make sure you have the correct access rights
npm ERR! and the repository exists.
npm ERR!
npm ERR! exited with error code: 128

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/mansona/.npm/_logs/2018-04-21T07_58_49_792Z-debug.log
Install for github:create-stonecircle/create-ember-app failed with code 1
```

I then tried `npx npm@v6.0.0-next.2 init github:stonecircle/create-ember-app my-app` to see if the `github:` previx made a difference but then got this output: 

```
➜  what npx npm@v6.0.0-next.2 init github:stonecircle/create-ember-app my-app
npx: installed 680 in 13.275s
ENOENT: no such file or directory, stat 'create-github:stonecircle/create-ember-app'
```

Is this something that npm is likely wanting to support? I am reluctant to create a throw-away npm package just to test if my patch fixes the issue upstream

### supporting information:

 - `npm -v` prints: N/A using `npx npm@v6.0.0-next.2`
 - `node -v` prints: v8.11.1
 - `npm config get registry` prints: https://registry.npmjs.org/
 - Windows, OS X/macOS, or Linux?: macOS

#### Is this a feature you're prepared to implement, with support from the npm CLI team?

[...]

PR #20403, opened 2018-04-22
feat: add npm init support of git, path, url, and version names

jdalton:
This PR addresses #20399.

### DIFF
--- a/lib/init.js
+++ b/lib/init.js
@@ -4,9 +4,11 @@ module.exports = init
 
 var path = require('path')
 var log = require('npmlog')
+var npa = require('npm-package-arg')
 var npm = require('./npm.js')
 var npx = require('libnpx')
 var initJson = require('init-package-json')
+var isRegistry = require('./utils/is-registry.js')
 var output = require('./utils/output.js')
 var noProgressTillDone = require('./utils/no-progress-while-running').tillDone
 var usage = require('./utils/usage')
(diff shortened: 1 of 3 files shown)


lib/init.js
```diff
@@ -15,17 +17,26 @@ init.usage = usage(
   'init',
   '\nnpm init [--force|-f|--yes|-y|--scope]' +
   '\nnpm init <@scope> (same as `npx <@scope>/create`)' +
-  '\nnpm init [<@scope>/]<pkg> (same as `npx [<@scope>/]create-<pkg>`)'
+  '\nnpm init [<@scope>/]<name> (same as `npx [<@scope>/]create-<name>`)' +
+  '\nnpm init <pkg> (same as `npx <pkg>`)'
```
zkat: This line is kinda confusing, tbh. I'm not really sure at first glance how it might be different than the previous one (since it's not clear this is about non-registry packages).

lib/init.js
```diff
@@ -15,17 +17,26 @@ init.usage = usage(
...
-    var packageName = /^@[^/]+$/.test(initerName)
-      ? initerName + '/create'
-      : initerName.replace(/^(@[^/]+\/)?/, '$1create-')
-
+    var packageName = initerName
+    if (/^@[^/]+$/.test(initerName)) {
+      packageName = initerName + '/create'
+    } else {
+      var req = npa(initerName)
+      if (isRegistry(req)) {
```
zkat: I wonder if we could do more here. Should we also, maybe, handle hosted git deps? `npm init zkat/templ` -> `npx github:zkat/create-templ`? What's the expectation around this? What makes these others any better than doing `npx file:some/initializer`?

zkat (commented):
Thanks for taking the time to keep improving this feature, @jdalton! I've made a first pass at this PR.

lib/init.js
jdalton: I think I saw `<pkg>` used in another place to mean `<folder|tarball|git url|user/repo>`. Maybe a way to say everything else is just forwarded to `npx`.

lib/init.js
jdalton: > What makes these others any better than doing npx file:some/initializer?

At the moment they aren't any better than using `npx`. We could keep it simple. When a user provides a non registry name to `npm init` it could just recommend using `npx` instead. Something like
```shell
$ npm init file:some/initializer
> Unrecognized initializer: file:some/initializer.
> For more package binary executing power check out `npx`:
> https://www.npmjs.com/package/npx

Comment (jdalton) [reply lib/init.js]: Ok, I've updated the PR to more gracefully punt on git, path, and url names for now instead of just erring out.

File: lib/init.js [hunk 3]
```diff
@@ -15,17 +17,32 @@ init.usage = usage(
...
+        packageName = req.name.replace(/^(@[^/]+\/)?/, '$1create-')
+        if (req.rawSpec) {
+          packageName += '@' + req.rawSpec
+        }
+      } else {
+        log.warn(
```
zkat: This should probably be an actual error. Maybe with `err.code === 'EUNSUPPORTED'` or something? 🤔

jdalton: Ah ok!

jdalton:
Made the `warn` and error and added support for expanding git names.

zkat (approved):
I'm v happy with this. Thanks, @jdalton!

jdalton:
Thank you!

PR #20407, opened 2018-04-23
docs: add from field back into git dependencies

watilde:
An update of the spec doc follows a change that the "from" field was returned.

Refs: https://github.com/npm/npm/pull/20384


zkat (approved):

iarna:
Merged as b7fca1084, thank you, this was a good catch! =)

Issue #16224, opened 2017-04-03
npm outdated: clarify package name colors in output


tunderdomb:
#### I'm opening this issue because:

  - [ ] npm is crashing.
  - [ ] npm is producing an incorrect install.
  - [x] npm is doing something I don't understand.
  - [ ] Other (_see below for feature requests_):

#### What's going wrong?

Running `npm outdated` in a color enabled terminal prints you a nice color coded table of your outdated packages. However it's not clear what red and yellow package names mean. 
Reading the [docs](https://docs.npmjs.com/cli/outdated) doesn't reveal the difference either. The only place I found mention of this is #4176.

My first impression was this: 

 - red: the package is really-really outdated, and updating it would probably bring in a breaking change
 - yellow: the package is outdated, but updating is fine

Actually it means the opposite.

Highlighting this in the docs would help, would you accept a PR with some text about the output colors?

#### How can the CLI team reproduce the problem?

run `npm outdated` in a package where dependencies are out of date.

batjko:
There's also a related SO question which is only taking wild guesses so far:
http://stackoverflow.com/questions/31559390/npm-outdated-output-colour-coding-meaning

tdd:
It seems to me that red is 'there's a newer version matching your semver requirements, so you should update now' while yellow is 'there's a newer version above your semver requirements (usually new major, or new 0.x minor) so proceed with caution'.

oshalygin:
This is confirmed, thanks @tdd 

We should document this somewhere.

teameh:
I had to find this issue to get info on the coloring, not ideal. 

Added a PR to make your life easy and to fix this. Please feel free to edit / amend the commit and to change it to whatever you like.

PR #20408, opened 2018-04-23
Fix 16224 - docs/cli/outdated - added colors explanation

teameh:
Fix for https://github.com/npm/npm/issues/16224


iarna:
This was merged into `release-next` as 7a6555e61, thank you!

PR #20441, opened 2018-04-25
Remove "request" dependency and add "node-fetch-npm"

southpolesteve:
On Twitter, a discussion started around the increasing size of npm as it is bundled with node.js. It appears that `request` is one of the biggest contributors as it bundles a lot of functionality (like signing AWS requests) that npm does not use. This PR removes the top level dependency on `request`. It still may exist in sub-dependencies, but this is a first step toward a decreased size

Questions:
- Should I be committing package-lock.json changes?
- Should I be committing `node-fetch-npm` as a bundled dependency?


zkat:
We already have a PR going for the process of extracting `request` properly from our dependencies, which involves removing npm-registry-client. See #20270. We'll also be doing other things to reduce the bundle size eventually, but it's fairly low priority. A huge reason for this also has to do with legacy upgrade concerns, because npm itself uses an installation style that reduces deduplication across the tree in some cases.

Since the goal of this PR actually involves a much bigger process (and switching to a different client), I'm going to close it. Thanks for taking the time to put this together, since it's definitely trying to fix an existing pain point. 💚

PR #20487, opened 2018-05-02
cross-platform shell for npm run scripts

gucong3000:
Closes: #18580

### DIFF
--- a/node_modules/npm-lifecycle/index.js
+++ b/node_modules/npm-lifecycle/index.js
@@ -262,7 +262,7 @@ function runCmd_ (cmd, pkg, env, wd, opts, stage, unsafe, uid, gid, cb_) {
 
   if (customShell) {
     sh = customShell
-  } else if (process.platform === 'win32') {
+  } else if (opts.isWindowsShell == null ? process.platform === 'win32' : opts.isWindowsShell) {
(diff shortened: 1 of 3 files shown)


node_modules/npm-lifecycle/index.js
```diff
@@ -262,7 +262,7 @@ function runCmd_ (cmd, pkg, env, wd, opts, stage, unsafe, uid, gid, cb_) {
 
   if (customShell) {
     sh = customShell
-  } else if (process.platform === 'win32') {
```
iarna, line 265: This file is in a dependency. To patch this you need to PR this change to [npm-lifecycle](https://github.com/npm/npm-lifecycle).

I think this is the right approach though.  When you PR this to `npm-lifecycle` make sure you include a default value for the `isWindowsShell` option that's compatible.

iarna (changes requested):

node_modules/npm-lifecycle/index.js
gucong3000: done.

gucong3000:
@iarna Is there anything else I need to do?

zkat:

zkat:
p.s. You need to update `npm-lifecycle` in its own repo and not include the patch directly like this.

PR #20495, opened 2018-05-02
build: bundle the npm CLI itself for release

zkat:
THIS IS A BIG DIFF for dependency reasons. You only really need to look at [`scripts/webpack-bundle.js`](https://github.com/npm/npm/blob/zkat/bundled-cli/scripts/webpack-bundle.js) and changes inside `lib/`, pretty much.

This makes it so we can bundle the npm CLI into a handful of files, instead of hundreds of thousands, with some advantages:

* Smaller distribution and installed size
* Faster use through `npx`
* We'll be able to use scoped dependencies again (currently only needed for legacy upgrades)
* We can fully flatten the npm dependency tree (currently needed because of a Node.js installer bug)

So far, I've tried `browserify`, `rollup`, and even `parcel`, and `webpack` seems to be the closest thing to working, but it's not quite there yet.

Still TODO:

* [ ] There are a number of dynamic requires throughout npm and its dependencies. I'm not sure how to resolve this.
* [ ] On that topic, `update-notifier` uses `require-lazy` for its dependencies, and that doesn't get picked up by webpack _at all_.
* [ ] There are _three_ workers that get loaded through `child_process.spawn/exec`. Those need to be standalone entry points. The workers are the extraction worker and the metrics worker on the npm installer, and the extraction worker in `libcipm` (a dependency).
* [ ] It's unclear whether this will seriously impact startup performance. Big slowdowns are no good.


DrewML:
> There are a number of dynamic requires throughout npm and its dependencies. I'm not sure how to resolve this.

Speaking specifically about the dynamic requires in `lib/npm.js`, I _think_ you'd be able to resolve that with the [ContextReplacementPlugin](https://webpack.js.org/plugins/context-replacement-plugin/). I suspect that webpack's context feature isn't making any assumptions about what `__dirname` means in that case, but it should be able to just collect all files matching `*.js` under `__dirname` into the bundle.

evenstensberg:
@zkat Does this help?
https://webpack.js.org/plugins/split-chunks-plugin/#optimization-splitchunks
https://gist.github.com/sokra/1522d586b8e5c0f5072d7565c2bee693
https://medium.com/webpack/webpack-4-code-splitting-chunk-graph-and-the-splitchunks-optimization-be739a861366

I think you should set the upper `chunk` to async. If I've gotten the idea right, you want to have a commons chunk where all the same modules are gathered, and use that bundle? If so, you need to set those chunks up manually

evenstensberg:
Meaning you should split per entry point, assign those chunks to unique names and set `reuseChunks: true` ( https://github.com/webpack/webpack-cli/issues/404 ) Also, if you do `webpack-cli init` ( version 2.0.14/15, you should have a nice starter config 

https://github.com/rollup/rollup-starter-code-splitting might also be useful

evenstensberg:
Runtimechunks would be good for npm, but it creates some thrash folders, don't know if you want that

sokra:
Replace `require(path.join(__dirname, a + '.js'))` with `require('./' + a + '.js')` and optionally use the ContextReplacementPlugin as @DrewML proposed.

The `require(onload)` is an issue. You could replace it with `__non_webpack_require__(onload)` but that would not allow running it without webpack.

We could add `require(/* webpackIgnore: true */ onload)` to webpack like with `import()`.

sokra:
`optimization.splitChunks.chunks: "all"` is not supported for `target: "node"` and doesn't make sense.

`entry` also takes relative requests relative to `context`

zkat:
This was moved to https://github.com/npm/cli/pull/4

PR #20504, opened 2018-05-03
feat(webpack): this adds some ezpz externals and whitelisting support for webpacking npmcli

TheLarkInn:
This commit does the following: 

* Ignores _truly dynamic requires_ by wrapping them in `new Function`
* Expands the reach of `shebang-loader` (therefore any js file with the env/shebang can be consumed)
* adds webpack-node-externals and adds the "bundledDependencies" list to the whitelist with a few hacked on dependencies missing (yargs, update-notifier, etc.)
* makes a local "dynamic" require a context module

### DIFF
--- a/package.json
+++ b/package.json
@@ -279,6 +279,7 @@
     "webpack": "^4.6.0"
   },
   "scripts": {
+    "bundle": "node scripts/webpack-bundle.js",
     "dumpconf": "env | grep npm | sort | uniq",
     "prepare": "node bin/npm-cli.js --no-audit --no-timing prune --prefix=. --no-global && rimraf test/*/*/node_modules && make -j4 doc",
     "preversion": "bash scripts/update-authors.sh && git add AUTHORS && git commit -m \"update AUTHORS\" || true",


scripts/webpack-bundle.js
```diff
@@ -40,6 +42,128 @@ function bundler (opts) {
...
+          "which",
+          "wrappy",
+          "write-file-atomic",
+          "safe-buffer",
+          "tiny-relative-date",
+          "cli-columns",
+          "node-gyp"
+        ]
+      })
+    ],
```
zkat: Can you make this `require('../package.json').bundleDependencies.filter(...)` so we know which specific packages are being blacklisted?

lib/npm.js
```diff
@@ -124,7 +124,7 @@
       npm.command = c
       if (commandCache[a]) return commandCache[a]
 
-      var cmd = require(path.join(__dirname, a + '.js'))
+      var cmd = require("./" + path.join(__dirname, a) + ".js");
```
zkat: This needs to be `''` and drop the `;`, because `standard`.

scripts/webpack-bundle.js
```diff
@@ -19,7 +21,7 @@ if (require.main === module) {
 }
 
 module.exports = bundler
-function bundler (opts) {
+function bundler(opts) {
```
zkat: Put this space back, because `standard`.

scripts/webpack-bundle.js
```diff
@@ -31,7 +33,7 @@ function bundler (opts) {
     node: false,
     module: {
       rules: [
-        {test: /np.-cli.js$/, loader: 'shebang-loader'}
+        { test: /.js$/, loader: 'shebang-loader' }
```
zkat: I guess this is probably reasonable tbh.

scripts/webpack-bundle.js
zkat: Also, does this track transitive dependencies?

zkat (changes requested):

lib/npm.js
zkat: Oh and this needs to be `'./' + a + '.js'`. Drop the __dirname.

TheLarkInn: perfect!! Will change.

scripts/webpack-bundle.js
TheLarkInn: Oh yeah thats a great idea! I'll do so.

zkat:
wait what why did it close D:

zkat:
I guess it got merged. Heh.

PR #20513, opened 2018-05-04
Fix nested aliases, restore alias 'sit': 'cit'

mmermerkaya:
I investigated why `npm sit` was failing and fixed nested aliases not being dereferenced.

I see that the issue was already fixed in https://github.com/npm/npm/commit/6644cdb1fef9941b05f1415ae7fe86429fecefa8 but I'm opening this PR to prevent similar issues in the future, and to restore the better looking `'sit': 'cit'` :)

### DIFF
--- a/lib/config/cmd-list.js
+++ b/lib/config/cmd-list.js
@@ -46,7 +46,7 @@ var affordances = {
   'rm': 'uninstall',
   'r': 'uninstall',
   'rum': 'run-script',
-  'sit': 'install-ci-test',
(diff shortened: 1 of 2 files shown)


iarna:
I'm closing this for a moment, but I'll reopen it. I'm just working around a github bug.

zkat (approved):
Only thing I can think of this doing is creating cycles but, tbh, if we introduce one it'll be very obvious, very quickly. LGTM

PR #20536, opened 2018-05-07
update-notifier: remove colon + support canary

zkat:
Noticed a couple of things after getting an update notification for `canary` that looked like this:

screen shot 2018-05-07 at 12 04 54

Things should look better now.


iarna (approved):

iarna:
LGTM!

PR #20568, opened 2018-05-10
audit: support for --json output

### LINKED ISSUE #20587: make npm audit parseable for CI/CD
#### What's the feature?

Create a flag `npm audit --ci` or `npm audit --output-json`, which will output a machine parseable string of the information, e.g. in JSON.

The general availability of `npm audit` in npm 6 is awesome and will help make applications build with node a lot more secure. However, the current output of npm is not really machine parseable and hinders the usage in CICD pipelines.

Yes. The summary is currently saved as an artifact in our CICD-pipeline, but we are reluctant to parse the human readable output. We would like to make sure, that no high-priority security problems are present and would like the build to break in that case. 

A possible solution with the JSON output would allow us to use tools like `jq` to effectively query and act upon in scripts and would allow users very fine-grained options.

An easier solution (but not as powerful) would be a switch like `npm audit --error-severity high`, where npm would exit with a POSIX error code of not 0, if a security problem of severity greater or equal than [low|medium|high] was found.

Yes - other security scanners provide machine-readable output options.

We could help with that.

finnp:
Makes use of the `json` formatter of `npm-audit-report` when using the `--json` option.

resolves #20587

### DIFF
--- a/lib/install/audit.js
+++ b/lib/install/audit.js
@@ -105,7 +105,7 @@ function printInstallReport (auditResult) {
 function printFullReport (auditResult) {
   return auditReport(auditResult, {
     log: output,
-    reporter: 'detail',
+    reporter: npm.config.get('json') ? 'json' : 'detail',
     withColor: npm.color,
     withUnicode: npm.config.get('unicode')
   }).then(result => output(result.report))


naugtur:
the failing check is on node6, npm audit with details reporter fails on node6 because it uses Object.entries

zkat:
just rebase onto `latest`. It's already been fixed.

finnp:
I was up-to-date with `latest`. But I rebased onto `release-next`. That's the right branch, right?

zkat:
that should be fine. And it looks like you're green now, so 👍

zkat (approved):
looks great!

PR #20569, opened 2018-05-10
audit: add subcommand to automatically fix vulns

zkat:
Example, combined with https://github.com/npm/npm-audit-report/pull/12 to give a better idea of what it'll look like with both changes (that PR is aesthetic).

screen shot 2018-05-09 at 17 30 41

### DIFF
--- a/lib/audit.js
+++ b/lib/audit.js
@@ -1,18 +1,24 @@
 'use strict'
-const fs = require('graceful-fs')
+
 const Bluebird = require('bluebird')
+
 const audit = require('./install/audit.js')
-const npm = require('./npm.js')
-const log = require('npmlog')
-const parseJson = require('json-parse-better-errors')
+const fs = require('graceful-fs')
+const Installer = require('./install.js').Installer
 const lockVerify = require('lock-verify')
+const log = require('npmlog')
+const npa = require('npm-package-arg')
+const npm = require('./npm.js')
+const output = require('./utils/output.js')
+const parseJson = require('json-parse-better-errors')
 
 const readFile = Bluebird.promisify(fs.readFile)
 
 module.exports = auditCmd
 
 auditCmd.usage =
-  'npm audit\n'
+  'npm audit\n' +
+  'npm audit fix\n'
 
 auditCmd.completion = function (opts, cb) {
   const argv = opts.conf.argv.remain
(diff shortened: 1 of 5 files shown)


lib/audit.js
```diff
@@ -1,18 +1,24 @@
 'use strict'
-const fs = require('graceful-fs')
-const Bluebird = require('bluebird')
+
+const BB = require('bluebird')
```
iarna: Changing this is ... very weird. Don't change it back now, but next time consider… not changing it? I don't go around changing `BB` to `Bluebird whenever I see it…

lib/config/cmd-list.js
```diff
@@ -47,7 +47,8 @@ var affordances = {
   'r': 'uninstall',
   'rum': 'run-script',
   'sit': 'cit',
-  'urn': 'run-script'
+  'urn': 'run-script',
+  'audi': 'audit'
```
iarna: This alias is unnecessary as automatic command shortening would add it regardless.

lib/install/save.js
```diff
@@ -48,7 +47,7 @@ function saveShrinkwrap (tree, next) {
   if (!npm.config.get('shrinkwrap') || !npm.config.get('package-lock')) {
     return next()
   }
-  createShrinkwrap(tree, {silent: false}, next)
+  require('../shrinkwrap.js').createShrinkwrap(tree, {silent: false}, next)
```
iarna, line 50: Why is this requiring this inline?

lib/audit.js
```diff
@@ -25,6 +31,63 @@ auditCmd.completion = function (opts, cb) {
   }
 }
 
+class Auditor extends Installer {
+  constructor (where, dryrun, args, opts) {
+    super(where, dryrun, args, opts)
+    this._deepArgs = (opts && opts.deepArgs) || []
```
iarna: If this ability is folded into the main installer it shouldn't be underscore prefixed. (Doesn't matter much in a subclass, but I haven't been using that convention elsewhere.)

lib/audit.js
```diff
@@ -25,6 +31,63 @@ auditCmd.completion = function (opts, cb) {
...
+    this._runId = opts.runId || ''
+    this.audit = false
+  }
+
+  loadAllDepsIntoIdealTree (cb) {
+    BB.fromNode(cb => super.loadAllDepsIntoIdealTree(cb)).then(() => {
+      if (this._deepArgs && this._deepArgs.length) {
+        this._deepArgs.forEach(arg => {
+          arg.reduce((acc, child, i) => {
```
iarna: Point of style, so please feel free to disregard: Personally I avoid single letter var names if they aren't used on the same line or next line, because they make search much harder. I would use `ii` here.

lib/audit.js
```diff
@@ -25,6 +31,63 @@ auditCmd.completion = function (opts, cb) {
...
+            if (!acc) {
+              // We might not always be able to find `target` through the given
+              // path. If we can't we'll just ignore it.
+              return
+            }
+            const spec = npa(child)
+            const target = (
+              acc.requires.find(n => n.package.name === spec.name) ||
+              acc.requires.find(
+                n => audit.scrub(n.package.name, this._runId) === spec.name
```
iarna: Long term we'll be undoing scrubbing in `submitForFullReport`, but this is fine till then, and conveniently compatible.

iarna (approved):
LGTM

Some style notes that I leave to your judgement.

And a question about the bare require in `lib/install/save.js`

lib/audit.js
zkat: Sorry! I was grabbing code from elsewhere and writing a bunch of it, so I just changed it to this out of habit. I didn't mean for it to come off as some pass-agg thing 😅 . Noted for the future -- I'll be more careful <3

lib/install/save.js
zkat: This is how I resolved that awful circular require thing I think I mentioned to you a while back :\

Since it's really only ever called once, I didn't bother with the usual conditional require+caching thing.

lib/audit.js
zkat: I agree!

lib/audit.js
zkat: noted :)

lib/audit.js
zkat: that'll be nice!

PR #20570, opened 2018-05-10
install: include number of audited packages in output

zkat:
Expects https://github.com/npm/npm-audit-report/pull/12 for the full effect but can land without it. This bit only makes the installer print out the number of audited packages.

screen shot 2018-05-09 at 15 23 19


legodude17:
Wouldn't it make more sense for `npm install` to only audit installed packages?

zkat:
no, because previously-installed packages may end up getting vulnerability notices long after having been installed. You'll want to find out when that happens.

legodude17:
Ah that makes sense.

iarna (approved):
:sheep:

PR #20594, opened 2018-05-11
audit: ignore devDependencies in production

### LINKED ISSUE #20564: Allow npm audit to ignore dev dependencies
#### What's the feature?
Add a flag to ignore dev dependencies when running `npm audit`.
Maybe it could ignore them by default and only check them with a flag.

Only receive a report back for true application dependencies.

No.

Snyk ignores dev dependencies by default: https://snyk.io/docs/snyk-for-nodejs#testing-node-js-projects

I'll assist as much as I can.

welwood08:
Fixes #20564

### DIFF
--- a/lib/audit.js
+++ b/lib/audit.js
@@ -67,10 +67,14 @@ function auditCmd (args, cb) {
     } else if (shrinkwrap && lockfile) {
       log.warn('audit', 'Both npm-shrinkwrap.json and package-lock.json exist, using npm-shrinkwrap.json.')
     }
+    const includeDev = npm.config.get('dev') ||
+          (!/^prod(uction)?$/.test(npm.config.get('only')) && !npm.config.get('production')) ||
+          /^dev(elopment)?$/.test(npm.config.get('only')) ||
+          /^dev(elopment)?$/.test(npm.config.get('also'))
     const requires = Object.assign(
       {},
       (pkgJson && pkgJson.dependencies) || {},
-      (pkgJson && pkgJson.devDependencies) || {}
+      (includeDev && pkgJson && pkgJson.devDependencies) || {}
     )
     return lockVerify(npm.prefix).then(result => {
       if (result.status) return audit.generate(sw, requires)


nicolasnoble:
It'd be great to make sure that for consistency, this is also honoring the `NODE_ENV` environment variable, just like what `npm install` does. I'm not certain this PR does that at the moment.

Also it'd be great to amend the documentation to reflect this change.

welwood08:
I successfully tested `NODE_ENV=production npm audit`, so it does work somehow. I'm not sure what the valid values are for NODE_ENV or where it gets translated into the config setting, but `NODE_ENV=prod` does not work.

westy92:
This is great! Should there be unit tests?

zkat (changes requested):
I would much prefer if this filtering happened on the reporting side. This strategy won't really change what data we upload (because all the data will still be sent as part of the overall thing). Would you be willing to change this PR to do that instead?

welwood08:
@zkat The goal was never to change the uploaded data of course, but now you point it out it does make more sense to modify the output in the reporter. It would unfortunately become a more invasive change, but if that's the preferred direction I'll give it a shot.

For changes to `npm-audit-report`, I assume I should send a PR to the upstream repo rather than modify the in-tree copy?

zkat:
@welwood08 yup, that's correct. Please PR it over in `npm-audit-report` :)

zkat:
heads-up: I'm starting the release today, so if this isn't ready by EOD today, it won't go into this week's release (and it'll be two more weeks until the next npm release)

welwood08:
Apologies for the inactivity, I made a start the other week but was not satisfied with the code I came up with or the size of the diff so I discarded it. Since I'm not directly affected by this change, it hasn't been a priority to me. In the meantime I noticed too late there are already PRs for upstream `npm-audit-result` touching some of the code I will also need to touch, so I'd like to avoid further duplication of effort and wait until they merge.

There will still need to be changes in `npm` to pass appropriate configuration to the reporter, so if you don't mind leaving this PR open I'll force push new commits here when it's ready or I'll send a new PR if you'd rather close this.

zkat:

PR #20629, opened 2018-05-15
Make tone in "publishConfig" docs more neutral

jeremyckahn:
Hello!  This information was not immediately obvious to me, so I removed the superfluous "of course" bit of the sentence.

Thanks for the great documentation!


iarna (approved):
Thanks, this looks good!

PR #20715, opened 2018-05-22
ci: pause log before logging summary

legodude17:
From [here](https://npm.trydiscourse.com/t/npm-ci-progress-bar-not-properly-removed/61/2), this will remove the progress bar before logging the summary, so that it doesn't log the summary over the progress bar and print something like `added 2 pacakges in 0.966sractTree: extractTree`.

### DIFF
--- a/lib/ci.js
+++ b/lib/ci.js
@@ -29,9 +29,12 @@ function ci (args, cb) {
   })
     .run()
     .then(
-      (details) => console.error(`added ${details.pkgCount} packages in ${
-        details.runTime / 1000
-      }s`)
+      (details) => {
+        npmlog.pause()
+        console.error(`added ${details.pkgCount} packages in ${
+          details.runTime / 1000
+        }s`)
+      }
     )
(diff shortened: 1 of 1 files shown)


lib/ci.js
```diff
@@ -29,9 +29,12 @@ function ci (args, cb) {
   })
     .run()
     .then(
-      (details) => console.error(`added ${details.pkgCount} packages in ${
-        details.runTime / 1000
-      }s`)
+      (details) => {
+        npmlog.pause()
```
zkat: This should be `.disableProgress()`

zkat (changes requested):

lib/ci.js
legodude17: Ah thanks.

zkat (approved):
Looks good!

legodude17:
@zkat Fixed!

legodude17:
@zkat Great!

legodude17:
🎉

PR #20750, opened 2018-05-25
config: Enable config for suppressing update-notifier

travi:
based on conversation from https://npm.community/t/npm-side-config-for-suppressing-update-notifier/75

### DIFF
--- a/bin/npm-cli.js
+++ b/bin/npm-cli.js
@@ -74,7 +74,10 @@
   conf._exit = true
   npm.load(conf, function (er) {
     if (er) return errorHandler(er)
-    if (!unsupported.checkVersion(process.version).unsupported) {
+    if (
+      npm.config.get('update-notifier') &&
+      !unsupported.checkVersion(process.version).unsupported
+    ) {
       const pkg = require('../package.json')
       let notifier = require('update-notifier')({pkg})
       if (
--- a/doc/misc/npm-config.md
+++ b/doc/misc/npm-config.md
@@ -1130,6 +1130,14 @@ Set to true to suppress the UID/GID switching when running package
 scripts.  If set explicitly to false, then installing as a non-root user
 will fail.
 
+### update-notifier
+
+* Default: true
+* Type: Boolean
+
+Set to false to suppress the update notification when using an older
+version of npm than the latest.
+
 ### usage
 
 * Default: false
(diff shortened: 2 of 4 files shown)


travi:
i'm happy to make any adjustments necessary, but this should be close. thanks for being open to the contribution

zkat (approved):
This looks pretty much about right! Thank you!

bin/npm-cli.js
```diff
@@ -78,6 +78,7 @@
       const pkg = require('../package.json')
       let notifier = require('update-notifier')({pkg})
       if (
+        npm.config.get('update-notifier'),
```
zkat: Actually, I'd like this to be further up so you don't even require the notifier package at all?

travi:
looks like i missed a simple syntax issue that `npm t` would have pointed out to me had i run it before pushing :\

ill have a fix up in a min

bin/npm-cli.js
travi: sure, that makes sense. how about combined w/ the conditional on line 77?

travi:
i'm having some trouble getting the tests to run correctly locally, but the travis output gives me proper insight with a longer feedback loop. i'll wrap those up tomorrow.

travi:
ah, it looks like [a few tests](https://github.com/npm/npm/blob/4c65cd952bc8627811735bea76b9b110cc4fc80e/test/tap/config-meta.js#L111-L131) that ensure the new config var is used somewhere in the code, but it [doesnt seem to check files under `bin/`](https://github.com/npm/npm/blob/4c65cd952bc8627811735bea76b9b110cc4fc80e/test/tap/config-meta.js#L38-L52). do you want me to update the file search to include `bin/` in addition to `lib/`?

zkat:
Gonna cc @iarna on that q (I'm off work today)

travi:
sure thing. i'm happy to make the updates if thats best, but want to focus on what solution you see as most valuable. 

enjoy the day off @zkat

travi:
i went ahead and included `bin/` in the search for config usages and it does get the tests passing again. i made that change as a separate commit so i can easily back that out if you want to take a different approach. if you want to move forward with this approach, i can squash the two together, if you'd like.

travi:
anything else needed on my side to move this forward?

zkat:
Everything looks good :)

travi:
thanks!

PR #20762, opened 2018-05-25
Add synopsis to audit brief help and suppress trailing newline

wyardley:
I realize these are documented in `npm help audit`, but this brings `npm audit --help` a bit more in line with say, `npm install`, and makes it clear that there are other arguments.


zkat (approved):
Very nice! Thank you!

PR #20886, opened 2018-06-04
Fix typo in npm-scope.md

smfoote:


legodude17:
I am pretty sure this is intentional. `points` is being used as a third person noun.

smfoote:
You're totally right. I need to stop making pull requests so early in the morning :p

PR #20920, opened 2018-06-07
Republish waiting period

gmattie:
Explicitly state that republishing an unpublished package requires a 24h waiting period.


zkat:
While this is accurate, we're gonna be landing a change soon that changes this, so I'm gonna hold off on it. :)

jefflembeck:
Ahem, make that 72 hours, all around.

PR #20948, opened 2018-06-10
gyp: fix regex to match multi-digit versions

Trott:
Update to node-gyp@3.6.3 to accommodate LLVM 10.0.0. Earlier versions of
node-gyp had a regular expression that did not match the version string
'10.0.0'. node-gyp@3.6.3 fixes this bug.

Refs: https://github.com/nodejs/node/pull/21239
Refs: https://github.com/nodejs/node-gyp/issues/1454
Refs: https://github.com/nodejs/node-gyp/pull/1455
Refs: https://github.com/nodejs/node-gyp/commit/293092c362febffe19f72712467565045e08e8f1

Commit message from original patch:

    gyp: fix regex to match multi-digit versions
    
    This fixes the regular expression matching in `xcode_emulation`
    to also handle version numbers with multiple-digit major versions
    which would otherwise break under use of XCode 10
    
    Fixes: https://github.com/nodejs/node-gyp/issues/1454

### DIFF
--- a/node_modules/node-gyp/gyp/pylib/gyp/xcode_emulation.py
+++ b/node_modules/node-gyp/gyp/pylib/gyp/xcode_emulation.py
@@ -1262,7 +1262,7 @@ def XcodeVersion():
   except:
     version = CLTVersion()
     if version:
-      version = re.match(r'(\d\.\d\.?\d*)', version).groups()[0]
+      version = re.match(r'(\d+\.\d+\.?\d*)', version).groups()[0]
     else:
       raise GypError("No Xcode or CLT version detected!")
     # The CLT has no build information, so we return an empty string.


bnoordhuis:
In node-gyp's v3.x branch we pinned request to <2.82.0 for compatibility with old Node.js versions.  I'd suggest floating a patch here that relaxes that constraint so it can share a copy with npm (which is at 2.86.0, I think?)

Trott:
@bnoordhuis Certainly makes for a much easier-to-review change set. Done.

zkat (approved):
I'm going to accept this patch while https://github.com/nodejs/node-gyp/pull/1484 or similar lands. I really don't want a bunch more copies of `request` in our tree.

bnoordhuis:
That PR isn't going to land in node-gyp's v3.x branch so I closed it out, see discussion in that issue.  This PR is a good way forward for now.

PR #20955, opened 2018-06-11
Update `cli-table2` to `cli-table3`

Turbo87:
This PR updates the `cli-table2` dependency to `cli-table3`, which fixes one of the two remaining `npm audit` warnings :)

https://github.com/npm/npm-audit-report/pull/24 fixes the other warning.

`cli-table2` (like `cli-table` itself) is no longer maintained. In https://github.com/jamestalmage/cli-table2/issues/43 a couple of people have offered to take over maintenance but the current maintainer did not respond so as a result the project was forked to https://github.com/cli-table/cli-table3.


zkat:
I did this upgrade by hand on this repo (accepted the PR in `npm-audit-report`). We're all set here. Thanks! So happy to get rid of All Of `lodash`™. Saved like 300kb+ from the tarball itself (much more after extracting).

PR #20981, opened 2018-06-13
Add "Since" to the docs

luislobo:
There are several times I need to check if a certain commands is available for the npm version I'm using.


luislobo:
@iarna Are you interested in having this information?
If yes, I can go forward and add it to all the commands

zkat:
@luislobo I'm gonna /cc @neverett on this. They're in charge of docs now, and I know we all have plans to add more version-awareness to npm docs, make multiple versions available, etc. If and when we do that, I think this sort of entry would not be the exact way we do that, and I'd rather do it all in one fell swoop, as you suggest.

As such, I'm inclined to turn this specific one down, but I'm hoping we get around to making this sort of information available at some point in the future as more public docs work ramps up!

luislobo:
LGTM!

PR #20992, opened 2018-06-14
audit: allow the audit failure level to be configured

lennym:
`npm audit` currently exits with exit code 1 if any vulnerabilities are found of any level.

Add a flag of `--audit-level` to `npm audit` to allow it to pass if only vulnerabilities below a certain level are found.

Example: `npm audit --audit-level=high` will exit with 0 if only low or moderate level vulns are detected.

### DIFF
--- a/doc/misc/npm-config.md
+++ b/doc/misc/npm-config.md
@@ -164,6 +164,14 @@ When "true" submit audit reports alongside `npm install` runs to the default
 registry and all registries configured for scopes.  See the documentation
 for npm-audit(1) for details on what is submitted.
 
+### audit-level
+
+* Default: `"low"`
+* Type: `'low'`, `'moderate'`, `'high'`, `'critical'`
+
+The minimum level of vulnerability for `npm audit` to exit with
+a non-zero exit code.
+
 ### auth-type
 
 * Default: `'legacy'`
--- a/lib/audit.js
+++ b/lib/audit.js
@@ -252,11 +252,11 @@ function auditCmd (args, cb) {
         })
       })
     } else {
-      const vulns =
-        auditResult.metadata.vulnerabilities.low +
-        auditResult.metadata.vulnerabilities.moderate +
-        auditResult.metadata.vulnerabilities.high +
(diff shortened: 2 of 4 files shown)


lennym:
Use case: we'd like to include `npm audit` in our CI pipelines because it's a _brilliant_ feature, but also we don't want to worry about low level vulnerabilities because if we did then we'd never actually ship anything.

We can work around it by parsing the json and working it out for ourselves, but this PR was easier than doing that.

legodude17:
@lennym You also need to document this, add it to the types object in `lib/config/defaults.js`, and add some tests.

lennym:
I've updated to include the types.

Before I go too deep into writing tests and docs it's be good to get some steer as to whether this is a feature that might get merged _in principle_ as is (docs and tests pending). 

I'm not super keen to go and spent a few hours coding on this if it's not going to get merged or it's just going to end up going own a rabbit hole of expanding features. Obviously _very_ happy to do so if it results in getting the feature in a future version though.

legodude17:
In that case, I would suggest you drop something in https://npm.community/c/ideas about this.

lennym:
One step ahead... https://npm.community/t/allow-a-configurable-vuln-level-to-make-npm-audit-fail/245

I did try to post a reply pointing to this PR, but I've been put in time out there for some reason (I assume because I was freshly signed up and posted too many links - no complaint from me in that regard)

naugtur:
Shameless advertising:
Have you seen https://www.npmjs.com/package/npm-audit-resolver ?
It's meant for enabling audits use in CI

lennym:
Apologies for the radio silence on this. I've added some docs and tests in the hope that this might make it into a future release.

I couldn't find any obvious tests for the basic `audit` command - only `audit fix` so I created a new test file and attempted to reverse engineer the `audit fix` tests as best I could, so it might be worth someone who really understands how both the test runner and the audit command work better than I do taking a look over my attempts.

Cheers.

zkat:

PR #21002, opened 2018-06-16
Add npm audit resolve command

naugtur:
Add means for a human to resolve issues if they can't be fixed and interactively make decisions about each issue.

See https://npm.community/t/interactive-tool-to-manage-audit-findings-npm-audit-resolve/197

I'm still hoping to discuss this, as I assume I'm not aware of all npm-cli features and modes of operation.

TODO:

 - [ ] add handling of npm configuration like `prefix` that I don't know how to do as of now
 - [ ] make npm audit use the audit-resolv.json if present 
 - [ ] fix node6 support (a dependency is using `...` syntax)


zkat:
Hey, just dropping in a note so you know we saw this: I've been out sick for 3 weeks, and this week has been npm's yearly all-hands, so we haven't had a lot of time to be looking at these things.

I do have some thoughts about this off the top of my head, but you can wait until I do a more thorough review: my gut feeling is that I find the `fix`/`resolve` pairing confusing, as I would expect both of those to be aliases for one another.

That said, I _really_ like the idea of having an _interactive_ fix that has the ability to mute certain notices. Muting is definitely on our roadmap, and I think finding a good way to manage those mutes is probably a good first step. The big question is where to record it: I think writing it into `package-lock.json` itself would be good, for example.

I'm gonna /cc @npm/security-product to get their thoughts on this approach. Thanks so much for taking the time to put this together!

evilpacket:
Initial observations while checking this out. I think this is really cool and that an interactive mode could be useful for people or teams that want some hand holding or more information along the process. I do think some details could be improved and maybe slightly adjusted for long term product plans for things like (remind me in 24 hours).

- I don't like experience of the status bar being in the way of user input. I think if we are going to do something interactive that status bar has to change / not be in the way.

- What's the spec for the audit-resolv.json file? is that available someplace? I don't mind an additional file being added but want to make sure this file and format lines up with how we want to implement muting / exception on the registry side of things, because this won't just have impacts on the client side. nsp used to use .nsprc for this, but we have plans to improve upon this and doing so on the registry side vs in a file could have a nice network effect across the registry for reducing false positives.

`del) Remove all listed dependency paths`
- Does this mean to uninstall the vulnerable dep?

naugtur:
Let me reply in order

1. Making fix do the interactive part as well could be a way too. I assumed fix would stay as is, so it can work unattended for some use cases. Not my case really. I don't have a strong opinion on one vs two commands. A separate command seemed easier to explain and later ship as unstable feature until it matures. `npm audit fix` feels pretty solid the way it is.
  
2. There's more to the interactivity than just muting. If a fix is not available there's an option to investigate. I started working on some options there, including recommending where to go to github and PR a fix.

3. I don't think this should go to any of the existing files. Here's why: 
- package.json is already overburdened with all kinds of stuff 
- package-lock is not for that and there's plenty of usecases where deleting package-lock is a standard thing to do (my team does it every now and then). Also, package-lock is so big, noone's looking at its diffs in git history and audit-resolv.json is a nice track record of who decided to mute/postpone/fix what and when
- npmrc or alike - it'd have to apply to projects separately, also git history and visibility of the file
The feature of having audit-resolv be a track record for decisions is important (to my team as an example)

4. "remind me in 24h" - I introduced that as a way to unblock the CI pipeline if a fix is not there yet, but not a way to keep ignoring an issue, but I assume long term it's not enough. People will want choice :)

5.  I didn't expect it to collide with the progress bar in any way, definitely not intended that way.

6. There's no official spec for audit-resolv, but the format is a key-value map with keys identifying an issue in a particular dependency and value being how you decided to address it. The important part is if the same vulnerability shows up in another dependency you might use in a different way, it doesn't automatically get ignored. So if you decided to ignore something because it's a dependency of a tool you use for your build you don't want the same vulnerability to be ignored in dependencies of your production code if it shows up there later.

7. yes, del just uninstalls the top dependency where issue was found. I added it later, after I found I need it. I noticed some of the vulns were in old tools we no longer needed (like a dependency graph generator we didn't really need)

zkat:

PR #21005, opened 2018-06-17
docs: tips on cb() never called issues

jcrben:
This is definitely a bit of a stab in the dark, but I want to get this one started. Hopefully someone can build on top of this.

I haven't seen this error myself in a long time.

I see a lot of issues for this - a search for ["cb() never called"](https://github.com/npm/npm/issues?page=2&q=is%3Aissue+%22cb%28%29+never+called%22+sort%3Acomments-desc+is%3Aopen) pulls up over 400 issues - but hardly any demonstrate any real knowledge of why or even seem to do a bit of investigation or speculation. The only one that suggested any fundamental understanding, I've referenced here. Not sure if that's relevant to the others at all.

Perhaps there's room for more granular errors along the path to this particular error.

### DIFF
--- a/TROUBLESHOOTING.md
+++ b/TROUBLESHOOTING.md
@@ -26,6 +26,7 @@ For more details of the content and formatting of these entries, refer to exampl
 * [Cannot find module](#cannot-find-module)
 * [Shasum Check Fails](#shasum-check-fails)
 * [No Git](#no-git)
+* [npm ERR! cb() never called!](#cb()-never-called)
 
 ## Upgrading npm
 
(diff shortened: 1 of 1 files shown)


zkat:
Heya! I've started moving all the troubleshooting bits over to https://npm.community/c/support/troubleshooting and in the next couple of weeks, we'll be removing this TROUBLESHOOTING.md file. Do you mind putting this content in a topic over there, instead? Or if you'd rather I do it, I can copy-paste it. You can see the format for the posts over there.

Thank you for this, though! It's true that we seriously need to make sure folks are clear that this is a generic error.

PR #21019, opened 2018-06-20
Update npm-link.md shorthand comment

davidgilbertson:
I wasn't clear on the exact behaviour with the shorthand, but I think this change is correct.


zkat (approved):
Thanks for this clarification! `npm link` is so confusing. Can't wait to make it better soon :)

PR #21026, opened 2018-06-20
Fixes error code when no `test` has been defined

deiga:
When no `test` script has been defined `npm run test` returns an exit code of 0, which is confusing and makes working with other scripts prone to errors.

---
This change is [Reviewable](https://reviewable.io/reviews/npm/npm/21026)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,4 @@ npm-debug.log
 /node_modules/standard/
 /node_modules/tap/
 /node_modules/tacks/
+/test/npm_cache
--- a/lib/run-script.js
+++ b/lib/run-script.js
@@ -136,7 +136,7 @@ function run (pkg, wd, cmd, args, cb) {
   } else {
     if (!pkg.scripts[cmd]) {
       if (cmd === 'test') {
-        pkg.scripts.test = 'echo \'Error: no test specified\''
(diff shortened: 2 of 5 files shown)


deiga:
There seems to be a few flickering tests in `test/tap/shrinkwrap-save-with-existing-dev-deps.js` https://travis-ci.org/npm/npm/jobs/394524898#L1003

zkat:
As benign as it looks, this is definitely a breaking change, and I'm inclined not to take it. I think "no tests exist" being exit code 0 is pretty acceptable as far as default behavior goes. A lot of tooling might be relying on this behavior and doing automatic `npm t` in their CI scripts or such, and expect initial pushes of a repo to not fail the build.

I think @iarna and I can probably talk about this more while it's available, but I just wanna point out this is not a trivial one to accept, even if we do include it.

deiga:
@zkat I do understand that it's not a trivial change as it touches quite a core part of NPM.

But I do feel that `npm t` being empty just returning 0 is very wrong in a Unix way. Because `npm foo` returns 1 and `npm run foo` returns 1 if not defined.

And let's not forget that `npm init` creates a `test` script with `exit 1`

zkat:

PR #21030, opened 2018-06-20
config: Only translate environment variables up to the first colon

### LINKED ISSUE #15565: How to configure npm token via npm_config_* env var?

  - [ ] npm is crashing.

#### What's going wrong?

npm seems to ignore my `npm_config_//registry.npmjs.org/:_authToken` environment variable.

I'm trying to restrict the environments where my npm login token is stored. I tried using an environment variable in my `.npmrc` as described in #8356 but adding that config line makes npm (e.g. `npm --help`) crash with `Error: Failed to replace env in config: ${NPM_TOKEN}` in all shells where the var isn't set.
Luckily, ["npm is extremely configurable"](https://docs.npmjs.com/cli/npm#configuration), so I tried the "environment variables" approach. Using the `export` command as suggested in the docs didn't work because:

```text
bash: export: `npm_config_//registry.npmjs.org/:_authToken=00000000-0000-0000-0000-000000000000': not a valid identifier
```

… so instead I used `env` to spawn a new bash that would inherit my variable. Inside it, I used node.js to verify it is set correctly, then tried to publish:

```bash
$ env 'npm_config_//registry.npmjs.org/:_authToken=00000000-0000-0000-0000-000000000000' bash
$ # we're now in the inner shell
$ nodejs -p process.env | grep Token
  'npm_config_//registry.npmjs.org/:_authToken': '00000000-0000-0000-0000-000000000000',
$ npm pub .
npm ERR! Linux 3.13.0-108-generic
npm ERR! argv "/usr/bin/nodejs" "/usr/bin/npm" "pub" "."
npm ERR! node v6.9.4
npm ERR! npm  v3.10.10
npm ERR! code ENEEDAUTH
…
```

(Of course I used a token that worked in my `.npmrc` just seconds ago, so that should not be the problem.)

With the commands above, hopefully.

### supporting information:

 - `npm -v` prints: 3.10.10
 - `node -v` prints: v6.9.4
 - `npm config get registry` prints: http://registry.npmjs.org/
 - Windows, OS X/macOS, or Linux?: Linux (Ubuntu 14.04.5 LTS trusty)
 - Network issues:
   - [x] I use a proxy to connect to the npm registry.
   - [x] I use a proxy to connect to the web.
   - [x] I use a proxy when downloading Git repos.
 - Container:

mkhl:
As discussed on [npm.community][1], the fact that
npm registry authentication tokens
cannot be defined using environment variables
does not seem justified anymore.

The restriction is caused by the config loader translating

* all `_` to `-`
* the whole variable name to lowercase

while the credential checker expects a key ending in `:_authToken`.

As suggested, this change fixes the problem
by limiting the translation by the config loader
to the part *before* the first colon.

Closes #15565

[1]: https://npm.community/t/cannot-set-npm-config-keys-containing-underscores-registry-auth-tokens-for-example-via-npm-config-environment-variables/233

### DIFF
--- a/lib/config/core.js
+++ b/lib/config/core.js
@@ -346,11 +346,13 @@ Conf.prototype.addEnv = function (env) {
     .forEach(function (k) {
       if (!env[k]) return
 
-      // leave first char untouched, even if
-      // it is a '_' - convert all other to '-'
-      var p = k.toLowerCase()
+      // leave first char untouched, even if it is a '_'
+      // convert all other to '-', up to the first ':'
+      var ps = k.split(':')
+      ps[0] = ps[0].toLowerCase()
         .replace(/^npm_config_/, '')
         .replace(/(?!^)_/g, '-')
+      var p = ps.join(':')
       conf[p] = env[k]
     })
   return CC.prototype.addEnv.call(this, '', conf, 'env')


mkhl:
An alternative more limited approach would be to duplicate [this block of code that looks for the `…:_authToken` config key](https://github.com/npm/npm/blob/7e679fdc862a7176c290b51db91b189f5c6a0c95/lib/config/get-credentials-by-uri.js#L31-L35) and letting the copy look for `…:-authtoken` (lowercase and with a hyphen instead of an underscore).

zkat:

PR #21041, opened 2018-06-21
deps: update node-gyp to 3.7.0

MylesBorins:
new version with many fixed bugs 🎉


mistydemeo:
This includes the fix for Xcode 10, so it would be really great to see this make it into npm.

MylesBorins:
I believe that npm 6.2.0 is floating this as a patch.

MylesBorins:
Floating the fix to xcode that is

zkat:
I did this by hand and managed to do some `request` juggling to eliminate the weird duplication.

PR #21079, opened 2018-06-27
Fixed npm CLI version call format.

wstuckey:
npm did not show the 'v' in front of the version (as shown below); this change adds the 'v' in front of the returned value.

versioncall

### DIFF
--- a/lib/version.js
+++ b/lib/version.js
@@ -205,7 +205,7 @@ function dump (data, cb) {
 
   if (npm.config.get('json')) v = JSON.stringify(v, null, 2)
 
-  output(v)
+  output('v' + v)
   cb()
 }
 


zkat:
While I appreciate the desire to be consistent here, I don't think we should land this change: While it's a relatively minor fix, it has potentially very serious side-effects for the larger community. Anyone who has written a regex to parse the current npm version, ever, will now potentially break when consuming the output of `npm -v`. It's also worth noting that while _node_ made the choice to include the `v`, it's pretty common across the ecosystem (for example `npm view <pkg> version`) to not include that `v`. Partly _because_ it makes it more direct to parse using something like [`semver`](https://npm.im/semver).

Thanks for the time to put this patch together, and I understand where you're coming from with this, but I believe the risk vs reward ratio is too high to accept this. Cheers!

PR #21105, opened 2018-06-28
Add warning text about using temporary cache for debugging

karanjthakkar:
Inspired from a discussion with Kat on the WeAllJS slack.


zkat (approved):
lgtm 👍

PR #21119, opened 2018-07-01
Better explanation on OTP

jdeniau:
Use the defaut OTP explanation everywhere except when the context is "OTP-aware" (like when setting double-authentication)

Follows #19580

### DIFF
--- a/lib/auth/legacy.js
+++ b/lib/auth/legacy.js
@@ -52,7 +52,7 @@ function login (conf) {
     })
     .catch((err) => {
       if (err.code !== 'EOTP') throw err
-      return read.otp('Authenticator provided OTP:').then((otp) => {
+      return read.otp().then((otp) => {
         conf.auth.otp = otp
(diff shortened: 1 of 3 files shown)


zkat (approved):
👍 thanks for the follow-up!

lib/auth/legacy.js
```diff
@@ -52,7 +52,7 @@ function login (conf) {
     })
     .catch((err) => {
       if (err.code !== 'EOTP') throw err
-      return read.otp('Authenticator provided OTP:').then((otp) => {
+      return read.otp().then((otp) => {
```
zkat, line 55: In retrospect: I don't think we should change this one. `npm login` is intended to be a pretty interactive workflow and I don't think it's useful to tell people this is an error or dump a bunch of text on them. I think we can leave this one as-is.

zkat (changes requested):

zkat:

Issue #21114, opened 2018-06-30
npm@6.2.0-next.0 fails on postinstall script


simonua:
#### I'm opening this issue because:

  - [x] npm is crashing.
  - [ ] npm is producing an incorrect install.
  - [ ] npm is doing something I don't understand.
  - [ ] npm is producing incorrect or undesirable behavior.
  - [ ] Other (_see below for feature requests_):

#### What's going wrong?
Failed at the npm@6.2.0-next.0 postinstall script.
This leaves a broken npm install. I'll need to install it via the node installer and upgrade now, I think.

#### How can the CLI team reproduce the problem?
Attempt to run `npm i -g npm@6.2.0-next.0` on a Windows 7 64-bit machine.

[https://gist.github.com/simonua/be475b58b14967c5cfc3ca78bd97988a](https://gist.github.com/simonua/be475b58b14967c5cfc3ca78bd97988a)

![image](https://user-images.githubusercontent.com/8005145/42125601-2a0deebc-7c48-11e8-94da-0be91b8f14ae.png)

### supporting information:

 - `npm -v` prints: Fails as the installation is broken but previous version was 6.1.0
 - `node -v` prints: v10.5.0
 - `npm config get registry` prints: Fails as the installation is broken
 - Windows, OS X/macOS, or Linux?: Windows 7 64-bit
 - Network issues:
   - Geographic location where npm was run:
   - [ ] I use a proxy to connect to the npm registry.
   - [ ] I use a proxy to connect to the web.
   - [ ] I use a proxy when downloading Git repos.
   - [ ] I access the npm registry via a VPN
   - [ ] I don't use a proxy, but have limited or unreliable internet access.
 - Container:
   - [ ] I develop using Vagrant on Windows.
   - [ ] I develop using Vagrant on OS X or Linux.
   - [ ] I develop / deploy using Docker.
   - [ ] I deploy to a PaaS (Triton, Heroku).

zkat:
Oh.

Yeah that's borked. Thanks for trying the prerelease and spotting this 😅

By the way, in the future, can you post these to npm.community?

simonua:
Hi Kat,

Yes, definitely. How do I do that, please?

zkat:
@simonua https://npm.community/c/bugs

We're moving our issue tracking over there very soon :)

zkat:
@simonua #21129 was merged. Can you confirm this fixed it for you?

zkat:
(As in, I published to `npm@next`)

[...]

PR #21129, opened 2018-07-02
misc: remove postinstall script that broke stuff

zkat:
Fixes: https://github.com/npm/npm/issues/21114

/cc @simonua

### DIFF
--- a/package.json
+++ b/package.json
@@ -275,7 +275,6 @@
   },
   "scripts": {
     "dumpconf": "env | grep npm | sort | uniq",
-    "postinstall": "node scripts/gen-dev-ignores.js",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -275,7 +275,6 @@
   },
   "scripts": {
     "dumpconf": "env | grep npm | sort | uniq",
-    "postinstall": "node scripts/gen-dev-ignores.js",
```
iarna, line 278: This needs to be in prepare. We can't just delete it.

zkat:
update: talked with @iarna about it and I'm gonna merge this as-is. We will later add a script somewhere that gives us a better check and acts like a good check for this where appropriate.

PR #21165, opened 2018-07-07
Remove back-ticks not being parsed as markdown

hchiam:
I believe it should be:
`npm install foo`
instead of:
`npm install `foo``


zkat (approved):
👍 hah! Nice catch!

PR #21188, opened 2018-07-11
Use https when possible.

XhmikosR:
Also, fix a few redirects.


XhmikosR:
Oops, just noticed the move notice. Should I redo this there?

zkat:
Yep

PR #21199, opened 2018-07-12
Colonarie- test

karomaie:
First trial at node, hope that this attempt is on the right track


zkat:
I'm not sure what the goal here was, but this isn't really a sort of patch we're looking to accept -- npm has a more thorough test suite over at `test/tap/*.js`, and making a PR for this repo as your first experimental attempt at poking at node is probably not the best way to go about contributing. Cheers!

PR #21209, opened 2018-07-12
End of an Era

zkat:
This is it. This removes all our source code from this repro as part of [the migration to npm.community](https://blog.npmjs.org/post/175587538995/announcing-npmcommunity).

The diff is largely code removal, but you can see [the new README.md rendered](https://github.com/npm/npm/blob/zkat/doooooooom/README.md).


zkat:
So long, and thanks for all the fish. 🐬

